PR #35, opened 2014-10-17
Added Cookie interceptor.

petromykhailysyn:


twilson650:
What about a generic header interceptor? An implementation might require "x-api-version" or others.

PR #78, opened 2015-01-07
Update RestfulServer.java

ahenket:
I believe that in line 538 you are falsely assuming that multiple _* parts in the path could not happen. As of DSTU2 there's _meta/_delete. Unfortunately the DSTU for Comment erroneously lists _meta-delete.


jamesagnew:
[Jenkins Automated Message] Should I test this patch?

ahenket:
It is just a comment that says "FIXME + text". Testing unnecessary

PR #110, opened 2015-02-19
mysql configs and DSTU2 security labels

mochaholic:
Added (commented-out) mysql configs to hapi-fhir-server-database-config.xml and the corresponding class to facilitate mysql-based deployment.


jamesagnew:
[Jenkins Automated Message - This is not a real comment from James] OK TO TEST?

mochaholic:
Yes. This is a very small change.

mochaholic:
My latest commit includes support for DSTU-2 style security labels for both XML and JSON parsers. I ended up adding a method to the FHIR Context to return the suitable implementation of CodingDt based on the FHIR version. 
Note that I avoided modifying BaseCodingDt lest it will break other people's codes. Therefore, I only support Code, System and Display but I have added todo note and commented-out for supporting Version.

jamesagnew:
Ok, I'm merging this now. Nice work on it, thanks for the contribution!

I made a couple of tweaks to get it to support other attributes in CodingDt (version, active, etc.) and added a unit test, but otherwise it's going in as it.

PR #121, opened 2015-03-09
Added net.sf.saxon:saxon-he

esteban-aliverti:
Not sure if you are interested in this fix, but hapi-fhir-structures-hl7org-dstu2 was not compiling because saxon dependency was missing.

### DIFF
--- a/hapi-fhir-structures-hl7org-dstu2/pom.xml
+++ b/hapi-fhir-structures-hl7org-dstu2/pom.xml
@@ -33,6 +33,11 @@
 		<artifactId>xpp3_min</artifactId>
 		<version>1.1.4c</version>
 	</dependency>
+	<dependency>
+		<groupId>net.sf.saxon</groupId>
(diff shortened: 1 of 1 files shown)


esteban-aliverti:
I just saw that there is already a pull request regarding this issue. You can close this one after https://github.com/jamesagnew/hapi-fhir/pull/118 is merged.

jamesagnew:
[Jenkins Automated Message - This is not a real comment from James] OK TO TEST?

jamesagnew:
Thanks for the pull request!

Will merge this, looks like it's a newer version of Saxon than in #118 so might as well. :) As I mentioned in #118: FYI the HL7.org structures library isn't a part of the default HAPI build because we're still working on getting these structures integrated with the rest of the library. The goal is to have that work done in the next few weeks, but it won't be until after 0.9 is released which should be this week.

esteban-aliverti:
Sounds good. Thanks!

PR #153, opened 2015-04-09
Read server interceptors from the spring context.

mochaholic:


jamesagnew:
Thanks for submitting! Merging this now.

PR #200, opened 2015-07-27
Update vagrant provisioning with apt recipe to ensure repository package lists are up-to-date

corbinbs:
Hi,

I encountered errors running `vagrant up` as described here:  http://jamesagnew.github.io/hapi-fhir/doc_vagrant.html    It seems like the provisioning was attempting to run with out-dated repository package lists so I added the `apt` recipe at the beginning of the provisioning process to ensure `apt-get update` would run prior to downloading dependencies like mysql.   This allows me to `vagrant up` successfully and I can access the HAPI FHIR JPA Server via assigned_ip:8080/hapi-fhir-jpaserver/
Thought I'd pass this along in case other folks run into similar errors when running `vagrant up` to evaluate the FHIR server.  

Thanks,

Brian


jamesagnew:
This is awesome, thanks for the contribution!

PR #223, opened 2015-09-12
Adjust the checkstyle location to play nice on Win

SingingTree:
Removed the file:// prefix on the checkstyle config location. This appeared to be breaking the site build on Windows. I've also tested this on Linux (OpenSuse 13.2) and believe that change has not broken anything. However it would be good if someone else could make sure the site still builds with this change on Linux.

If there are reasons not to remove the file:// prefix, or issues with doing so, please let me know, and I'll attempt to remedy the build issues on Windows via other means.


jamesagnew:
Thanks for the submission!

PR #235, opened 2015-09-30
Change android build to work without having to "hack" the jar

botunge:
Made changes to include files in the android jar, also added the standalone android dstu, dstu2 and shaded jar to distribution, instead having an almost empty distribution in 1.2 release


jamesagnew:
Thanks for the submission! Merging now.

PR #239, opened 2015-10-05
Make dependencies truly optionel and clean up android pom.

botunge:
Phloc and Schematron has been made truly optionel and removed from the android jars to make them smaller.
I tried to remove javax.servlet and some of the server side classes from the client, but failed. So i undid the change but left some outcommented code in the pom.xml for android.

Im trying to make it easier to use the android jar in android and (using j2objc) in Ios.


botunge:
hmm..not sure i can do much about the decrease in coverage..

jamesagnew:
Don't sweat the coverage thing. :) Coveralls is a bit flaky and tiny variances are usually just noise.

This looks awesome, I'm planning on giving it a good review tonight or tomorrow and I'll merge it then.

botunge:
Thanks that sounds great... BTW.. i think you received an email today from Tomas at Systematic.. I'm on the guys he was writing about :)

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/BeanUtils.java
```diff
@@ -20,49 +20,37 @@
...
 
-import java.beans.BeanInfo;
-import java.beans.IntrospectionException;
-import java.beans.Introspector;
-import java.beans.PropertyDescriptor;
 import java.lang.reflect.Method;
 
+import ca.uhn.fhir.util.reflection.IBeanUtils;
+
 public class BeanUtils {
```
jamesagnew: FYI I've actually deleted this class in the latest trunk. Early versions of HAPI FHIR used method accessors via reflection to work with the model, but we've since moved to using field level accessors, which are somewhat faster. In the medium-term I'd like to look at coming up with something even faster.. I made one pass at trying to get ASM working to create dynamic accessors and mutators but my profiling showed that this wasn't actually any faster. I guess bottom line, more to come here....

hapi-fhir-base/pom.xml
```diff
@@ -132,6 +132,20 @@
 	</dependencies>
 
 	<build>
+		<plugins>
```
jamesagnew: The source plugin actualy gets invoked for all relevant modules as a part of [hapi-deployable-pom/pom.xml](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-deployable-pom/pom.xml), but only when HAPI is built with the `mvn -P DIST` argument/profile in order to be a bit faster.

Would it make sense to similarly make generating an android source JAR a part of the same profile?

jamesagnew:
I've merged this in now. Thanks for the pull!

This seems like a very positive improvement for the android library.. Would you be interested in joining the project as a committer to maintain this stuff more directly?

botunge:
Thank you for merging. 

I will take a look at the android sources jar and see if i can generate it using the same profile as the rest..

I might be able to contribute to (at least the android module) more directly..Currently i'm doing this "off work" even though we use it at work, so its hobbby based commits.

PR #244, opened 2015-10-15
Added example using embedded Jetty and Guice - meaning its way more e…

jkiddo:
Embedded Jetty - gotta lova that ;)


jamesagnew:
Very cool! Thanks for the contrib! Merging it now.

jkiddo:
awesome

PR #245, opened 2015-10-21
Fixed first server ignored in test overlay

samlanfranchi:


jamesagnew:
Thanks for the pull request! What an embarrassing fix to need on my part :)

PR #251, opened 2015-11-06
Server with Jax-Rs

petervanhoute:
When using a JEE container, we want to benefit from the integration capabilities provided b y JAX-RS. This fork add some interfaces around the classes such as RestfulServer, RequestDetails and ResponseDetails in order to remove the dependencies to the servlet-api. 
By making use of these interfaces, it become possible to create a jax-rs server. This fork includes a working jax-rs server base, jax-rs server example and the necessary changes to the hapi-base to make it work.


jamesagnew:
Hi Peter- Just wanted to mention that I'm not ignoring this. I am in the process of releasing HAPI 1.3 and I figured it would be lower risk to land this after that release was out the door.

Will look at merging as soon as 1.3 is done, which will be very soon.

jamesagnew:
Hi @petervanhoute ,

I've merged this now. Looks amazing! I've also added you in as a committer so you can do any maintenance to it you need.

In merging, I made one change to it. I don't think paging was working correctly as the patch stood. It required the URL to have the name of the resource in it, e.g.
`http://example.com/Patient?_getpages=foo`
instead of
`http://example.com?_getpages=foo`

A bunch of other things depend on the paging requests working from the root of the server so I fixed that.. but would you be able to look at what I did and see if it makes sense?

petervanhoute:
Hello @jamesagnew ,

I'm glad you like it! Thanks for merging and adding me as a commiter.

I was struggling with the paging and the URL's (self, next, resource references, ..). It's good you found an elegant, working solution for the paging and the url's. 

You can remove the commented line 389 in Dstu2BundleFactory.java:
//			theServerBase = theCompleteUrl.contains("?") ? theCompleteUrl.substring(0, theCompleteUrl.indexOf("?")) : theCompleteUrl;

Kind regards,

Peter

PR #271, opened 2015-12-15
Refactored RP and DAO layers after upstream merge

bdenton:
James..

I think this should be the correct pull request for this refactoring

cheers,
Bill Denton


hapi-fhir-dao-base/build.properties
```diff
@@ -0,0 +1,3 @@
+source.. = src/main/java/
```
jamesagnew: What's this file for?

bdenton: James…

It is an artifact of Eclipse Plug-In projects.. Have not really found a good use for it.. I have a suspicion that maybe it would get used if these were real Eclipse plugins…. Eclipse just always creates it

Cheers, Bill

hapi-fhir-dao-base/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,19 @@
+Manifest-Version: 1.0
```
jamesagnew: If these are just here so that you can open the project in eclipse, can we add them to the `.gitignore`? We do that with the Eclipse `.project` and `.classpath` files already.

Alternately I suppose we could try using `${project.version}` instead of `1.4.0` and Maven would substitute the correct version in.

I know they are important to OSGi and I'm sure we can find a way to get useful ones into the final artifacts, but I'm just really nervous about leaving hardcoded versions in these files if they are checked into Git. I need to be conscious of making Maven handle the entire build process without needing any manual file tweaking a part of the process. Having managed the HAPI for HL7v2 project for roughly a decade now, if there's one thing I'm sure of it's that I'll miss manual steps sometimes and get tons of angry emails about it.... That part of running an open source project can be annoying....

bdenton: I am definitely on the same page as you about hard-coded stuff

The thing is that without the Manifest, you can’t open and compile the project in Eclipse… The META-INF/MANIFEST.MF is not included in the resulting JAR so not so sure how the ${} substitution would work out.. The generated JAR has just the manifest generated by the maven-bundle-plugin.. that has hardcoded stuff as well that I am still working to reduce.. particularly the ca.uhn package versions.. Is there a Maven variable that does not have the SNAPSHOT suffix?

Also, Maven is not great at handling OSGi dependencies.

1)      Wants to put “bracketed” versions on all imports.. I mentioned about that in a previous post

2)      Does not pick up secondary dependencies.. there are always package imports that are not discovered until you try to deploy the bundle.. to be fair, the same problem exists in Eclipse’s dependency tracking. It always takes a few tries to get a new bundle to deploy in an OSGi container for the first time… deploy, resolve errors, adjust manifest, retry

jamesagnew: Oh, wait.. So these files aren't actually a part of the final build. I hadn't noticed that. In that case I don't think it's such a big deal...

bdenton: Yes.. just for Eclipse IDE use.. same as .project, .classpath and .settings directory

The “real” manifest is generated by maven-bundle-plugin.. the pom has a bunch of hardcoded stuff that I am working on reducing

bdenton:
I pushed my updates to the root project POM.. the travis build should move a bit farther

bdenton:
James.. 

I am going to close this pull request. I have been doing some more thinking and will give this another shot from a different direction. I'll send you a new pull request in a couple of weeks.. Meanwhile, I have one or two other OSGi things I'll send your way for merging

cheers, Bill Denton

PR #279, opened 2016-01-07
Resolves #278: Update JpaServerDemo to also support DSTU 2.1.

### LINKED ISSUE #278: The hapi-fhir-jpaserver-example project's JpaServerDemo supports DSTU1 and DSTU2, but not DSTU2.1
This class has a number of `if ... else` blocks that cope with the differences between DSTU1 and DSTU2, but does not yet include any logic for DSTU2.1.

Not a big deal, but would be nice.

karlmdavis:
Pretty minor set of changes, but makes it somewhat easier for folks to support DSTU2.1 when using hapi-fhir-jpaserver-example.


jamesagnew:
Thanks for the contribution! Merging it now.

PR #288, opened 2016-01-20
Exposing parameters for HTTP client request pool max connections in RestfulClientFactory

matt-blanchette:
The current RestfulClientFactory uses the default HTTP client PoolingHttpClientConnectionManager settings for max number of connections.
The defaults for these are 20 max total connections and 2 max per route.

These changes expose parameters on RestfulClientFactory to allow overriding these defaults.

The default for max connections per route is also changed here from 2 to 20 (using the same value as max total connections).

Other possible approaches:
* Expose HttpClientBuilder in RestfulClientFactory, allow subclasses to override settings
  * Expose a setter for IRestfulClientFactory in FhirContext to allow subclassing


jamesagnew:
Thanks! This looks awesome, merging it now.

PR #289, opened 2016-01-22
ServletContextPath can start with servletPath, it should be taken into consideration

petromykhailysyn:
… root) we should take it into consideration while determinating server base.


jamesagnew:
Thanks Petro! Looks great, I'll merge this now.

Issue #291, opened 2016-02-03
Enhance HAPI RestfulServer to support adding resource provider as CDI bean in JBoss enviroment


fw060:
We have encountered an issue in RestfulServer.bindMethod when we trying to add a resource provider using CDI in JBoss Wildfly environment (see code below). 

In this setup, the resource provider passed in to RestfulServer is not PatientProvider class, but a proxy class generated by JBoss. Then bindMethod would have trouble inspect getPatient method in PatientProvider. After some investigating, we find out that this is caused by the issue that Jboss Proxy loses generic type information (see https://issues.jboss.org/browse/WELD-1539). This will lead to error when bindMethod trying to inspect getPatient method which would return a List of Patient. bindMethod can only see List without any generic type so a null pointer exception would happen. 

This is NOT an issue in HAPI, however, maybe solving this issue in HAPI is the cleanest way since other CDI environment might have similar problem. 

Our current thought is to create a new method:
public void addResourceProvider(IResourceProvider theResourceProvider, Class theProviderClass) 
which would allow designating the Provider class rather than deceiving provider class from the provider instance itself. 

I will be more than happy to provide any more detailed information. If anyone has a better/cleaner solution, please also let us know. I will provider the pull request hopefully today or tomorrow. 

Thanks!
Fei

public class FhirRestfulServlet extends RestfulServer {
	@Inject
	PatientProvider patientProvider = null;
	
	public FhirRestfulServlet() {
		super(FhirContext.forDstu2());
	}
	
	@Override
	protected void initialize() throws ServletException
	{
		/*
		 * The servlet defines any number of resource providers, and
		 * configures itself to use them by calling
		 * setResourceProviders()
		 */
		List<IResourceProvider> resourceProviders = new ArrayList<IResourceProvider>();
		resourceProviders.add(this.patientProvider);
		setResourceProviders(resourceProviders);
	}
}

public class PatientProvider implements IResourceProvider {

    @Search()
    public List<Patient> getPatient(@RequiredParam(name = Patient.SP_IDENTIFIER) StringParam identifier) {
   
}
}

jamesagnew:
Hi Fei,

Interesting problem :)

Is the issue that because of the proxy that Weld creates, the server sees
```java
@Search()
public List getPatient(@RequiredParam(name = Patient.SP_IDENTIFIER) StringParam identifier) {
```
instead of 
```java
@Search()
[...]

PR #292, opened 2016-02-04
Enhance HAPI RestfulServer to support adding resource provider as CDI…

fw060:
… bean in JBoss enviroment #291


jamesagnew:
Resolved per #291

PR #293, opened 2016-02-05
Allow loading from local file/caching

adam-carbone:
I have added the ability to load data zip files from a local file/directory 

- If the path passed in points to a local zip file it will load the resources from inside.
- If the path points to a local directory it will look inside for zip files in that directory and load the resources

Also used a temporary file in the ~/.hapi-fhir-cli/cache as a temporary file for the downloaded example file, this allowed for usage of a FileInputStream instead of a ByteArrayInputStream, thus allowing us to handle bigger files because they are not loaded completely into memory. 

Also provided for automatic caching of that local temp file for up to 12 hours using the -c option.


adam-carbone:
Hmmmm that coverage change is talking about a class that I didn't touch... @jamesagnew What can I do to fix the change?

adam-carbone:
The coverage change is indicating a drop in coverage in classes that I didn't even touch, I only changed one class. @jamesagnew  hat can I do to help to help resolve that?

jamesagnew:
The coveralls check is probably nothing. Coveralls often seems to get it wrong for pull requests, I've never figured out why.

This looks great, thanks for the contribution. Merging it now!

PR #303, opened 2016-02-25
Jaxrs client

petervanhoute:
When using HAPI-FHIR in a JEE container we want to have as little external dependencies as possible, given that they might conflict with other versions. 
For the http requests, we would like to use JaxRs, as this is available on each JEE server. In order to avoid dependencies to apache core and client, I've added API's around apache's client and core classes: IHttpClient, IHttpRequest, IHttpResponse. This allows to use alternative client technologies. 

This branch contains these API classes together with 2 implementations:
- **Apache** in the module base _hapi-fhir-base_ and package _ca.uhn.fhir.rest.client.apache_

- **jax-rs** in the module _hapi-fhir-jaxrsserver-base_ and package _ca.uhn.fhir.jaxrs.client_

The dependencies to apache have been exclude from the pom.xml of hapi-fhir-jaxrsserver-base. 
The tests ensure that these apache artifacts are not required on run-time as well.

### DIFF
--- a/examples/src/main/java/example/ClientExamples.java
+++ b/examples/src/main/java/example/ClientExamples.java
@@ -3,11 +3,11 @@ package example;
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.rest.client.IGenericClient;
 import ca.uhn.fhir.rest.client.IRestfulClientFactory;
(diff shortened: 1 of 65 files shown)


jamesagnew:
This has been merged. Just looking into the issue you pointed out by email where binaries and param requests don't get headers. I believe this is probably a bug in the original, so I'm going to fix it. Will update then.

PR #321, opened 2016-03-24
Tinder Plugin Enhancements

bdenton:
This pull request adds some new functions to the HAPI Tinder Plugin:

1) The ability to use TInder to generate resource-related files using an external template. This allows TInder to be used from new projects without having to add the template(s) to the hapi-tinder-plugin project.

2) Allow Tinder to be used from Ant. This makes the Tinder plugin accessible from non-Maven build streams.


jamesagnew:
This is cool!

Merging now.

PR #359, opened 2016-05-04
Fix: Add null check in JsonParser, Relax property handling introduced in Jaxp 1.5

subhrajyotim:
Hello James,
While developing with HAPI, based on the data I am using, I have made 2 modifications:
-  to handle Null check in case of arrays in JsonParser
- To relax the property settings which are only supported by Jaxp1.5 only, some of the jvms being used still now have Jaxp1.4, to maintain backward compatibility, in SchemaBaseValidator

Please review, and see if this could help in anyways.

Thanks,
Subhro.


jamesagnew:
Thanks for the contribution! Looks good, merging it now.

PR #361, opened 2016-05-05
Wrong comment, unable to set type in new versions

Virdulys:


jamesagnew:
Hi @Virdulys, thanks for the patch, I'll merge it now.

I should point out though that if you're doing new development you should consider migrating to the DSTU2/STU3 Bundle resources instead of the DSTU1 Bundle that this patch fixes. This class will be removed at some point now that DSTU1 is starting to go away.

PR #367, opened 2016-05-18
Jaxrs server evolution

SRiviere:
Modification inside the Jax-rs framework
- enable the conditional update and delete
- creation of a bundle provider, and support of the @Transaction
- bug fix on the exceptions handling as some exceptions throw outside bean context were not intercept.
- add the possibility to have the stacktrace in the jaxrsException


jamesagnew:
This looks excellent, thanks for the contribution. Merging it now!

PR #404, opened 2016-07-13
Fix missing ID with JAX-RS and DSTU2 RI

mion00:
While using DSTU2 reference implementation, the ID was not correctly initialized in the request object


jamesagnew:
Hi Carlo, Thanks for the contribution! Looks great, merging it now.

PR #424, opened 2016-08-13
Parse/encode Json-like data structures

bdenton:

1) When building an equivalent of the jpaserver implemented in native MongoDB (no JPA or JDO), it was necessary to be able to serialize/deserialize FHIR resources via BSONReader and BSONWriter. This is possible (with some help) with Json-P but not with Gson.

2) We have build a general purpose data mapping engine that is able to map from a set of customer/application-specific sets od data into FHIR resources. The output of this mapping engine is a JSON-like data structure implemented in Java Maps, Lists, and scalars.

This pull request generalizes the IParser interface and, specifically, its JSONParser implementation so that any sort of Json-like data structure or framework can be integrated in the future.


bdenton:
James..

I notices that the Travis build for thus pull request has failed on the jpaserver-base project. However, I do not see the same failure on my Windows-10 system at the same code base. I also cannot see how the SQL constraint violation is related to this commit.

Thoughts on the Travis error or on the pull request in general??

cheers,
Bill Denton
Akana, Inc.

bdenton:
James..
I am going to cancel this pull request. I messed up the last commit somehow. I meant to just merge/sync your latest update to BaseParser but it ended up pushing 2001 files... I'll get it worked out and resubmit next week.

cheers, Bill Denton

PR #432, opened 2016-08-30
Parse/encode Json-like data

bdenton:
(This is resubmission after fixing the bad merge in my fork)

1) When building an equivalent of the jpaserver implemented in native MongoDB (no JPA or JDO), it was necessary to be able to serialize/deserialize FHIR resources via BSONReader and BSONWriter. This is possible with a lot of work with Json-P but not with Gson.

2) We have built a general purpose data mapping engine that is able to map from a set of customer/application-specific sets of data into FHIR resources. The output of this mapping engine is a JSON-like data structure implemented in Java Maps, Lists, and scalars.

This pull request generalizes the IParser interface and, specifically, its JSONParser implementation so that any sort of Json-like data structure or framework can be integrated in the future.


bdenton:
James...

1) Have you had a chance to look over this pull request? If this is something you are interested in, I would like to get it merged in some time soon.

2) I don't know how to kick off the CI test that was failing before your latest commit. Can you tell me how I can resubmit the Travis build for the pull request without having to commit code changes in my fork?

cheers,
Bill Denton
Akana, Inc.

jamesagnew:
Hi Bill! Finally getting round to reviewing this one.

I'm good with this in principle, but I'm not sure I follow why `IParser` needs to change. I feel like this will be confusing to people using the interface, since that's basically the first thing new users of the API play with generally. And IParser has no notion of JSON or XML today, it's intended to be implementation agnostic.

Eventually we'll be adding a third IParser implementation for the turtle format, and I don't see any reason why a BSON one couldn't just be a top-level parser type too.

Is the reason that things worked before for JSONP but don't work now for Gson because of Gson's not using interfaces instead of concrete classes?

bdenton:
James..

I updated the interface mainly because I needed to invoke the parser/encoder using something other than Reader and Writer. This is the same reason why BaseParser was updated. I had started out just adding the new method signatures to JsonParser but, then. Consumers would have to cast the results from the FhirContext parser factory methods to a concrete implementation class… I did not feel too comfortable doing that from outside the HAPI projects.

The fact that my Mongo stuff stopped working with the switch to Gson was only part of the motivation for these changes. The real motivation was that I wanted to be able to use HAPI’s understanding of how to build and navigate resources to interface with new data structures based on java Maps, Lists, and scalar objects. So I am successfully using the JsonLikeStruture interfaces to do a bunch of things with this Map/List structure including JsonPath navigation

We are building a data mapping layer between customer-supplied non-FHIR data sources and HAPI-hosted FHIR resources. Our first approach started out looking very much like what was already being done by the HAPI parser classes. We thought it would be better to leverage what is already there rather than reinventing all of the annotation processing and detailed knowledge of how to build FHIR data types.

BTW, I also have a branch in my GIT fork that implements the same stuff for the Json-P parser in HAPI 1.6

Cheers, Bill

jamesagnew:
Hmm.. Do you think it would be realistic to put the required hooks into JsonParser and then have your BsonParser that extends JsonParser?

I'm thinking something along the lines of you have your BsonParser with a constructor that takes in a FhirContext and handles the casting stuff internally. Users of your API would then do something like `new BsonParser(ctx);`

That way the change would be invisible to users of "plain HAPI" but would be pleasant for people using your.

bdenton:
James..

My thinking had not really involved having a BsonParser but just have an alternative source/target data representation that would be produces and consumed by the “plain HAPI” parser. Some of the key points I was trying to accomplish were:

·         Have parser/encoder sources and targets other than Reader and Writer

·         Change the JSON parser to be implementation agnostic by wrapping the implementation (Json-P, Gson, etc)

·         Still allow parser/encoder consumer to deal with just interfaces and let FhirContext figure out the correct implementation

So, I guess I am not crazy about the BsonParser concept for two reasons:

1)      Bson is not an alternate structure like turtle would be. It is just an alternate representation of Json.

2)      Not keen on having the consumer directly instantiate the parser implementation class

Cheers, Bill

jamesagnew:
Maybe this will make more sense if you show an example of how this would actually get used. I'm just not understanding what kind of code would be able to take advantage of the new "JsonLikeStructure" API without needing anything custom. 

I'm sorry for being difficult here- do understand though that I (and my colleagues here) are the front door for every support request that comes in, and if we start adding weird methods right to the parser API I already know that it will generate a whole slew of emails and support problems from people who don't understand what this is for and try to use it for weird stuff. Such is the nature of running an open source project...

I want to accomodate the needs of your implementation for sure, but I also want to find a way to do it without cluttering the core API with single-use hooks any more than needed.

bdenton:
James..

Each implementation would have a custom implementation of the JsonLikeStructure and, maybe, the JsonLikeWriter. These would then get passed into the IParser (maybe) interface instead of a simple Java Reader or Writer.

The default implementation (via the methods that take Reader and Writer) instantiates the Gson implementation of the JsonLikeStructure and JsonLikeWriter. The idea is to change the JsonParser to be able to create and consume any Json-like data representation to create or serialize resources.

Below is a simple code snippet of from my Mongo adaptation of the JPA server. This has the Resource stored as an opaque object inside the Mongo document.

       protected Map<String,Object> getAdapterResults (MongoResourceEntity<T> entity, Set<String> requestedFields) {
              JsonLikeMapWriter jsonLikeWriter = new JsonLikeMapWriter();
              jsonLikeWriter.setResultMap(new EntryOrderedMap<String,Object>());

              IParser jsonParser = getFhirContext().newJsonParser();
              try {
                     jsonParser.encodeResourceToJsonLikeWriter(entity.getResource(), jsonLikeWriter);
              } catch (DataFormatException | IOException e) {
                     throw new ValidationException("Error extracting data from MongoDB", e);
              }

              Map<String,Object> result = jsonLikeWriter.getResultMap();
              result.put(IInputSource.NATIVE_PRIMARY_KEY, entity.getNativeId());
              Map<String,Object> metadata = buildMetadata(entity);
              if (metadata != null) {
                     result.put(IInputSource.METADATA, metadata);
              }

              return result;
       }

MongoResourceEntity is the MongoDB document that has the resource embedded (entity.getResource() method)
EnryOrderedMap is exactly what it says.. it is a Map that returns Entry elements in the order they were added
The buildMetadata() method is similar to the equivalent in the jpaserver

I have attached the JsonLikeMapWriter and JsonLikeMapStructure for your reference. I have also included an implementation of the JsonProvider that allows JsonPath processing of this Java Map-based Json structure

I guess the real objective is to be able to leverage the existing HAPI parser/encoder to process source and target representations other that strings.. Maybe we could generalize this to include XML or other structures? I don’t know. My feeling is that the “structure” is what is unique and, at least for Json,  I was trying to separate that from its representation as a string.

Cheers,
Bill Denton

jamesagnew:
Ok, I get the intent here. I think really this could probably be a way to get Turtle structures written too at some point which would be cool.

I'll give this a closer look this weekend.

bdenton:
James..

Yes, you could easily implement turtle structures using this framework.. The idea was to decouple the building of the representation from the parser logic.. with the new framework. You should be able to support anything that even remotely smells like Json without needing to touch the parser classes. If I had submitted this pull request much earlier, the conversion from Json-P to Gson could have been mucheasier and not need anything done in the parser package.

Cheers, Bill

bdenton:
I am going to close this PR and resubmit one that does not involve changes to `IParser` or `BaseParser`

PR #451, opened 2016-09-17
Added the possibility to search for reference using OR search

botunge:
I've opened up, so that the ReferenceClientParam, can take a list of ids, used in an OR search and added a couple of tests displaying that it works in DSTU2 and STU3


jamesagnew:
Thanks for the contribution! Merging it now

jamesagnew:
Ok, this has been merged. Thanks for the contribution!

FYI I have renamed the method from `ids(..)` to `hasAnyOfIds` just to keep consistent with the other methods in that class, but otherwise it's been added as is.

PR #455, opened 2016-09-23
DSTU1: Add entry author to encoded xml bundle

hnnesv:
It seems the XML encoder for DSTU1 wasn't adding in the author for bundle entry; this just adds the functionality.


jamesagnew:
Thanks for the submission! Merging it now!

PR #456, opened 2016-09-23
added search parameters on Observation for bodySite and interpretation

Robbert1:
Hi,

I'd like to be able to search Observations using the following search parameters. Currently I have to create a subclass as well as a custom non-generated ResourceProvider and override a lot of spring configuration in order to get this working. Which is pretty nice already.

But this pull request would allow out-of-the-box index based searching using the HAPI-FHIR JPA server impl.

I know that the org.hl7 classes seem directly related to the spec but I'm not sure where else to start when requesting a change. Furthermore I'm not exactly sure if there are any particular criteria for a field to be added as a search parameter or not?

Anyway this pull request at least visualized the exact change that I'd like to see made (together with regeneration of the ObservationResourceProvider), hope its clear.


jamesagnew:
Hi Robert,

Unfortunately this one can't be merged..

The model classes are actually code generated based on the HL7 specification itself, so even if I do merge this it will be overwritten the next time I sync in the generated model classes.

My hope is that at some point we'll have an easy mechanism for adding custom search parameters to the JPA server, since it's perfectly acceptable to do this per the FHIR spec. It never really occurred to me that one workaround one could use to accomplish this would be a fork of HAPI with custom params added right to the resources (as you've done). That's a pretty elegant way of doing it actually.

I guess in the end, you have a few options: You can certainly propose that your params be added to the core spec. I have no idea whether they would be more broadly accepted or not, but it certainly doesn't hurt to try. Probably asking a question on the implementer stream of http://chat.fhir.org would be a good place to start that conversation if you wanted to..

In the more immediate term, I'd probably just say your easiest approach is to keep maintaining your fork with the extra params added and keep it in sync with HAPI releases. At some point we'll have a "native" way of adding custom params.

Cheers,
James

PR #467, opened 2016-10-13
fix resources links being correctly indexed when linking eachother as part of a single transaction bundle

Robbert1:
When creating a Patient with Id and an Observation with a reference to that patient in a single Bundle (transaction). The logic seems to persist/merge the entities multiple times which causes their Links to be reset (due to LAZY loading of the relation). This does not actually trigger a missing link, but it does fail to actually index that link using lucene.


jamesagnew:
Hi Robert,

Don't suppose you have any code that illustrates the issue? This looks like a pretty low risk change so I certainly don't object to merging it but I wouldn't mind having a unit test to make sure we don't cause regressions in the future if that's possible.

jamesagnew:
BTW, thanks for the contribution :)

PR #469, opened 2016-10-15
Support JSON-like resource structures

bdenton:
(resubmitting after recovering my branch from a couple of bad merges)

1) When building an equivalent of the jpaserver implemented in native MongoDB (no JPA or JDO), it was necessary to be able to serialize/deserialize FHIR resources via BSONReader and BSONWriter. This is possible (with some pain) with Json-P but is not at all possible with Gson.

2) We have built a general purpose data mapping engine that is able to map from a set of customer/application-specific sets of data into FHIR resources. The output of this mapping engine is a JSON-like data structure implemented in Java Maps, Lists, and scalars.

This pull request extends the IParser interface and updates the JSONParser to implement the sub-interface so that any sort of Json-like data structure or framework can be integrated in the future.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/IJsonLikeParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/IJsonLikeParser.java
@@ -1,0 +1,101 @@
+package ca.uhn.fhir.parser;
+
+/*
+ * #%L
+ * HAPI FHIR - Core Library
+ * %%
+ * Copyright (C) 2014 - 2016 University Health Network
+ * %%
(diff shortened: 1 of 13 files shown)


bdenton:
James.. 
are you OK with this commit even though I can't seem to get Coveralls to give me a good score?
cheers, Bill

jamesagnew:
Bill, this is fantastic. I know it was painful to get it resubmitted with git being the nuisance it can be... but this looks awesome now. I'm gonna merge it now.

jamesagnew:
Oh, and yeah.. Coveralls can be weird sometime. It doesn't appear that coverage changed in any meaningful way so I'm not to worried about that.

PR #470, opened 2016-10-16
add the possibility to search for resources using OR profile search

botunge:
The second place where i've made it possible to do OR searches using HAPI Fhir.. Now on the profile, including a JPA server test.

I would like to open up on Date for allowing for OR searches.. But i'm having trouble finding an approach for a nice fluent API, that still supports the "old/existing" way of writing Date searches.. Update - The second commit shows a way it could be done.. You can include the second commit if you like it, or leave it if you dont :)


jamesagnew:
Thanks for the contribution! Merging now

PR #474, opened 2016-10-24
Fix formatting for LOINC example

vadi2:
At least I hope I do - don't know how to render it, so I'm just going off the SNOMED example above.


vadi2:
I don't know did it decrease - I did not modify any code. Seems to me it is counting wrong.

jamesagnew:
Hi @vadi2 , thanks for the PRs! Yeah, you can definitely ignore the coveralls warning.. I haven't figured out why but it seems to detect tiny changes in coverage even in changes that don't touch the code at all. I keep meaning to carve out some time to look at that.

Either way, will try to review these soon :)

PR #489, opened 2016-11-09
Added new independant overlay build for example usage and lenient handling

jkiddo:


jkiddo:
Added lenient handling on server

jamesagnew:
Hi Jens,

Thanks for the contribution- I'm going to merge it now, and move it into a subdirectory off of the root project (the root is getting a bit unweildy, I want to organize the example projects a bit).

Would you be willing to maintain this example? My main apprehension in taking on sample projects is that they get stale over time.. The JPA Vagrant server is a good example of that, it no longer works and I have no idea how to fix it, and the submitter hasn't kept it up to date. If you're up for it, I'd love to give you committer access so you can just keep it up and running yourself :)

jkiddo:
It would be an honour ;)

PR #493, opened 2016-11-10
Fixed typo in valueset constraint

jimsteel:


jamesagnew:
Thanks for the submission, will merge now. I think this may be a symptom of a bigger issue though, since this file is sync'ed out of the core spec. I think that sync may not have worked correctly if we have an out-of-date schematron.. will investigate.

jamesagnew:
Oh actually scratch that.. This is a bug in the source definition.

jamesagnew:
Fixed in FHIR SVN commit 10156

Issue #517, opened 2016-11-25
new RestfulClient with old (pre CapabilityStatement) stu3 structures fails


lawley:
When running with latest hapi-fhir-base (2.2-SNAPSHOT) and older hapi-fhir-structures-dstu3 (before switch from Conformance to CapabilityStatement), validateServerBase throws a DataFormatException because the CapabilityStatement resource definition is unknown.
```
ca.uhn.fhir.parser.DataFormatException: Unknown resource name "CapabilityStatement" (this name is not known in FHIR version "DSTU3")
	at ca.uhn.fhir.context.FhirContext.getResourceDefinition(FhirContext.java:433)
	at ca.uhn.fhir.rest.client.RestfulClientFactory.validateServerBase(RestfulClientFactory.java:290)
	at ca.uhn.fhir.rest.client.RestfulClientFactory.validateServerBaseIfConfiguredToDoSo(RestfulClientFactory.java:203)
	at ca.uhn.fhir.rest.client.BaseClient.invokeClient(BaseClient.java:189)
	at ca.uhn.fhir.rest.client.BaseClient.fetchResourceFromUrl(BaseClient.java:121)
	at au.csiro.ontoserver.web.fhir.ConformanceTest.testRead(ConformanceTest.java:78)
```

It looks like there is fallback code to handle the case where the client knows about CapabilityStatement but the server, while still being FHIR 1.6.0, doesn't.  However this doesn't handle the case where the client doesn't yet know about CapabilityStatement.

jimsteel:
See #518

jamesagnew:
Closing as the PR has been merged.

lawley:
Looking a little deeper at this, I see that `FhirVersionEnum` is part of `hapi-fhir-base` and includes the specific fhirVersion numbers, but I think these should really come from the specific `hapi-fhir-structures-dstu3` so that it matches, for example, `org.hl7.fhir.dstu3.model.Constants.VERSION`

This is probably yak shaving though, since the FHIR practice of not distinguishing the fixed version (eg final 1.0.2) from a moving version (as 1.6.0 currently is) makes this hard.  If FHIR used a suffix like 1.6.0-SNAPSHOT to distinguish rolling versions, we might be better able to handle this stuff in code (or at least know when we're on shaky ground).

PR #518, opened 2016-11-25
#517 Add fallback for for pre-CapabilityStatement STU3 client

lawley:
Handle case where stu3 structures don't contain CapabilityStatement and fall back to Conformance


jamesagnew:
This looks good, merging now. Thanks for the pull request!

Issue #516, opened 2016-11-24
$expand with an invalid filter returns 500 instead of 422


jimsteel:
POSTing a ValueSet to $expand, where the valueset uses an invalid filter operator, results in an IllegalArgumentException. It would be good if this could be caught somewhere and thrown as a UnprocessableEntityException, so it comes out as a 422 response.

Example follows.

`curl -X POST -H "Content-Type: application/json+fhir" -H "Cache-Control: no-cache" -d '{
  "resourceType": "ValueSet",
  "url": "http://sample/ValueSet/education-levels",
  "version": "1",
  "name": "Education Levels",
  "status": "draft",
  "compose": {
    "include": [
      {
        "filter": [
          {
            "property": "n",
            "op": "n",
            "value": "365460000"
          }
        ],
        "system": "http://snomed.info/sct"
      }
    ],
    "exclude": [
      {
        "concept": [
          {
            "code": "224298008"
          },
          {
            "code": "365460000"
          },
          {
            "code": "473462005"
          },
          {
            "code": "424587006"
          }
        ],
        "system": "http://snomed.info/sct"
      }
    ]
  },
  "description": "A selection of Education Levels",
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\"><h2>Education Levels</h2><tt>http://csiro.au/ValueSet/education-levels</tt><p>A selection of Education Levels</p></div>"
  },
  "experimental": true,
  "date": "2016-07-26"
}' "http://fhirtest.uhn.ca/baseDstu3/ValueSet/%24expand"`

jamesagnew:
Thanks for reporting! Checking in a fix now.

FYI the fix for #525 had the side effect that this invalid code would now be ignored entirely and the server would just ignore the filter. With the checkin about to come, the server return an HTTP 400 when it receives an invalid code (unless the parserErrorHandler has been configured to ignore this problem)

PR #525, opened 2016-12-06
#516 Report invalid values for enums through parser error handler

lawley:
This patch catches the low-level IllegalArgumentException from the model code and instead reports an invalidValue through the parser's error handler.
Unit test included


jamesagnew:
Thanks for the contribution!

PR #533, opened 2016-12-29
correct ordering of codesystem dao types

Robbert1:
The latest version of Spring is no longer able to run the JPA server due to an issue with the ordering of types. This pull request should solve the issue.


jamesagnew:
Thanks for the contribution! 

Man.. I had no idea what was keeping Spring from booting properly with the latest versions. I'm glad you have sharper eyes than I do! :)

PR #543, opened 2017-01-19
Jaxrs sever evolution

SRiviere:
Like discussed in Amsterdam FHIR dev, i added the possibility to use the configuration of the baseServerUrl to generate the extension url, via a key word: %BASE_SERVER_URL%, which is replace during the parsing of the resource if present.

I also modify the validation to use the datatype profile when available in the structure definition instead of always the default one. Like that if a datatype is redefined it can be validate.

And i made little modifications in the Jax-rs module: enable the @Validate for the server, and switch to CapabilityStatement for DSTU3


hapi-fhir-structures-dstu2/src/test/java/ca/uhn/fhir/parser/CustomPatientDstu2.java
```diff
@@ -0,0 +1,29 @@
...
+import ca.uhn.fhir.model.primitive.BooleanDt;
+
+@ResourceDef(name = "Patient", profile = "Patient")
+public class CustomPatientDstu2 extends Patient {
+
+
+    private static final long serialVersionUID = 1L;
+
+    @Child(name = "homeless", order = 1)
+    @Extension(url = "%BASE_SERVER_URL%/StructureDefinition/homeless", definedLocally = true, isModifier = false)
```
jamesagnew: This is really neat! Sorry for taking a while to review it, life's been crazy lately..

Really the only question I have is whether we actually need the substitution variable in this.

Would it make sense to just let you put either an absolute or a relative URL in the extension url? In other words, set it up so you could either do:
```java
    @Extension(url = "http://example.com/StructureDefinition/homeless", definedLocally = true, isModifier = false)
```
or you could do:
```java
    @Extension(url = "StructureDefinition/homeless", definedLocally = true, isModifier = false)
```

..and the parser could just assume you want the substitution if there is no base. Would that make sense?

SRiviere: Indeed like is done for the resource profile, i should have think that before :). I'll make the modifications.

PR #547, opened 2017-01-27
Fix #546 CapturingInterceptor will now buffer response 

### LINKED ISSUE #546: CapturingInterceptor response is already consumed
When using the CapturingInterceptor with the HAPI Client, if you attempt to retrieve the response body after the client is finished, an exception will always be thrown because the body has already been consumed.

daliboz:
Added tests and updated CapturingInterceptor to buffer the response if the entity was not repeatable.

Fixes #546


jamesagnew:
This is awesome! Thanks for the PR.

I kinda can't believe that's all it took to fix that... :)

Issue #567, opened 2017-02-23
org.hl7.fhir.dstu2016may.hapi.validation.DefaultProfileValidationSupport - Unable to load profiles


lawley:
The log message is:
`org.hl7.fhir.dstu2016may.hapi.validation.DefaultProfileValidationSupport - Unable to load resource: /org/hl7/fhir/instance/model/dstu3/profile/profiles-resources.xml`

The problem seems to be the `dstu3` in the path which should instead be `dstu2016may`

jamesagnew:
PR merged in. Thanks for the contribution!

PR #568, opened 2017-02-23
Correct paths to profiles and valuesets/codesystems

lawley:
This is a patch for #567


lawley:
Note, by fixing the paths to profiles and valuesets/codesystems for dstu2.1, this enabled additional parts of the validation code to run and then uncovered bugs due to lack of (implemented) valueset expand support and problems with FHIRPath strings in some resource profiles.

I've fixed these as best I can, but am not a FHIRPath expert so my fixes may not be accurate.

Issue #551, opened 2017-02-06
Parsing of XML content with input fields > 512kb fails


jodue:
I have test xml data with a binary resource which is > 512kb. When HAPI-FHIR tries to parse this an exception is thrown because the utilized XML woodstox library has 512kb as a default max size for a single attribute.

This issue has already been fixed some time ago but now this happens again because there is a check if the WstxInputFactory is used that does not work correctly any more:

XMLUtil.java line 1618:
`if (inputFactory instanceof com.ctc.wstx.stax.WstxInputFactory)`
This does evaluate to false as inputFactory is wrapped "__redirected.__XMLInputFactory@614ab92" but the real factory object is indeed "actual: WstxInputFactory"

jamesagnew:
Hi @jodue , would you be able to share more information about your setup?

This is working fine on my end (and I've committed a unit test just to be sure). I'm wondering if you have a different version of Woodstox on your classpath or something like that?

jodue:
Hi James,

thanks for looking into this! Woodstox is added as a maven dependency by HAPI-FHIR, woodstox-core-asl-4.4.1.jar so this should be correct i guess. 

We are running HAPI_FHIR in an JBoss EAP 7 which seems to somehow wrap this factory class which is why the call to XMLInputFactory.newInstance(); returns this "__redirected.__XMLInputFactory". As far as i could find out this should affect all JBoss versions and not only EAP 7.

An easy solution to this would be to remove the instanceof check as adding the two properties to other implementations would not really have any downside. 

Since the HAPI-FHIR build currently seems to be broken i am unable to fix/test this but if you wish i could look into this when the build is running again!

Thank you!
Johannes

jamesagnew:
Hi Johannes,

Unfortunately we can't set those properties blindly.. we used to do exactly that but this caused issues on Glassfish, where a non-woodstox StAX provider is used by default (SJSXP).. on that platform having unrecognized properties caused the parser to refuse to work.

I wonder if there is any way other than the `instanceof` test that could be used to detect that the parser is woodstox. It might be worth asking on one of the JBoss list I suppose, I'm not really sure why they are using a repackaged woodstox in the first place.

One option might be something like:

```java
boolean isWoodstox = inputFactory.getProperty("org.codehaus.stax2.implVersion") != null;
```

You'd want to wrap that in a try-catch, but that could work.

Ps, build is passing again!

jodue:
Hi James,

[...]

PR #571, opened 2017-02-27
Pull request for issue #551 to fix parsing of input fields > 512kb with JBoss

jodue:
Hi James,

as discussed in #551 the parsing of XML fields > 512kb fails in JBoss because the instanceof check for Woodstox failes due to some JBoss magic. I implemented and tested your proposed fix to also check the version property, which solves this issue!


hapi-fhir-base/src/main/java/ca/uhn/fhir/util/XmlUtil.java
```diff
@@ -1615,7 +1615,20 @@ private static XMLInputFactory getOrCreateInputFactory() throws FactoryConfigura
...
-				if (inputFactory instanceof com.ctc.wstx.stax.WstxInputFactory) {
+				boolean isWoodstox = inputFactory instanceof com.ctc.wstx.stax.WstxInputFactory;
+				if ( !isWoodstox )
+				{
+					// Check if implementation is woodstox by property since instanceof check does not work if running in JBoss
+					try
+					{
+						isWoodstox = inputFactory.getProperty( "org.codehaus.stax2.implVersion" ) != null;
+					}
+					catch ( IllegalArgumentException e )
```
jamesagnew: To be extra safe, would it make sense to just catch `Exception` here?

jodue: According to the javadoc of XMLInputFactory [1] only an IllegalArgumentException should be thrown but if we want to be on the safe side we probably could just catch Exception.

[1] https://docs.oracle.com/javase/7/docs/api/javax/xml/stream/XMLInputFactory.html#getProperty(java.lang.String)

jamesagnew: Cool! Thanks for fixing that, this looks good to me. Will merge a bit later.

jodue: Perfect, thank you!

jamesagnew:
Thanks again for the PR! Merging it now.

PR #576, opened 2017-03-01
revise developer info for joelsch

joelsch:


jamesagnew:
BTW, if you want an organization or any other details listed, happy to add them :)

PR #602, opened 2017-03-20
Added Prefer Header for CORS preflight

elnin0815:

PR #603, opened 2017-03-20
Added Prefer Header for CORS Preflight

elnin0815:


jamesagnew:
Closing as duplicate of #602

elnin0815:
Please reopen. This is not a duplicate.
This is for the HAPI CLI, which currently doesn't support CORS with preflight. Which is necessary for POST requests.

jamesagnew:
Ahh oops. I stared at it to find a difference, but I guess I didn't stare long enough. :)

PR #610, opened 2017-03-28
Fix FhirContext initialisation race condition.

dangerousben:


jamesagnew:
This looks great! I'm going to merge it now.

jamesagnew:
FYI I've added a double-check in `FhirContext#validateInitialized` just to keep performance up. That method gets called multiple times during a parse cycle so it needs to be pretty fast. Let me know if you have any concerns with the implementation! The test does still pass so that seems like a good sign.

PR #613, opened 2017-04-03
Changing with stackTrace paramter without server restart

SRiviere:
As the ExceptionHandler was static, it was impossible to change the configuration,  with stacktrace or not, wihtout restart of the server.


jamesagnew:
This looks great, merging now!

PR #623, opened 2017-04-12
Accept header and MIME type fixes for DSTU2_1

lawley:
Fixes for the following problems:

1. The Capability resource incorrectly specifies the new MIME types as does the Accept header issued in client requests.
2. Further, FhirVersionEnum does not account for FhirVersionEnum#isEquivalentTo in the implementations of FhirVersionEnum#isNewerThan and FhirVersionEnum#isOlderThan


jamesagnew:
Thanks for the PR, merging now!

PR #624, opened 2017-04-13
Feature make bundle fullurl override configurable

CarthageKing:
This PR makes the overriding of resource ids when a Bundle.entry.fullUrl is specified a configurable behavior.

When this behavior is disabled via the added option in the FhirContext or the IParser, then Bundle.entry.fullUrl will not replace whatever resource id is set for resource entry in the bundle. This is useful for implementations such as ours that want to do some further checks/resolutions against the resource id and fullUrl and do some proper error reporting.


jamesagnew:
Thanks for the PR! Merging now.

PR #632, opened 2017-04-24
Use the charset from the content-type of the HttpEntity instead of always using the default charset.

gijsbert802:
We ran into a character encoding issue on a system where the default charset was not UTF-8. I believe this change fixes our issue.


jamesagnew:
Hi @gijsbert802 , thanks for the PR!

Is that CI build failure related to the change, or was that happening before you started your changes?

gijsbert802:
The CI build test failure is caused by this change, I'll push a fix.

gijsbert802:
I'm closing the PR for now, the getRequestBodyFromStream method that I "fixed" is only used for logging request bodies, so it will never have a Content-Type header (and cannot be the cause of the issue I had).
I'll do some more debugging.

jamesagnew:
To be honest this seems like a worthwhile change to me.. It may not be the source of the issue you're trying to fix, but if you added a test for null to handle the times when there was no encoding in the response I'd love to merge it. :)

gijsbert802:
Well, yes I agree, but the change was still not good :)
I tested it and it doesn't work because HttpEntity doesn't have content-type set. But there is a header in the request, I'm working on a new PR.

PR #633, opened 2017-04-25
Use the charset from the content-type header of the HttpRequest instead of always using the default charset.

gijsbert802:
Second attempt. The Content-type header from the HttpRequest is now used instead of the Content-type of HttpEntity (which apparently is always null).
All tests pass now.


jamesagnew:
Looks great! Merging now, thanks. :)

PR #636, opened 2017-04-28
Fix parsing of hexadecimal encoded entities.

gijsbert802:
I noticed this code was probably copied from another repo at http://gforge.hl7.org. Is this the correct place to fix this? Or should this (also) be fixed there?


jamesagnew:
Looks good, thanks! Merging now.

PR #649, opened 2017-05-16
Null access fix for BaseClient exception handlers.

maclema:
Fix for null access on httpRequest in DataFormatException and IOException handlers - passing UNKNOWN as the verb and URI when the httpRequest is null, this is better than a null access being thrown when calling .execute()


jamesagnew:
Thanks for the contribution!! Merging it now.

PR #656, opened 2017-05-29
Improve SearchBuilder query generation for code:modifier searches

lawley:
Uses disjunction of AND and IN for code:modifier searches.

Without this, OR conditions with too many disjuncts are generated which can lead to
failures in the underlying database code generation.


jamesagnew:
This is super cool. Merging now, thanks!

PR #660, opened 2017-06-01
Hacky but workable fix for race condition surfaced by If-None-Exists creates

karlmdavis:
Per our discussion here: https://github.com/jamesagnew/hapi-fhir/commit/3db40916010bcf0c4fd48230625c26d8f37517e5

Some notes from other attempts I made to fix this in a less hacky way:

* Tried to `@Autowire` `myResourceDaos` from a setter (rather than using the annotation on a field), and initialize `myResourceTypeToDao` in that setter, instead. Couldn't get it to work: Spring started throwing odd bean dependency errors. Don't really understand why.
* Tried to move the exceptions being thrown on `null` `getDao(Class)` results into that method, but that was breaking a test case. Didn't investigate why.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
@@ -636,19 +636,21 @@ public abstract class BaseHapiFhirDao<T extends IBaseResource> implements IDao {
 	@SuppressWarnings("unchecked")
 	public <R extends IBaseResource> IFhirResourceDao<R> getDao(Class<R> theType) {
 		if (myResourceTypeToDao == null) {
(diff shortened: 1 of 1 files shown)


karlmdavis:
Confirmed fix: with this in place, I've got an in-progress benchmark run that has made it through over 3M `Bundle`s, whereas before this change it was dying after a few thousand.

@jamesagnew: I don't think the code coverage difference here is meaningful, but it's your project, not mine. In general, a regression test case for this issue would need to push... 30K or so resources into a FHIR server asynchronously, using... at least dozens of threads. (Like I said, our benchmarks were dying after a few thousand requests, but they were using 390 threads on giant EC2 instances. We just weren't seeing this in our local dev builds/runs.) Thoughts?

jamesagnew:
This doesn't even seem hacky to me. :) 

Thanks for the investigation and fix! Merging it now.

karlmdavis:
It just feels wrong to let it enter that block and redo that work if initialized by multiple threads at once. But I couldn't get Spring to play nice, so oh well.

PR #662, opened 2017-06-02
Unit Test for search with valid chain on a reference param

nagrajdivvala:
Fix needed to allow chained parameters on a reference param.

Unit test added as per discussion on below link
https://groups.google.com/forum/#!topic/hapi-fhir/sdePyR_p87c


jamesagnew:
Hi @nagrajdivvala,

In order to use a chained method parameter, you need to use a reference parameter.

So, instead of 

```java
@OptionalParam(name = Patient.SP_GENERAL_PRACTITIONER) TokenAndListParam theGPIdentifiers) {
```

It should be:
```java
@OptionalParam(name = Patient.SP_GENERAL_PRACTITIONER) ReferenceAndListParam theGPIdentifiers) {
```

PR #686, opened 2017-07-10
Fix STU3 validation

SRiviere:
find correct extension profile, use custom datatype profile, allow to disable terminology checks

### DIFF
--- a/hapi-fhir-structures-dstu3/src/main/java/org/hl7/fhir/dstu3/elementmodel/Property.java
+++ b/hapi-fhir-structures-dstu3/src/main/java/org/hl7/fhir/dstu3/elementmodel/Property.java
@@ -2,6 +2,7 @@ package org.hl7.fhir.dstu3.elementmodel;
 
 import java.util.ArrayList;
 import java.util.List;
+import org.apache.commons.lang3.StringUtils;
 
 import org.hl7.fhir.dstu3.conformance.ProfileUtilities;
 import org.hl7.fhir.dstu3.context.IWorkerContext;
@@ -203,7 +204,7 @@ public class Property {
     ElementDefinition ed = definition;
     StructureDefinition sd = structure;
     List<ElementDefinition> children = ProfileUtilities.getChildMap(sd, ed);
-    if (children.isEmpty()) {
+    if (children.isEmpty()  || isElementWithOnlyExtension(ed, children)) {
       // ok, find the right definitions
       String t = null;
       if (ed.getType().size() == 1)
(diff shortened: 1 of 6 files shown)


hapi-fhir-structures-dstu3/src/main/java/org/hl7/fhir/dstu3/elementmodel/Property.java
```diff
@@ -203,7 +204,7 @@ public boolean isChoice() {
     ElementDefinition ed = definition;
     StructureDefinition sd = structure;
     List<ElementDefinition> children = ProfileUtilities.getChildMap(sd, ed);
-    if (children.isEmpty()) {
+    if (children.isEmpty()  || isElementWithOnlyExtension(ed, children)) {
```
jamesagnew: I'm not totally clear on what this is for..

Would you be able to include a unit test that demonstrates what this part is fixing?

SRiviere:
Sure sorry i forget it. In summary is in a resource when you specify a list of possible extension you may have a element like that:

        <element>
            <path value="Patient.extension"/>
            <slicing>
                <discriminator>
                    <type value="value"/>
                    <path value="url"/>
                </discriminator>
                <rules value="open"/>
            </slicing>
            <definition value="May be used to represent additional information that is not part of the basic definition of the element."/>
        </element>

And this element was generating "types == 0, and no children found" when it try to validate an extension

SRiviere:
Sure i'll add the tests. In summary this is to manage a element like that:
<element>
   <path value="Patient.extension"/>
   <slicing>
     <discriminator>
       <type value="value"/>
       <path value="url"/>
     </discriminator>
     <rules value="open"/>
   </slicing>
   <definition value="May be used to represent additional information that is not part of the basic definition of the element."/>
</element>

And avoid a this error: "types == 0, and no children found", because ther is no type define when the extension is defined in is own element so the search should continue

hapi-fhir-structures-dstu3/src/main/java/org/hl7/fhir/dstu3/elementmodel/Property.java
SRiviere: added

jamesagnew:
Thanks @SRiviere , I've merged this!

jamesagnew:
By the way, I'm crediting this to @SRiviere in the changelog. Please let me know if you would like me to use another name.

jamesagnew:
(And I'm also forward-porting this fix to the R4 validator)

PR #688, opened 2017-07-10
Correctly serialize baseurl/FooResource references

### LINKED ISSUE #659: Reference url using only 'baseurl/FooResource' is incorrectly serialized to either XML or JSON
Hi,

Given the following code below:
```java
		String refVal = "http://localhost:8090/FooBar";
		FhirContext fctx = FhirContext.forDstu3();
		// test xml parser
		{
			Patient fhirPat = new Patient();
			fhirPat.addExtension().setUrl("x1").setValue(new Reference(refVal));
			String output = fctx.newXmlParser().encodeResourceToString(fhirPat);
			System.out.println("output: " + output);
		}

		// test json parser
		{
			Patient fhirPat = new Patient();
			fhirPat.addExtension().setUrl("x1").setValue(new Reference(refVal));
			String output = fctx.newJsonParser().encodeResourceToString(fhirPat);
			System.out.println("output: " + output);
		}

		// test xml parser
		{
			Patient fhirPat = new Patient();
			fhirPat.setManagingOrganization(new Reference(refVal));
			String output = fctx.newXmlParser().encodeResourceToString(fhirPat);
			System.out.println("output: " + output);
		}

		// test json parser
		{
			Patient fhirPat = new Patient();
			fhirPat.setManagingOrganization(new Reference(refVal));
			String output = fctx.newJsonParser().encodeResourceToString(fhirPat);
			System.out.println("output: " + output);
		}
```

With HAPI FHIR 1.4, it will output the following:
```sh
output: <Patient xmlns="http://hl7.org/fhir"><extension url="x1"><valueReference><reference value="http:/localhost:8090/FooBar"/></valueReference></extension></Patient>
output: {"resourceType":"Patient","extension":[{"url":"x1","valueReference":{"reference":"http://localhost:8090/FooBar"}}]}
output: <Patient xmlns="http://hl7.org/fhir"><managingOrganization><reference value="http:/localhost:8090/FooBar"/></managingOrganization></Patient>
output: {"resourceType":"Patient","managingOrganization":{"reference":"http:/localhost:8090/FooBar"}}
```
As noted above, only the second output from the JSON parser seems to be correct in that the reference URL is correctly serialized.

However, starting with HAPI FHIR 1.5 until the latest snapshot (i.e. 2.5-SNAPSHOT as of this time), the output is consistent:
```sh
output: <Patient xmlns="http://hl7.org/fhir"><extension url="x1"><valueReference><reference value="http:/localhost:8090/FooBar"/></valueReference></extension></Patient>
output: {"resourceType":"Patient","extension":[{"url":"x1","valueReference":{"reference":"http:/localhost:8090/FooBar"}}]}
output: <Patient xmlns="http://hl7.org/fhir"><managingOrganization><reference value="http:/localhost:8090/FooBar"/></managingOrganization></Patient>
output: {"resourceType":"Patient","managingOrganization":{"reference":"http:/localhost:8090/FooBar"}}
```
I don't think the serialized URL value is correct. Specifically, there should have been one more '/' after 'http:/'.

Is this valid behavior?

Note that if you use 'baseurl/FooResource/something', then the reference URL is correctly serialized from HAPI FHIR 1.4 until the latest snapshot.

CarthageKing:
This PR fixes #659.

The cause was found to be in the parsing of reference values as IdDt or IdType objects. Given an example reference value of 'http://my.org/Foo', 'my.org' was treated as the resource type while 'Foo' was treated as the resource id, which I believe is wrong.

setValue() was hence modified such that when it encounters an url of the aforementioned form, it'll treat 'Foo' as the resource type rather than the id. It also sets a number of internal variables such that a call to getValue() will give the correct output.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/IdDt.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/IdDt.java
@@ -305,11 +305,16 @@ public class IdDt extends UriDt implements /*IPrimitiveDatatype<String>, */IIdTy
 				b.append(myResourceType);
 			}
 
-			if (b.length() > 0) {
+			if (b.length() > 0 && isNotBlank(myUnqualifiedId)) {
 				b.append('/');
 			}
 
-			b.append(myUnqualifiedId);
+			if (isNotBlank(myUnqualifiedId)) {
+				b.append(myUnqualifiedId);
+			} else if (isNotBlank(myUnqualifiedVersionId)) {
+				b.append('/');
+			}
+
 			if (isNotBlank(myUnqualifiedVersionId)) {
 				b.append('/');
 				b.append(Constants.PARAM_HISTORY);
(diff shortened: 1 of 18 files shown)


jamesagnew:
Thanks for the PR! Merging now

PR #689, opened 2017-07-12
CLI: command-line option to configure reuse of search results

eug48:
It'd be nice to be able to turn off reuse of search results without having to rebuild the CLI. I've had a go at some possible wording.


jamesagnew:
Super cool, thanks for the PR!

PR #701, opened 2017-08-01
Add qualifiers for token search, fix javascript bug for selecting text qualifier

dconlan:
Wanted to add :above :below qualifiers to code search criteria.

In the course of making the change also noticed that changing the text qualifier to exact would not allow changing back to matches, so I fixed that issue.


jamesagnew:
this is awesome, thanks!

I'll merge shortly. Please let me know if you'd like me to credit another name than `@dconlan`

PR #706, opened 2017-08-06
Fixes #705 - Client correctly validates the server's conformance statement FHIR version

### LINKED ISSUE #705: Client's FHIR version check of server's conformance statement doesn't work as expected
In #88 there was a change to have clients check the server conformance statement the first time you connect to a server to see whether you have the right version of the client for the version of the server. When setting up a simple project of my own (https://github.com/InfiniteLoop90/fhir-client-example)  to play with the HAPI FHIR client code I noticed this statement logged at at the DEBUG level:
```
DEBUG ca.uhn.fhir.rest.client.RestfulClientFactory - Server conformance statement indicates unknown FHIR version: 1.0.2
```

This wasn't expected since 1.02 is the FHIR version for DSTU2.
In looking at the `RestfulClientFactory` class, the code hasn't been updated in a while and is checking for incorrect FHIR version strings when validating the server's conformance statement.

Update the code in `RestfulClientFactory` to check against correct FHIR version strings.

InfiniteLoop90:
Fixed #705 by changing the FHIR version string comparisons that the client side code does when validating the server's conformance statement.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirVersionEnum.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirVersionEnum.java
@@ -42,7 +42,7 @@ public enum FhirVersionEnum {
 
 	DSTU2_1("org.hl7.fhir.dstu2016may.hapi.ctx.FhirDstu2_1", null, true, new Version("1.4.0")),
 
-	DSTU3("org.hl7.fhir.dstu3.hapi.ctx.FhirDstu3", null, true, new Dstu3Version());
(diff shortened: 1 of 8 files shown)


jamesagnew:
This looks great, thanks for the PR!

I'm going to merge, although I'm reverting one small bit: the removal of `Dstu3Version.java`. The reason we use this instead of hardcoding the string is that when we update the model classes (which may happen again if there are any further patch releases to FHIR DSTU3 itself) it's good to pull the version from the structures themselves to ensure we're outputting the correct one.

PR #708, opened 2017-08-09
🔗 Fix broken usage patterns links.

psbrandt:
The links to the usage patterns pages on the homepage are broken. I guess because the SVG wasn't updated when the switch was made to the [`hapifhir.io`](hapifhir.io) domain name.


jamesagnew:
Oh nice, thanks for catching that! Merging now.

Issue #675, opened 2017-06-16
rest-hook subscriptions are not becoming active


javajeff:
I have created a rest-hook subscription using HAPI-FHIR 2.6 SNAPSHOT, but the subscription is not activating.  It is staying in the requested state.

javajeff:
Bug found.  In the RestHookSubscriptionDstu3Interceptor.resourceCreated() the if statement should check for Subscription.Status.REQUESTED instead of ACTIVE.   When creating a subscription, it can't have the active status, and must first be requested.  Then we can change the status to ACTIVE.

mscaldas2012:
I'm seeing the same issue with subscription of type message. status is stuck on "requested"

sidharthramesh:
I am facing the same issue. I'm using HAPI FHIR v5.4.0 and I also tried the latest docker image. The subscriptions don't change the status to active, and the HTTP requests don't go through when a resource is created or edited.

PR #710, opened 2017-08-10
Fix for rest-hook subscriptions are not becoming active #675

javajeff:
Currently you cannot create rest hook subscriptions, because they must start off with a Requested state.  However the current code looks for new rest hook subscriptions in the Active state.  Changed the if statement to look for the Requested state.  Then changed the subscription state to Active.


jamesagnew:
Looks good! Merging now.

javajeff:
Thanks!

PR #711, opened 2017-08-13
Added R4 to the client's server version check. Also tweaked the ClientServerValidation* tests to be a little more predictable

InfiniteLoop90:
Added R4 to the client's server version check. Also tweaked the ClientServerValidation* tests to be a little more predictable.


InfiniteLoop90:
This is related to issue #706. Now that the R4 structures are merged into `master`, add R4 to the client's server FHIR version check.

PR #712, opened 2017-08-18
Add imports for org.slf4j.Logger and org.slf4j.LoggerFactory

rhausam:
Add imports for org.slf4j.Logger and org.slf4j.LoggerFactory to resolve compile errors.


jamesagnew:
Thanks Rob! Merging now.

PR #717, opened 2017-08-24
fix bugs with inverted test for If-Modified-Since (impl and tests)

lawley:
The implementation of support for If-Modified-Since and the corresponding unit tests are broken.

The implementation has the timestamp comparison test backwards and the unit tests have the wrong expectations.

Implementation updated to use correct test, and unit tests have been updated to 1. fix the expectations, and 2. also cover the case where the header timestamp is identical to the modified timestamp.


jamesagnew:
Man, that's quite the issue. Thanks for the contribution, merging now!

PR #718, opened 2017-08-24
fixed CORS preflight

patrick-werner:
CORS preflight was not working because of missing Allowed-Header "Prefer"


jamesagnew:
Good catch, looks good to me. Merging now.

Issue #734, opened 2017-09-22
Binary.securityContext is not being encoded


malcolmm83:
I've got a breakpoint in BaseHapiFhirDao.populateResourceIntoEntity, the resource is a Binary and it has securityContext set, but  the encoded version created by the parser is missing the securityContext field.  There is code to parse this field, so I believe this is a bug.

PR #736, opened 2017-09-25
Fix for Binary.securityContext is not being encoded #734

malcolmm83:
And unit test.


malcolmm83:
The patch to JasonParser is in hapi-fhir-base which doesn't have access to the DSTU3 structures where Binary lives (getSecurityContext isn't defined on BaseBinary or IBaseBinary).  My solution to this was to use reflection.  I don't like this because it relies on exception handing if it is called with a resource that has no getSecurtityContext.  This seems non-ideal.  It would be better to either add this to the IBaseBinary interface, or to refactor the parser so it is easy to override this one behavior.  The encodeResourceToJsonStreamWriter is quite long and I'd hate to duplicate most of that code in order to override the behavior for Binary.  That was more than I wanted to take on, so I used reflection to dodge the issue.

jamesagnew:
This is cool! Thanks for the PR, merging now.

PR #743, opened 2017-09-30
Add Spring Boot Starter

mouellet:
Provides spring-boot-starter with auto-configurations for Rest server,
Jpa server, FHIR Validation and clients.

Default configurations overridable by configuration properties and 
rest-server customizer.


jamesagnew:
Hi Mathieu, this is cool! Thanks for the contribution!

Merging it now. Note that I'm going to move it into `example-projects/` within the root HAPI project just to keep things organized, but that's the only change. Hope that's ok.

jamesagnew:
Hi @mouellet, FYI I ended up not moving anything into the examples directory, on further reflection it is very nicely organized already :)

I have one question for you, as I don't know Spring Boot very well. As a result of merging #747 in, I had to add [this line](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-spring-boot/hapi-fhir-spring-boot-autoconfigure/src/test/java/ca/uhn/fhir/spring/boot/autoconfigure/FhirAutoConfigurationTest.java#L80) to the JPA test to get it to work.

That property is going to be required by anyone who uses this and they aren't likely to want to change it. Is there any way of making this the default?

PR #747, opened 2017-10-05
Analyzer def using search mapping

JiajingLiang:
This is a separation of analyser definitions from entity class in order to provide more flexibility on search engine configuration. For example, replacing embedded Apache Lucene by ElasticSearch


jamesagnew:
Wow, will review in depth a bit later but that is super cool!

PR #767, opened 2017-10-22
Fixes #764 to allow the AbstractJaxRsConformanceProvider to support the DSTU2_HL7ORG, DSTU2_1, and R4 FhirContexts

### LINKED ISSUE #764: AbstractJaxRsConformanceProvider doesn't work with DSTU2_HL7ORG, DSTU2_1, or R4 FhirContexts
I tried to create a JAX-RS server using HAPI FHIR quite similar to the example in HAPI FHIR here https://github.com/jamesagnew/hapi-fhir/tree/master/hapi-fhir-jaxrsserver-example but I am having trouble getting the conformance provider setup. In looking at the `AbstractJaxRsConformanceProvider` class in the current development version of HAPI FHIR (3.1.0-SNAPSHOT), it looks like this class only supports the `DSTU2` and `DSTU3` FhirContexts. It doesn't support `DSTU2_HL7ORG` (the one I'm trying to use) nor does it support `DSTU2_1` or `R4`.

Could this be updated to support the `DSTU2_HL7ORG`, `DSTU2_1`, and `R4` contexts?

InfiniteLoop90:
Adds support for the DSTU2_HL7ORG, DSTU2_1, and R4 FhirContexts in `AbstractJaxRsConformanceProvider`

### DIFF
--- a/hapi-fhir-jaxrsserver-base/pom.xml
+++ b/hapi-fhir-jaxrsserver-base/pom.xml
@@ -66,6 +66,11 @@
 			<artifactId>hapi-fhir-structures-dstu3</artifactId>
 			<version>${project.version}</version>
 		</dependency>
+		<dependency>
+			<groupId>ca.uhn.hapi.fhir</groupId>
+			<artifactId>hapi-fhir-structures-r4</artifactId>
+			<version>${project.version}</version>
+		</dependency>
 
 		<dependency>
 			<groupId>javax.ws.rs</groupId>
--- a/hapi-fhir-jaxrsserver-base/pom.xml
+++ b/hapi-fhir-jaxrsserver-base/pom.xml
@@ -53,6 +53,11 @@
 		</dependency>
 		<dependency>
 			<groupId>ca.uhn.hapi.fhir</groupId>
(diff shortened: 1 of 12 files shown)


InfiniteLoop90:
It would be kind of nice if there could be a Conformance Provider class in each of the various resource structure modules. That way, the `hapi-fhir-jaxrsserver-base` module could have the dependencies on the resource structure modules marked as optional so that a project that pulls in the hapi-fhir-jaxrsserver-base artifact wouldn’t be forced to pull in _all_ of the resource structures transitivily. But I think unfortunately breaking changes would be required to make that happen.

jamesagnew:
This looks great- merging now.

I'm not sure I understand your comment though:

> It would be kind of nice if there could be a Conformance Provider class in each of the various resource structure modules. 

There is a conformance provider in each structure module. E.g. [this one](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-structures-hl7org-dstu2/src/main/java/org/hl7/fhir/instance/conf/ServerConformanceProvider.java) for hl7org-dstu2. What am I missing?

Issue #768, opened 2017-10-22
Unable to exclude stacktrace from DataFormExceptions in OperationOutcomes when using JAX-RS providers


InfiniteLoop90:
In issue #111, stacktraces were supposed to not be included by default. In doing some digging, it looks like `AbstractJaxRsProvider#createOutcome(DataFormatException)` always includes the stacktrace in the details of the OperationOutcome. So if an invalid date format is sent for a `DateParam`, the whole stracktrace will be included in the response. It seems like it'd be idea if `DataFormatException`s would be handled like other exceptions by having the stacktrace be included if the Exception type is included as such in `ExceptionHandlingInterceptor`.
(Note, this issue still exists in the latest 3.1.0-SNAPSHOT build).

See screenshot from Postman for an example:
<img width="1074" alt="screen shot 2017-10-22 at 1 18 28 pm" src="https://user-images.githubusercontent.com/20246655/31864974-9d9329b0-b72c-11e7-9821-09ac6e3dab42.png">

InfiniteLoop90:
Closing since this was fixed in HAPI FHIR 3.1.0.

PR #769, opened 2017-10-22
Fixes issue #768 to conditionally include stacktraces for DataFormatExceptions in OperationOutcomes when using AbstractJaxRsProviders

InfiniteLoop90:
When `DataFormatException`s are thrown when using `AbstractJaxRsProvider`s, the stacktrace will be included in the OperationOutcome details only if `AbstractJaxRsProvider#withStrackTrace()` is configured to return `true`.

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptor.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptor.java
@@ -32,6 +32,8 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import ca.uhn.fhir.parser.DataFormatException;
+import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.hl7.fhir.instance.model.api.IBaseOperationOutcome;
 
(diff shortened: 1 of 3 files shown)


InfiniteLoop90:
@jamesagnew In my second commit I took a different approach to solving the problem. Let me know if you have any concerns with it or any preferred recommendations on a better way to handle the problem. Thanks!

InfiniteLoop90:
@jamesagnew This is ready for review. Thanks!

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptor.java
```diff
@@ -99,7 +101,10 @@ public Object handleException(RequestDetails theRequestDetails, BaseServerRespon
 	@Override
 	public BaseServerResponseException preProcessOutgoingException(RequestDetails theRequestDetails, Throwable theException, HttpServletRequest theServletRequest) throws ServletException {
 		BaseServerResponseException retVal;
-		if (!(theException instanceof BaseServerResponseException)) {
+		if (theException instanceof DataFormatException) {
+			// Wrapping the DataFormatException as an InvalidRequestException so that it gets sent back to the client as a 400 response.
+			retVal = new InvalidRequestException(theException);
```
jamesagnew: I'm trying to decide if this makes sense.. It seems a bit odd to me, but then again I can't think of any valid reasons for a server to be throwing DataFormatException that wouldn't be caused by client error. And I especially don't see any reason that this would be worse than the current 500 error. So I guess it is better. :)

jamesagnew:
Looks great, merging now!

PR #774, opened 2017-10-30
Removed "hapi-fhir" from 4x Learn More links

jamesdaily:
The "Learn More" links within hapi_usage_patterns.svg are dead. I've removed "/hapi-fhir" from each path to correct this.


jamesagnew:
Thanks for catching and fixing this!! Merging now.

PR #783, opened 2017-11-12
Added cqf-ruler project 

c-schuler:
I have packaged the [cqf-ruler](https://github.com/DBCG/cqf-ruler) project as a HAPI example module. This should provide a good example of how to provide decision support using HAPI's JPA server.


brynrhodes:
Recommend packaging the code that would be common in every implementation and then referencing it as dependencies.

PR #786, opened 2017-11-15
ensure to use resourceType properly when it is explicitly provided in a chained reference param

CarthageKing:
This PR ensures that when the resourceType is explicitly defined in the reference search parameter, that it gets used properly and also that the value also gets parsed properly.

Before this fix, specifying a search like Observation?subject:Patient.identifier=http://a.b/c/d|123, the 'Patient' resourceType would be overwritten with a path from the provided url value, and the value would be incorrect. This is due to the invocation of setValue(), which internally calls myId.setValue().


jamesagnew:
Looks great, thanks for the PR!

Issue #799, opened 2017-11-28
Unable to use @Read annotation with the HAPI FHIR Jax-RS server base with R4 and DSTU2_1 Fhir contexts


InfiniteLoop90:
I'm unable to use `@Read` annotation with the HAPI FHIR Jax-RS server base with R4 and DSTU2_1 Fhir contexts.
I get the following stacktrace:
[jaxrs-read-r4-error.txt](https://github.com/jamesagnew/hapi-fhir/files/1511282/jaxrs-read-r4-error.txt)

It looks like the `JaxRsRequest` class wasn't updated to support the R4 and DSTU2_1 contexts.

jamesagnew:
Fix merged.

PR #800, opened 2017-11-28
Fixes issue #799 by adding logic to handle the R4 and DSTU2_1 contexts in ca.uhn.fhir.jaxrs.server.util.JaxRsRequest.

InfiniteLoop90:
Fixes issue #799 by adding logic to handle the R4 and DSTU2_1 contexts in ca.uhn.fhir.jaxrs.server.util.JaxRsRequest.

Also changed the locations in the hapi-fhir-jaxrsserver-base module where it was doing if-else branches based on the FhirVersionEnum to now use switch statements instead.


jamesagnew:
Looks great, merging now!

Issue #778, opened 2017-11-03
AuthorizationInterceptor not working for Collection of IIdType (theOwners)


santoshkumar2016:
I am implementing AuthorizationInterceptor for my fhir server.
The interceptor works fine for a single owner:
IdDt userIdPatientId = new IdDt("Patient", 2L);
new RuleBuilder().allow().read().allResources().inCompartment("Patient", userIdPatientId )
 However when I am setting the owners as a collection of ids its failing with 
ERROR ca.uhn.fhir.rest.server.interceptor.ExceptionHandlingInterceptor - ExceptionHandlingInterceptor.createOperationOutcome Failure during REST processing: java.lang.NullPointerException java.lang.NullPointerException
	at ca.uhn.fhir.rest.server.interceptor.auth.RuleImplOp.applyRule(RuleImplOp.java:252):

List<IdDt> idtList = new ArrayList<>();
................
new RuleBuilder().allow().read().allResources().inCompartment("Patient", idtList)

While checking code for ca.uhn.fhir.rest.server.interceptor.auth.RuleBuilder class I found that it is not setting myInCompartmentName and myInCompartmentOwners in method inCompartment(String theCompartmentName, Collection<? extends IIdType> theOwners) :
@Override
public IAuthRuleBuilderRuleOpClassifierFinished inCompartment(String theCompartmentName, Collection<? extends IIdType> theOwners) {
	Validate.notBlank(theCompartmentName, "theCompartmentName must not be null");
	Validate.notNull(theOwners, "theOwners must not be null");
	Validate.noNullElements(theOwners, "theOwners must not contain any null elements");
	for (IIdType next : theOwners) {
				validateOwner(next);
			}
	myClassifierType = ClassifierTypeEnum.IN_COMPARTMENT;
				return finished();
			}

@Override
public IAuthRuleBuilderRuleOpClassifierFinished inCompartment(String theCompartmentName, IIdType theOwner) {
Validate.notBlank(theCompartmentName, "theCompartmentName must not be null");
Validate.notNull(theOwner, "theOwner must not be null");
validateOwner(theOwner);
**_myInCompartmentName = theCompartmentName;
myInCompartmentOwners = Collections.singletonList(theOwner);_**
myClassifierType = ClassifierTypeEnum.IN_COMPARTMENT;
		return finished();
				}

jamesagnew:
Fix has been merged

PR #806, opened 2017-12-14
fix inCompartment for a collection of owners

JiajingLiang:
concerning issue: #778


JiajingLiang:
Added some test to ensure that inCompartment with a collection of owners works. 
But I don't understand why we have exactly the some tests for Dstu3 and R4. I mean why separate the tests according to fhir context version? @jamesagnew

jamesagnew:
Looks great, merging now!

Issue #809, opened 2017-12-22
Resources sometimes not searchable by date after update


darktyco:
Resources are searchable by date after they are initially created. After the resource is updated the date search no longer works. Every update to the resource causes the behavior to alternate back and forth.

darktyco:
I have a fix locally, will open a PR in a moment.

jamesagnew:
Fix merged.

PR #810, opened 2017-12-22
Fixes issue #809 where date search params get deleted upon resource update

darktyco:
The equals function in ResourceIndexedSearchParamDate doesn't work as expected when the Dates in a persistent entity are java.sql.Timestamp objects but the dates in a transient entity are java.util.Date objects. Given a pair of equivalent Date and Timestamp objects, Date.equals(Timestamp) will be true but Timestamp.equals(Date) will not be true.

Updating the equals and hashcode method to compare the underlying time value instead. Added tests.


jamesagnew:
Great catch, thanks! Merging now

PR #812, opened 2017-12-23
Added ability to create a resthook subscription without returning a p…

javajeff:
…ayload using an HTTP POST for notification

I removed a validation for a blank payload, removed setting a default payload, and added the ability to send a HTTP POST notification to the subscriber's server if there is no payload.


jamesagnew:
Looks good, merging now!

PR #817, opened 2018-01-05
Adding new module - hapi-fhir-jpaserver-dynamic

anoush-bch:
This project has been built with hapi-fhir-jpaserver-example as a base. It has been made more dynamic by replacing web.xml with `ca.uhn.fhir.jpa.demo.WebInitializer` class which extends Spring `org.springframework.web.WebApplicationInitializer` class and loads application contexts in a dynamic manner, so that based on environment and/or property variables it can be started either as dstu2 or dstu3 version of HAPI-FHIR JPA Server. Some of the classes have been also refactored to make them more generic.


jamesagnew:
Hi Anoush. Thanks for the contribution! FYI I'm going to merge this, but also move it to the examples folder just to try and keep the number of root modules manageable.

anoush-bch:
Thank you!

PR #836, opened 2018-01-24
Fix problem with _include=MedicationRequest:medication

RuthAlk:
A query for MedicationRequest that should include medication
did not work, because the SearchParamDefinition defined the
path as MedicationRequest.medication.as(Reference)

Enhanced the matchesInclude method to allow extra parts after
the search path


jamesagnew:
Thanks for the PR! It has been merged.

Please let me know if you would like your real name added to the credits

PR #852, opened 2018-02-19
Sources and resource generator mojo which doesn't use value sets.

splatch:
I've attempted to reorganize a little bit the way how configuration is being processed because there was a lot of duplicated code. It is still far from being perfect but it works.


jamesagnew:
Hi @splatch - What is the purpose of this? I'm not really clear on why one would need a version of the structure generator that doesn't use valuesets?

splatch:
Hey James,
This particular generator works without issues with STU3 resources. Due to changes in value set handling tinder doesn't work with STU3 and is compatible only with DSTU2.

jamesagnew:
Hi @splatch - I'm merging this as it seems non-intrusive and I assume you have a use for it. I would recommend if you can that you submit a further PR with a unit test though to avoid this being broken in the future.

splatch:
@jamesagnew thank you very much for including this one. We stayed with @hzbarcea on 3.2.x since this PR was made. We needed it for custom code generators based only on resource definitions. Despite of support for custom code generation available in FHIR spec we could get it over.

hzbarcea:
Thanks @jamesagnew and @splatch !

PR #857, opened 2018-02-21
Support of method chaining for DateRangeParameter

Tastelezz:
- according to the design of most other `IQueryParameterXXX` implementations I changed the methods `void setUpperBound()` and `DateRangeParam setLowerBound()` to `DateRangeParam setUpperBound()` and `void setLowerBound()` supporting method chaining.
- changed the order in which the state of a `DateRangeParam` instance got first modified and then verified.
- renamed the private method `validateAndThrowDataFormatExceptionIfInvalid()` to `validateAndSet(DateParam, DateParam)`.
- replaced the methods `boolean haveLowerBound()` and `boolean haveUpperBound()` with the stateless method `boolean hasBound(DateParam)`.


jamesagnew:
Looks great! Thanks! Merging now.

Issue #11, opened 2014-08-14
FhirContext  xml pretty print issue


t106uhn:
This is low priority.  When printing the contents of a resource using "FhirContext#newXmlParser().setPrettyPrint(true).encodeResourceToString()" several duplicate resources appear in the "contained" element when they should not be there.

patrick-werner:
couldn't reproduce in Dstu3 & R4, added Tests. 
Probably can be closed

PR #858, opened 2018-02-21
added Test for #11 to Dstu3 and R4, fixed #853

### LINKED ISSUE #853: Resource validation throws an Error when an Observation has multiple performers
Observations go through this validation
[https://github.com/jamesagnew/hapi-fhir/blob/89a834723c87939843aa72c3140e356ba933f82b/hapi-fhir-validation/src/main/java/org/hl7/fhir/r4/validation/InstanceValidator.java#L3541](url)

which ends up throwing an Error when there are more than one Performer for an Observation. However, Performer is a 0...* element in FHIR.

[https://www.hl7.org/fhir/observation.html](url)

patrick-werner:
removed TestTest in Dstu3


patrick-werner:
sorry for mixing two issues, will do feature branches in the future

jamesagnew:
Cool!

Issue #826, opened 2018-01-14
Modular OSGi Support


fmayoral:
Hello!
We are currently moving towards supporting STU3 on an OSGi environment and after the changes on package names in HAPI we think this is feasible without any need to repackage.
I have noticed hapi-fhir-osgi-core puts everything together but it's not the best approach given that we could have it modular. It is a bit hard to tell what should be put in each manifest but we'll give it a go; However I must ask first, is there any known reason why this wouldn't work or create some issues?
Thanks!

johnpoth:
Hi @fmayoral it seems from #784 that hapi-fhir-osgi-core was created because there were split packages which don't seem to exist anymore.
Are you still having a go at implementing this? I'd be happy to help out.

fmayoral:
I'm at it right now, just for a few bundles I need to have (base, utilities, server, client and dstu3); 
I've mostly used the maven bundle plugin to get it going, with some tweaking here and there; Ideally, I'd go module by module doing this correctly but the work I'm doing with this wouldn't allow it because of time/priorities.

Due to the reflective behaviour of some flows there seem to be some "soft dependency" (not really a dependency but not sure how to call it) from some modules onto others, this is fine when no strict modularity is applied, but the moment I create OSGi manifests I start getting a lot of issues with class resolution as the bundles try to access elements that are not present in the manifest, hence unavailable to them.
For this to work I need to enable the dynamic import package header which defeats the modular nature of OSGi, the alternative is to start working on a different architecture that doesn't rely on reflection (or if it does in a way that doesn't create circular "soft dependencies") to handle these scenarios in an appropriate way;

The ones I've identified so far are:
hapi-fhir-base needs to have access to:
```
// all the contexts
ca.uhn.fhir.model.dstu2;resolution:=optional,
org.hl7.fhir.instance;resolution:=optional,
org.hl7.fhir.dstu2016may.hapi.ctx;resolution:=optional,
org.hl7.fhir.dstu3.hapi.ctx;resolution:=optional,
org.hl7.fhir.r4.hapi.ctx;resolution:=optional,
// required when building servers
ca.uhn.fhir.rest.api.server;resolution:=optional,
ca.uhn.fhir.model.api;resolution:=optional
```
additionally I've enabled dynamic package import package as follows, so there might be more scenarios uncovered by previous imports:
```
<DynamicImport-Package>
ca.uhn.fhir.*;version="3.2.0", 
org.fhir.*;version="3.2.0", 
[...]

PR #864, opened 2018-02-26
Avoid unecessary reference to javassist.

splatch:
By accident I've found that hapi-fhir-base pulls in javassist without need. This might help (a little) with #826. :-)


jamesagnew:
Looks good, merging now! Thanks!

PR #868, opened 2018-02-27
equals() and hashCode() for DateRangeParam and DateParam

Tastelezz:
- added guava-testlib as test dependency to modules hapi-fhir-structures-dstu2 and hapi-fhir-structures-dstu3
- added DateRangeParam.equals(Object) and DateRangeParam.hashCode()
- added DateParam.equals(Object) and DateParam.hashCode()

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/param/BaseParamWithPrefix.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/param/BaseParamWithPrefix.java
@@ -107,5 +107,4 @@ public abstract class BaseParamWithPrefix<T extends BaseParam> extends BaseParam
 		myPrefix = thePrefix;
 		return (T) this;
 	}
-
 }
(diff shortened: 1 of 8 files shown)


Tastelezz:
I added some test samples and wonder why on my machine the prefix `APPROXIMATE` matches but not on the CI? The error message is:

`[ERROR] Failures: 
[ERROR]   DateParamTest.testEqualsAndHashCode:129 DateParam[prefix=APPROXIMATE,value=2011-11-11T11:11:11.111-11:11] [group 5, item 1] must be Object#equals to DateParam[prefix=APPROXIMATE,value=2011-11-11T12:12:12.123-12:12] [group 5, item 2]`

Any ideas? The second test sample is of now real importance, but I am curious why this happens, because I can't see how the prefix and the dates provided could depend on the execution environment.

jamesagnew:
Might have been a fluke or something.. Looks like it's passing now.

jamesagnew:
Thanks for the contribution! It has been merged.

PR #869, opened 2018-02-27
Fix xpath error in questionnaire schematron file

hdconradi:
A typo in the questionnaire schematron file for R4  caused an xpath error:

Bad assert: XPath syntax error. Unclosed parenthesis. Suggestion: add ')'.
      not(f:option) or not(count(f:*[starts-with(local-name(.), 'initial')]|)

Changed: )]|) -> )]))


hdconradi:
This fix raises the question of what should happen in a case when the schematron file is corrupt.
Currently, the exception is swallowed and the any input is returned with 'No issues'.
Is that what is really desired here, or is this a case where a ServerException is appropriate?

hdconradi:
has been obsoleted by commit: 07290ae8a0

PR #870, opened 2018-02-27
Fix questionnaire schematron for que-7

hdconradi:
Without this fix, the 'hasAnswer' element of a questionnaire item is ignored.
This in turn permits that one can have both an 'answer' and a 'hasAnswer' item.

A fix is applied for both DSTU3 and R4 and tests are provided.

The tests for R4 are currently ignored, since they require a valid schematron,
e.g. by merging #869

### DIFF
--- a/hapi-fhir-validation-resources-dstu3/src/main/resources/org/hl7/fhir/dstu3/model/schema/fhir-invariants.sch
+++ b/hapi-fhir-validation-resources-dstu3/src/main/resources/org/hl7/fhir/dstu3/model/schema/fhir-invariants.sch
@@ -4503,7 +4503,7 @@
       <sch:assert test="not((f:type/@value='group' and not(f:item)) or (f:type/@value='display' and f:item))">que-1: Group items must have nested items, display items cannot have nested items</sch:assert>
     </sch:rule>
     <sch:rule context="f:Questionnaire/f:item/f:enableWhen">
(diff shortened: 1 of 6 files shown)


hdconradi:
After the last round of commits for R4 - in particular 07290ae - all the R4 stuff here is obsolete.
Should I open a new pull request for the STU3 changes in isolation?

jamesagnew:
Hi @hdconradi, I'm sorry to have to say this but those files are code-generated so pull requests against them won't have any effect the next time we sync from the FHIR sources.

To file change requests against the validation artifacts you'll need to file a tracker item in HL7's tracker here: https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemBrowse&tracker_id=677 (we are moving to JIRA soon for HL7 stuff so the process will get better, but for now gforge is still the place)

jamesagnew:
... that said, your tests are much appreciated and I'm going to merge them.

hdconradi:
Hi @jamesagnew,
two questions:
- you pulled the change for the dstu3 questionnaire schematron file. Is that change then going get away then in the future, or did you propagate the change the original FHIR sources.
- I have another change on my todo list concerning recursive constraints in the Questionnaire and QuestionnaireResponse (dstu3 and r4). These then should directly go to the original FHIR sources?

And one comment:
- Due to the recent structure changes in R4, my tests are meaningless there and should probably be removed. Should I open a PullRequest to do that?

Thanks

PR #874, opened 2018-03-01
Validation server exception when local reference absent

hdconradi:
Currently an invalid local reference causes an InternalErrorException caused by: FHIRException: Resource resolution services not provided.
The validation should instead return normally and contain an error message.

This pull request fixes this error by allowing `fetcher==null` in case of local references.

Unfortunately, there are now two error messages with the same meaning:
- one from the ProfileValidation
- one from the code in this location.


hdconradi:
Rebased to the state after the R4 sync.

hdconradi:
Rebased after 3.3

jamesagnew:
Thanks for the PR! This looks great.

It has been merged and synced upstream.

PR #875, opened 2018-03-05
Fix error in narrative for CodeableConcept

RuthAlk:
There was an error in the html-template for the narrative of a CodeableConcept. It displayed 'true' instead of the value.


jamesagnew:
Thanks for the PR! Merged.

PR #876, opened 2018-03-05
Allow slotting in own IMessageResolver

RuthAlk:
This is useful in case we want to define our own way of translating
the codes in the thymeleaf templates.


jamesagnew:
Hi @RuthAlk - This looks potentially good, but would you be able to add a unit test showing how this would be used?

Issue #878, opened 2018-03-06
Resources in a bundle are updated even if their contents don't change


darktyco:
Resources in a bundle are updated even if their contents don't change. Steps to reproduce:

1) POST a bundle containing a resource:

```json
{
  "resourceType": "Bundle",
  "type": "batch",
  "entry": [
    {
      "resource": {
        "resourceType": "Patient",
        "id": "TestBundleUpate",
        "name": [
          {
            "use": "official",
            "family": [
              "LastName"
            ]
          }
        ]
      },
      "request": {
        "method": "PUT",
        "url": "Patient/TestBug"
      }
    }
  ]
}
```
2) POST the bundle a second time without modifying any data.
3) Observe in the response that the etag has incremented.

The issue appears to be happening due to the 2-pass technique that the code uses to save batch/transaction bundles (I debugged this in FhirSystemDaoDstu2.java). In the first pass a meta block is being added to the resource which then is saved to the database in the serialized json. Every subsequent update to that resource will have a different meta block causing the code to update the resource each time the bundle is processed.

darktyco:
Added PR that fixes the issue- please inform if it looks like it will cause any unintended side effects.

PR #879, opened 2018-03-06
Fixes issue #878 where resources in a bundle are updated even when no…

darktyco:
… data changes.

It looks like the issue is being caused by a misconfigured element exclusion list that is passed to the parser. "meta" was being passed in but the parser actually looks for "<resource_type>.meta" or "*.meta" when checking if it should serialize the meta block (see "shouldEncodeResourceMeta" in BaseParser.java). Updating the meta element to *.meta so the meta block will be excluded for all resource types resolves the issue.


jamesagnew:
Looks great, merging now!

PR #881, opened 2018-03-08
Two typos in the InstanceValidator

hdconradi:
After the rebase the title is no longer true - there is only one typo left: a wrong javadoc parameter name.


hdconradi:
Rebased after 3.3

jamesagnew:
Merged and synced upstream.

Thanks!

PR #886, opened 2018-03-13
Instance validator NPE on itemtype

hdconradi:
When a questionnaire choice (or openchoice) item has an embedded item a NPE can be triggered.


hdconradi:
Rebased after 3.3

PR #892, opened 2018-03-20
Allow text items in questionnaire

hdconradi:
Currently it is not possible to have a validated QuestionnaireResponse for a Questionnaire with a text item:
According to the specification (https://www.hl7.org/fhir/valueset-item-type.html, or the current version http://build.fhir.org/valueset-item-type.html), the answer has to be a valueString. The InstanceValidator does currently not validate such a request and insists that the answer has to be a 'text'.

This patch-set contains two patches with add tests for all item types for both DSTU3 and R4.
The other patch fixes the issue by adapting the value to compare against.

Furthermore, the patches fix a NPE which I stumbled against when creating the test. Unfortunately, I did forget to record the exact circumstances.


hdconradi:
Rebased after 3.3

PR #903, opened 2018-03-26
Fix #902, IndexOutOfBoundsException for DSTU2 QuestionnaireResponses with groups

### LINKED ISSUE #902: IndexOutOfBoundsException for DSTU2 QuestionnaireResponses with groups
I'm able to consistently get an IndexOutOfBoundsException when trying to create a DSTU2 QuestionnaireResponse. After some digging I've discovered the cause in org.hl7.fhir.instance.validation.QuestionnaireResponseValidator on line 248.

`QuestionnaireResponseValidator.validateGroups` loops through the answer groups and calls `validateGroup` on each one, starting on line 247. Line 248 gets an index to use in `thePathStack` for reporting errors. However instead of getting the index of `nextAnswerGroup` like the previous version of `validateGroups` did, it gets the index of `answerGroups.get(1)`. If answerGroups should happen to only have one group in it, which is the expected behavior of *all* groups that don't repeat, then this line of code will throw an IndexOutOfBoundsException every time a group is validated.

The fix is to revert line 248 to get the index of `nextAnswerGroup`, not of `answerGroups.get(1)`.

dgileadi:
The PR also adds a test that demonstrates the bug when the fix isn't applied, as well as exercises org.hl7.fhir.instance.validation.QuestionnaireResponseValidator which wasn't exercised by the existing tests.

### DIFF
--- a/hapi-fhir-validation/src/main/java/org/hl7/fhir/instance/validation/QuestionnaireResponseValidator.java
+++ b/hapi-fhir-validation/src/main/java/org/hl7/fhir/instance/validation/QuestionnaireResponseValidator.java
@@ -245,7 +245,7 @@ public class QuestionnaireResponseValidator extends BaseValidator {
 				}
 			}
 			for (org.hl7.fhir.instance.model.QuestionnaireResponse.GroupComponent nextAnswerGroup : answerGroups) {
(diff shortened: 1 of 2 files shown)


jamesagnew:
Fix merged. Thanks for the contribution!

PR #909, opened 2018-03-28
moved OWASP check to a profile

patrick-werner:
fixes build, currently the CDS Server has dependencies with CVSS Score > 8

```
[ERROR] One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '8.0':
[ERROR] 
[ERROR] commonj.sdo-2.1.1.jar: CVE-2015-7082, CVE-2017-14867
[ERROR] eclipselink-2.6.0.jar: CVE-2017-14867
```


jamesagnew:
I see that Travis failed this.. I think we've just grown to a number of modules where it it just plain not possible for the entire HAPI FHIR build to complete in under 40 minutes (which seems to be Travis's limit for cutting off.

I wonder what's the right thing to do here.. There is a lot of duplication in the JPA tests (i.e. same test on different versions of FHIR) and they are about 20 minutes of that time, so there are some savings to be had there. They are also probably the most important tests though since so many people use that module.

Maybe creating a profile for Travis that doesn't include the OSGI or Spring Boot stuff for now would be the best course of action..

jamesagnew:
Incidentally I still think this is a good idea, even if it didn't fix the build. :)

patrick-werner:
`[INFO] BUILD SUCCESS`

ran it locally

PR #917, opened 2018-04-09
Extension domains configuration for the fhr instance validator

hdconradi:
Creating larger questionnaires with extension domains will currently produce lot's of information messages with regard `Unkown extensions` which might drown the more critical warnings and errors.
The `InstanceValidator` already has a data structure which is used for registering `extension domains`, but this data structure is not exposed to client code.

The first of patch for this branch add the necessary functionality for the `FhirInstanceValidator` in DSTU3 and R4 and the second patch add two tests.

The third and fourth patch are not really about this functionality and can in priniciple remove from this branch and/or moved in their own branches.
The third patch fixes and obvious bug when intialising the `FhirInstanceValidator` and the forth one tackles a variety of different less important issues.

### DIFF
--- a/hapi-fhir-validation/src/main/java/org/hl7/fhir/dstu3/hapi/validation/FhirInstanceValidator.java
+++ b/hapi-fhir-validation/src/main/java/org/hl7/fhir/dstu3/hapi/validation/FhirInstanceValidator.java
@@ -121,17 +121,13 @@ public class FhirInstanceValidator extends BaseValidatorBridge implements IValid
 	}
 
 	private String determineResourceName(Document theDocument) {
-		Element root = null;
-
 		NodeList list = theDocument.getChildNodes();
 		for (int i = 0; i < list.getLength(); i++) {
 			if (list.item(i) instanceof Element) {
-				root = (Element) list.item(i);
-				break;
+				return list.item(i).getLocalName();
 			}
 		}
-		root = theDocument.getDocumentElement();
-		return root.getLocalName();
+		return theDocument.getDocumentElement().getLocalName();
 	}
 
 	private ArrayList<String> determineIfProfilesSpecified(Document theDocument) {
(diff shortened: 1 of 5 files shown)


hapi-fhir-validation/src/main/java/org/hl7/fhir/dstu3/hapi/validation/FhirInstanceValidator.java
```diff
@@ -70,18 +71,24 @@ public FhirInstanceValidator(IValidationSupport theValidationSupport) {
 		myValidationSupport = theValidationSupport;
 	}
 
-	private String determineResourceName(Document theDocument) {
-		Element root = null;
+	public FhirInstanceValidator setExtensionDomains(List<String> extensionDomains) {
```
jamesagnew: This is very cool. Would you mind to add some javadoc to the `setExtensionDomains` indicating what the values would be and what the impact of setting them is? I'm assuming that the actual term "extension domain" comes from InstanceValidator so it's probably worth preserving that term, but it's not very intuitive what it would be. :)

hdconradi: Sorry, I somehow missed your comment.
And, yes I will of course do it. I will just take a few hours since at the moment I am a bit swamped.
But it should only be hours ...

hdconradi:
I have added a bit javadoc and renamed those methods from `setExtensionDomains` -> `setCustomExtensionDomains`.
Furthermore I have discovered a (probable) bug for the predefined fhir StructureDefinitions and corrected it in the very first patch.
The tests have been adapted accordingly.

hdconradi:
Rebase after 3.4

hdconradi:
Rebase after 3.5.0

hdconradi:
Rebase after 3.6

jamesagnew:
Hi @hdconradi - Sorry this took so long to merge, and many thanks for keeping it up to date. I'm merging it now so you don't need to do this any longer. :)

PR #918, opened 2018-04-09
Delete unused dstu3 InstanceValidator class

hdconradi:
According to the 3.3 release notes the dstu3 `InstanceValidator` class is not used any more.
And indeed the class can simply be deleted.
It has been deleted once already in commit 8d468de5515c74bc7a7b59977bd2559ede0ff7b7, but was reintroduced in 07290ae8a0181c4c9d0ba863cf72f447065e66a1 (probably by accident).
This patch simply removed the class again together with the `IResourceValidator`.


hdconradi:
Already partially done in the trunk.

PR #928, opened 2018-05-03
R4 support for Spring Boot

t4deon:
R4 support added to Spring Boot distribution


patrick-werner:
thanks for added R4 Config support

PR #932, opened 2018-05-07
A generic version of IResourceProvider

jkiddo:


jamesagnew:
This is a neat idea, but I feel like it just trades putting the resource name in one place (the getResourceType method) for putting it in another place (the generic type), which is a few fewer keystrokes but also risky if the class has implemented more than one typed interface (and if history has shown  me anything it's that someone will do just that and then have incomprehensible errors on the mailig list :) )

This might be better as an example in the fhirstarters project or somewhere else I'm thinking..

jkiddo:
right - fair enough. It's purely thought as a shorthand notation-thingy - and as with all other code, it should be used carefully.

PR #969, opened 2018-05-30
Fix client upload-terminology for r4.

rhausam:
Allows client to successfully upload SNOMED CT and LOINC to R4 server.


jamesagnew:
Merged. Thanks Rob! :)

PR #974, opened 2018-05-31
Fixes #972 Update Spring Data 1.11.6.RELEASE to 2.0.7.RELEASE

### LINKED ISSUE #972: Spring data migration to 2.x
Is spring data migration to 2.x (that brakes binary compatibility) planned any time soon? If yes I can start working on that and get a pull request ready. 

P.S. thanks for the awesome software!

Romanow88:
I was not sure what to do with the orElse part of the Optional that Spring Data now returns for .findBy... , so I used a simple IllegalStateException. If you think we should use something else - let me know.


Romanow88:
@jamesagnew is the CI valid? One failed with a test that the second one passed only to fail on another.

Running the build locally produces a success.

jamesagnew:
No, the CI is in a bad state right now due to HAPI's build being so big that it regularly fails for no good reason on both Travis and Appveyor. I'll be landing the `java9` branch ASAP which gets it back in good state.

PR #990, opened 2018-06-07
Adding the conversion of the Procedure resource

pz12:
This is a change in the conversion between dstu3 and r4


jamesagnew:
Hi @pz12 - For some reason this is showing as a diff that changes every line of the file. GitHub won't even show me the changes because it's too big.

I'm assuming you only added one method- Are you able to just submit that method?

pz12:
The changes seem to be out of date and will probably be even more out of date as R4 progresses. I would just ignore this pull request.

patrick-werner:
> The changes seem to be out of date and will probably be even more out of date as R4 progresses. I would just ignore this pull request.

closed

Issue #1006, opened 2018-06-20
Writing structured Binary resources fails with charset being set in Content-Type header


ohr:
If a Binary resource is uploaded using the Content-Type application/fhir+json;charset=UTF-8 or application/fhir+xml;charset=UTF-8, it is not detected as being a resource upload. Instead, the JSON/XML is treated as Binary content.
Apparently. ResourceParameter#parseResourceFromRequest does a strict check on application/fhir+json or application/fhir+xml only. Everything else is treated as binary upload. 

However, https://www.hl7.org/fhir/http.html#mime-type states:
> FHIR uses UTF-8 for all request and response bodies. Since the HTTP specification (section 3.7.1) defines a default character encoding of ISO-8859-1, requests and responses SHALL explicitly set the character encoding to UTF-8 using the charset parameter of the MIME-type in the Content-Type header.

patrick-werner:
you are right - despite your use case doesn't seem makes sense to me. If you have a FHIR resource there is no benefit of saving it as a Binary (maybe i can't see the benefit).
But according to the spec the implementation was wrong. I've commited a fix, will do a pull request later

patrick-werner:
fixed, see pull request.

PR #1007, opened 2018-06-21
1006 writing structured binary

patrick-werner:
Binary resource upload using _application/fhir+json;charset=UTF-8_ or 
 _application/fhir+xml;charset=UTF-8_ works now.
see #1006


patrick-werner:
builds without errors locally and with appveyor

PR #1015, opened 2018-06-27
avoid potential short circuit during saveDeferred processing of uploaded (e.g. LOINC) terminology

jschneid-nmdp:
The BaseHapiTerminologySvcImpl saveDeferred method's early return when myDeferredConcepts and myConceptLinksToSaveLater are both empty has the potential to cause some value sets and concept maps to not be saved.

Prior to making the change illustrated by this PR, testing of upload-terminology for LOINC 2.64 indicated 270 deferred ValueSet resources were not saved.  Some of the not-saved value sets were:  LL736-0, LL999-4, loinc-document-ontology, top-2000-lab-observations-us, and loinc-universal-order-set-vs.


jamesagnew:
Looks good, thanks! This has been merged.

PR #1047, opened 2018-08-01
Fix NullPointerExceptions in DateRangeRaram

hdconradi:
There a possible `NullPointerExceptions` in the `DateRangeParam` class which can be triggered e.g. when a search query for `_lastUpdated` is not properly filled (examples include `"", "lt", "gt"`).
The patch consits of two commits. The first adds adds junits tests which demonstrate the problem.
The second commit proposes a solution.


jamesagnew:
Thanks for the PR! Merging before 3.6.0 release.

Issue #1048, opened 2018-08-01
FhirInstanceValidator::validate() not looking in meta/profile section of request prior to calling InstanceValidator::validate()


anthonys123:
When a StructureDefinition is loaded on demand by FhirInstanceValidator::validate(), the code should inspect the meta/profile section in the request to see if profile(s) are specified.  If so, then the meta.profile value should be used instead of the resource name to pass to InstanceValidator::validate().

Also, multiple profiles can be specified in that section, which would mean all profiles have to be validated.

Our profiles are named differently from the standard FHIR ones, this is how we hit the problem.  For instance, our practitioner profile is named "ocp-practitioner".  The current code sends the FHIR validator the resource type name, "practitioner", which means that the "ocp-practitioner" profile is never retrieved by the validator.

I've tested a fix in FhirInstanceValidator::validate() for the JSON processing of a single profile.  With the changes below, the FHIR validator picks up our "ocp-practitioner" profile and validates the request correctly.  Later this week I'll re-factor this to handle multiple profiles in the meta section.  I'll also implement and test the fix for XML.  Initial JSON fix below:

		} else if (theEncoding == EncodingEnum.JSON) {
			Gson gson = new GsonBuilder().create();
			JsonObject json = gson.fromJson(theInput, JsonObject.class);

			String resourceName = null;
			try {
				String[] components = json.getAsJsonObject("meta").get("profile").getAsString().split("/");
				resourceName = components[components.length - 1];
			} catch (Exception e) {
				resourceName = json.get("resourceType").getAsString();
			}
			StructureDefinition profile = findStructureDefinitionForResourceName(theCtx, resourceName);
			if (profile != null) {
				try {
					v.validate(null, messages, json, profile.getUrl());
				} catch (Exception e) {
					throw new InternalErrorException("Unexpected failure while validating resource", e);
				}
			}
		} else {
			throw new IllegalArgumentException("Unknown encoding: " + theEncoding);
		}

anthonys123:
James - feel free to assign this one to me.

jamesagnew:
Sounds good!

anthonys123:
Hi James - created pull request this morning.  Thanks.

anthonys123:
Hi James - I just realized that there are two more FhirInstanceValidator.java files I need to commit these changes to also.  I'll do that now and create another pull request.

anthonys123:
[...]

PR #1050, opened 2018-08-06
Fixes for issue #1048.  FhirInstanceValidator::validate() now looking…

anthonys123:
… in meta/profile section of request prior to calling InstanceValidator::validate().


anthonys123:
Hi James - I just realized that there are two more FhirInstanceValidator.java files I need to commit these changes to also.  I'll do that now and create another pull request.

jamesagnew:
Hi Anthony,

After merging this in, we had a few complaints that the new requirement for SD URLs to contain the string StructureDefinition was breaking existing working code. I have rolled this back. Any objections?

anthonys123:
No objections.  Thanks for inquiring.

anthonys123:
Hi James - quick follow-up question (also e-mailed you).  Did you mean that you rolled the entire implementation for this issue out or only the code block that looks for the "StructureDefinition" text?  I interpreted your comment to mean the latter - was that correct?

PR #1051, opened 2018-08-07
head operation should return 200OK response code instead of 400 BadRequest

Cory00:
Hello,

Based on the FHIR specifications (http://hl7.org/fhir/2018May/http.html#head) :

"Anywhere that a GET request can be used, a HEAD request is also allowed. HEAD requests are treated as specified in HTTP: same response as a GET, but with no body."

Right now, the head operation returns 400 Bad Request with no body and it should return the code 200 OK (same as for a normal GET). 

The head operation works as excepted only for _history operations.


stevelle:
* good change for the project
* implementation is reasonable size and shape
* adds & updates appropriate testing

Nice 👍

jamesagnew:
This looks great, merging now. Thanks for the PR!

Please let me know if you'd like a particular name/organization credited in the project developer list.

Cory00:
Hello, 
Please add Corina Dalaban as the name and Cerner Corporation as organization.
Thank you!

PR #1064, opened 2018-08-27
Issue 430: Generate OperationDefinitions for named queries, fix minor…

### LINKED ISSUE #430: DSTU3 - The named query (search) does not been generated in ServerConformanceProvider
Hello,
   I found that the advanced query which was defined by named query - "@Search(name=xxx)" does not been generated in ServerConformanceProvider.
  http://www.hl7.org/implement/standards/fhir/search.html#advanced, the official document said that "Servers can define their own additional named queries to meet their own uses using an OperationDefinition.", So when there are named queries used in resource providers, then the metadata should generate the OperationDefinition like as the "@Operation" if I understood correctly.
   @jamesagnew, can you check that?

srdo:
… errors in generation of OperationDefinition for operations. Closes #430.

See https://github.com/jamesagnew/hapi-fhir/issues/430

This fixes the following issues:
* Generate OperationDefinition for named queries. Added tests for queries at type level and at server level.
* Fix potential NPE in BaseMethodBinding when the method return type is a raw typed collection.
* When an operation appears at system level with no IdType parameter, the generated definition specified that the operation could be invoked at both system and type level.


srdo:
Tests are failing, will take a look.

PR #1088, opened 2018-10-03
Issue-1077: Make CapabilityStatementProvider use the closest common s…

srdo:
…uperclass of provided resources when generating rest.resource.profile, instead of always using the base definition.

This resolves https://github.com/jamesagnew/hapi-fhir/issues/1077.

The capabilitystatement.rest.resource.profile link should point to a StructureDefinition that applies to all profiles of that particular resource name. The current implementation simply points to the base definitions for the name, since these profiles are guaranteed to be adhered to by profiles extending the base definitions.

The changes here try to narrow down the right StructureDefinition link further. For example, if a server only provides a Patient profiling called MyPatient, there's no reason not to point to the MyPatient StructureDefinition in capabilitystatement.rest.resource.profile.

The approach here is to look at all the resource providers for a given resource name, and find the "most specific" resource superclass that is a supertype of all profiles specified via getResourceType. The profile for that type should apply to all subclasses as well, so it's the most specific profile that still applies to all resources with that name on the server.

For example, say the hierarchy for Patient is
```
Patient -> MyPatient -> MyPatient2
                     -> MyPatient3
```

If the server has two resource providers for Patient, with one supplying MyPatient2 and one supplying MyPatient3, we can see that the common supertypes for these are Patient and MyPatient. Since MyPatient is the most specific profiling that still applies to all the provided resources, it should be the one linked to in capabilitystatement.rest.resource.profile.

### DIFF
--- a/hapi-fhir-jaxrsserver-base/src/main/java/ca/uhn/fhir/jaxrs/server/AbstractJaxRsConformanceProvider.java
+++ b/hapi-fhir-jaxrsserver-base/src/main/java/ca/uhn/fhir/jaxrs/server/AbstractJaxRsConformanceProvider.java
@@ -46,6 +46,7 @@ import ca.uhn.fhir.rest.api.server.IRestfulResponse;
 import ca.uhn.fhir.rest.server.*;
 import ca.uhn.fhir.rest.server.method.BaseMethodBinding;
 import ca.uhn.fhir.util.ReflectionUtil;
+import java.util.stream.Collectors;
 
 /**
  * This is the conformance provider for the jax rs servers. It requires all providers to be registered during startup because the conformance profile is generated during the postconstruct phase.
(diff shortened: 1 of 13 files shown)


srdo:
I'm happy to make the same changes to the R4 capability statement provider, but I'd like feedback on this before I spend more time on it.

jamesagnew:
Hi @srdo - Apologies for the slow review on this. The approach makes lots of sense to me. If you're interested in making the equivalent R4 change I'm definitely happy to merge.

srdo:
@jamesagnew Rebased and added R4 support. Please let me know when you've reviewed, and I will squash to one commit.

srdo:
Rebased to fix conflicts.

srdo:
@jamesagnew Friendly reminder, this should be ready to merge once you have time to look at it.

srdo:
Implemented this for R5 as well, also switched all tests to use the non-deprecated ServerCapabilityProvider constructor.

srdo:
@jvitrifork If this looks fine, could you merge it? I'd like to not have to keep fixing conflicts with master.

jkiddo:
@srdo Im afraid that my priviledges are the same as yours. You need the attention of @jamesagnew  I guess ;)

srdo:
Ah, sorry. Assumed that you had merge access due to the approval.

jkiddo:
My hope was that 'my approval' could help push this on its way

(@jvitrifork == @jkiddo)

jamesagnew:
Sorry for the slow merge... We've been backlogged while trying to get infrastructure issues solved (CI and coverage, etc.). Now that we're past that, it's time for digging out!

srdo:
Thanks

PR #1100, opened 2018-10-24
changed offset for links to use the number of resources that match th…

papagator:
…e request resource name


papagator:
Seems that this fixes one issue but causes another where the next links have the wrong page offset. This is due to the fact that the next link does not have a resourceType. This means the offset never increases.

jamesagnew:
Hi @papagator - I'm not clear on what this is actually fixing. Would you be able to add a unit test that demonstrates the issue?

PR #1105, opened 2018-10-27
Make test observation lists thread safe

fil512:
Small fix to address intermittent test failures.


fil512:
Merging at James' request

PR #1112, opened 2018-11-08
Enhancements to tinder generic Mojo and Ant tasks

bdenton:
This PR allows the generic Mojo and Ant tasks in the Tinder plug-in to generate resource-based artifacts based on the model structures (hapi-fhir-structures-xxx) in addition to the existing FHIR spreadsheets. This is to align generic source generation with the same approach currently used in the JPA providers.

This PR also adds FHIR-R4 support to the generic Mojo and Ant tasks

(Note, not sure why my commits resulted in whole file replacement and failed to properly show the actual diffs.)

cheers,
Bill Denton


bdenton:
Any objections to merging this PR? It is pretty non-intrusive.
cheers, Bill

PR #1117, opened 2018-11-15
Fixed small log formatting error in ExceptionHandlingInterceptor.java

magnuswatn:
Hi,

Came across this small log formatting error in ExceptionHandlingInterceptor.java, and thought it should be fixed.

(formatting anchor is not needed when logging a throwable)


jamesagnew:
Looks great, thanks!

PR #1139, opened 2018-12-07
issue-1107: make v3.3.0 JDK7

### LINKED ISSUE #1107: Hapi V3.3.0 is not JDK 7 compatible
https://github.com/jamesagnew/hapi-fhir/wiki/Getting-Help

**Describe the bug**
The changelogs for Hapi v3.3.0 mention that it is JDK 7 compatible, but it requires a version of Guava that is past JDK 7 support (Guava 23.0). You can't simply exclude this if you're using OSGI like I am because that version is still trying to be pulled in through the manifest. Is there a way we can exclude this from the manfiest classpath or downgrade to 21.0 which is the last supported version of Guava to support JDK 7?

**To Reproduce**
1. Try to run the application on JRE 7.
2. Get major.minor version error 52.0 on Guava

**Expected behavior**
Should work on JRE/JDK 7

 - HAPI FHIR Version: 3.3.0
 - OS: Windows 10

**Additional context**
If there were a solution to excluding the import-package from the manifest I would take that too, but as it stands 3.3.0 is not technically JDK 7 compatible.

restevez-chs:
# Help. I Cannot create PR against tag
I have tried to submit a Pull Request against v3.3.0 tag, [but every time I try](https://github.com/jamesagnew/hapi-fhir/compare/4c0d5b49d9b24e8459d89b6533b4034f60d25082...restevez-chs:issue-1107-v3.3.1-release-candidate), I do not get a "Create Pull Request" form. I see a diff which is correct, but no way to create a Pull Request. That said, the fix here is simple. I did checkout the v3.3.0 tag first, then made the POM file changes. I built locally with `mvn clean install` and it was successful.

I am happy to submit another PR, but I think @jamesagnew there needs to be a branch that you or another HAPI FHIR contributor creates off the v3.3.0 tag. Then I can create a PR to that new branch. Or its a permissions issue?

# About
[Google Guava release 21](https://github.com/google/guava/wiki/Release21) requires JDK8. HAPI FHIR v3.3.0 was previously specifying a newer Guava version which requires JDK8 rather than JDK7, which HAPI FHIR v3.3.0 is [advertised to support](http://hapifhir.io/download.html).

# Changes
Update POM to specify the "android" version which is [recommended](https://github.com/google/guava/wiki/Release22#new-android-version) by Google for JDK7. Using `23.0-android` is the latest Guava available.

# References
* https://github.com/google/guava/wiki/Compatibility
* https://github.com/google/guava/wiki/Release23
* https://github.com/google/guava/wiki/Release22
* https://search.maven.org/artifact/com.google.guava/guava/23.0-android/bundle

#Fix #1107

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -12,7 +12,7 @@
 	<groupId>ca.uhn.hapi.fhir</groupId>
 	<artifactId>hapi-fhir</artifactId>
 	<packaging>pom</packaging>
-	<version>3.3.0</version>
+	<version>3.3.1</version>
 	<name>HAPI-FHIR</name>
 	<url>https://hapifhir.io</url>
 
@@ -443,7 +443,8 @@
 		<commons_lang3_version>3.6</commons_lang3_version>
 		<derby_version>10.14.1.0</derby_version>
 		<error_prone_annotations_version>2.0.18</error_prone_annotations_version>
-		<guava_version>23.0</guava_version>
+		<!-- guava_version: the android flavor is needed for JDK7 (see https://github.com/google/guava/wiki/Compatibility) -->
+		<guava_version>23.0-android</guava_version>
 		<gson_version>2.8.1</gson_version>
 		<jaxb_api_version>2.3.0</jaxb_api_version>
 		<jaxb_core_version>2.3.0</jaxb_core_version>


pom.xml
```diff
@@ -443,7 +443,8 @@
 		<commons_lang3_version>3.6</commons_lang3_version>
 		<derby_version>10.14.1.0</derby_version>
 		<error_prone_annotations_version>2.0.18</error_prone_annotations_version>
-		<guava_version>23.0</guava_version>
+		<!-- guava_version: the android flavor is needed for JDK7 (see https://github.com/google/guava/wiki/Compatibility) -->
+		<guava_version>23.0-android</guava_version>
```
jamesagnew: This change would affect all of HAPI FHIR (including non-android builds). Can you set this change up as a separate variable so that it can be used only by the hapi-fhir-android module?

patrick-werner:
JDK level already changed, not relevant anymore

PR #1152, opened 2018-12-19
Allow to set additional Http headers directly on the IClientExecutable

ohr:
I'd like to propose an extension to the IClientExecutable interface that allows to set arbitrary HTTP headers. This is an alternative to using AdditionalRequestHeadersInterceptor, with one important difference: it makes the client stateless so it can even be reused with header values like OAuth2 tokens that change with every request. 
And I think it's simply more straightforward to use.


jamesagnew:
Is there a reason you wouldn't just use [AdditionalRequestHeadersInterceptor](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/interceptor/AdditionalRequestHeadersInterceptor.java) to add arbitrary headers?

PR #1154, opened 2018-12-21
Subscription module startup changes

fil512:
This PR makes a few minor changes to how the subscription registry works in HAPI FHIR.

The main change is that subscription interceptors are now marked @Lazy and are only explicitly loaded by SubscriptionInterceptorLoader.registerInterceptors().  This method first checks DaoConfig to see if any types of Subscriptions are supported and then loads all active subscriptions into the SubscriptionRegistry before loading the interceptors.

Also I changed the SubscribableChannelFactory interface so that it more naturally supports the different kinds of channel factories that need to be supported.


fil512:
Also this PR adds support for R4 subscriptions

PR #1165, opened 2019-01-12
Resilient searchparamregistry

fil512:
SearchParamRegistry and SubscriptionRegistry now both poll retrying connection to the server.  This will help in particular in the scenario where the subscriptions are managed in a standalone server that depends on a FHIR Server being available to pull search parameters and subscriptions from.

### DIFF
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/retry/Retryer.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/retry/Retryer.java
@@ -1,0 +1,46 @@
+package ca.uhn.fhir.jpa.searchparam.retry;
+
+import org.apache.commons.lang3.time.DateUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.function.Supplier;
+
+public class Retryer<T> {
+	private static final Logger ourLog = LoggerFactory.getLogger(Retryer.class);
+
+	private final Supplier<T> mySupplier;
+	private final int myMaxRetries;
+	private final int mySecondsBetweenRetries;
+	private final String myDescription;
+
+	public Retryer(Supplier<T> theSupplier, int theMaxRetries, int theSecondsBetweenRetries, String theDescription) {
+		mySupplier = theSupplier;
+		myMaxRetries = theMaxRetries;
+		mySecondsBetweenRetries = theSecondsBetweenRetries;
+		myDescription = theDescription;
+	}
+
+	public T runWithRetry() {
+		RuntimeException lastException = null;
+		for (int retryCount = 0; retryCount < myMaxRetries; ++retryCount) {
+			try {
+				return mySupplier.get();
+			} catch(RuntimeException e) {
+				ourLog.info("Failed to {}.  Attempt {} / {}: {}", myDescription, retryCount, myMaxRetries, e.getMessage());
+				lastException = e;
+				try {
+					Thread.sleep(mySecondsBetweenRetries * DateUtils.MILLIS_PER_SECOND);
+				} catch (InterruptedException ie) {
+					lastException = null;
+					Thread.currentThread().interrupt();
+					break;
+				}
+			}
+		}
+		if (lastException != null) {
+			throw lastException;
+		}
+		return null;
+	}
+}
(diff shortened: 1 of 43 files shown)


fil512:
@jamesagnew ready to me merged to master pending code review.

fil512:
There's a lot of noise in this PR with the regsitry -> registry typo fix.  Search for "Retryer" to find the material code changes.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/retry/Retryer.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class Retryer<T> {
+	private static final Logger ourLog = LoggerFactory.getLogger(Retryer.class);
+
+	private final Supplier<T> mySupplier;
+	private final int myMaxRetries;
+	private final int mySecondsBetweenRetries;
+	private final String myDescription;
+
+	public Retryer(Supplier<T> theSupplier, int theMaxRetries, int theSecondsBetweenRetries, String theDescription) {
```
jamesagnew: I guess the name of this class is a made up word anyway... but i feel like it would be spelled Retrier :) A google search seems to argue in favour of that spelling: 57k results for "Retrier" and 8k results for "Retryer"

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/retry/Retryer.java
```diff
@@ -0,0 +1,46 @@
...
+					lastException = null;
+					Thread.currentThread().interrupt();
+					break;
+				}
+			}
+		}
+		if (lastException != null) {
+			throw lastException;
+		}
+		return null;
```
jamesagnew: Does it really make sense to clear the exception if the thread is interrupted, and return null here?

At least my naive expectation of a class with an API like this would be that it will always either return the results of the supplier, or throw an exception.. So if the supplier returns null as a meaningful result I'd be confused to get null back because of an interrupt.

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/module/cache/SubscriptionLoader.java
```diff
@@ -45,46 +46,52 @@
 @Lazy
 public class SubscriptionLoader {
 	private static final Logger ourLog = LoggerFactory.getLogger(SubscriptionLoader.class);
+	@VisibleForTesting
+	public static final int INITIAL_SECONDS_BETWEEN_RETRIES = 5;
+	private static final int MAX_RETRIES = 60; // 5 minutes
```
jamesagnew: Only because I spent way too long pondering how 60 = 5 mins, could you beef this comment up to say something like "60 * 5 seconds = 5 mins"

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/retry/Retryer.java
fil512: done

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/module/cache/SubscriptionLoader.java
fil512: done

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/retry/Retryer.java
fil512: The interrupted case is a special case, almost always a shutdown scenario.  I returned null to usher in a smooth shutdown.  I see your point though, the caller might not be prepared for that.  I'll just rethrow the last exception.

fil512: Fixed.  Also added some test cases.

PR #1169, opened 2019-01-16
Subscription startup side-effects

fil512:
- Changed SubscriptionMatchingIntercepter so that you must call start() on it to start the subscription matching jms channel (before it auto-started in `@PostConstruct`)
- SubscriptionTriggeringSvcImpl now depends on an IResourceModifiedConsumer as opposed to explicitly using a SubscriptionMatcherInterceptor to send subscriptions.  This allows users to create their own interceptors that send resources to a standalone subscription matching server.


fil512:
Note a bunch of changes here were IntelliJ recommendations that I accepted.  Not sure if they’re all improvements (esp the ones that result in uneven visibility)

PR #1170, opened 2019-01-17
fixed null subscription id

fil512:
changed subscription match api to take a CanonicalSubscription instead of a criteria String


jamesagnew:
@fil512 looks good, merge away!

PR #1187, opened 2019-01-31
Add JSON-like tests for DSTU3 and R4

bdenton:
Add tests of the JSON-Like features of the JSON Parser to the DSTU3 and R4 structures projects. All the other JSON parser tests will also exercise the JSON-Like features used by the default GSON implementation.


jamesagnew:
Much appreciated.

PR #1191, opened 2019-02-02
Standalone subscription delete fail

fil512:
Found a bug where the standalone subscription message handler couldn't handle delete messages (because they have no payload).  This PR includes a test that failed before I made the fix.


fil512:
@jamesagnew please review

jamesagnew:
Looks good- Merging now.

PR #1204, opened 2019-02-10
Subscriptions are not enabled in JPA Server Example 3.7.0.

volsch:
Unlike in JPA Server Example 3.6.0 and previous versions, subscriptions are not enabled in JPA Server Example 3.7.0 (after the change of subscription handling). This pull request enables the same subscription types that were also enabled in JPA Server Example 3.6.0.


jamesagnew:
HI @volsch - I'm definitely happy to merge this but FYI we're deprecating this sample project in favour of this one: https://github.com/hapifhir/hapi-fhir-jpaserver-starter

PR #1205, opened 2019-02-12
no subs check interceptor calls on DELETE

fil512:
don't call SUBSCRIPTION_BEFORE_PERSISTED_RESOURCE_CHECKED and SUBSCRIPTION_AFTER_PERSISTED_RESOURCE_CHECKED for DELETE operations (since we don't currently match subscriptions for DELETE operations)


fil512:
@jamesagnew please review

fil512:
A couple of projects were marked as "failed" by sonar, so I cleaned up the code that was triggering the failures.  The ResourceHistoryTable fix looks like it was a legitimate bug that Sonar found...

Sorry to muddy the waters on this little PR, but I like seeing those green "PASSED" marks in Sonar.

PR #1208, opened 2019-02-15
Reuse subscription channels

fil512:
Subscription Registry reuses subscription channel and handlers if channel type didn't change when a subscription is updated


fil512:
@jamesagnew ready for review

PR #1211, opened 2019-02-17
changes needed to support different number of consumers in matching vs delivery

fil512:


jamesagnew:
Fyi will review this in 2 hrs

sent from my phone.

PR #1214, opened 2019-02-20
Improve SQL IN(..) queries

jamesagnew:
@fil512 code review please


fil512:
@jamesagnew This looks good to me.  My only question is do we need to test this against different databases (Oracle, Postgres, etc.) . How confident are we in the JPA mapper for this?

jamesagnew:
@fil512 very. :)

Issue #1042, opened 2018-07-25
QuestionnaireResponse validator doesn't consider enableWhen


eevaturkka:
We have a questionnaire with required element that has "enable when" set, see: https://simplifier.net/Testproject25/Questionnaire-example-duplicate-2 

The standard says: "This element is a modifier because if enableWhen is present for an item, "required" is ignored unless one of the enableWhen conditions is met." (http://hl7.org/fhir/questionnaire-definitions.html#Questionnaire.item.enableWhen)

Here is an QuestionnaireResponse that should be valid, because of the conditions of the enableWhen are not met: https://simplifier.net/testproject25/questionnaireresponse-example-duplicate-6 josta HAPI valittaa ”No response found for required item 2”.

I found no references to the enableWhen in the code so maybe this is not implemented yet?

`				rule(theErrors, IssueType.BUSINESSRULE, thePathStack, !(theQuestion.getRequired() && responseQuestion.getAnswer().isEmpty()), "Missing answer to required question with linkId[{0}]", linkId);
`
This is reproducible with Hapi 3.2

eevaturkka:
Tested now and is still reproducible with Hapi 3.5

eevaturkka:
@jamesagnew  we are planning to fix this, so could you assing this? We can have a chat about the intended solution if you have any thoughts about how this should be done, one requirement we have for the solution is that we can inject our own optional modifier extension behaviour to dstu3 validation of enable when in addition to the default behaviour described in the standard.

PR #1228, opened 2019-03-06
QuestionnaireResponses should be able to have Items without Answers in disabled Questions

mattiuus:
Related to issue  #1042 

This augments our earlier pull request with Enablewhen support. Without these fixes the validation fails with answers that have items, but no answers – which should be allowed in R4.


mattiuus:
Working on the conflicts

mattiuus:
Merge conflicts resolved. Unfortunately rest of the code is not compiling but I can't get the master to compile either.

hapi-fhir-converter/pom.xml
```diff
@@ -84,7 +84,7 @@
 		<dependency>
 			<groupId>ca.uhn.hapi.fhir</groupId>
 			<artifactId>hapi-fhir-client</artifactId>
-			<version>${project.version}</version>
+			<version>3.6.0</version>
```
jamesagnew: Why is this being hardcoded to a previous version of the client rather than the current one?

hapi-fhir-validation/src/main/java/org/hl7/fhir/r4/validation/InstanceValidator.java
```diff
@@ -2506,7 +2506,7 @@ public void setAllowXsiLocation(boolean allowXsiLocation) {
     this.allowXsiLocation = allowXsiLocation;
   }
   
-  public void setEnableWhenEvaluator(IEnableWhenEvaluator myEnableWhenEvaluator) {
+  public void setMyEnableWhenEvaluator(IEnableWhenEvaluator myEnableWhenEvaluator) {
```
jamesagnew: Following javabean conventions, this should be `setEnableWhenEvaluator`

mattiuus: will fix

hapi-fhir-converter/pom.xml
mattiuus: will fix

mattiuus:
I made the fixes requested in the review. Thr PR is ready for a new review.

PR #1232, opened 2019-03-09
Has param no in

fil512:
@jamesagnew this is ready for review


hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/param/HasParam.java
```diff
@@ -33,7 +33,8 @@
 public class HasParam extends BaseParam implements IQueryParameterType {
 
 	private static final long serialVersionUID = 1L;
-	
+
+	// FIXME KHS rename this
```
jamesagnew: Presumably this should be resolved..

fil512: done

PR #1236, opened 2019-03-11
Pointcut parameters and Retrier log

fil512:
Fixed a pointcut parameter list and log an exception that was swallowed by Retrier

### DIFF
--- a/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/interceptor/api/Pointcut.java
+++ b/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/interceptor/api/Pointcut.java
@@ -136,7 +136,7 @@ public enum Pointcut {
 	 * Hooks may accept the following parameters:
 	 * </p>
 	 * <ul>
-	 * <li>java.lang.Exception - The exception that caused the failure</li>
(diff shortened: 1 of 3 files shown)


fil512:
@jamesagnew please review & merge this small PR when you get a chance

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/interceptor/api/Pointcut.java
```diff
@@ -137,8 +137,8 @@
 	 * </p>
 	 * <ul>
 	 * <li>java.lang.Exception - The exception that caused the failure</li>
-	 * <li>ca.uhn.fhir.jpa.subscription.module.CanonicalSubscription</li>
-	 * <li>ca.uhn.fhir.jpa.subscription.module.subscriber.ResourceDeliveryMessage</li>
+	 * <li>ca.uhn.fhir.jpa.subscription.module.subscriber.ResourceDeliveryMessage - the message that triggered the exception</li>
+	 * <li>java.lang.Exception</li>
```
jamesagnew: I'd be inclined to make this explicitly declare `BaseServerResponseException` just to be generally consistent with how HAPI works. Anything the client will throw will already extend this, and you could add a catch-all if necessary that rethrows non-BaseServerResponseException as an `InternalErrorException` if you need to.

Documentation should probably point out that IO failures will be thrown as a `FhirClientConnectionException` (as opposed to an IOException)

fil512: Per our discussion, leaving as is with further clarification in the documentation for that pointcut.

PR #1248, opened 2019-03-21
Added possibility for composition-wise construction of resource provi…

jvitrifork:
…ders through use of easy-to-use default interface methods. Changes to ReflectionUtil was needed as default methods pr. current implementation was not handled

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/ReflectionUtil.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/ReflectionUtil.java
@@ -19,10 +19,12 @@ package ca.uhn.fhir.util;
  * limitations under the License.
  * #L%
  */
+import java.lang.invoke.MethodType;
 import java.lang.reflect.*;
-import java.util.LinkedHashSet;
-import java.util.List;
+import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 import org.apache.commons.lang3.Validate;
 
@@ -37,7 +39,9 @@ public class ReflectionUtil {
 
 	public static LinkedHashSet<Method> getDeclaredMethods(Class<?> theClazz) {
 		LinkedHashSet<Method> retVal = new LinkedHashSet<Method>();
-		for (Method next : theClazz.getDeclaredMethods()) {
+
+		Collection<Method> methods = getAllMethods(theClazz, false, false);
+		for (Method next : methods) {
 			try {
 				Method method = theClazz.getMethod(next.getName(), next.getParameterTypes());
 				retVal.add(method);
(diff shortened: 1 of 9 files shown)


jamesagnew:
This seems like a a neat concept- It's not clear to me what the advantage is though. Is this just about whitelisting operations in the JPA server?

jvitrifork:
Yes, it makes it possible to eg. have a signature like the following:

```
class PatientResourceProvider extends TrimmedJpaResourceProviderDstu3<Patient> implements IPatch, IHistory
```

- Which could come in handy if eg. your patient creation should be managed through a custom operation that looks up all the proper demographics that should not be overwritten by clients. As the methods on the interfaces are implemented as default, one do not have to do anything else than declaring that it implements the interfaces. If one needs to override eg. the IPatch behaviour, one simply overwrites that implementation.

jamesagnew:
Got it- I'm ok with adding this, but it would definitely need comprehensive tests and documentation before this could be merged.

jvitrifork:
How many permutations do you require?

jamesagnew:
I don't think there is any hard and fast rule- I definitely don't think every last branch needs to be tested but certainly for something as fundamental as resourceprovider class scanning I guess it'd be important to exercise the various error conditions and ways people could go wrong to make sure that they all result in either a working system or an error message that gives a clear indication of what's wrong.

My one nervousness with this kind of thing is always support- the volume of requests across the various channels has already vastly outstripped our ability to keep up so I try to put a lot of emphasis on documentation, tests, and helpful error messages.

jvitrifork:
Fair enough ... I'll see what I can do to 'ease the pain', - FWIW, its primarily extraction of the frameworks own existing code. The primary addition is actually the ReflectionUtil scanning part that now can pick up default methods

stefan-lindstrom:
@jamesagnew 
Hi James, 

here's the simplified example you wanted:
`
We created a normal interceptor with an annotation and registered in beans.xml, e.g:

Annotation:
`package com.foo;

@Inherited
@InterceptorBinding
@Retention(RUNTIME)
@Target({ METHOD, TYPE})
public @interface Xyzzy {
  @Nonbinding
  String m() default = "";
}`

Interceptor:
`package com.foo;

@Xyzzy
@Interceptor
XyzzyInterceptor {
  @AroundInvike
  public Object handle(InvocationContext context) throws Exception {
	return context.proceed();
  }
}`

beans,xml:
  `  <interceptors>
        <class>com.foo.Xyzzy</class>
    </interceptors>`
 
Then the provider is basically:

`package com.foo;
public class PRProvider implements IResourceProvider {
  @Override
  public Class<PractitionerRole> getResourceType() {
    return PractitionerRole.class;
  }
  
  @Search
  @Xyzzy(m = "Abandon hope all ye who enter here")
  public List<PractitionerRole> search@IncludeParam(allow = {"PractitionerRole:practitioner"}) Set<Include>  includes) throws BaseServerResponseException {
	return Collections.emptyList();
  }
}`

When scanning PRProvider the scan will be done on the weld proxy class, and when doing the check for triple-nested generic collections, the proxy does not have any generic parameterized info for the includes set, and will return a null on first test, which will trigger NPE on the second test:

`if (Collection.class.isAssignableFrom(parameterType)) {
  innerCollectionType = (Class<? extends java.util.Collection<?>>) parameterType;
  parameterType = ReflectionUtil.getGenericCollectionTypeOfMethodParameter(theMethod, paramIndex);
}
if (Collection.class.isAssignableFrom(parameterType)) { <--NPE
`

I hope this helps, And apologies in advance for the bad formatting.

BR
/Stefan

PR #1251, opened 2019-03-26
Fix a couple of bugs

fil512:
Fixed a bug in CanonicalSubscription.equals(): email subscriptions
Fixed a bug in SubscriptionMatcherHandler that only handled ids in Patient/1 format
Added support for Dstu2


fil512:
@jamesagnew this is ready for review

PR #1261, opened 2019-04-01
JPA Performance Interceptor Framework

jamesagnew:
This change does a couple of things:

* Interceptors can now be registered to the registry as ThreadLocals, making it much easier to attach an interceptor to an individual thread to trace performance on things, etc. I was worried about the performance implications of doing this, but there is a microbenchmark in this PR showing that it is not an issue.. woohoo
* A new interceptor was added that logs performance stuff for the JPA server

The next step in this train of thought is the following:

* I want to come up with some way of having a handoff so that any ThreadLocal interceptors on HTTP threads get handed off to the SearchCoordinator worker thread executing a search
* This should ultimately make it possible to log and maybe return as HTTP headers details about performance relating to individual queries


hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/interceptor/executor/InterceptorService.java
```diff
@@ -41,22 +39,30 @@
...
 
 public class InterceptorService implements IInterceptorRegistry, IInterceptorBroadcaster {
-	private static final Logger ourLog = LoggerFactory.getLogger(InterceptorService.class);
 	private final List<Object> myInterceptors = new ArrayList<>();
-	private final ListMultimap<Pointcut, BaseInvoker> myInvokers = ArrayListMultimap.create();
+	private final ListMultimap<Pointcut, BaseInvoker> myGlobalInvokers = ArrayListMultimap.create();
 	private final ListMultimap<Pointcut, BaseInvoker> myAnonymousInvokers = ArrayListMultimap.create();
 	private final Object myRegistryMutex = new Object();
 	private String myName;
+	private final ThreadLocal<ListMultimap<Pointcut, BaseInvoker>> myThreadlocalInvokers = new ThreadLocal<>();
```
fil512: I recommend creating a new class that encapsulates ListMultimap<Pointcut, BaseInvoker> and collects into it all the behaviour you need for the way that map is used in this class.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/interceptor/executor/InterceptorService.java
```diff
@@ -87,6 +93,37 @@ public void clearAnonymousHookForUnitTest() {
...
+			return false;
+		}
+		ListMultimap<Pointcut, BaseInvoker> invokers = getThreadLocalInvokerMultimap();
+		scanInterceptorAndAddToInvokerMultimap(theInterceptor, invokers);
+		return !invokers.isEmpty();
+
+	}
+
+	@Override
+	public void unregisterThreadLocalInterceptor(Object theInterceptor) {
```
fil512: If people don't remember to explicitly call this, does it risk introducing a memory leak?  Do you automatically unregister all threadlocals just before the request goes out of scope?  Wouldn't want these to linger in the http threadpool.

hapi-fhir-jpaserver-model/src/test/java/ca/uhn/fhir/jpa/model/interceptor/executor/InterceptorServiceTest.java
```diff
@@ -188,6 +206,113 @@ public void testValidateParamTypesWrongParam() {
...
+	 *    ThreadLocalEnabled=true - Performed 500000 loops in 8383.0ms - 0.017ms / loop
+	 *    ThreadLocalEnabled=false - Performed 500000 loops in 3743.0ms - 0.007ms / loop
+	 *    ThreadLocalEnabled=true - Performed 500000 loops in 6163.0ms - 0.012ms / loop
+	 *    ThreadLocalEnabled=false - Performed 500000 loops in 3487.0ms - 0.007ms / loop
+	 *    ThreadLocalEnabled=true - Performed 1000000 loops in 00:00:12.458 - 0.012ms / loop
+	 *    ThreadLocalEnabled=false - Performed 1000000 loops in 7046.0ms - 0.007ms / loop
+	 * </pre>
+	 */
+	@Test
+	@Ignore
```
fil512: Give `@Ignore` a reason e.g. `@Ignore("Performance Test")`

hapi-fhir-jpaserver-model/src/test/java/ca/uhn/fhir/jpa/model/interceptor/executor/InterceptorServiceTest.java
```diff
@@ -197,6 +322,11 @@ public void before() {
 		ourInvocations.clear();
 	}
 
+	@After
+	public void after() {
+		myInterceptorRegistry.setThreadlocalInvokersEnabled(new InterceptorService().isThreadlocalInvokersEnabled());
+	}
+
 	@Configuration
 	@ComponentScan(basePackages = "ca.uhn.fhir.jpa.model")
```
fil512: You can delete this `@ComponentScan` now since I added a mandatory name for `InterceptorRegistry`

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/interceptor/executor/InterceptorService.java
jamesagnew: It does- but i don't think there's an obvious way of dealing with this fact from within the registry. I'll add some cautionary docs at least..

Once the registry is bumped up to the regular server it'll be easy since we could have a pointcut for a request being done and clear the threadlocals then

hapi-fhir-jpaserver-model/src/test/java/ca/uhn/fhir/jpa/model/interceptor/executor/InterceptorServiceTest.java
jamesagnew: I never knew you could do that!

fil512:
I've reviewed https://github.com/jamesagnew/hapi-fhir/pull/1261/commits/7359b77d9963b00bd4520b8c7433f53016988317

PR #1263, opened 2019-04-03
Expunge testing

fil512:
This patch fixes two bugs.

The first bug is that the expunge operation wasn't bailing once it hit its limit because the return operation was inside a transaction.  The fix was to move it outside.

The second bug is that one case wasn't properly handled: when a resourceId with no version is provided.  This executed the case where only resource type is provided.


fil512:
@jamesagnew please review

PR #1269, opened 2019-04-10
Issue 1268: Fix date range missing STARTS_AFTER and ENDS_BEFORE handling

srdo:
Fixes https://github.com/jamesagnew/hapi-fhir/issues/1268

It looks like DateRangeParam handles "sa" and "eb" fine, the enum values just aren't included in the `addParam` method, so you can't use them. 

The other change here is to make lower bound and upper bound to instant handling more alike. When getting the lower bound as a Date in getLowerBoundAsInstant, and the prefix is "gt" or "sa", we add one unit of the precision to the lower bound. So e.g. if the range is gt19:00:58, and the precision is 1s, the lower bound instant would be 19:00:59.

The upper bound to instant handling was pretty different. When getting the upper bound and the prefix is "lt" or "eb", getUpperBoundToInstant subtracts 1 milliscond instead of using the specified precision. So upper bound for lt19:00:58 would be 19:00:57.999.

Additionally the handling for "eq" on upper bounds is weird. For lower bounds, we just pass the original timestamp through, so if you do "eq19:00:58", getLowerBoundAsInstant would return "19:00:58". For upper bounds, we instead add 1 unit of precision, then subtract a millisecond. If the precision is 1 second, we would therefore get that "eq19:00:58" works out to an upper bound instant of "19:00:57.999".

I've changed the upper bound handling to be like the lower bound handling.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/param/DateRangeParam.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/param/DateRangeParam.java
@@ -167,6 +167,7 @@ public class DateRangeParam implements IQueryParameterAnd<DateParam> {
 			switch (theParsed.getPrefix()) {
 				case GREATERTHAN:
 				case GREATERTHAN_OR_EQUALS:
+				case STARTS_AFTER:
 					if (myLowerBound != null) {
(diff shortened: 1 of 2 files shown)


srdo:
Closing for now, not sure this is a real issue

PR #1275, opened 2019-04-14
Websocket cleanup

fil512:
@jamesagnew please review and merge.  Changes include:

- made the retrier stop logging exceptions for every try.  we're calling retrier because we expect exceptions.
- for ActiveSubscription, separated out the concept of unsubscribe, destroy, and remove to support the different scenarios required by different kinds of subscriptions


fil512:
@jamesagnew this is not ready for review after all.  Still need to make a couple of changes.

fil512:
@jamesagnew please review and merge

jamesagnew:
@fil512 This needs a changelog I think (to mention the configurable websocket path) but other than that looks good. Please go ahead and merge.

PR #1287, opened 2019-04-24
Interceptor framework refactoring

jamesagnew:
* Move Interceptor framework from hapi-fhir-jpaserver-model into hapi-fhir-base and remove all dependency on Spring
* Sync the latest validator and FHIRPath engines from the core library and deal with all kinds of fallout from that
* A couple of bug fixes needed by the EMPI branch in CDR


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseHapiTerminologySvcImpl.java
```diff
@@ -1405,6 +1405,8 @@ public boolean supportsSystem(String theSystem) {
 					targets.add(scrollableResultsIterator.next());
 				}
 
+				scrollableResults.close();
```
fil512: Why not a try with resources here?  Might not close if there's an exception...

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/ResourceProviderFactory.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.function.Supplier;
+
+public class ResourceProviderFactory {
+
+	private List<Supplier<Object>> mySuppliers = new ArrayList<>();
+
+	public void addSupplier(@Nonnull Supplier<Object> theSupplier) {
```
fil512: No-one calls this method.  How do suppliers ever get added?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseHapiTerminologySvcImpl.java
jamesagnew: oh yeah, good call.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/ResourceProviderFactory.java
jamesagnew: Yeah good call- I think I had a use in mind but it never happened. Removing.

Issue #1291, opened 2019-04-26
`Parameters` schematron missing


MFAshby:
**Describe the bug**
Trying to validate a `Parameters` resource using the schematron validator fails due to missing schematron file. 

**To Reproduce**
See minimal reproduction project: https://github.com/MFAshby/parametersbug
The validation fails with an error `Failed to load schematron for resource 'Parameters'`
Note that validation of a different resource (e.g. Observation) works as expected, this bug appears to be restricted to Parameters resources only. 

**Expected behavior**
The Parameters resource should be validated according to the schematron (available here http://hl7.org/fhir/parameters.sch) 

**Screenshots**

**Environment (please complete the following information):**
 - HAPI FHIR Version = 3.7.9
 - OS: [e.g. iOS] = Linux
 - Browser [e.g. chrome, safari] = N/A

**Additional context**

MFAshby:
On further investigation, the file parameters.sch appears to be missing from the STU3 release of FHIR: http://hl7.org/fhir/STU3/parameters.sch returns 404, wheras other resource schematrons are present in STU3 http://hl7.org/fhir/STU3/observation.sch. 

The Parameters schematron is present in version 4 http://hl7.org/fhir/parameters.sch, might it be worth backporting this to 3?

PR #1294, opened 2019-04-29
Fix parameters schematron in r4. Add missing parameters schematron to STU3 resources.

MFAshby:
This schematron was broken/missing, so Parameters resources could not be validated using the Schematron validator as noted in #1291 

Corresponding change requests have been made on hl7.org to address the incorrect & missing files there: 
https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=21153&start=0
https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=21180&start=0


MFAshby:
This is receiving no attention, I guess it's not a problem for anyone else.

PR #1303, opened 2019-05-07
Removing not needed "super." when calling preProcessValues()

basecade:
preProcessValues() is not overridden in JsonParser og XmlParser so calling super.preProcessValues() doesn't make sense and it prevents inherited versions of JsonParser and XmlParser to override preProcessValues().


jamesagnew:
Thanks! Makes sense.

PR #1305, opened 2019-05-07
Renaming package name of test classes

zaewonyx:
Build of jpaserver-model was failing due to incorrect package name for two classes


jamesagnew:
Thanks for the contribution! Merging now.

PR #1310, opened 2019-05-16
Fixing bug in reference aggregation mode validation

basecade:
Correction to checkReference() method so the reference aggregation mode validation can actually fail when using an aggregation mode that is not supported according to the profile/structuredefinition.


jamesagnew:
Hi @basecade - Thanks for the pull request!

Unfortunately the InstanceValidator class is no longer maintained in this repository. Would you be able to submit a pull request against the R5 InstanceValidator (which is now used to validate all versions) in this repo: https://github.com/hapifhir/org.hl7.fhir.core/

PR #1321, opened 2019-05-28
RDF parser

uurl:
This PR is the first version of the RDFParser based on XmlParser.


hapi-fhir-base/pom.xml
```diff
@@ -69,6 +69,12 @@
 			<groupId>org.apache.commons</groupId>
 			<artifactId>commons-lang3</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.jena</groupId>
+			<artifactId>apache-jena-libs</artifactId>
+			<type>pom</type>
+			<version>3.11.0</version>
```
jamesagnew: Can this dependency be marked as `<optional>` (so that people don't need to include it unless they explicitly want to support RDF) and can the version be moved to the root POM in the `<dependencyManagement>` section to match how we handle other deps?

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/RDFParser.java
```diff
@@ -0,0 +1,566 @@
+package ca.uhn.fhir.parser;
```
jamesagnew: One overall comment- I don't see any tests in the PR for this new functionality. Can you please add tests that cover the functionality being added?

jamesagnew:
Hi Raul, this looks great! Thanks for the contribution. It has been merged.

PR #1336, opened 2019-06-08
Issue-1325: Let OS assign random ports to servers during testing, and…

srdo:
… let tests get the port of the running server once the OS has assigned a port. Delete PortUtil and similar utilities.

As discussed in https://github.com/jamesagnew/hapi-fhir/issues/1325, I've updated the tests to let the OS assign ports to test servers, rather than using the PortUtil and various other utilities that were similar.

The method for assigning ports outlined in the issue works everywhere, except for the following two places:

* RestHookWithInterceptorR4Test -  Wants to set a subscription endpoint that won't respond properly. Instead of picking a random port, I changed the test to use the running test server, with an invalid URL path.
* TestSubscriptionConfig - Due to Spring, the tests using this config need to supply an IGenericClient, that apparently isn't used. Instead of creating a real client pointing to a random port, I just made the config supply a mock. The client was previously pointing to a port where nothing was running, so I don't think this should make a difference. 

I've not updated the following modules, as they aren't attached to the build. I'm not sure why they aren't deleted to be honest:

* hapi-fhir-oauth2
* hapi-fhir-structures-dstu
* hapi-fhir-jaxrs-sse

I'll be happy to open a PR to delete these modules if they aren't needed.

Other changes here:
* Some tests were closing servers in `@AfterClass` methods, while also clearing all static fields in a different `@AfterClass`, using the `TestUtil.clearAllStaticFieldsForUnitTest()` method. As JUnit may invoke these methods in any order, this caused servers to leak, due to the field being nulled before the server was stopped. Fixed the instances I found by just combining the methods.
* Jetty is only capable of shutting down gracefully if the StatisticsHandler is added to the server. Made sure this happens. When Jetty shuts down gracefully, a `stop()` call will block until the server is done processing open requests. Without StatisticsHandler, the server just shuts down immediately, which was causing exceptions to be logged in some tests.
* Ensured tests call both `stop` and `destroy` on the Jetty server. As far as I can tell, `stop` doesn't do cleanup (the server could be started again later), and `destroy` should be called to make sure all handlers are properly destroyed once the server is stopped.
* Removed a ComponentScan that was trying to scan the "ca.uhn.fhir.jpa.model.interceptor.executor" package. This package doesn't seem to exist, and the tests run fine without it.

I think it would be good to write e.g. a JUnit rule to wrap server creation/destruction, so test writers won't have to remember to call the right test utility methods, and to properly clean up the server. I haven't done it here, since it's a lot of code to update, but it could be a good idea for new test code.

I've run the build a couple of times on a Linux VM, and I'm pretty much only seeing test failures in the JPA server module, and those failures don't seem related to which port the server is running on. I'll raise separate issue to track fixing the failures I saw.


srdo:
The JettyUtil class in hapi-fhir-base requires that I add jetty as a dependency. If there's a better module for the utility, I'm happy to move it.

jamesagnew:
Super cool. Definitely want to review this with a bit more detail a bit later, but one comment:

> Some tests were closing servers in @AfterClass methods, while also clearing all static fields in a different @AfterClass, using the TestUtil.clearAllStaticFieldsForUnitTest() method

That clearAllStaticFieldsForUnitTest method may well not be needed any more truthfully. It was added a long time ago, when the FhirContext consumed quite a bit of memory and the tests were running out. It probably makes sense to just remove it at this point...

fil512:
I wonder how much removing that method would speed up the build...

srdo:
I'm happy to try removing clearAllStaticFieldsForUnitTest, but as it is used in ~500 files, I'd rather do that in another PR.

hapi-fhir-base/pom.xml
```diff
@@ -85,6 +85,10 @@
 			<groupId>com.google.guava</groupId>
 			<artifactId>guava</artifactId>
 		</dependency>
+        <dependency>
+			<groupId>org.eclipse.jetty</groupId>
+			<artifactId>jetty-server</artifactId>
+		</dependency>
```
jamesagnew: I think we should avoid introducing this dependency- Given that all kinds of things depend on hapi-fhir-base, including Android, OSGI, client, etc., having a server dependency in the core JAR is likely to cause unintended issues.

My vote would be to introduce a new module called `hapi-fhir-test-utils` that hosts JettyTest that the tests depend on.

srdo: Makes sense. This change is in the latest commit for ease of review.

PR #1341, opened 2019-06-09
DeleteConflict hook, SearchableHashMapResourceProvider, and in-memory searching

fil512:
Added DeleteConflict hook.  Also some AnyResource classes that support LiveBundle work.


fil512:
@jamesagnew this PR is ready for review.

One thing I'm not sure about is whether we need to preserve deprecated dao delete method signatures for backwards compatibility.  The current PR changes the method signature without deprecated converters.

fil512:
Here are the change log entries for easy reference:
<action type="change">
  Moved in-memory matcher from Subscription module to SearchParam module and renamed the result type
  from SubscriptionMatchResult to InMemoryMatchResult.
</action>
<action type="add">
  Added some experimental version-independent model classes to ca.uhn.fhir.jpa.model.any.  They permit
  writing code that is version independent.
</action>
<action type="add">
  Added new subclass of HashMapResourceProvider called SearchableHashMapResourceProvider that uses the
  in-memory matcher to search the HashMap (using a full table scan).  This allows rudimentary testing
  without a database.
</action>
<action type="add">
  Added a new interceptor hook called STORAGE_PRESTORAGE_DELETE_CONFLICTS that is invoked when a
  resource delete operation is about to fail due to referential integrity conflicts.
	Hooks have access to the list of resources that have references to the resource being deleted and
  can delete them.  The boolean return value of the hook indicates whether the server should try 
  checking for conflicts again (true means try again). 
</action>

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/DaoRegistry.java
```diff
@@ -26,11 +26,11 @@
 import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import org.apache.commons.lang3.Validate;
 import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.jetbrains.annotations.Nullable;
```
jamesagnew: Should use `javax.annotation.Nullable` for consistency.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/DeleteConflictFinderService.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Service
+class DeleteConflictFinderService {
+	@PersistenceContext(type = PersistenceContextType.TRANSACTION)
+	protected EntityManager myEntityManager;
+
+	List<ResourceLink> findConflicts(ResourceTable theEntity, int maxResults) {
+		TypedQuery<ResourceLink> query = myEntityManager.createQuery("SELECT l FROM ResourceLink l WHERE l.myTargetResourcePid = :target_pid", ResourceLink.class);
+		query.setParameter("target_pid", theEntity.getId());
+		query.setMaxResults(maxResults);
```
jamesagnew: Good call setting maxresults here- This is the kind of thing that's easy to miss and have crazy memory errors...

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/DeleteConflictService.java
```diff
@@ -0,0 +1,116 @@
...
+	@Autowired
+	protected IResourceLinkDao myResourceLinkDao;
+	@Autowired
+	private FhirContext myFhirContext;
+	@Autowired
+	protected IInterceptorBroadcaster myInterceptorBroadcaster;
+
+	public int validateOkToDelete(DeleteConflictList theDeleteConflicts, ResourceTable theEntity, boolean theForValidate) {
+		DeleteConflictList newConflicts = new DeleteConflictList();
+		boolean tryAgain = findAndHandleConflicts(newConflicts, theEntity, theForValidate, MIN_QUERY_RESULT_COUNT);
```
jamesagnew: I'm having trouble understanding this bit.. What's the rationale for passing `1` as the first max and `60` later? Also why is the second count called `MAX_RETRY_COUNT`? I find that name confusing vs `MAX_RETRY_ATTEMPTS`.

fil512: It is a performance optimization.

In most cases, there will be no hooks, and so we only need to check if there is at least one conflict and return that.  Only in the case where there is a hook do we need to go back and get a bigger batch.

I'll rename the constants to try to make the meaning of them clearer.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/DeleteConflictFinderService.java
fil512: You had it in there before, just hard-coded to 1.  I made it configurable so when there is a hook we can go back for larger batches on subsequent retries.

jamesagnew:
That makes sense. Might make sense to add acomment to that effect too.

sent from my phone.

fil512:
I added the explanation

PR #1346, opened 2019-06-15
Issue-1345: Split HAPI FHIR travis build up, so we don't hit the Trav…

srdo:
…is time limit on build jobs

As mentioned in https://github.com/jamesagnew/hapi-fhir/issues/1345, it would be good to run more of the build concurrently on Travis, as we very often hit the job time limit.

Here's an example of how this could look. Jacoco doesn't work with this, as no part of the build runs a full set of tests. It may be possible to merge build results (https://www.eclemma.org/jacoco/trunk/doc/merge-mojo.html), but I didn't look into it too much.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,12 +8,14 @@ dist: trusty
 
 language: java
 jdk:
-    - openjdk11
-   #- oraclejdk9
-
+ - openjdk11
 env:
   global:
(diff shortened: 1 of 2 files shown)


srdo:
Current test failure is due to a flaky test. Passing build https://travis-ci.org/jamesagnew/hapi-fhir/builds/546120115

srdo:
I think this isn't necessary, as I get the impression that HAPI FHIR isn't going to keep using Travis. Closing.

PR #1357, opened 2019-06-21
Subscriptions correctly interpreting the payload

seanmcilvenna:
Currently the subscription framework in HAPI-FHIR does not correctly interpret the meaning of the Subscription.channel.payload property. The payload property is described by the FHIR spec as:

> The mime type to send the payload in - either application/fhir+xml, or application/fhir+json. If the payload is not present, then there is no payload in the notification, just a notification. The mime type "text/plain" may also be used for Email and SMS subscriptions.

Right now, HAPI-FHIR simply sends the raw text value of the Subscription.channel.payload property, rather than using the value of the payload property to determine how (or whether) to serialize the resource that triggered the subscription in the body of the notification (whether that be a rest-hook subscription or an email subscription).

This changes make it so that it HAPI tests the value of the Subscription.channel.payload and serializes the resource that triggered the subscription in the body of the notification, which should align with the FHIR specification's intent.


hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/module/subscriber/ResourceDeliveryMessage.java
```diff
@@ -89,9 +97,15 @@ public void setSubscription(CanonicalSubscription theSubscription) {
 		mySubscription = theSubscription;
 	}
 
-	public void setPayload(FhirContext theCtx, IBaseResource thePayload) {
+	public void setPayload(FhirContext theCtx, IBaseResource thePayload, Boolean isXml) {
```
jamesagnew: `s/isXml/theXml/` for style consistency.

pandapaul: `isXml` makes more sense for a Boolean, but I might be misunderstanding.

seanmcilvenna: This is a matter of preference. This is a boolean, so I prefer "isXml" as does @pandapaul... But, if you require that I change it to get the PR merged, I will :) Let me know @jamesagnew

jamesagnew: Much of the HAPI FHIR codebase uses the code style of:
- Parameters are named `theFoo`
- Fields are named `myFoo`
- Statics are named `ourFoo`
- Constants are named `FOO_CONSTANT`

We aren't strict on this at all, but I do think it makes sense to match the existing style of a class when adding to it.

seanmcilvenna: Will update to theXml in just a sec

jamesagnew:
FYI- This unfortunately seems to have caused a bunch of downstream regressions in other projects that were depending on the serialized resource being in the delivery message (i.e. the one in the queue, in ResourceDeliveryMessage) even if no payload type was specified. I'm going to see if I can tweak this to keep the right behaviour without dropping that functionality.

PR #1368, opened 2019-07-05
Prevent the check bundle.size() == 0 from throwing a NPE.

fil512:
I could have gone with the softer approach of changing the bodies of the existing size() methods to try to ensure they can't return null, but I chose the enforcement approach of changing the return type of IBundleProvider.size() from Integer to int.  I believe most people will assume size() can't return null and so this feels like the more helpful solution in the long run.

### DIFF
--- a/examples/src/main/java/example/PagingPatientProvider.java
+++ b/examples/src/main/java/example/PagingPatientProvider.java
@@ -39,7 +39,7 @@ public class PagingPatientProvider implements IResourceProvider {
       return new IBundleProvider() {
 
          @Override
-         public Integer size() {
+         public int size() {
             return matchingResourceIds.size();
          }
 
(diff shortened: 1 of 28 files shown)


jamesagnew:
I don't know about this change... `null` means "unknown" in this case, it's not the same thing as `0`. Arguably this method should return an Optional, but I don't think the solution is to return 0...

fil512:
People are so used to size() from Collections classes that can never return null.  I think in this case, even a -1 (a la indexOf) would be better than null.

fil512:
I see your point that this could be misleading though--where a user assumes that bundle.size() <= 0 means there were no matches from a search when in fact there were matches but their number is unknown.  What's a better way to determine whether a search returned matches?

jamesagnew:
Yeah, I guess I could see -1 being used... This would have to involve a careful analysis of anywhere we're currently checking for null though.

E.g. the following search includes no `"total"` element because it returns too many results for the search coordinator to know the total when the first page is loaded: http://hapi.fhir.org/baseDstu3/Patient

I'm assuming that with this patch as-is, it would include a misleading `"total": 0`

fil512:
I agree this PR should be rejected.  Would it be possible to implement IBundleProvider.isEmpty() somehow?

fil512:
This is not the right way to solve this.

PR #1373, opened 2019-07-09
[(servermethodbinding)] Fix method binding to handle special search n…

petromykhailysyn:
In ca.uhn.fhir.rest.server.method.SearchMethodBinding#incomingServerRequestMatchesMethod underscore parameters are always considered matched except for the parameters _id and _language. This is handled by the loop:
 
for (String next : theRequest.getParameters().keySet()) {
if (next.startsWith("_") && !SPECIAL_SEARCH_PARAMS.contains(next)) {
methodParamsTemp.add(next);
}
}
 
There is though a bug in this implementation, since if the request parameters are qualified (_id:exact e.g.), then it will be considered matched because SPECIAL_SEARCH_PARAMS.contains(next) is false. This leads to failure since a request with _id:exact parameters will actually match a method with no _id parameter.


petromykhailysyn:
hi @jamesagnew , on my build InterceptorUserDataMapDstu2Test has failed but locally it works. Could you please rerun build ?

dmitchellmim:
Would this also fix `_id:not`? I noticed that on the test server, it returns exactly the opposite of what you'd expect. Example: http://hapi.fhir.org/baseDstu3/Patient?_id:not=1509209

petromykhailysyn:
> Would this also fix `_id:not`? I noticed that on the test server, it returns exactly the opposite of what you'd expect. Example: http://hapi.fhir.org/baseDstu3/Patient?_id:not=1509209

My fix will not fix that issue because it's related to handle modifiers but I fixed registering provider methods issue.

petromykhailysyn:
@jamesagnew please review

PR #1375, opened 2019-07-10
Update identifier example type

nickrobison-usds:
This change updates the example definition for the `Identifier` datatype to use a String type, rather than URI. This should address an issue that we're facing when creating profiles on top of this definition:

`StructureDefinition.snapshot.element[16].example.valueUri,message=URI values cannot have whitespace`

I'm sure there are other changes that I'm missing, happy to expand the PR to address them.

Thanks for the excellent work!


jamesagnew:
Hi Nick,

Thanks for the PR- merging it now.

PR #1376, opened 2019-07-10
Resolve "experimental implementation for storing and expanding large ValuetSets cont'd..."

dmuylwyk:
See: https://github.com/jamesagnew/hapi-fhir/issues/1366


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseHapiTerminologySvcImpl.java
```diff
@@ -184,6 +194,18 @@ private void addCodeIfNotAlreadyAdded(ValueSet.ValueSetExpansionComponent theExp
 		}
 	}
 
+	private void addCodeIfNotAlreadyAdded(IValueSetCodeAccumulator theValueSetCodeAccumulator, Set<String> theAddedCodes, boolean theAdd, AtomicInteger theCodeCounter, String theCodeSystem, String theCode, String theDisplay) {
+		if (isNotBlank(theCode) && theAdd && theAddedCodes.add(theCode)) {
```
jamesagnew: I guess what we store in this map should probably be a composite key really, like `theSystem + " | " + theCode` since technically a VS expansion could contain two different codes where the code is the same but the system is different (this would be weird, but not illegal)

jamesagnew: Man, sometimes the little things make all the difference.. It's so much nicer in GitLab how they colour the code snippets red and cause them to not break over lines.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseHapiTerminologySvcImpl.java
```diff
@@ -184,6 +194,18 @@ private void addCodeIfNotAlreadyAdded(ValueSet.ValueSetExpansionComponent theExp
...
+			theValueSetCodeAccumulator.addCode(theCodeSystem, theCode, theDisplay);
+			theCodeCounter.incrementAndGet();
+		}
+
+		if (!theAdd && theAddedCodes.remove(theCode)) {
+			// FIXME: DM 2019-07-10 - Remove code from accumulator!
+			theCodeCounter.decrementAndGet();
```
jamesagnew: Heh, I forgot about excludes.. I think your instinct is right here, there should be a `removeCode` method too.

dmuylwyk: Added:

```
void excludeCode(String theSystem, String theCode);
```

Renamed `addCode(...)`:

```
void includeCode(String theSystem, String theCode, String theDisplay);
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseHapiTerminologySvcImpl.java
dmuylwyk: Fixed.

src/changes/changes.xml
```diff
@@ -284,6 +284,14 @@
 				profiles via the $snapshot operation, and will automatically generate a snapshot when
 				needed for validation.
 			</action>
+			<action type="fix">
```
jamesagnew: I consider these new features as opposed to bugfixes.

`s/fix/add/` and drop the corrected bit IMO

dmuylwyk: Fixed.

PR #1379, opened 2019-07-12
Expunge hooks

fil512:
- Added a new Pointcut STORAGE_PRESTORAGE_EXPUNGE_EVERYTHING that is called at the start of the expungeEverything operation.
- Added RequestDetails to all Expunge operations.
- Added missing lengths to string columns in tables recently added to the migrator
- Improved the error message when string column length is missing :-)


fil512:
@jamesagnew this PR is ready for your review.

src/changes/changes.xml
```diff
@@ -270,6 +270,14 @@
 				The <![CDATA[<code>_summary</code>]]> element was not always respected when encoding
 				JSON resources.
 			</action>
+			<action type="add">
+				Added a new Pointcut STORAGE_PRESTORAGE_EXPUNGE_EVERYTHING that is called at the start of
+				the expungeEverything operation.
+			</action>
+			<action type="change">
+				Added RequestDetails to all Expunge operations.
```
jamesagnew: I'd leave this second message out- It's more an implementation detail so I don't think it'll be meaningful to anyone reading it

fil512: done

PR #1393, opened 2019-07-25
Expunge resource hook

fil512:
Consolidate your expunge resource with mine.
Also a couple of other small tweaks my cdr branch needs.


fil512:
@jamesagnew this is ready for review and merge

PR #1404, opened 2019-07-30
Issue 1402: disambiguate @Path values

sethrylan:
Add historyForInstance and historyForType to AbstractJaxRsResourceProvider with non-ambiguous @Path values.


jamesagnew:
This is great, thanks for the submission!

PR #1405, opened 2019-07-31
couple of small changes

fil512:
Changed Stopwatch getThroughput param from int to long
Added FhirVersion accessor to AnyListResource


fil512:
@jamesagnew please review

PR #1410, opened 2019-08-01
fix log message

fil512:
When I first added this error message, I accidentally assumed it was a formatted string.  It's fixed here.


fil512:
@dmuylwyk please review & merge

PR #1415, opened 2019-08-06
in-memory id match

fil512:
There was a bug in the in-memory matcher that failed when trying to compare a long id to a string id.

This MR fixes that bug.


fil512:
@jamesagnew small bugfix MR ready for review

PR #1418, opened 2019-08-07
StopWatch API int to long

fil512:
Missed a couple the first time


fil512:
@jamesagnew another small MR to convert StopWatch from int to long

PR #1430, opened 2019-08-16
allow urls in batch transactions to start with /

fil512:


fil512:
@jamesagnew this is ready to review & merge

PR #1433, opened 2019-08-19
Fix deprecated references to versioned TerminologyUploaderProvider.

dmuylwyk:
@jamesagnew, I found no other breaking references to the deprecated classes. Not sure how to write a test to prevent this from happening in the future.


jamesagnew:
This needs a changelog, approved otherwise.

PR #1434, opened 2019-08-19
awaitility version bump

fil512:
running a test compile

### DIFF
--- a/hapi-fhir-test-utilities/pom.xml
+++ b/hapi-fhir-test-utilities/pom.xml
@@ -46,12 +46,6 @@
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
 		</dependency>
-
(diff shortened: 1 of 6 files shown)


hapi-fhir-test-utilities/pom.xml
```diff
@@ -50,6 +50,7 @@
 		<dependency>
 			<groupId>org.awaitility</groupId>
 			<artifactId>awaitility</artifactId>
+			<scope>test</scope>
```
jamesagnew: The test-utilities project is only ever imported itself as a test-scoped import, so I don't think its dependencies need to be test scoped. If anything you want them not to be so that they get imported transitively, also test scoped..

fil512:
mvn install completed successfully.  Merging.

PR #1435, opened 2019-08-19
Handle cascading deletes correctly with circular references

jamesagnew:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/CascadingDeleteInterceptor.java
```diff
@@ -102,6 +102,10 @@ public DeleteConflictOutcome handleDeleteConflicts(DeleteConflictList theConflic
 
 				IFhirResourceDao dao = myDaoRegistry.getResourceDao(nextSource.getResourceType());
 
+//				if (theConflictList.getResourcesMarkedForDeletion().contains(nextSource.toUnqualifiedVersionless())) {
```
fil512: remove comment

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/DeleteConflictList.java
```diff
@@ -30,6 +30,22 @@
 public class DeleteConflictList implements Iterable<DeleteConflict> {
 	private final List<DeleteConflict> myList = new ArrayList<>();
 
+	private List<String> myResourceIdsMarkedForDeletion = new ArrayList<>();
+
+	/**
+	 * Entries should have the form <code>Patient/123</code>
+	 */
+	public List<String> getResourceIdsMarkedForDeletion() {
```
fil512: Rather than leaking the list, it would be better to move the add() function into this class.  That would also allow you to move the toUnqualifiedVersionless calls here for extra safety

src/changes/changes.xml
```diff
@@ -44,6 +44,10 @@
 				SubscriptionDstu2Config incorrectly pointed to a DSTU3 configuration file. This
 				has been corrected.
 			</action>
+			<action type="fix">
+				Cascading deletes now correctly handle circular references. Previously this failed with
+				an HTTP 500 error.
```
fil512: I'm curious how circular references failed before.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/CascadingDeleteInterceptor.java
jamesagnew: done

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/DeleteConflictList.java
jamesagnew: ooh yeah good call, will do this.

src/changes/changes.xml
jamesagnew: Heh, it was a stackoverflow. Res A would cascade-delete Res B, but then that would try to cascade-delete Res A, and so on.

This change is basically just preventing us from cascading a delete if we've already decided to delete that resource.

PR #1480, opened 2019-09-09
%now and in-memory date compare

fil512:
@jamesagnew this is ready for review.

This MR adds support for comparing dates in-memory as well as a new %date constant that can be used as a date searchparam value that evaluates to the current DateTime.


jamesagnew:
I think we probably need to implement this for regular search too- It is confusing enough (but inevitable) that some features work in DB searches but not in in-memory ones. It would be doubly confusing that other features only work the other way.

fil512:
> I think we probably need to implement this for regular search too- It is confusing enough (but inevitable) that some features work in DB searches but not in in-memory ones. It would be doubly confusing that other features only work the other way.

I’m pretty sure it will work for regular search already since the searchparam code is shared.  But I’ll add a test tomorrow to confirm.

fil512:
Confirmed it works with regular search as well.

PR #1489, opened 2019-09-17
Feature/json performance

dmap:
Improve performance of JsonParser when encoding resources to JSON.  There are 3 main improvements:
 * Optimising the search for resources to contain by avoiding repeated traversals of the model
 * Caching the calculation of which composite children need encoding for a given composite
 * Optimise for encoding String primitive data types as these are much more common than the other primitive data types

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
@@ -27,6 +27,7 @@ import ca.uhn.fhir.model.primitive.IdDt;
 import ca.uhn.fhir.rest.api.Constants;
 import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
 import ca.uhn.fhir.util.UrlUtil;
+
 import com.google.common.base.Charsets;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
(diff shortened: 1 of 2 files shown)


jamesagnew:
@dmap this is super cool. Will give a detailed review tomorrow. 

Any sense on how much of an improvement it brings?

dmap:
It all depends on how you measure the improvement, in particular the scale of the improvement depends on the implementation of the `java.io.Writer` you are encoding to (because things like whether you are compressing the output or not have a large impact on throughput).  I'm sure the exact details of the model you are encoding make a difference as well.

Suffice to say that on my test case (encoding large dstu2 DiagnosticReport models) the throughput was approximately doubled.

dmap:
Apologies for the failing builds... I obviously didn't run them all properly before I created the PR.  I'll try to work out why they failed now.  The last commit in the PR isn't a performance change, it was just something I noticed in the code that I'm sure was a bug...  I'm sure you will have a better idea than me as to whether it really it is or not.

dmap:
FYI - there is also a PR I created for an improvement in the gson library that improves the performance: https://github.com/google/gson/pull/1576

dmap:
Sorry, I'm not sure why the builds are failing.  When I run locally the test that is currently failing the build works fine, but I have other unrelated tests failing in my local build.  My local failures I don't think are because of the changes I have made, they seem to be timezone related (I am based in NZ so UTC+12).

jamesagnew:
No worries about the failing build. Travis-CI has become very flaky lately, we really need to migrate off of it. The build for your branch passes locally on my machine.

Merging now. Thanks for the awesome PR!

PR #1496, opened 2019-09-20
Replaced migration task column length constants with literals because…

dmuylwyk:
… they are not necessarily constant across schema versions.

Also increased `val` column length for for TermConceptDesignation and TermValueSetConceptDesignation ([see this post](https://groups.google.com/forum/#!topic/hapi-fhir/OD96QMWSEhc)).


dmuylwyk:

jamesagnew:
Good call!

Issue #1482, opened 2019-09-10
Need to fix duplication of codes in some LOINC Answer Lists.


dmuylwyk:
For example:

All three codes are duplicated in `ValueSet.compose.include` (pre-expansion of ValueSets ensures the expansion will not include duplicates).

https://fhir-dev.loinc.org/ValueSet/LL2451-4/$expand

```json
{
  "resourceType": "ValueSet",
  "status": "active",
  "compose": {
    "include": [
      {
        "system": "http://loinc.org",
        "concept": [
          {
            "code": "LA14100-4",
            "display": "Undetermined"
          },
          {
            "code": "LA6576-8",
            "display": "Positive"
          },
          {
            "code": "LA6577-6",
            "display": "Negative"
          },
          {
            "code": "LA14100-4",
            "display": "Undetermined"
          },
          {
            "code": "LA6576-8",
            "display": "Positive"
          },
          {
            "code": "LA6577-6",
            "display": "Negative"
          }
        ]
      }
    ]
  },
  "expansion": {
    "identifier": "ba8a1dd2-dbb3-4ac9-bb5e-3d48d44cf9d1",
    "timestamp": "2019-09-10T18:26:47+00:00",
    "total": 3,
    "offset": 0,
    "parameter": [
      {
        "name": "offset",
        "valueInteger": 0
      },
      {
        "name": "count",
        "valueInteger": 1000
      }
    ],
    "contains": [
      {
        "system": "http://loinc.org",
        "code": "LA6576-8",
        "display": "Positive"
      },
      {
        "system": "http://loinc.org",
        "code": "LA6577-6",
        "display": "Negative"
      },
      {
        "system": "http://loinc.org",
        "code": "LA14100-4",
        "display": "Undetermined"
      }
    ]
  }
}
```

dmuylwyk:
Working branch: [1482-need-to-fix-duplication-of-codes-in-some-loinc-answer-lists](https://github.com/jamesagnew/hapi-fhir/tree/1482-need-to-fix-duplication-of-codes-in-some-loinc-answer-lists)

dmuylwyk:
Pull request: #1498

dmuylwyk:
Merged.

PR #1498, opened 2019-09-20
Resolve "Need to fix duplication of codes in some LOINC Answer Lists."

dmuylwyk:
See #1482


dmuylwyk:

PR #1505, opened 2019-09-24
Resolve "Need to implement the ancestor and descendant filters for LOINC."

### LINKED ISSUE #1454: Need to implement the ancestor and descendant filters for LOINC.
Need to implement the `ancestor` and `descendant` filters for LOINC.

Reference:

- http://hapi.fhir.org/baseDstu3/CodeSystem?url=http://loinc.org#L69

dmuylwyk:
Closes #1454 

This also includes changes to make sure we call the `updateEntity(...)` method on the correct DAO type when processing inside a transaction.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/DaoConfig.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/DaoConfig.java
@@ -149,21 +149,21 @@ public class DaoConfig {
 	private boolean myFilterParameterEnabled = false;
 	private StoreMetaSourceInformationEnum myStoreMetaSourceInformation = StoreMetaSourceInformationEnum.SOURCE_URI_AND_REQUEST_ID;
 	/**
-	 * EXPERIMENTAL - Do not use in production! Do not change default of {@code false}!
+	 * Do not change default of {@code true}!
 	 */
-	private boolean myPreExpandValueSetsExperimental = false;
+	private boolean myPreExpandValueSets = true;
 	/**
-	 * EXPERIMENTAL - Do not use in production! Do not change default of {@code 0}!
+	 * Do not change default of {@code 0}!
 	 */
-	private int myPreExpandValueSetsDefaultOffsetExperimental = 0;
+	private int myPreExpandValueSetsDefaultOffset = 0;
 	/**
-	 * EXPERIMENTAL - Do not use in production! Do not change default of {@code 1000}!
+	 * Do not change default of {@code 1000}!
 	 */
-	private int myPreExpandValueSetsDefaultCountExperimental = 1000;
+	private int myPreExpandValueSetsDefaultCount = 1000;
 	/**
-	 * EXPERIMENTAL - Do not use in production! Do not change default of {@code 1000}!
+	 * Do not change default of {@code 1000}!
 	 */
-	private int myPreExpandValueSetsMaxCountExperimental = 1000;
+	private int myPreExpandValueSetsMaxCount = 1000;
 
 	/**
 	 * Constructor
@@ -920,7 +920,7 @@ public class DaoConfig {
 	 * <p>
 	 * Default is {@literal true} beginning in HAPI FHIR 2.4, since this
 	 * feature is now specified in the FHIR specification. (Previously it
-	 * was an experimental/rpposed feature)
+	 * was an experimental/proposed feature)
 	 * </p>
 	 *
 	 * @since 1.5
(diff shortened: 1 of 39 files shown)


dmuylwyk:
Moving implementation of ancestor filter with IN operator for LOINC to #1512.

dmuylwyk:

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/DaoConfig.java
```diff
@@ -1720,83 +1692,104 @@ public boolean isStoreRequestId() {
...
+	}
+
+	/**
+	 * <p>
+	 * If set to {@code true}, ValueSets and expansions are stored in terminology tables. This is to facilitate
+	 * optimization of the $expand operation on large ValueSets.
+	 * </p>
+	 * <p>
+	 * The default value for this setting is {@code true}.
+	 * </p>
```
jamesagnew: I always like to include a @Since javadoc tag on new feature switches if I think of it... E.g.

```java
/**
 * ..other docs..
 * @since 4.1.0
 */
```

jamesagnew: I also like to capitalize since in a weird way... wtf

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/dstu3/FhirResourceDaoValueSetDstu3.java
```diff
@@ -310,13 +310,15 @@ private void addFilterIfPresent(String theFilter, ConceptSetComponent include) {
 
 		boolean haveIdentifierParam = theValueSetIdentifier != null && theValueSetIdentifier.isEmpty() == false;
 		ValueSet vs = null;
+		boolean isBuiltInValueSet = false;
 		if (theId != null) {
 			vs = read(theId, theRequestDetails);
 		} else if (haveIdentifierParam) {
 			vs = myValidationSupport.fetchResource(getContext(), ValueSet.class, theValueSetIdentifier.getValue());
```
jamesagnew: Kind of a subtle thing, but I think this is problematic: Technically myValidationSupport could fetch a VS from the database (JpaValidationSupport is on the chain, and it would do this). I think the best thing for here would be to have a built-in field of type DefaultProfileValidationSupport.

Unfortunately you can't currently autowire that kind of bean in.

You could though, do the following:

* Add a new bean of type DefaultProfileValidationSupport to `BaseDstu3Config`
* Autowire that bean into `JpaValidationSupportChainDstu3` instead of creating one there
* Autowire that bean into `FhirResourceDaoValueSetDstu3` and use it instead of myValidationSupport here

jamesagnew: Ps I guess this probably also applies to the equivalent R4/R5 classes

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/DaoConfig.java
dmuylwyk: Done.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/dstu3/FhirResourceDaoValueSetDstu3.java
dmuylwyk: @jamesagnew, should I really use DefaultProfileValidationSupport _instead_ of myValidationSupport here?

If there are cases where myValidationSupport would fetch a ValueSet from the database, we don't want to lose that.

I propose fetching with DefaultProfileValidationSupport first, and then trying with myValidationSupport if the first call doesn't yield a result - and that would be a clearer indication that we're working with a built-in ValueSet.

Does that make sense?

dmuylwyk: @jamesagnew, this is done (as I described). Please review again. Thanks!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/DaoConfig.java
```diff
@@ -1699,6 +1707,8 @@ public boolean isStoreRequestId() {
 	 * <p>
 	 * The default value for this setting is {@code true}.
 	 * </p>
+	 *
+	 * @since 4.1.0
```
jamesagnew: Nice!

PR #1511, opened 2019-09-26
[hapi-fhir-spring-boot] Fix fullUrls in response bundle

plchldr:
The property `hapi.fhir.server.path` is already used to map the servlet, resulting in fullUrls in response bundles like "/fhir/*/Patient/id" currently.
`hapi.fhir.server.url` seems to be the intended property to be used here.


tadgh:
Heya, sorry, This is now 3 years old and has gotten quite stale. I'm going to close this MR for now, if it is still needed, please resubmit, and ping me, I will get it looked at. Apologies.

PR #1516, opened 2019-09-30
Resolve "Need to implement the ancestor filter with IN operator for LOINC"

### LINKED ISSUE #1512: Need to implement the ancestor filter with IN operator for LOINC
Carried over from #1454 and #1505

Still figuring my way around Lucene.

I've got this:
```
theBool.must(theQb.keyword().onField("myParentPids").matching("" + code.getId()).createQuery());
```
where `myParentPids` is a field for a space-delimited string of entity IDs. In this case, I'm getting all of the parent IDs for a given code ID, which Lucene resolves to returning the actual parent codes. This works.

Now I want to get all of the parent IDs for a given *collection* of code IDs. Basically: "select c.myParentPids from TermConcept c where c.myId IN (1, 2, 3,...n);" except I don't care about the parent IDs themselves, Lucene will return the actual parent codes.

I've tried
```
theBool.must(theQb.keyword().onField("myParentPids").matching(values).createQuery());
```
where `values` is a String[] or List<String> but that's not working; I get
```
org.hibernate.search.bridge.BridgeException: Exception while calling bridge#objectToString
	entity class: ca.uhn.fhir.jpa.entity.TermConcept
	entity property path: myParentPids
	field bridge: TwoWayString2FieldBridgeAdaptor [stringBridge=org.hibernate.search.bridge.builtin.StringBridge@4e47b636]
```
I've also tried the string representation of the collection but that's equivalent to putting in a junk ID.

dmuylwyk:
Closes #1512


dmuylwyk:

PR #1518, opened 2019-09-30
Resolve "Need account for property values that exceed 500 characters in length for LOINC."

### LINKED ISSUE #1517: Need account for property values that exceed 500 characters in length for LOINC.
Specifically, we're concerned with the EXTERNAL_COPYRIGHT_NOTICE property value, which can be ~3000 characters.

dmuylwyk:
Closes #1517


dmuylwyk:

PR #1519, opened 2019-10-01
Fix migration task for TermConceptProperty; should be BLOB, not CLOB.…

dmuylwyk:
… Good job!


dmuylwyk:

PR #1526, opened 2019-10-04
Some work on jamesagnew/hapi-fhir/#1297

jelmerterwal:


jamesagnew:
Hi @jelmerterwal - Would you be able to provide some details about the issue that this PR addresses?

jelmerterwal:
Hi @jamesagnew - This is an initial attempt to fix https://github.com/jamesagnew/hapi-fhir/issues/1297. The example in that issue shows that it is not always possible to use equivalent Kotlin code in cases Java code would have worked.

jamesagnew:
@jelmerterwal ah, got it.

Looks like this change is breaking CI so it probably needs some work unfortunately. Do you think it would be realistic to include a Kotlin integration test in the main HAPI build so as to provide some kind of test coverage for these changes? I'm picturing a module like `hapi-fhir-server-kotlin-test` that reproduces the type of setup that is in the ticket you linked.

jelmerterwal:
@jamesagnew okay, think I see why I broke it. Hopefully this will fix the build and will add a test later today or tomorrow

jelmerterwal:
@jamesagnew can you have a look?

Notes;
- we (my colleague and I) removed some overrides since these blocked a working build, tests still work
- we fixed a problem with `CodeableConcept` as well

jamesagnew:
This is looking cool! Why did you scrap the `hapi-fhir-server-kotlin-test` though? I don't know about overloading hapi-fhir-jaxrsserver-example into being both a demo of the alternate JAX-RS server functionality and Kotlin support.. JAX-RS is not well supported as it is, and I'm worried that adding Kotlin into the mix there will further confuse people who are trying to use it as an example.

jelmerterwal:
I see, separated the example from the Kotlin test.

jamesagnew:
Looks great, I've merged it. Thanks for the PR!

PR #1535, opened 2019-10-09
Email subscription improvements

seanmcilvenna:
Adding interceptors for SUBSCRIPTION_BEFORE_EMAIL_DELIVERY and SUBSCRIPTION_AFTER_EMAIL_DELIVERY

Adding enhancements to email subscriptions to accept a more-complex payload mime-type that includes mime-type options of:

- bodynarrative=true|false
- bodytext=BASE64ENCODEDDATA

For example: "application/fhir+xml;bodynarrative=true" or "application/fhir+json;bodytext=dGhpcyBpcyBhIHRlc3QgbWVzc2FnZQ=="

When bodynarrative or bodytext is indicated, the resource XML/JSON is added to the email body as an attachment (a second "part" in the multipart body).

These changes are backwards compatible with previous functionality, with the exception that it _always_ creates a multipart message body, even when there is only a plain/simple/empty string for the message body that doesn't have any attachments.


jamesagnew:
Hey @seanmcilvenna - The last CI faliure was as a result of a bug unrelated to your change that has been fixed. Can you merge master into your branch once more?

PR #1538, opened 2019-10-11
Fix errors when viewing a Resource in testpage-overlay

jiaola:
Added R5 support for testpage-overly. The current fix doesn't work when a Resource on the left side of the page is clicked. It throws error: 

Caused by: java.lang.IllegalStateException: Unknown FHIR version: R5
    at ca.uhn.fhir.to.Controller.populateModelForResource (Controller.java:298)

This pull request will fix that. 

PS> I also had the same fix for R5 in pom.xml and hapi-fhir-testpage-overlay/src/main/java/ca/uhn/fhir/to/BaseController.java but I noticed that they're already fixed :)


jamesagnew:
Thanks for the PR! I have merged.

PR #1539, opened 2019-10-12
refactor "split" packages; create osgi bundle for server framework

bdenton:
combines Java packages that were split between hapi-fhir-base and hapi-fhir-server.
also adds OSGi bundle wiring to hapi-fhir-server


jamesagnew:
This has been merged. Thanks for the PR!

PR #1552, opened 2019-10-21
npe check for binary storage interceptor

fil512:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/binstore/BinaryStorageInterceptor.java
```diff
@@ -169,6 +169,9 @@ private void extractLargeBinaries(ServletRequestDetails theRequestDetails, IBase
 	@SuppressWarnings("unchecked")
 	@Hook(Pointcut.STORAGE_PRECOMMIT_RESOURCE_CREATED)
 	public void storeLargeBinariesBeforeCreatePersistence(ServletRequestDetails theRequestDetails, IBaseResource theResource, Pointcut thePoincut) throws IOException {
+		if (theRequestDetails == null) {
```
jamesagnew: I don't think this alone will cut it, there is also use of RequestDetails in the methods above.

Note you could probably easily replicate this in a unit test by copying the method `testCreateAndRetrieveBinary_ServerAssignedId_ExternalizedBinary` and passing null in for the RequestDetails instead of `mySrd` (which is a mock RequestDetails)

PR #1563, opened 2019-10-24
in-memory _source search

fil512:


hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/RestSearchParameterTypeEnum.java
```diff
@@ -89,13 +89,18 @@
...
 	 * Code Value: <b>number</b>
 	 *
 	 * Search parameter SHALL be a number (a whole number, or a decimal).
 	 */
 	SPECIAL("special", "http://hl7.org/fhir/search-param-type"),
 
+	/**
+	 * _source parameter
+	 */
+	SOURCE("string", "http://hl7.org/fhir/search-param-type"),
```
jamesagnew: The `_source` parameter is defined in the FHIR spec as being a token. You don't need to define anything new for it, just use it as a token.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/SearchBuilder.java
```diff
@@ -861,19 +861,10 @@ private Predicate addPredicateSource(List<? extends IQueryParameterType> theList
...
-			if (lastHashValueIndex == -1) {
-				sourceUri = nextParamValue;
-				requestId = null;
-			} else {
-				sourceUri = nextParamValue.substring(0, lastHashValueIndex);
-				requestId = nextParamValue.substring(lastHashValueIndex + 1);
-			}
-			requestId = left(requestId, Constants.REQUEST_ID_LENGTH);
-
+			// FIXME KHS this works, but is it right?
```
jamesagnew: Well... You are gonna need to parse the token value into the system and transaction id parts.. but I don't think it needs a new class for that. Have a look in `SearchBuilder`, specifically searching for `Constants.PARAM_SOURCE` to see how the regular DB search deals with this param. You could even refactor that parsing bit out of `addPredicateSource(...)` and into its own utility method so you could share.

hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/RestSearchParameterTypeEnum.java
fil512: fixed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/SearchBuilder.java
fil512: fixed

PR #1574, opened 2019-10-31
Composition-wise construction of resource providers

ttntrifork:
Through easy-to-use default interface methods. 

This PR supersedes #1248 (due to different source branch). 

The primary difference is that there are no longer changes to ReflectionUtil. 
Instead the changes are restricted to RestfulServer#findResourceMethods and RestfulServer#removeResourceMethods.


ttntrifork:
> [Build #20191031.1 had test failures](https://github.com/jamesagnew/hapi-fhir/pull/1574/checks?check_run_id=282986692)

Strange.. the tests do not fail when I run them on my machine

ttntrifork:
Superseeded by #1692

PR #1579, opened 2019-11-05
RestfulServer now picks of default methods

jvitrifork:
Adds the capability of picking of default interface methods in resource providers. This relates to https://github.com/jamesagnew/hapi-fhir/pull/1574 - but cherry picks the essense, in that the option of using default interface methods is now not part of the actual PR. This just enables the feature without explicitly adding any reference implementations - that part is now up to the developer him/herself. An example is made in the unit test


jkiddo:
Documentation:
The Restful server implementation can now pick up listed methods implemented in default (interface) methods. This means that if eg. a common custom operation should be present in multiple Provider classes, this can now be expressed with the use of eg. default methods instead of using inheritance. This means that the following example:
```
	interface SomeInterface {
		@Operation(
			name = "$fancyOmniPresentMethod",
			idempotent = true}
		)
		default Parameters fancyOmniPresentMethod(RequestDetails theRequestDetails,
			@OperationParam(name="code") TokenAndListParam theEnd) {
			return isCodeFancy(theEnd);
		}

                boolean isCodeFancy(TokenAndListParam theEnd);

	}
```
- can be listed as implemented by any resource provider and thereby having the custom operation generated in the CapabilityStatement automatically by the existing logic. Having it as a default interface also makes it easily possible to cherry pick which resource provider that should implement it and which shouldn't with minimal effort.

tadgh:
@jvitrifork @jkiddo Doing some spring cleaning of our pull requests over here. Is this one still valid? If so can it be resubmitted against current master? If it is, ping me when done and I can ensure it gets in. Going to close for now as it is 4 years old

PR #1586, opened 2019-11-09
Ks channel type

fil512:
Add a java type parameter to createChannel method so Kafka can support more than one kind of consumer


hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/module/channel/ISubscribableChannelFactory.java
```diff
@@ -23,7 +23,7 @@
 import org.springframework.messaging.SubscribableChannel;
 
 public interface ISubscribableChannelFactory {
-	SubscribableChannel createSubscribableChannel(String theChannelName, int theConcurrentConsumers);
+	SubscribableChannel createSubscribableChannel(String theChannelName, Class theMessageType, int theConcurrentConsumers);
```
jamesagnew: Ooh I like this, I wonder if we can do something to enforce this in non-kafka channels too at some point (like, just for catching bugs)

PR #1600, opened 2019-11-22
fix circular reference include everything

fil512:


jamesagnew:
@fil512 did a changelog get added for this?

PR #1607, opened 2019-11-26
Fix gettimezone performanceissue

abrsystematic:
TimeZone.getTimeZone() is static synchronized and can take significant time for "custom" time zone ids like "GMT+01:00" - see CustomID on https://docs.oracle.com/javase/7/docs/api/java/util/TimeZone.html. This means that it is a performance bottleneck on systems with high number of transactions. Fixed such that TimeZone is cached in BaseDateTimeDt.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/BaseDateTimeDt.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/BaseDateTimeDt.java
@@ -22,7 +22,11 @@ package ca.uhn.fhir.model.primitive;
 
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
-import java.util.*;
+import java.util.Calendar;
+import java.util.Date;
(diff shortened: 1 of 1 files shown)


abrsystematic:
The CustomID format is parsed via java.util.TimeZone#parseCustomTimeZone where the result is not chached by TimeZone.getTimeZone()

abrsystematic:
@jamesagnew Do you have time to look at this small PR?

abrsystematic:
@jamesagnew We would like to have this merged. We have to fork Hapi FHIR to build the version of Hapi FHIR we use because of this issue.

jamesagnew (approved):

jamesagnew:
Apologies for the slow reply- I didn't merge this only because I kept thinking about how we could fix this also for the other versions of FHIR in an organized way but never actually came to a conclusion. I'm going to merge.

abrsystematic:
Understandably :) Thanks.

PR #1608, opened 2019-11-27
Fixing so it is possible to search by versioned reference

basecade:
Fixed so it is possible to search by a versioned reference
https://groups.google.com/forum/#!forum/hapi-fhir


jamesagnew:
@basecade - Thanks for the PR. Two comments:

* The test failure in CI was an intermittent failure (unrelated to your change) that has been addressed in master. Could you merge master into your branch?
* Would it be possible to add a unit test that demonstrates this change?

PR #1611, opened 2019-11-27
Missing mandatory field in loinc.xml. The codesystem will fail the Ha…

basecade:
Missing mandatory field in loinc.xml. The codesystem will fail the Hapi validation if this is not set.

https://groups.google.com/forum/#!topic/hapi-fhir/Z2p5Uo1ds80


jamesagnew:
Thanks @basecade - FYI I have notified the LOINC team just to confirm that they're good with these changes.

PR #1637, opened 2019-12-17
Add index on CONCEPT_PID on TRM_CONCEPT_PROPERTY

jasmdk:
Index to optimize a full-reindex of lucene index which can be needed when running in a docker container where the lucene files are discarded on redeploy.
When we have a terminology server with both SNOMED and LOINC imported, then a full rebuild of the lucene index takes 54 hours while database CPU is running at 100%, but this index brings execution time down to 1.5 hours with only 15% CPU usage on the database server.
I acknowledge that using file based lucene index files are inappropriate when running in a docker container and we are working on setting up an ElasticSearch cluster as backend, but that will also require to build the index again from scratch.


jamesagnew:
Thanks @jasmdk ! This has been merged.

PR #1646, opened 2019-12-23
Resolve "Implement DateRangeParam.isDateWithinRange(Date theDate)."

dmuylwyk:
Closes #1645

Issue #1645, opened 2019-12-23
Implement DateRangeParam.isDateWithinRange(Date theDate).


dmuylwyk:
The JPA server translates DateRangeParam into SQL. Great! However, I am implementing some hybrid providers, and I'd like to easily match dates against DateRangeParam. I'm going to implement:

`ca.uhn.fhir.rest.param.DateRangeParam`
*  `public boolean isDateWithinRange(Date theDate)`

dmuylwyk:
Working branch: [1645-implement-daterangeparam-isdatewithinrange](https://github.com/jamesagnew/hapi-fhir/tree/1645-implement-daterangeparam-isdatewithinrange)

dmuylwyk:
Pull request: https://github.com/jamesagnew/hapi-fhir/pull/1646

PR #1653, opened 2019-12-30
Resolve "DateRangeParam.isDateWithinRange(Date theDate) needs to account for intervals."

### LINKED ISSUE #1652: DateRangeParam.isDateWithinRange(Date theDate) needs to account for intervals.
See #1645 / #1646

We're matching exactly but need to account for intervals such that

`[parameter]=eq2019-12-30`

would match

`2019-12-30T18:28:17.500-05:00`

See *Other Notes* section of spec here: https://www.hl7.org/fhir/search.html#date

dmuylwyk:
Closes #1652


dmuylwyk:
This and the previous related changes in #1645 / #1646 are unnecessary. Rolling them back.

Doing this via commit instead of a revert to preserve the fixed typos. : )

dmuylwyk:
@jamesagnew, please review. Thanks.

PR #1655, opened 2020-01-02
Fixed JSON generation for profiled resource.

mykhsystematic:
When max cardinality in profile is changed to 1 output JSON should still contain an array. http://hl7.org/fhir/profiling.html#cardinality

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
@@ -24,6 +24,8 @@ import ca.uhn.fhir.context.*;
 import ca.uhn.fhir.context.BaseRuntimeElementDefinition.ChildTypeEnum;
 import ca.uhn.fhir.model.api.*;
 import ca.uhn.fhir.model.api.annotation.Child;
+import ca.uhn.fhir.model.api.annotation.DatatypeDef;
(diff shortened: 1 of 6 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
```diff
@@ -581,6 +587,24 @@ private void encodeCompositeElementChildrenToStreamWriter(RuntimeResourceDefinit
 		}
 	}
 
+	private boolean isMultipleCardinality(int maxCardinality) {
+		return maxCardinality > 1 || maxCardinality == Child.MAX_UNLIMITED;
+	}
+
+	private Map<String, Child> getParentProfileElementMetadata(IBase theElement) {
```
jamesagnew: This seems like it would have pretty negative performance implications, as it introduces several reflection calls to every composite element we serialize.

Could this check be moved into the reflection checks that happen when the model is initially scanned? I.e. in the constructor to `BaseRuntimeChildDatatypeDefinition` or something like that?

mykhsystematic: Thank you for advise. Please have a look to commit 8961a0a036cad5cf7f46127da716a5b68abe6229 .

jamesagnew:
Nice work, this looks great!

PR #1656, opened 2020-01-02
Changed BaseresourceReturningMethodBinding to accept IBaseBundle as a…

j-chan:
Changed BaseResourceReturningMethodBinding to accept IBaseBundle as a bundle resource.

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
@@ -83,7 +83,7 @@ public abstract class BaseResourceReturningMethodBinding extends BaseMethodBindi
 
 		} else if (IBaseResource.class.isAssignableFrom(methodReturnType)) {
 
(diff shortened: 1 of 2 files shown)


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
```diff
@@ -81,7 +82,8 @@ public BaseResourceReturningMethodBinding(Class<?> theReturnResourceType, Method
 			}
 
 		} else if (IBaseResource.class.isAssignableFrom(methodReturnType)) {
-			if (Modifier.isAbstract(methodReturnType.getModifiers()) == false && theContext.getResourceDefinition((Class<? extends IBaseResource>) methodReturnType).isBundle()) {
+
+			if ((Modifier.isAbstract(methodReturnType.getModifiers()) == false && theContext.getResourceDefinition((Class<? extends IBaseResource>) methodReturnType).isBundle()) || IBaseBundle.class.isAssignableFrom(methodReturnType)) {
```
jamesagnew: You could actually only test for IBaseBundle (ie just keep the one you added and drop the other one). That should cover all cases actually. I don't think the IBaseBundle resource existed when this code was written..

j-chan: updated

PR #1664, opened 2020-01-10
Fixing bug in BaseParser that expects all resources to have an "extension"

seanmcilvenna:
I'm trying to import data into my new 4.1 installation and I've hit a NullPointerException error.
the exception is coming from BaseParser.shouldBeEncoded():1208
this.myDef is null
shouldBeEncoded() is being called by JsonParser.isEncodeExtension()
BaseRuntimeChildDefinition childDef = definition.getChildByName("extension");
CompositeChildElement c = new CompositeChildElement(this, theParent, childDef, theEncodeContext);
retVal = c.shouldBeEncoded(theContainedResource);
it is hitting this exception when dealing with a contained Binary resource
the problem, I believe, is that the Binary resource is derived from "Resource", not by "DomainResource". Resource does not have "extension"
So, definition.getChildByName("extension") is returning null
and childDef gets passed as myDef, and there's no null-checking on it in shouldBeEncoded()

This PR is compatible with the v4.1 tag. If possible, please update the 4.1 release with this patch fix so that I can use it in my 4.1 environment.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
@@ -1198,14 +1198,15 @@ public abstract class BaseParser implements IParser {
 
 		public boolean shouldBeEncoded(boolean theContainedResource) {
 			boolean retVal = true;
+			String elementName = myDef != null ? myDef.getElementName() : "";
 			if (myEncodeElements != null) {
 				retVal = checkIfParentShouldBeEncodedAndBuildPath();
 			}
 			if (retVal && myDontEncodeElements != null) {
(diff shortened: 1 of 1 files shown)


jamesagnew:
Would it be possible to add a unit test that demonstrates this fix?

seanmcilvenna:
@jamesagnew Point me to where and I'd be happy to. Not ssure where this type of test should get added.

jamesagnew:
@seanmcilvenna JsonParserR4Test might be a good candidate

tadgh:
Heya @seanmcilvenna we are doing some spring cleaning over here and I see this PR is 2 years old. I am going to close it for now as our codebase has heavily changed, but if it is still needed, please feel free to re-implement against master, and ping me on the pull request. I can try to ensure it gets through.

PR #1666, opened 2020-01-13
Added TLS parameters to Java Mail Sender

fil512:


jamesagnew:
@fil512 does this need a changelog?

PR #1673, opened 2020-01-16
[dev] Integration of a Jackson serializer and deserializer

bogdansolga:
An initial attempt to integrate a [Jackson](https://github.com/FasterXML/jackson) based serializer and deserializer, as an alternative to the current [Gson](https://github.com/google/gson) based serializer and deserializer.

A few notes and details about the integrated classes:
- the main serializer is implemented in the `JacksonSerializer` class, which extends the `StdSerializer` abstract class from the Jackson library. It uses an internal structure similar to the `GsonWriter`:
  - an `enum` of the used block types
  - a `LinkedList` for keeping the stack of blocks
  - a  `Writer` object for writing the actual objects
- the serializer is used through the `JacksonWriter` class, which extends the HAPI `JsonLikeWriter` abstract class and implements the abstract methods from it
- the `JacksonWriter` is used through a `JacksonStructure` class, which _for now_ uses an internal structure similar with the `GsonStructure` class. I have copy & pasted the content of the `GsonStructure` class just to see if the `JacksonSerializer` starts to write the serialized objects.

The current status - when attempting to serialize a FHIR resource, the serialization operation fails with a `JSONException` which says "Cannot start an object, expecting field name". I have tried to (first of all) understand why the issue happens and then to fix it, but to no avail.

Any feedback is welcome, both on why the exception occurs and if the integration approach is correct.


bogdansolga:
We have fixed the serializing issues and we have done some further cleanups.

With the current codebase, the serializing of the FHIR resources is working properly. I have done some initial benchmarking and comparisons with the `GsonWriter` and the results seem promising - the performance of the `JacksonWriter` are ~25-30% better than the performance of the `GsonWriter`, for a few `Bundle` resources which contain 400 and 800 `Observation` entries.

The performance of the `JacksonWriter` can be further improved, first of all by replacing the Gson classes from the `JacksonStructure` with their corresponding Jackson classes. I will further investigate that.

@jamesagnew - any feedback is more than welcome.

bdenton:
As the original author of the JsonLikeXxxx stuff, I am pleased to see someone extending it to use other Json implementations like Jackson (which would have been my first choice of implementations except HAPI was already using Gson because of some sort of Android considerations).. Just a couple of comments...

1) Why bother with the JsoNSerializer. None of its capabilities are being used. I would suggest just going directly with the JsonGenerator and use the other "write" methods like writeNumber()

2) On the JsonLikeStructure side, I agree that the Gson stuff should all go.. Just use ObjectMapper.readTree() to deserialize and just walk the JsonNode structures

cheers,
Bill Denton

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/json/jackson/JacksonSerializer.java
```diff
@@ -0,0 +1,75 @@
...
+public class JacksonSerializer extends StdSerializer<Object> {
+
+    private JsonGenerator jsonGenerator;
+
+    public JacksonSerializer(Writer writer) {
+        super(Object.class);
+        try {
+            this.jsonGenerator = new JsonFactory().createGenerator(writer);
+        } catch (IOException e) {
+            e.printStackTrace();
```
jamesagnew: This is dangerous- Should probably rethrow this as an unchecked exception (ConfigurationException?) if it's not possible to jut throw it directly from here.

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/json/jackson/JacksonStructure.java
```diff
@@ -0,0 +1,347 @@
...
+                        pbr.unread(nextInt);
+                        break;
+                    }
+                    throw new DataFormatException("Content does not appear to be FHIR JSON, first non-whitespace character was: '" + (char)nextInt + "' (must be '{' or '[')");
+                }
+                throw new DataFormatException("Content does not appear to be FHIR JSON, first non-whitespace character was: '" + (char)nextInt + "' (must be '{')");
+            }
+
+            //TODO remove the Gson references
+            Gson gson = new GsonBuilder().disableHtmlEscaping().create();
```
jamesagnew: I'm not sure I get what this is doing.. Why do we need Gson if we're using Jackson?

bogdansolga: @jamesagnew - we don't need Gson, I intend to remove it ASAP. I have initially included it for two reasons:
1. to see if I can create the initial structure of the Jackson parser
2. because I didn't know what are the corresponding classes from the Jackson library

Now that the parser is working properly, I will replace all the `Gson` classes with their corresponding classes from Jackson.

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/json/jackson/JacksonSerializer.java
bogdansolga: @jamesagnew - understood, I will throw a `ConfigurationException` from there.

bogdansolga:
> As the original author of the JsonLikeXxxx stuff, I am pleased to see someone extending it to use other Json implementations like Jackson (which would have been my first choice of implementations except HAPI was already using Gson because of some sort of Android considerations).. Just a couple of comments...
> 
> 1. Why bother with the JsoNSerializer. None of its capabilities are being used. I would suggest just going directly with the JsonGenerator and use the other "write" methods like writeNumber()
> 2. On the JsonLikeStructure side, I agree that the Gson stuff should all go.. Just use ObjectMapper.readTree() to deserialize and just walk the JsonNode structures
> 
> cheers,
> Bill Denton

Thank you for your comments and recommendations, @bdenton.
For 1 - I will try to no longer use the `JsonSerializer` and just use the `JsonGenerator`
For 2 - as I have already commented, my intent was from the beginning to not use `Gson` at all. Now that I have seen that the parser is working, I will clean it from the code ASAP. Thank you for the references to the `ObjectMapper.readTree()`, I'll proceed from there.

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/json/jackson/JacksonStructure.java
bogdansolga: @jamesagnew - I have replaced the Gson references with Jackson only classes. I will further test to see if all is well.

Please have a look and tell me if you see any easy to spot issues.

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/json/jackson/JacksonSerializer.java
bogdansolga: @jamesagnew  - done.

bogdansolga:
> As the original author of the JsonLikeXxxx stuff, I am pleased to see someone extending it to use other Json implementations like Jackson (which would have been my first choice of implementations except HAPI was already using Gson because of some sort of Android considerations).. Just a couple of comments...
> 
> 1. Why bother with the JsoNSerializer. None of its capabilities are being used. I would suggest just going directly with the JsonGenerator and use the other "write" methods like writeNumber()
> 2. On the JsonLikeStructure side, I agree that the Gson stuff should all go.. Just use ObjectMapper.readTree() to deserialize and just walk the JsonNode structures
> 
> cheers,
> Bill Denton

@bdenton - I have integrated the changes recommended by you. When you'll have some time... please review them.

bogdansolga:
@jamesagnew and @bdenton - I have finalised the parser integration, as far as I understand the framework and as I have tested.

Please let me know your feedback and if you see any easy to spot issue.

jamesagnew:
Hi @bogdansolga - FWIW my stress tests show an improvement stemming from this change. It's not actually a huge difference, but it's not nothing either.

Before this change:

```
[StressTestParserTest.java:62] Serialized JSON Bundle with 1167 entries 199 times - 119ms/pass
```

After this change:

```
[StressTestParserTest.java:62] Serialized JSON Bundle with 1167 entries 199 times - 106ms/pass
```

I do still think the switch to Jackson would be worthwhile- It would be great to be able to take advantage of the better pretty-printing, line number capturing, and other features that Jackson brings.. And a performance boost doesn't hurt, even if it's not a massive gain.

I'm curious how the corresponding switch to Jackson for parsing will affect performance on that side. Do you think that will be much effort based on the experience so far?

jamesagnew:
Incidentally- one thing I do notice is that pretty-printing is broken with this change.

bogdansolga:
> Incidentally- one thing I do notice is that pretty-printing is broken with this change.

I have removed the pretty-printing (for now), as I wasn't sure if that will affect anything related to the parser usage. I can re-enable it anytime, it's just one line of code.

bogdansolga:
> And a performance boost doesn't hurt, even if it's not a massive gain.
I think the performance might be further improved, especially by someone who knows the HAPI framework and the serializing and deserializing logic.
 
> I'm curious how the corresponding switch to Jackson for parsing will affect performance on that side. Do you think that will be much effort based on the experience so far?
I'm not sure I understand your question. If by 'parsing' you are referring to the JSON --> FHIR resource deserializing - that is already working. In terms of the deserializing performance difference (between the Gson-based and the Jackson-based processing) - I haven't done any measurements, but I presume it is also improved.
Please let me know if my understanding of your question is incorrect.

jamesagnew:
@bogdansolga 

Maybe I'm misunderstanding? In JsonParser I only see changes to the [createJsonWriter](https://github.com/jamesagnew/hapi-fhir/pull/1673/files#diff-ec3e8375f2700726da976df0c89f0760L150) method, but I don't see any corresponding change in [the parse method](https://github.com/jamesagnew/hapi-fhir/pull/1673/files#diff-ec3e8375f2700726da976df0c89f0760L175) - It's still creating an instance of GsonStructure. Doesn't that mean it's still using Gson?

bogdansolga:
> @bogdansolga
> 
> Maybe I'm misunderstanding? In JsonParser I only see changes to the [createJsonWriter](https://github.com/jamesagnew/hapi-fhir/pull/1673/files#diff-ec3e8375f2700726da976df0c89f0760L150) method, but I don't see any corresponding change in [the parse method](https://github.com/jamesagnew/hapi-fhir/pull/1673/files#diff-ec3e8375f2700726da976df0c89f0760L175) - It's still creating an instance of GsonStructure. Doesn't that mean it's still using Gson?

Thank you for pointing out an issue - I forgot to commit the updated `JsonParser` file. I have commited and pushed it now, hopefully all is well now.

jamesagnew:
Hmm.. Ok, with the serializer integrated I have run performance tests. In this case, for whatever reason the performance is actually a bit worse with Jackson.

```
	 * Pre #1673
	 * 15:22:30.758 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 1700 passes - 12ms / pass - 79.3 / second
	 * 15:22:31.968 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 1800 passes - 12ms / pass - 79.5 / second
	 * 15:22:33.223 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 1900 passes - 12ms / pass - 79.5 / second
	 * 15:22:34.459 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 2000 passes - 12ms / pass - 79.6 / second
	 * 15:22:35.696 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 2100 passes - 12ms / pass - 79.7 / second
	 * 15:22:36.983 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 2200 passes - 12ms / pass - 79.6 / second
	 * 15:22:38.203 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 2300 passes - 12ms / pass - 79.7 / second
	 * 15:22:39.456 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 2400 passes - 12ms / pass - 79.7 / second
	 * 15:22:40.699 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 2500 passes - 12ms / pass - 79.7 / second
	 * 15:22:42.135 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:638] - Parsed 2600 passes - 12ms / pass - 79.3 / second

	 * Post #1673
	 * 15:25:25.247 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:646] - Parsed 1800 passes - 14ms / pass - 70.5 / second
	 * 15:25:27.061 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:646] - Parsed 1900 passes - 14ms / pass - 69.5 / second
	 * 15:25:28.644 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:646] - Parsed 2000 passes - 14ms / pass - 69.1 / second
	 * 15:25:29.942 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:646] - Parsed 2100 passes - 14ms / pass - 69.5 / second
	 * 15:25:31.284 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:646] - Parsed 2200 passes - 14ms / pass - 69.7 / second
	 * 15:25:34.249 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:646] - Parsed 2300 passes - 15ms / pass - 66.6 / second
	 * 15:25:36.378 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:646] - Parsed 2400 passes - 15ms / pass - 65.5 / second
	 * 15:25:37.702 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:646] - Parsed 2500 passes - 15ms / pass - 65.8 / second
```

I notice that I can get the performance back up to par with the old code (maybe even a bit faster) by getting rid of the two LinkedHashMap instances entirely from JacksonStructure. They seem like they are intended to be a cache to improve performance, but linkedhashmaps are pretty slow. Just not caching the `get()` methods seems to be faster. (@bdenton curious if you have any thoughts)

```
15:37:01.416 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:655] - Parsed 2000 passes - 11ms / pass - 85.1 / second
15:37:02.630 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:655] - Parsed 2100 passes - 11ms / pass - 85.0 / second
15:37:03.815 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:655] - Parsed 2200 passes - 11ms / pass - 85.0 / second
15:37:05.028 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:655] - Parsed 2300 passes - 11ms / pass - 84.9 / second
15:37:06.231 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:655] - Parsed 2400 passes - 11ms / pass - 84.8 / second
15:37:07.446 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:655] - Parsed 2500 passes - 11ms / pass - 84.7 / second
15:37:08.656 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:655] - Parsed 2600 passes - 11ms / pass - 84.6 / second
15:37:09.977 [main] INFO  ca.uhn.fhir.parser.JsonParserR4Test [JsonParserR4Test.java:655] - Parsed 2700 passes - 11ms / pass - 84.2 / second
```

bogdansolga:
> I notice that I can get the performance back up to par with the old code (maybe even a bit faster) by getting rid of the two LinkedHashMap instances entirely from JacksonStructure. They seem like they are intended to be a cache to improve performance, but linkedhashmaps are pretty slow. Just not caching the `get()` methods seems to be faster. (@bdenton curious if you have any thoughts)

I can either:
- remove the usage of the `LinkedHashMap`
- replace it with a `Stack`

if that will further improve the performance. 

@bdenton - I'm looking forward for your thoughts.

bdenton:
I agree that the LinkedhashMaps are of questionable value.. the intent was
to avoid recreating the JsonLikeValue instances when an array/object member
was accessed more than once. In looking more closely at the details of
JsonParser, it would seem that accessing an array/object member more than
once is quite rare. like it never happens? That being the case, I vote for
just removing that caching functionality.

Overall, I would be surprised if there is going to be much difference in
Gson vs Jackson vs javax.json vs anything else as long as we stay away from
any of the fancy data type/class mapping things most of the libraries have.
There are only so many ways to do the simple processing to go to/from what
is basically a Map/List/scalar structure and they all mostly share similar
code paths

Finally, before jumping on board the Jackson train, we should find out what
Android users think about Jackson since their voice was what prompted the
move from javax.json to Gson. I know that Jackson can be somewhat heavy
from a JAR size viewpoint but, if just simple functions are used, its
memory footprint is not terrible.

cheers, Bill

bogdansolga:
I have removed the `LinkedList` from the `JacksonWriter` class.

@bdenton - regarding the performance of the various JSON solutions (Gson, Jackson, javax.json): at a high level, I agree they are doing pretty much the same processing. However, that processing yields in different results in terms of performance, especially when the performance is evaluated from small to big payloads.

I have made a search for some JSON libraries performance benchmarks and found [these](https://interviewbubble.com/performance-comparison-of-json-libraries-jackson-vs-gson-vs-fastjson-vs-json-simple-vs-jsonp/) [two](https://blog.overops.com/the-ultimate-json-library-json-simple-vs-gson-vs-jackson-vs-json/) articles. The 2nd one is more recent, and [the referenced Github page](https://github.com/fabienrenaud/java-json-benchmark) contains more details on the tested libraries. Based on their benchmarks, the '[dsl-json](https://github.com/ngs-doo/dsl-json)' library seems to be the fastest. Given the overall differences (compared to the Gson and Jackson libraries), I would give it a try as an alternate (3rd) parser.

@jamesagnew and @bdenton - any thoughts on [dsl-json](https://github.com/ngs-doo/dsl-json)?

jamesagnew:
@bogdansolga I don't think dsl-json is an option here- I remember investigating it once when I saw a benchmark that showed it being very fast. It's a specialized POJO binding framework that relies on its own annotations being present on model objects so it would probably require a significant rewrite of the model classes themselves. Not something I am up to taking on in the short term.

I think this branch is about ready to merge though. I'm going to do some final testing and see about getting it in so that we have lots of time to let it be tested in the 4.3.0 cycle.

bdenton:
If we are all sure that replacing the GSON parser/serializer with Jackson
is the thing to do, then I'm all for it. Personally, I would be concerned
about the footprint of Jackson versus GSON. I seem to recall that it was a
concern of the Android developers.

Once that merge is complete, I will port my base class changes and re-push
PR #1688 (the BSON stuff)

cheers,
Bill Denton

jamesagnew:
These changes were pulled into #1733 and merged. Thanks for the copntribution @bogdansolga , this is awesome stuff!

Closing this MR now.

Issue #1674, opened 2020-01-16
remove transient fields from migration task hashCode() and equals()


fil512:
e.g. We need to remove both `myIsOkayIfNeitherColumnExists` and 
`myDeleteTargetColumnFirstIfBothExist` from `RenameColumnTask.hashCode()` and `RenameColumnTask.equals()`.

Only those fields directly pertinent to what's being migrated should be included in the hashCode.  Flyway uses this hashCode to validate the migration hasn't been "tampered with" since it was last run.

jamesagnew:
ooh good call.

I guess we should get a rush on this since it will be a nuisance to implement once 4.2.0/2020.02 are out.

fil512:
Thanks for doing this @jaferkhan !

jaferkhan:
@fil512 You're most welcome ^_^

PR #1679, opened 2020-01-20
Remove transient fields from migration task

jaferkhan:
Hello people,

This pull request fixes issue #1674, which asks for the removal of transient fields from the migration task named `RenameColumnTask`, as these fields could cause un-intended side effects whilst running this task.

Please let me know if there is any room for improvement.

Thanks,
Jafer


jamesagnew:
Thanks for doing this! Merging.

CI failure appears to be unrelated.

PR #1689, opened 2020-01-29
updated custom narrative generation documentation

dionmcm:
I found that the documented pattern of
 [name].class=[fully qualified classname]
No longer works. The pattern of
 [name].resourceType=[ResourceType]
worked for me, I updated the datatype example as well in kind, but I've not used it that way.

Maybe some extended examples for the other property type extensions would be good - e.g. .profile, .style, .contextPath, .narrative, .title


jamesagnew:
Thanks for the contribution! Looks good to me.

PR #1696, opened 2020-02-03
Drop unnecessary index

jamesagnew:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ITermValueSetConceptDesignationDao.java
```diff
@@ -37,6 +35,4 @@
 	@Modifying
 	void deleteByTermValueSetId(@Param("pid") Long theValueSetId);
 
-	@Query("SELECT vscd FROM TermValueSetConceptDesignation vscd WHERE vscd.myConcept.myId = :pid")
```
dmuylwyk: I'm curious why you got rid of this query. Don't we still need it to populate designations in an expansion?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/entity/TermValueSetConceptDesignation.java
```diff
@@ -33,9 +33,7 @@
 import static org.apache.commons.lang3.StringUtils.left;
 import static org.apache.commons.lang3.StringUtils.length;
 
-@Table(name = "TRM_VALUESET_C_DESIGNATION", indexes = {
-	@Index(name = "IDX_VALUESET_C_DSGNTN_VAL", columnList = "VAL")
```
dmuylwyk: Yep, this index was definitely wrong. We *eventually* need indexes on `designation.language` and `designation.use` to support the `designation` in parameter of the `$expand` operation; however, we definitely do not need an index on `designation.value`. Glad this came to light.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ITermValueSetConceptDesignationDao.java
jamesagnew: That query was not used at all, I only deleted it because IJ flagged it as an unused method

PR #1698, opened 2020-02-03
Remove searchLastReturned column

fil512:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ISearchDao.java
```diff
@@ -38,16 +37,12 @@
 	@Query("SELECT s FROM Search s LEFT OUTER JOIN FETCH s.myIncludes WHERE s.myUuid = :uuid")
 	Optional<Search> findByUuidAndFetchIncludes(@Param("uuid") String theUuid);
 
-	@Query("SELECT s.myId FROM Search s WHERE (s.mySearchLastReturned < :cutoff) AND (s.myExpiryOrNull IS NULL OR s.myExpiryOrNull < :now)")
+	@Query("SELECT s.myId FROM Search s WHERE (s.myCreated < :cutoff) AND (s.myExpiryOrNull IS NULL OR s.myExpiryOrNull < :now)")
 	Slice<Long> findWhereLastReturnedBefore(@Param("cutoff") Date theCutoff, @Param("now") Date theNow, Pageable thePage);
```
jamesagnew: A couple of thoughts here: 

* We should probably rename this method to reflect its new criteria
* Created doesn't have an index and it's being used as the primary search criteria - Obviously we don't expect this table to grow too big but we also know it does for some people.. I'd be inclined to create a new index on that column

fil512: done

Issue #1300, opened 2019-05-05
Pagination bug: retrieving last page with _count>number of items to return results in server timeout error


imranmoezkhan:
This looks like a bug to me (let me know if its not and I'm doing something wrong):
I have some data on my server with 1835 observations of a particular loinc code:
/baseDstu3/Observation?code=http://loinc.org%7C2093-3&_summary=count
output bundle has -> "total": 1835

I do a normal search for the code using default "_count" and keep going to the next page until i reach the last page:
/baseDstu3?_getpages=8a158809-fae8-4fb5-a00b-2d2631b0bc96&_getpagesoffset=1820&_count=10&_pretty=true&_bundletype=searchset

On this page, the url for the next page is:
/baseDstu3?_getpages=8a158809-fae8-4fb5-a00b-2d2631b0bc96&_getpagesoffset=1830&_count=10&_pretty=true&_bundletype=searchset

Trying to retrieve that url results in a timeout error:
{
  "resourceType": "OperationOutcome",
  "issue": [
    {
      "severity": "error",
      "code": "processing",
      "diagnostics": "Request timed out after 60385ms"
    }
  ]
}

When I change the url to have "_count=5" or less, I get the remaining observations bundle.

jamesagnew:
What version of hapi are you seeing this on? Can you confirm whether you
see it on current 3.8.0 snapshot builds?

J

sent from my phone.

imranmoezkhan:
Yes, its HAPI FHIR Server - 3.8.0-SNAPSHOT
I used the hapi-fhir-cli project to deploy the server, changed the db to postgres and used the DatabaseBackedPagingProvider class to set the paging provider. I also configured some extra hibernate properties (mostly the c3p0 pooling ones) - not sure if this is causing the issue:

extraProperties.put("hibernate.dialect", org.hibernate.dialect.PostgreSQL94Dialect.class.getName());
extraProperties.put("hibernate.jdbc.batch_size", "20");
extraProperties.put("hibernate.cache.use_query_cache", "false");
extraProperties.put("hibernate.cache.use_second_level_cache", "false");
extraProperties.put("hibernate.cache.use_structured_entries", "false");
extraProperties.put("hibernate.cache.use_minimal_puts", "false");
extraProperties.put("hibernate.search.model_mapping", LuceneSearchMappingFactory.class.getName());
extraProperties.put("hibernate.search.default.directory_provider", "filesystem");
[...]

PR #1712, opened 2020-02-12
remove distinct from search

fil512:
A large enterprise installation reported that adding distinct to the main hapi-fhir search function had serious performance consequences for them.  Commented out with comment for now.

Also improved migration logging.


fil512:
Large enterprise installation tested and confirmed this fix resolved the issue for them.

jamesagnew:
Fixing this breaks a test for #1300 - Seeing if I can find another way to fix that

PR #1716, opened 2020-02-17
Improve search provider method binding accuracy

mykhsystematic:
SearchMethodBindin.incomingServerRequestMatchesMethod handle search parameters which start with "_" (for example: _lastUpdated) in wrong way. (If exists search method without _lastUpdated which match search request it could be taken).

Fix for " _lastUpdated" and "_has".


mykhsystematic:
@jamesagnew could you please have a look.

tadgh:
Heya @mykhsystematic  we are doing some spring cleaning over here and I see this PR is 2 years old. I am going to close it for now as our codebase has heavily changed, but if it is still needed, please feel free to re-implement against master, and ping me on the pull request. I can try to ensure it gets through.

Issue #1499, opened 2019-09-21
Date match bug


fil512:
See https://github.com/jamesagnew/hapi-fhir/tree/date-match-bug for a failing test that demonstrates the error.

The source of the bug is the timezone shift calculation in DateRangeParam.getUpperBoundAsInstant()

Recommended fix:
Add two new columns to ResourceIndexedSearchParamDate that store the low and high date (with no time) as an integer in the format 20190921.  When the search is for a date without a time, use this column for the search instead of the date column.

Romanow88:
Just encountered this as well just now. 

I do not know why this was implemented in this way. I agree that the date range should account for time, but rather then trying to play with timezone shifts, it's better to just unwrap the date into low-high via appending 00:00:00.0000000 and 23:59:59.9999999 accordingly.

Romanow88:
@jamesagnew  I'm willing to submit a PR if you think this is valid.

jamesagnew:
Wouldn't that approach suffer from the same issue? You'd have to pick a timezone to apply that 00:00 and 23:59 to..

Romanow88:
I would use the same time zone that is used to compute the date of creation of the resource. That way what we display on the response is in sync with the query.

Without knowing the users timezone we will net be able to determine the correct time that they would need, but at least we would be consistent with what is in the creation date.

Right now when I request resources for 21-02-2020, the current implementation unwraps that into 20-02-2020 12:30:00 - 21-02-2020 13:30:00 for the server and client in the same timezone.

I would propose to unwrap it to 21-02-2020 00:00:00 - 21-02-2020 23:59:59 using the server timezone, if no timezone is provided.

jamesagnew:
You mean like you'd just use the underlying OS timezone and just choose a
24-hour period that aligns with that timezone?

I'm not 100% sure but I feel like that's already what we're doing....

Romanow88:
We are doing that, true, but in a weird way. What we do is: Take a date, and make it a time by just taking the midnight. Then to get the time range we create two times based on that, one is -12 hours, other is +12 hours. The problem is that if the resource was created after midday on that date, it will not be included in the search. 

Example:
1. We have a resource a resource that was created at 21-02-2020 17:00:00
2. A search is triggered with the date param set to 21-02-2020 
3. 21-02-2020 is unwrapped to 21-02-2020 00:00:00
[...]

PR #1730, opened 2020-02-25
Add integer date fields to ResourceIndexedSearchParamDate

tadgh:
This addresses #1499. By adding two integer fields representing the low and high date of the DateParam, which we can then use  during searches which need only DAY-precision granularity. 

* Add `SP_VALUE_LOW_DATE_ORDINAL` and `SP_VALUE_HIGH_DATE_ORDINAL` to ResrouceIndexedSearchParamDate
* Add flag to DaoConfig controlling this behaviour.
* Add database migration task. This includes a mild rework of `CalculateHashesTask`, in which I extract out the common functionality to `BaseColumnCalculatorTask`. 
* Modify broken test to reflect new reality.
* Add changelog entry.

**Before**:
Depending on time of ingestion, queries for `Observation?date=2020-02-25` could miss some observations. 
**After**:
Anything incoming with date `2020-02-25` at time of ingestion will show up in the search results.


Romanow88:
Is having the data additional with a lower precision the best way forward? Why not create the search predicate so that the results are returned correctly?

jamesagnew:
@Romanow88 I don't think there is any way of achieving full compliance with the FHIR spec without adding this lower precision mechanism. Consider this case:

Server is in Toronto CA (UTC-04) and an Observation gets created in Sydney AU (UTC+11) with an effectiveDate of `2019-02-27T13:00:00+11:00`.

With what you're proposing, we'll store this as a range of `2019-02-26T00:00:00-04:00` through `2019-02-26T23:59:59-04:00`

Then if a user performs the query `Observation?date=ge2019-02-26` they will get this result back, and if they perform the query `Observation?date=ge2019-02-27` they will not.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
```diff
@@ -150,67 +152,111 @@ private Predicate createPredicateDate(IQueryParameterType theParam,
...
 		Predicate ub = null;
 
 		if (operation == SearchFilterParser.CompareOperation.lt) {
-			if (lowerBound == null) {
+			if (lowerBoundInstant == null) {
 				throw new InvalidRequestException("lowerBound value not correctly specified for compare operation");
 			}
-			lb = theBuilder.lessThan(theFrom.get("myValueLow"), lowerBound);
+			//im like 80% sure this should be ub and not lb, as it is an UPPER bound.
+			if (isOrdinalComparison) {
```
jamesagnew: I feel like instead of having a whole bunch of `if (isOrdinalComparison)` for each of these, you could have one at the top where you say 
```java
if (isOrdinalComparison) {
   lowFieldName = "myValueLowDateOrdinal";
} else {
   lowFieldName = "myValueLow";
```

... and avoid most of the duplication below.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
```diff
@@ -221,8 +267,10 @@ private Predicate createPredicateDateFromRange(CriteriaBuilder theBuilder,
 			throw new InvalidRequestException(String.format("Unsupported operator specified, operator=%s",
 				operation.name()));
 		}
-
-		ourLog.trace("Date range is {} - {}", lowerBound, upperBound);
+		if (isOrdinalComparison) {
+			ourLog.trace("Ordinal date range is {} - {} ", lowerBoundAsOrdinal, upperBoundAsOrdinal);
+		}
```
jamesagnew: this should probably be an `else`

hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/BaseColumnCalculatorTask.java
```diff
@@ -0,0 +1,271 @@
...
+		initializeExecutor();
+
+		try {
+
+			while(true) {
+				MyRowCallbackHandler rch = new MyRowCallbackHandler();
+				getTxTemplate().execute(t -> {
+					JdbcTemplate jdbcTemplate = newJdbcTemplate();
+					jdbcTemplate.setMaxRows(100000);
+					String sql = "SELECT * FROM " + getTableName() + " WHERE " + getColumnName() + " IS NULL";
```
jamesagnew: I'm actually not sure this will work for the dates.. The reason is that it's possible for one of the dates (either `SP_VALUE_HIGH` or `SP_VALUE_LOW` to be saved as null even if the other isn't. For example if we were indexing a Period with a start date but no finish date...

It's probably gonna require something like `SELECT FROM HFJ_SPIDX_DATE WHERE SP_VALUE_LOW IS NOT NULL AND SP_VALUE_LOW_ORDINAL IS NULL`

hapi-fhir-jpaserver-base/src/test/resources/logback-test.xml
```diff
@@ -27,7 +27,7 @@
 		<appender-ref ref="STDOUT" />
 	</logger>
 
-	<logger name="ca.uhn.fhir.jpa.dao" additivity="false" level="info">
+	<logger name="ca.uhn.fhir.jpa.dao" additivity="false" level="trace">
```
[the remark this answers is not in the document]
tadgh: addressed in df7a9916c39d66422eb8a67f7d3fa4ebf4accaf3

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
tadgh: Addressed in 6123e4b189e450952043cf29c705ccda0ad1a12e

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
tadgh: Addressed in a5105439339d6cb94ed22308c97822932d20ad0e

hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/BaseColumnCalculatorTask.java
tadgh: Attempted to address this in 8b223a1, but this issue probably merits a minor refactor to more easily allow calculators, or table tasks, to determine their WHERE clauses. This is a hacky solution but it'll do until we can figure out a minor rework.

jamesagnew:
Looks great! Lert's merge this as soon as we've produced the next interim build.

PR #1733, opened 2020-02-28
Implement Jackson parser/serializer

jamesagnew:
Merging branch for #1673


hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/json/jackson/JacksonWriter.java
```diff
@@ -21,7 +19,6 @@
     }
 
     private BlockType blockType = BlockType.NONE;
-    private final List<BlockType> blocks = new LinkedList<>();
```
bdenton: Rather than removing LinkedList, I would suggest using a Stack like the GsonWriter uses. This is necessary in order for endBlock() to work properly if called twice in a row

jamesagnew:
@bdenton 

> Removing methods from the JsonLikeWriter interface is a breaking change that could impact some users. Even though the JsonParser might not use those methods, other user implementations might.

I'm not a big fan of having unused and untested code in the core library- We'll be soliciting community input about this change anyhow, we can ask anyone negatively affected by this change to speak up

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/json/jackson/JacksonWriter.java
jamesagnew: All tests are passing without having this stack- Truthfully I don't actually see where endBlock() is actually called either, other than in tests. What is its purpose?

bdenton: When it was written, there were a couple of places in the Json serialization process where the closing of nested arrays & objects was unclear... could need to close a nested array or the surrounding object. Rather than adding logic to JsonParser to keep track of what needed closing, seemed to make more sense to have the JsonLikeWriter keep track

jamesagnew: hmm, i wonder if they were refactored out as a part of other fixes. The good news is that they don't seem to be present any more.

PR #1741, opened 2020-03-02
Create custom span for Search Threads for Elastic APM

tadgh:
Currently in elastic APM, since searches occur on background threads, the Transactions that are reported caused by http requests are lacking the span information of the searches. 

To fix this, and ensure that any search time also counts towards the transaction's time, create a custom span in the task, while specifying the HTTP request transaction as the parent. This causes the search span to be associated to the original request when viewing APM data.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
@@ -56,6 +56,9 @@ import ca.uhn.fhir.rest.server.servlet.ServletRequestDetails;
 import ca.uhn.fhir.rest.server.util.ICachedSearchDetails;
 import ca.uhn.fhir.util.AsyncUtil;
 import ca.uhn.fhir.util.StopWatch;
+import co.elastic.apm.api.ElasticApm;
+import co.elastic.apm.api.Span;
+import co.elastic.apm.api.Transaction;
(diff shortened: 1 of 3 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
```diff
@@ -614,6 +617,7 @@ protected SearchTask(Search theSearch, IDao theCallingDao, SearchParameterMap th
 			mySearchRuntimeDetails = new SearchRuntimeDetails(theRequest, mySearch.getUuid());
 			mySearchRuntimeDetails.setQueryString(theParams.toNormalizedQueryString(theCallingDao.getContext()));
 			myRequest = theRequest;
+			myParentTransaction = ElasticApm.currentTransaction();
```
jamesagnew: I can't believe this is all that took....... hot damn

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
```diff
@@ -841,7 +845,8 @@ void requestImmediateAbort() {
 		@Override
 		public Void call() {
 			StopWatch sw = new StopWatch();
-
+			Span span = myParentTransaction.startSpan("db", "H2", "search");
```
jamesagnew: What are these strings? Presumably h2 refers to the DB, which would be misleading in any environment that isn't using H2....

PR #1750, opened 2020-03-05
extend Encounter/[<instance>/]$everything

gematik-fue:
Extends Encounter/\<instance>/$everything and Encounter/$everything to behave like Patient's $everything operation, so a ConsentService can decide within context of RequestDetails what resources to reveal.


gematik-fue:
hey @jamesagnew can you merge this?

It just passes down RequestDetails on Encounter/<instance>/$everything as additional parameter, just as $everything does on Patient/<instance/$everything

PR #1754, opened 2020-03-09
Optmize forced ID resolution

jamesagnew:
This PR significantly optimizes forced ID resolution during resource creation, as well as during searches.

It reduces 1-4 SQL statements during resource creation, and adds an optional caching layer that can further reduce calls on servers where deletion is disabled.

It also reduces resolution calls during searches in some cases.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IForcedIdDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IForcedIdDao.java
@@ -47,13 +47,17 @@ public interface IForcedIdDao extends JpaRepository<ForcedId, Long> {
 	void deleteByPid(@Param("pid") Long theId);
 
 	/**
-	 * This method returns an object array, where the order matters. Be careful if you change this query in any way.
+	 * This method returns a Collection where each row is an element in the collection. Each element in the collection
+	 * is an object array, where the order matters (the array represents columns returned by the query). Be careful if you change this query in any way.
 	 */
 	@Query("SELECT f.myForcedId, f.myResourcePid FROM ForcedId f WHERE myResourceType = :resource_type AND myForcedId IN ( :forced_id )")
 	Collection<Object[]> findByTypeAndForcedId(@Param("resource_type") String theResourceType, @Param("forced_id") Collection<String> theForcedId);
 
 	/**
-	 * This method returns an object array, where the order matters. Be careful if you change this query in any way.
+	 * Warning: No DB index exists for this particular query, so it may not perform well
+	 *
+	 * This method returns a Collection where each row is an element in the collection. Each element in the collection
+	 * is an object array, where the order matters (the array represents columns returned by the query). Be careful if you change this query in any way.
 	 */
 	@Query("" +
 		"SELECT " +
@@ -61,10 +65,11 @@ public interface IForcedIdDao extends JpaRepository<ForcedId, Long> {
 		"FROM ForcedId f " +
 		"JOIN ResourceTable t ON t.myId = f.myResourcePid " +
 		"WHERE f.myForcedId IN ( :forced_id )")
-	Collection<Object[]> findAndResolveByTypeAndForcedId(@Param("forced_id") Collection<String> theForcedIds);
+	Collection<Object[]> findAndResolveByForcedIdWithNoType(@Param("forced_id") Collection<String> theForcedIds);
 
 	/**
-	 * This method returns an object array, where the order matters. Be careful if you change this query in any way.
+	 * This method returns a Collection where each row is an element in the collection. Each element in the collection
+	 * is an object array, where the order matters (the array represents columns returned by the query). Be careful if you change this query in any way.
 	 */
 	@Query("" +
 		"SELECT " +
@@ -72,5 +77,5 @@ public interface IForcedIdDao extends JpaRepository<ForcedId, Long> {
 		"FROM ForcedId f " +
 		"JOIN ResourceTable t ON t.myId = f.myResourcePid " +
 		"WHERE f.myResourceType = :resource_type AND f.myForcedId IN ( :forced_id )")
-	Collection<Object[]> findAndResolveByTypeAndForcedId(@Param("resource_type") String theResourceType, @Param("forced_id") Collection<String> theForcedIds);
+	Collection<Object[]> findAndResolveByForcedIdWithNoType(@Param("resource_type") String theResourceType, @Param("forced_id") Collection<String> theForcedIds);
 }
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IForcedIdDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IForcedIdDao.java
@@ -2,6 +2,7 @@ package ca.uhn.fhir.jpa.dao.data;
 
 import java.util.Collection;
 import java.util.List;
+import java.util.Optional;
 
 /*
  * #%L
@@ -35,8 +36,11 @@ public interface IForcedIdDao extends JpaRepository<ForcedId, Long> {
 	@Query("SELECT f.myResourcePid FROM ForcedId f WHERE myForcedId IN (:forced_id)")
 	List<Long> findByForcedId(@Param("forced_id") Collection<String> theForcedId);
 
-	@Query("SELECT f.myResourcePid FROM ForcedId f WHERE myResourceType = :resource_type AND myForcedId IN (:forced_id)")
-	List<Long> findByTypeAndForcedId(@Param("resource_type") String theResourceType, @Param("forced_id") Collection<String> theForcedId);
+	@Query("SELECT f.myResourcePid FROM ForcedId f WHERE myResourceType = :resource_type AND myForcedId = :forced_id")
+	Optional<Long> findByTypeAndForcedId(@Param("resource_type") String theResourceType, @Param("forced_id") String theForcedId);
+
+	@Query("SELECT f.myResourcePid FROM ForcedId f WHERE myResourceType = :resource_type AND myForcedId IN ( :forced_id )")
+	Collection<Long> findByTypeAndForcedId(@Param("resource_type") String theResourceType, @Param("forced_id") Collection<String> theForcedId);
 
 	@Query("SELECT f FROM ForcedId f WHERE f.myResourcePid = :resource_pid")
 	ForcedId findByResourcePid(@Param("resource_pid") Long theResourcePid);
@@ -44,4 +48,20 @@ public interface IForcedIdDao extends JpaRepository<ForcedId, Long> {
 	@Modifying
 	@Query("DELETE FROM ForcedId t WHERE t.myId = :pid")
 	void deleteByPid(@Param("pid") Long theId);
+
+	@Query("" +
+		"SELECT " +
+		"   f.myResourceType, f.myResourcePid, f.myForcedId, t.myDeleted " +
+		"FROM ForcedId f " +
+		"JOIN ResourceTable t ON t.myId = f.myResourcePid " +
+		"WHERE f.myForcedId IN ( :forced_id )")
+	Collection<Object[]> findAndResolveByTypeAndForcedId(@Param("forced_id") Collection<String> theForcedIds);
+
+	@Query("" +
+		"SELECT " +
+		"   f.myResourceType, f.myResourcePid, f.myForcedId, t.myDeleted " +
+		"FROM ForcedId f " +
+		"JOIN ResourceTable t ON t.myId = f.myResourcePid " +
+		"WHERE f.myResourceType = :resource_type AND f.myForcedId IN ( :forced_id )")
+	Collection<Object[]> findAndResolveByTypeAndForcedId(@Param("resource_type") String theResourceType, @Param("forced_id") Collection<String> theForcedIds);
 }
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IForcedIdDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IForcedIdDao.java
@@ -39,15 +39,15 @@ public interface IForcedIdDao extends JpaRepository<ForcedId, Long> {
 	@Query("SELECT f.myResourcePid FROM ForcedId f WHERE myResourceType = :resource_type AND myForcedId = :forced_id")
 	Optional<Long> findByTypeAndForcedId(@Param("resource_type") String theResourceType, @Param("forced_id") String theForcedId);
 
-	@Query("SELECT f.myResourcePid FROM ForcedId f WHERE myResourceType = :resource_type AND myForcedId IN ( :forced_id )")
-	Collection<Long> findByTypeAndForcedId(@Param("resource_type") String theResourceType, @Param("forced_id") Collection<String> theForcedId);
-
 	@Query("SELECT f FROM ForcedId f WHERE f.myResourcePid = :resource_pid")
 	ForcedId findByResourcePid(@Param("resource_pid") Long theResourcePid);
 
 	@Modifying
 	@Query("DELETE FROM ForcedId t WHERE t.myId = :pid")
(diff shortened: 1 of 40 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -22,132 +22,222 @@
...
+		ListMultimap<String, String> typeToIds = organizeIdsByResourceType(theIds);
+
 		for (Map.Entry<String, Collection<String>> nextEntry : typeToIds.asMap().entrySet()) {
 			String nextResourceType = nextEntry.getKey();
 			Collection<String> nextIds = nextEntry.getValue();
 			if (isBlank(nextResourceType)) {
 
 				StorageProcessingMessage msg = new StorageProcessingMessage()
 					.setMessage("This search uses unqualified resource IDs (an ID without a resource type). This is less efficient than using a qualified type.");
+				ourLog.debug(msg.getMessage());
```
tadgh: Will having this warning message _inside_ the loop cause performance issues here?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -22,132 +22,222 @@
...
 				HookParams params = new HookParams()
 					.add(RequestDetails.class, theRequest)
 					.addIfMatchesType(ServletRequestDetails.class, theRequest)
 					.add(StorageProcessingMessage.class, msg);
-				JpaInterceptorBroadcaster.doCallHooks(theInterceptorBroadcaster, theRequest, Pointcut.JPA_PERFTRACE_WARNING, params);
+				JpaInterceptorBroadcaster.doCallHooks(myInterceptorBroadcaster, theRequest, Pointcut.JPA_PERFTRACE_WARNING, params);
```
tadgh: Same with here. Is it maybe worthwhile to pull this out and just broadcast _once_ if the list contained a type-less resource PID? Or is this message valuable for every instance of a pid that was processed?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -22,132 +22,222 @@
...
+							retVal.add(new ResourcePersistentId(nextCachedPid));
+						}
+					}
+				}
+
+				if (nextIds.size() > 0) {
+					Collection<Object[]> views = myForcedIdDao.findByTypeAndForcedId(nextResourceType, nextIds);
+					for (Object[] nextView : views) {
+						String forcedId = (String) nextView[0];
+						Long pid = (Long) nextView[1];
```
tadgh: This is essentially an undocumented data structure, and its a bit tough to read unless you know the full context. Is it worthwhile maybe creating a small data class that can consume one of these known object arrays and making it more readable? If this is super one-off, maybe just comment this block of code

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -156,17 +246,136 @@ public String translatePidIdToForcedId(String theResourceType, ResourcePersisten
 		}
 	}
 
+	private ListMultimap<String, String> organizeIdsByResourceType(Collection<IIdType> theIds) {
+		ListMultimap<String, String> typeToIds = MultimapBuilder.hashKeys().arrayListValues().build();
+		for (IIdType nextId : theIds) {
+			if (myDaoConfig.getResourceClientIdStrategy() == DaoConfig.ClientIdStrategyEnum.ANY || !isValidPid(nextId)) {
```
tadgh: Question: Why is this check necessary? Seems as though this function should organize IDs by type regardless of the ResourceClientIdStrategy, or if the pid is valid or not. If that _isnt_ what this function is supposed to do, (e.g. it should only work on non-valid pids), then this function should be renamed.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -156,17 +246,136 @@ public String translatePidIdToForcedId(String theResourceType, ResourcePersisten
...
+
+	private Long resolveResourceIdentity(String theResourceType, String theId) {
+		Long retVal;
+		retVal = myForcedIdDao
+			.findByTypeAndForcedId(theResourceType, theId)
+			.orElseThrow(() -> new ResourceNotFoundException(new IdDt(theResourceType, theId)));
+		return retVal;
+	}
+
+	private Collection<IResourceLookup> translateForcedIdToPids(RequestDetails theRequest, Collection<IIdType> theId) {
```
tadgh: Suggestion: `theIds`

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -156,17 +246,136 @@ public String translatePidIdToForcedId(String theResourceType, ResourcePersisten
...
+				for (Iterator<String> forcedIdIterator = nextIds.iterator(); forcedIdIterator.hasNext(); ) {
+					String nextForcedId = forcedIdIterator.next();
+					String nextKey = nextResourceType + "/" + nextForcedId;
+					IResourceLookup cachedLookup = myResourceLookupCache.getIfPresent(nextKey);
+					if (cachedLookup != null) {
+						forcedIdIterator.remove();
+						retVal.add(cachedLookup);
+					}
+				}
+			}
```
tadgh: So this only works in case where Delete is disabled? Is that because it would be too much of a pain to maintain the cache in a scenario where resource deletion is possible?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -156,17 +246,136 @@ public String translatePidIdToForcedId(String theResourceType, ResourcePersisten
...
+
+			if (nextIds.size() > 0) {
+				Collection<Object[]> views;
+				if (isBlank(nextResourceType)) {
+
+					StorageProcessingMessage msg = new StorageProcessingMessage()
+						.setMessage("This search uses unqualified resource IDs (an ID without a resource type). This is less efficient than using a qualified type.");
```
tadgh: This is the 2nd or 3rd time I've seen this chunk of code. Maybe a constant for this string, or rip the whole Interceptor Broadcast out into a helper function

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
```diff
@@ -164,6 +156,177 @@ public void testCreateWithClientAssignedId() {
...
+		observation = new Observation();
+		observation.getSubject().setReference("Patient/P");
+		myObservationDao.create(observation);
+		// select: lookup forced ID
+		assertEquals(1, myCaptureQueriesListener.countSelectQueriesForCurrentThread());
+		assertEquals(0, myCaptureQueriesListener.countDeleteQueriesForCurrentThread());
+		// insert to: HFJ_RESOURCE, HFJ_RES_VER, HFJ_RES_LINK
+		assertEquals(3, myCaptureQueriesListener.countInsertQueriesForCurrentThread());
+		assertEquals(0, myCaptureQueriesListener.countUpdateQueriesForCurrentThread());
+
```
tadgh: Out of curiosity, why create two here?

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
```diff
@@ -164,6 +156,177 @@ public void testCreateWithClientAssignedId() {
...
+		/*
+		 * Add another
+		 */
+
+		myCaptureQueriesListener.clear();
+		observation = new Observation();
+		observation.getSubject().setReference("Patient/P");
+		myObservationDao.create(observation);
+		// select: no lookups needed because of cache
```
tadgh: ah I now see why

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IForcedIdDao.java
```diff
@@ -35,13 +36,41 @@
...
 	@Query("SELECT f FROM ForcedId f WHERE f.myResourcePid = :resource_pid")
 	ForcedId findByResourcePid(@Param("resource_pid") Long theResourcePid);
 
 	@Modifying
 	@Query("DELETE FROM ForcedId t WHERE t.myId = :pid")
 	void deleteByPid(@Param("pid") Long theId);
+
+	/**
+	 * This method returns an object array, where the order matters. Be careful if you change this query in any way.
+	 */
```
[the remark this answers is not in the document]
jamesagnew: Updating to:

> /**
> * This method returns a Collection where each row is an element in the collection. Each element in the collection 
> * is an object array, where the order matters (the array represents columns returned by the query). Be careful if you change this query in any way.
> */

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoResourceLinkResolver.java
```diff
@@ -64,16 +64,15 @@
 	private DaoRegistry myDaoRegistry;
 
 	@Override
-	public ResourceTable findTargetResource(RuntimeSearchParam theNextSpDef, String theNextPathsUnsplit, IIdType theNextId, String theTypeString, Class<? extends IBaseResource> theType, IBaseReference theReference, RequestDetails theRequest) {
-		ResourceTable target;
-		ResourcePersistentId valueOf;
+	public IResourceLookup findTargetResource(RuntimeSearchParam theNextSpDef, String theNextPathsUnsplit, IIdType theNextId, String theTypeString, Class<? extends IBaseResource> theType, IBaseReference theReference, RequestDetails theRequest) {
```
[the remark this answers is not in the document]
jamesagnew: sounds good, will do.

```java
	/**
	 * This method resolves the target of a reference found within a resource that is being created/updated. We do this
	 * so that we can create indexed links between resources, and so that we can validate that the target actually
	 * exists in cases where we need to check that.
	 * <p>
	 * This method returns an {@link IResourceLookup} so as to avoid needing to resolve the entire resource.
	 *
	 * @param theSearchParam      The param that is being indexed
	 * @param theSourcePath       The path within the resource where this reference was found
	 * @param theSourceResourceId The ID of the resource containing the reference to the target being resolved
	 * @param theTypeString       The type of the resource being resolved
	 * @param theType             The resource type of the target
	 * @param theReference        The reference being resolved
	 * @param theRequest          The incoming request, if any
	 */
	IResourceLookup findTargetResource(RuntimeSearchParam theSearchParam, String theSourcePath, IIdType theSourceResourceId, String theTypeString, Class<? extends IBaseResource> theType, IBaseReference theReference, RequestDetails theRequest);
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoResourceLinkResolver.java
```diff
@@ -64,16 +64,15 @@
...
+		IResourceLookup valueOf;
 		String idPart = theNextId.getIdPart();
 		try {
-			valueOf = myIdHelperService.translateForcedIdToPid(theTypeString, idPart, theRequest);
+			valueOf = myIdHelperService.resolveResourceIdentity(theTypeString, idPart, theRequest);
 			ourLog.trace("Translated {}/{} to resource PID {}", theType, idPart, valueOf);
 		} catch (ResourceNotFoundException e) {
 
-			Optional<ResourcePersistentId> pidOpt = createPlaceholderTargetIfConfiguredToDoSo(theType, theReference, idPart);
+			Optional<ResourceTable> pidOpt = createPlaceholderTargetIfConfiguredToDoSo(theType, theReference, idPart);
```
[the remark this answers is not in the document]
jamesagnew: done

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IForcedIdDao.java
```diff
@@ -35,13 +36,41 @@
...
+	@Query("" +
+		"SELECT " +
+		"   f.myResourceType, f.myResourcePid, f.myForcedId, t.myDeleted " +
+		"FROM ForcedId f " +
+		"JOIN ResourceTable t ON t.myId = f.myResourcePid " +
+		"WHERE f.myForcedId IN ( :forced_id )")
+	Collection<Object[]> findAndResolveByTypeAndForcedId(@Param("forced_id") Collection<String> theForcedIds);
```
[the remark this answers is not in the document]
jamesagnew: Nope. Renaming...

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoResourceLinkResolver.java
```diff
@@ -89,36 +88,29 @@ public ResourceTable findTargetResource(RuntimeSearchParam theNextSpDef, String
 			valueOf = pidOpt.get();
 		}
 
-		target = myEntityManager.find(ResourceTable.class, valueOf.getIdAsLong());
-		RuntimeResourceDefinition targetResourceDef = myContext.getResourceDefinition(theType);
-		if (target == null) {
-			String resName = targetResourceDef.getName();
-			throw new InvalidRequestException("Resource " + resName + "/" + idPart + " not found, specified in path: " + theNextPathsUnsplit);
+		ourLog.trace("Resource PID {} is of type {}", valueOf, valueOf.getResourceType());
```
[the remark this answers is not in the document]
jamesagnew: God, not to mention that the variable is called `valueOf`, which is horrible.

Fixing.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
jamesagnew: good call, done

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
jamesagnew: per below

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
jamesagnew: The idea here is that it's still an improvement even if delete is enabled (some operations dropped from  4 or 5 SELECTS to 2 or 3) but it's an even bigger improvement if we have deletes disabled since then the cache can drop to 0 to 2.

The reason is that we need to check whether a resource is deleted if there is a reference targeting that resource (i.e. you can't create a resource with a reference to a deleted resource). We bundle that check into the same lookup so as to reduce the number of DB round trips, but then we can't trust the cache if things might have been deleted..

The reason deletes matter is that we cache the deletion status along with

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
jamesagnew: Done

PR #1755, opened 2020-03-09
Drop anyresource classes

jamesagnew:


tadgh:
Easy enough to review :) But why the change? Just cruft that is no longer referred to?

jamesagnew:
@tadgh These were in this repo but were only really used in CDR, so they were both confusing and a drain on our test coverage. I've moved 'em over there.

PR #1770, opened 2020-03-23
Fix parse issue for references to contained resources

augla:
Do not override the references with global contained resources if they have already been resolved


jamesagnew:
Hi @augla - Does this bug also affect the JSON parser?

augla:
> Hi @augla - Does this bug also affect the JSON parser?

Yes, added test that shows the same error for json (if changed mainline code is commented out)

augla:
Thanks!

PR #1782, opened 2020-03-29
Incorrect Docs Extension

yashrajbothra:
Changed Docs Reference Link from `.html` to `.md`


jamesagnew:
Hi @yashrajbothra - Thanks for the contribution! However, our docs rendering engine renders .md files into .html, so the links need to point to html or the published site won't work.

PR #1791, opened 2020-04-06
Fix GraphQL Error

ibrohimislam:
GraphQL Issues

there are 3 things here:
1. graphql search argument with a dash (`-`) on every resource causes an error. eg. search param appointment-type represented in graphql as `appointment_type`, but the query `{ AppointmentList(appointment_type:FOLLOWUP) {id}}` resulting an error.
2. graphql error message not helpful on an invalid argument. it shows "Unable to execute GraphQL Expression: null" message. e.g. `{ AppointmentList(test: ibrohim) {id}}`
3. graphql search argument _filter doesn't work. e.g. `{ AppointmentList(_filter: "(service-type eq doctor)") {id}}`

for tackling the first problem. `JpaStorageServices` now implement name mapping `graphqlArgumentToSearchParam` and `searchParamToGraphqlArgument`.

for the second problem. `JpaStorageServices` mimicking `BaseHapiFhirResourceDao`'s `translateRawParameters` to filter invalid parameter.

for the third problem. `PARAM_FILTER` treated as special case.


jamesagnew:
Hi @ibrohimislam ,

Thanks for the contribution! Would you be able to add a test that demonstrates and verifies the fix?

ibrohimislam:
ok. I'll add it.

ibrohimislam:
I've added test implementation on 1a6f96e.

ibrohimislam:
is the test good enough? @jamesagnew

PR #1810, opened 2020-04-21
use bootstrap "text-danger" style for "This is not a production server!" warning (instead of "color: red")

jschneid-nmdp:


jamesagnew:
Thanks Joel! Merging now.

PR #1811, opened 2020-04-22
Fix compilation error

fitzoh:
It appears that https://github.com/jamesagnew/hapi-fhir/pull/1791 was merged with a failing check due to a bad import.

This PR fixes that.


jamesagnew:
Thanks @fitzoh !

PR #1830, opened 2020-05-04
Clean up hash code calculation

jamesagnew:


hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceIndexedSearchParamCoords.java
```diff
@@ -68,21 +76,14 @@ public ResourceIndexedSearchParamCoords(PartitionSettings thePartitionSettings,
 		setParamName(theParamName);
 		setLatitude(theLatitude);
 		setLongitude(theLongitude);
+		calculateHashes();
 	}
 
 	@Override
-	@PrePersist
 	public void calculateHashes() {
-		if (myHashIdentity == null && getParamName() != null) {
```
fil512, line 76: Trying to understand this change.  Are you removing transitional code we had in there to bridge the pre-hash installs to the post-hash installs?  Or is this a performance optimization that's no longer possible due to changes in the way hashing works?

jamesagnew: The general idea here is that we calculated the hashes via 2 different ways, both of which were kind of haphazard:

- We calculated then in a `@PrePersist` hook as the data was sent to the database
- We calculated them if someone called `hashCode()`

This probably made sense originally where we only ever created them and put them in the DB and that was that.. But increasingly we're using these structures for all kinds of other stuff (e.g. in memory matching, reusing rows in the index tables rather than deleting one and adding a new one, etc.) and this haphazard logic was causing unexpected bugs when unpredicted situations arose (such as a stored row having no hash, whcih is the actual thing we're fixing here).

What this change does, more importantly, is remove chance from the equation by just explicitly calcuating the hashes as soon as possible.

fil512: nice

PR #1839, opened 2020-05-10
resolves #1838

### LINKED ISSUE #1838: Add %today date option that works like %now but has DAY precision

fil512:
Added %today that works like %now but with date precision instead of dateTime precision.

Resolves https://github.com/jamesagnew/hapi-fhir/issues/1838


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/4_3_0/1838-add-today-date-option.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
jamesagnew, line 1: Presumably we'll merge post 5.0.0 - Can you put this in a new directory called 5_1_0

hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/BaseDateTimeDt.java
```diff
@@ -638,6 +639,9 @@ public void setValueAsString(String theValue) throws DataFormatException {
 
 		if (NOW_DATE_CONSTANT.equalsIgnoreCase(theValue)) {
 			super.setValueAsString(ourXmlDateTimeFormat.format(new Date()));
+		} else if (TODAY_DATE_CONSTANT.equalsIgnoreCase(theValue)) {
+			super.setValueAsString(ourXmlDateTimeFormat.format(new Date()));
+			setPrecision(TemporalPrecisionEnum.DAY);
```
tadgh, line 644: I have a small concern/question here, as %today could be wrong if we rely on `new Date()`. Consider this scenario: 

It is 2020-05-11T07:51:00-0700 in Vancouver.
It is 2020-05-12T00:50:00+1000 in Melbourne.

A request coming in from a client in Melbourne using `%today` would expect it to resolve to the ordinal `20200512`, but with current implementation, it would resolve to `20200511`. However there is no easy way to know tz on these requests. Perhaps modify the documentation to indicate that `%today` is for server date? If i'm mistaken and this is already spelled out, thats fine.

PR #1923, opened 2020-06-14
Bug with Patient compartment user permission

RuslanZinovyev:


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
```diff
@@ -143,6 +144,13 @@ public Verdict applyRule(RestOperationTypeEnum theOperation, RequestDetails theR
 									appliesToResourceId = null;
 								}
 							}
+							if (theRequestDetails.getParameters().containsKey("_identifier")) {
```
jamesagnew, line 147: I don't understand this fix at all. What is `_identifier` (there is no search parameter with that name)? and if it's present why is it ok to assume that the request pertains to the first permitted compartment owner?

RuslanZinovyev: I see, my previous commit was in a fuss, it looks like there is no bug in Authorization interceptor. I found your java doc about this flag NO_NOT_PROACTIVELY_BLOCK_COMPARTMENT_READ_ACCESS and I believe it will solve the issue if we set it for AuthorizationInterceptor. The only problem that I didn't find in documentation how to set it.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
```diff
@@ -111,7 +112,7 @@ public Verdict applyRule(RestOperationTypeEnum theOperation, RequestDetails theR
 							}
 							break;
 						case SEARCH_TYPE:
-							if (theFlags.contains(AuthorizationFlagsEnum.NO_NOT_PROACTIVELY_BLOCK_COMPARTMENT_READ_ACCESS)) {
+							if (theFlags.contains(AuthorizationFlagsEnum.NO_NOT_PROACTIVELY_BLOCK_COMPARTMENT_READ_ACCESS) || myClassifierType == IN_COMPARTMENT && !theRequestDetails.getParameters().containsKey("_id")) {
```
jamesagnew, line 115: Curious about why the _id exclusion is needed.. Can you add a test to demonstrate what that is there for?

RuslanZinovyev: The only reason why I added this condition because I didn't want to break search by _id param implementation. You can check this behaviour in testReadByCompartmentReadByIdParam(). Since we already have this implemented, from my perspective was a good idea to leave it as is.

RuslanZinovyev: It helps us to block search by id operations to the method handler if we don't have access to this resource id in the compartment.

PR #1928, opened 2020-06-16
Im 20200615 mysql migration syntax

IanMMarshall:
Fix/enable migration tasks for MySQL. Ensured that these now work for both MySQL 5.7 and 8.0.

### DIFF
--- a/hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/RenameColumnTask.java
+++ b/hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/RenameColumnTask.java
@@ -23,6 +23,7 @@ package ca.uhn.fhir.jpa.migrate.taskdef;
 import ca.uhn.fhir.jpa.migrate.DriverTypeEnum;
 import ca.uhn.fhir.jpa.migrate.JdbcUtils;
 import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang3.Validate;
 import org.apache.commons.lang3.builder.HashCodeBuilder;
 import org.slf4j.Logger;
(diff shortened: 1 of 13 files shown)


hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/RenameColumnTask.java
```diff
@@ -82,6 +84,20 @@ public void doExecute() throws SQLException {
 					throw new SQLException("Can not rename " + getTableName() + "." + myOldName + " to " + myNewName + " because both columns exist and data exists in " + myNewName);
 				}
 
+				if (getDriverType().equals(DriverTypeEnum.MYSQL_5_7)) {
+					// Some DBs such as MYSQL require that foreign keys depending on the column be dropped before the column itself is dropped.
```
fil512, line 88: In the rename scenario, don't we need to add those constraints back in after the rename?

hapi-fhir-jpaserver-migrate/src/test/java/ca/uhn/fhir/jpa/migrate/taskdef/DropColumnTest.java
```diff
@@ -34,4 +36,39 @@ public void testDropColumn() throws SQLException {
 
 	}
 
+	@Test
+	public void testDropForeignKeyColumn() throws SQLException {
```
fil512, line 40: I recommend adding a test for the rename column scenario to validate we don't lose the fk constraints on rename

hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/RenameColumnTask.java
IanMMarshall: Good question. In this case, no I do not think there is a need to add the constraint back after rename. The constraint being dropped in this case depends on a column that is also being dropped (an existing column being replaced by renaming another existing column). As the renamed column would potentially have different data values, there is no guarantee that the old constraint will still be relevant.  

For non-MySQL databases, it appears that these types of constraints are dropped automatically when the column is dropped. This does not happen for MySQL and so the change ensures that the constraint is similarly dropped in the case of MySQL.

hapi-fhir-jpaserver-migrate/src/test/java/ca/uhn/fhir/jpa/migrate/taskdef/DropColumnTest.java
IanMMarshall: Good call. I have added tests for various scenarios involving renaming of columns that are used in foreign key constraints to confirm the existing behaviour and to ensure that the behaviour for MySQL use cases effectively matches that for non-MySQL DBs.

PR #1933, opened 2020-06-22
Fix ValueSet$expand with offset and count parameters.

dmuylwyk:
`BaseValidationSupportWrapper.expandValueSet(...)` and `ValidationSupportChain.expandValueSet(...)` were incorrectly replacing expansion options (i.e. offset and count) with `null`.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r5/ResourceProviderR5ValueSetTest.java
+++ b/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r5/ResourceProviderR5ValueSetTest.java
@@ -318,27 +318,20 @@ public class ResourceProviderR5ValueSetTest extends BaseResourceProviderR5Test {
 
 		String resp = myFhirCtx.newXmlParser().setPrettyPrint(true).encodeResourceToString(expanded);
 		ourLog.info(resp);
-		assertThat(resp, stringContainsInOrder(
-			"<total value=\"24\"/>\n",
-			"<offset value=\"1\"/>\n",
-			"<parameter>\n",
-			"<name value=\"offset\"/>\n",
-			"<valueInteger value=\"1\"/>\n",
-			"</parameter>\n",
-			"<parameter>\n",
-			"<name value=\"count\"/>\n",
-			"<valueInteger value=\"1000\"/>\n",
-			"</parameter>\n",
-			"<contains>\n",
-			"<system value=\"http://acme.org\"/>\n",
-			"<code value=\"11378-7\"/>\n",
-			"<display value=\"Systolic blood pressure at First encounter\"/>\n",
-			"</contains>\n",
-			"<contains>\n",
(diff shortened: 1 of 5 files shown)


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r5/ResourceProviderR5ValueSetTest.java
```diff
@@ -301,6 +301,122 @@ public void testExpandByIdWithPreExpansion() throws Exception {
...
+			.operation()
+			.onInstance(myExtensionalVsId)
+			.named("expand")
+			.withParameter(Parameters.class, "count", new IntegerType(1))
+			.execute();
+		ValueSet expanded = (ValueSet) respParam.getParameter().get(0).getResource();
+
+		String resp = myFhirCtx.newXmlParser().setPrettyPrint(true).encodeResourceToString(expanded);
+		ourLog.info(resp);
+		assertThat(resp, stringContainsInOrder(
```
jamesagnew, line 362: I think a better assert here would be to test that `expanded.getContains().size()` has a size of 1

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/BaseValidationSupportWrapper.java
```diff
@@ -70,7 +70,7 @@ public boolean isValueSetSupported(ValidationSupportContext theValidationSupport
 
 	@Override
 	public IValidationSupport.ValueSetExpansionOutcome expandValueSet(ValidationSupportContext theValidationSupportContext, ValueSetExpansionOptions theExpansionOptions, IBaseResource theValueSetToExpand) {
-		return myWrap.expandValueSet(theValidationSupportContext, null, theValueSetToExpand);
+		return myWrap.expandValueSet(theValidationSupportContext, theExpansionOptions, theValueSetToExpand);
```
jamesagnew, line 73: doh.....

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r5/ResourceProviderR5ValueSetTest.java
dmuylwyk: Agreed. I wrote those tests to be similar to the other tests in that class. I've since improved the new tests; however, I haven't gone back to change all of the old tests.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/BaseValidationSupportWrapper.java
dmuylwyk: Mistakes happen. : )

I'm surprised we didn't already have tests for the offset and count parameters. ¯\_(ツ)_/¯

PR #1938, opened 2020-06-23
$validate: support for profile parameter with ValidationOptions

oliveregger:
The $validate Operation can have a profile parameter. [FHIR R4: If this is nominated, then the resource is validated against this specific profile.](https://www.hl7.org/fhir/operation-resource-validate.html)

This PR checks if the profile RequestParameter is set and adds it to the ValidationOptions that the profile is enforced. The method has been changed in visibility to protected that it can be overwritten also from other packages.


jamesagnew:
Hi @oliveregger - Thanks for the PR!

I'm having a bit of trouble following though- The RequestValidatingInterceptor isn't responsible for handling the `$validate` operation, the resource providers are, e.g. https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/r4/JpaResourceProviderR4.java#L181

oliveregger:
Hi @jamesagnew  Thank you for your comment! I'm using the validation infrastructure in a non jpa environment. Thats why I extended the BaseValidatingInterceptor<T> with the ValidateOperationInterceptor with the @Hook(Pointcut.SERVER_INCOMING_REQUEST_POST_PROCESSED). https://github.com/ahdis/matchbox/blob/6ea32f147975ea56a4ffe65935fcd2374a1d7976/src/main/java/ch/ahdis/matchbox/interceptor/ValidateOperationInterceptor.java#L76. Would there be a better way to run the validation?

jamesagnew:
Hmm, this is interesting.. The intent with that interceptor is definitely to provide mandatory validation of resources as they pass through the system for "normal operations" (reads, searches, etc) as opposed to supplying the functionality for the `$validate` operation.

I think the issue with this implementation is that it would affect operations other than $validate... E.g. suppose someone using this interceptor performed a search for `OperationDefinition?profile=http://foo` this would actually modify the behaviour of the validation when that parameter was actually intended as a search parameter.

It'd probably make more sense to implement a resource provider specifically for the validate operation.. it could take in the profile param specifically for that operation and apply is correctly

PR #1945, opened 2020-06-26
Use read partition for finding update candidate on upsert

jamesagnew:


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/PartitioningR4Test.java
```diff
@@ -2073,12 +2082,24 @@ public void testSearch_RefParam_TargetForcedId_SearchDefaultPartition() {
 
 	}
 
+	@Test
+	public void testUpdate_ResourcePreExistsInWrongPartition() {
+		IIdType patientId = createPatient(withPutPartition(null), withId("ONE"), withBirthdate("2020-01-01"));
```
fil512, line 2087: This is the opposite scenario of what FMC requires.  They are trying to do a PUT with a non-null partition in the request when the pre-existing resource had a null partition.

I tried adding the following test, but it failed with an error:
```
	@Test
	public void testUpdate_ResourcePreExistsInNullPartition() {
		IIdType patientId = createPatient(withPutPartition(1), withId("ONE"), withBirthdate("2020-01-01"));
		addReadDefaultPartition();
		addCreateDefaultPartition();

		Patient p = new Patient();
		p.setId(patientId.toUnqualifiedVersionless());
		p.setGender(Enumerations.AdministrativeGender.MALE);
		myPatientDao.update(p);
	}
```

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_1_0/1945-use-read-partition-for-upsert-read.yaml
```diff
@@ -0,0 +1,7 @@
+---
+type: change
+issue: 1945
+title: "When performing a JPA 'upsert' (a PUT to an ID that may or may not already exist) on a partitioned system,
```
fil512, line 4: This reads different from the FMC scenario, but maybe it amounts to the same thing

 FMC is not using an interceptor; all partition id requests are coming in through http headers.  They can't add partitions to pre-existing resources because there's too much data, so they're leaving all pre-existing resources in the null partition.  New resources coming in (that may pre-exist) have a non-null requested partition id in the header of the request.  This currently throws an error when the resource already exists in the null partition.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/PartitioningR4Test.java
jamesagnew: I think the existing test is the scenario you describe and the test you added is the opposite actually...

Precondition: A resource exists, and it's in the 

```java
IIdType patientId = createPatient(withPutPartition(null), withId("ONE"), withBirthdate("2020-01-01"));
```

Test:  Have a resource come in with a PUT where the ID already exists. User has read-all-partitions permission, but no write permission at all (since none is needed- nothing will be written).

```java
addReadAllPartitions();

Patient p = new Patient();
p.setId(patientId.toUnqualifiedVersionless());
p.setGender(Enumerations.AdministrativeGender.MALE);
myPatientDao.update(p);
```

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_1_0/1945-use-read-partition-for-upsert-read.yaml
jamesagnew: It's not possible to use partitioning without an interceptor. They have to be using one unless they're doing URL based multitenancy (whcih they aren't).

Their interceptor should be allowing all partitions for all reads, and setting the write partition based on the incoming header.

fil512: I just spoke with Rakesh and he confirmed his interceptor works as you describe.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/PartitioningR4Test.java
fil512: Ah ok.  Looks like I read the test backwards.  Sounds like it's all good.

PR #1958, opened 2020-07-02
Im 20200630 code system load

IanMMarshall:
Fixed a race condition that can occur when code system version entity is updated (e.g. indexed) after term concepts have been queued for deferred save.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermDeferredStorageSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermDeferredStorageSvcImpl.java
@@ -22,6 +22,7 @@ package ca.uhn.fhir.jpa.term;
 
 import ca.uhn.fhir.jpa.api.config.DaoConfig;
 import ca.uhn.fhir.jpa.dao.data.ITermCodeSystemDao;
+import ca.uhn.fhir.jpa.dao.data.ITermCodeSystemVersionDao;
 import ca.uhn.fhir.jpa.dao.data.ITermConceptDao;
 import ca.uhn.fhir.jpa.dao.data.ITermConceptParentChildLinkDao;
 import ca.uhn.fhir.jpa.entity.TermCodeSystem;
@@ -62,7 +63,8 @@ public class TermDeferredStorageSvcImpl implements ITermDeferredStorageSvc {
 	@Autowired
 	protected ITermCodeSystemDao myCodeSystemDao;
 	@Autowired
-	protected PlatformTransactionManager myTransactionMgr;
+   protected ITermCodeSystemVersionDao myCodeSystemVersionDao;
+   @Autowired	protected PlatformTransactionManager myTransactionMgr;
 	private boolean myProcessDeferred = true;
 	private List<TermCodeSystem> myDefferedCodeSystemsDeletions = Collections.synchronizedList(new ArrayList<>());
 	private List<TermConcept> myDeferredConcepts = Collections.synchronizedList(new ArrayList<>());
@@ -142,7 +144,12 @@ public class TermDeferredStorageSvcImpl implements ITermDeferredStorageSvc {
 		ourLog.info("Saving {} deferred concepts...", count);
 		while (codeCount < count && myDeferredConcepts.size() > 0) {
 			TermConcept next = myDeferredConcepts.remove(0);
-			codeCount += myCodeSystemStorageSvc.saveConcept(next);
+			if(myCodeSystemVersionDao.findById(next.getCodeSystemVersion().getPid()).isPresent()) {
+         	codeCount += myCodeSystemStorageSvc.saveConcept(next);
+			} else {
+				ourLog.warn("Unable to save deferred TermConcept {} because Code System {} version PID {} is no longer valid. Code system may have since been updated.",
+					next.getCode(), next.getCodeSystemVersion().getCodeSystemDisplayName(), next.getCodeSystemVersion().getPid());
+			}
 		}
 
 		if (codeCount > 0) {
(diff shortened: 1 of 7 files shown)


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.hl7.fhir.r4.model.CodeType;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class TermCodeSystemStorageSvcTest extends BaseJpaR4Test {
+
+	public static final String URL_MY_CODE_SYSTEM = "http://example.com/my_code_system";
+
+	private CodeSystem createCodeSystemWithMoreThan100Concepts() {
```
fil512, line 15: nitpick: Convention is to put helper methods below the place where they are first used.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+		ResourceTable codeSystemResourceEntity = (ResourceTable)myCodeSystemDao.create(upload, mySrd).getEntity();
+
+		runInTransaction(() -> myTermCodeSystemStorageSvc.storeNewCodeSystemVersionIfNeeded(upload, codeSystemResourceEntity));
+
+		myTerminologyDeferredStorageSvc.setProcessDeferred(true);
+		myTerminologyDeferredStorageSvc.saveDeferred();
+		myTerminologyDeferredStorageSvc.saveDeferred();
+
+		assertEquals(125, myTermConceptDao.count());
```
fil512, line 40: Some comments in this test would help clarify the race condition it's testing

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermDeferredStorageSvcImpl.java
```diff
@@ -142,7 +145,12 @@ private void processDeferredConcepts() {
 		ourLog.info("Saving {} deferred concepts...", count);
 		while (codeCount < count && myDeferredConcepts.size() > 0) {
 			TermConcept next = myDeferredConcepts.remove(0);
-			codeCount += myCodeSystemStorageSvc.saveConcept(next);
+			if(myCodeSystemVersionDao.findById(next.getCodeSystemVersion().getPid()).isPresent()) {
```
fil512, line 148: This doesn't eliminate the race condition, but greatly reduces it.  If both the findById and save are in the same transaction with the right isolation level, that last gap could be closed.  Or it might be cleaner to just put the save in a try block so one failure doesn't interfere with the following saves in the loop.

I don't think it's worth the effort, but if you really wanted to test this final tiny race condition gap, you could create a version of your test where your mock returns a valid findById but then by the time the save is called it is no longer valid.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/term/TermDeferredStorageSvcImplTest.java
```diff
@@ -38,16 +44,45 @@ public void testSaveDeferred_Concept() {
...
+	@Test
+	public void testSaveDeferred_Concept_StaleCodeSystemVersion() {
+		TermConcept concept = new TermConcept();
+		concept.setCode("CODE_A");
+
+		TermCodeSystemVersion myTermCodeSystemVersion = new TermCodeSystemVersion();
+		concept.setCodeSystemVersion(myTermCodeSystemVersion);
+
+		TermDeferredStorageSvcImpl svc = new TermDeferredStorageSvcImpl();
+		svc.setTransactionManagerForUnitTest(myTxManager);
```
fil512, line 75: Another option you can use for this pattern is rather than using setXYZForUnitTest is you can use the MockBean pattern.  See `SearchParamRegistryImplTest` for an example.  Totally optional, but what I like about this pattern is it means you can remove all the `setXYZForUnitTest` methods.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
IanMMarshall: Good point - forgot my Clean Code principles here...

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
IanMMarshall: Good point. Done.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermDeferredStorageSvcImpl.java
IanMMarshall: Good call. I will use the try block per your suggestion. It's a little more blunt than what I had, but it should be cleaner and perform better than my original solution.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/term/TermDeferredStorageSvcImplTest.java
IanMMarshall: I like this idea, but found that it got complicated rather quickly (a surprisingly large number of dependencies that would need to be mocked here to make this work). Will note this for later.

PR #1966, opened 2020-07-04
Fix GraphQL Array of Argument should be ORed

ibrohimislam:
the fix provide IQueryParameterOr instead of IQueryParameterType to params.add.


jamesagnew:
Hi @ibrohimislam - Thanks for the pull request! Can you please add a unit test (presumably to `JpaGraphQLR4ProviderTest.java`) verifying the fix?

ibrohimislam:
I've added some tests. adding test to `JpaGraphQLR4ProviderTest.java` didn't test the `JpaStorageServices`, it have its own `JpaGraphQLR4ProviderTest.MyStorageServices`.

but, I still add `testGraphSystemArrayArgumentList` on it to ensure the argument parse is correct.
and, I add integration test on `JpaStorageServicesTest` to ensure `JpaStorageServices` querying to dao using OR instead of AND.

jamesagnew:
@ibrohimislam this is awesome, appreciate you adding the tests!

It looks great, I'm merging it now.

PR #1971, opened 2020-07-07
Avoid version conflicts on current writes

jamesagnew:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -932,6 +964,7 @@ protected void preDelete(T theResourceToDelete, ResourceTable theEntityToDelete)
 	}
 
 	@Override
+	@Transactional
```
tuomoa, line 967: @jamesagnew If i'm not wrong the read methods in the DAO could be `@Transactional(readonly = true)` which would allow to do reads from replica database nodes? WDYT?

jamesagnew: Makes sense to me

(Although I have to say it's just chance I happened to see this message, I don't typically see comments on 2-year-old PRs :) Definitely recommend posting to the mailing list)

PR #1979, opened 2020-07-13
Add migration for package change

tadgh:
Bulk Export package change caused an error in migration in Quartz. Quartz maintains class names of its jobs in persisted form, so when a job changes packages, it is necessary to migrate any data to reflect this new package. This results in 3 tables being modified: 
* `QRTZ_JOB_DETAILS`
* `QRTZ_SIMPLE_TRIGGERS`
* `QRTZ_TRIGGERS`

Which each reference that fully qualified class name in some way.


tadgh:
This ended up being solved elsewhere, not in HAPI.

PR #1985, opened 2020-07-16
Update to EMPI Docs

tadgh:
Small update to EMPI docs to reflect what the code actually does.


tadgh:
This document will be superseded by the MDM refactor

PR #1994, opened 2020-07-23
Fix mssql errors

michelgleeson:


hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/DropIndexTask.java
```diff
@@ -117,7 +117,7 @@ protected void generateHashCode(HashCodeBuilder theBuilder) {
 				case DERBY_EMBEDDED:
 				case ORACLE_12C:
 				case MSSQL_2012:
-					sql.add("alter table " + theTableName + " drop constraint " + theIndexName);
+					sql.add("drop index " + theIndexName + " on " + theTableName);
```
michelgleeson, line 120: The old drop index statement mixed with the new method to create unique indexes removes the fact that we also create a unique constraint on the column too

michelgleeson:
> Looks good to me- Would be good to add a changelog entry for this (have a look [here](https://github.com/jamesagnew/hapi-fhir/tree/master/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_1_0) for examples of how that works)

Cool! anything special I need to consider if this resolves CDR issues?

jamesagnew:
Nope, even just "An issue in the migrator when applying migrations against
some versions of MSSQL was resolved" is probably good enough

PR #1997, opened 2020-07-24
Allow empty candidateSearchParams in EMPI module config. 

### LINKED ISSUE #1996: $match operation on EMPI with missing candidateSearchParams
**Describe the bug**
When using EMPI module, if the `candidateSearchParams` are left empty in the JSON rules, the `$match` operation returns zero results. Empty `candidateSearchParams` should cause an empty search parameter map to be used.

**To Reproduce**
*TBD*

**Expected behavior**
`$Match` successfully matches. 

 - 5.1.0-SNAPSHOT
 - OS: Manjaro 19
- OpenJDK 11.0.6

tadgh:
Allows EMPI module to have an empty `candidateSearchParams` section of config. Closes #1996 

The original issue was that if empi-rules.json contained an empty candidateSearchParams, field, like so: 
```json
	"candidateSearchParams": []
```
the `EmpiCandidateSearchCriteriaBuilderSvc#buildResourceQueryString` function would simply skip performing any queries at all, as the original logic looped over them for populating candidates. 

Solution: 
1. Logical flow checking for this edge case, and when we detect it, we execute a single search using only the filter criteria. 
2. Do not return early from `buildResourceQueryString` if we detect an empty/null EmpiResourceSearchParamJson. 
3. Added two tests to ensure that null candidate search params do not return empty optionals, but instead return a FHIR query with only the filters as query parameters. e.g. `Patient?active=true`
4. As an aside, fixed a potential NPE with the EmpiResourceSearchParamJson.

However this raises a question: 
What scenario would be actually want to return this empty optional. In the case where the resource being matched doesnt have any of the values that the EmpiResourceSearchParamJson wants? e.g. if i have a patient with name 'gary', but no birthdate, and i have a search param saying

```json
{
	"resourceType": "Patient",
	"searchParams": ["birthdate"]
}
do I actually want it to return Zero candidates? if so, the existing early-exit logic is valid. However, what if I still want to match that person? Will they be unmatchable since they have no birthdate? Something to consider. What do you think, @fil512

### DIFF
--- a/hapi-fhir-jpaserver-empi/src/main/java/ca/uhn/fhir/jpa/empi/svc/candidate/EmpiCandidateSearchCriteriaBuilderSvc.java
+++ b/hapi-fhir-jpaserver-empi/src/main/java/ca/uhn/fhir/jpa/empi/svc/candidate/EmpiCandidateSearchCriteriaBuilderSvc.java
@@ -46,16 +46,17 @@ public class EmpiCandidateSearchCriteriaBuilderSvc {
 	public Optional<String> buildResourceQueryString(String theResourceType, IAnyResource theResource, List<String> theFilterCriteria, EmpiResourceSearchParamJson resourceSearchParam) {
 		List<String> criteria = new ArrayList<>();
 
-		resourceSearchParam.iterator().forEachRemaining(searchParam -> {
-			//to compare it to all known PERSON objects, using the overlapping search parameters that they have.
-			List<String> valuesFromResourceForSearchParam = myEmpiSearchParamSvc.getValueFromResourceForSearchParam(theResource, searchParam);
-			if (!valuesFromResourceForSearchParam.isEmpty()) {
-				criteria.add(buildResourceMatchQuery(searchParam, valuesFromResourceForSearchParam));
-			}
-		});
-		if (criteria.isEmpty()) {
-			return Optional.empty();
+		//If there is no candidateSearchParams, then we want to just use the filters.
+		if (resourceSearchParam != null) {
+			resourceSearchParam.iterator().forEachRemaining(searchParam -> {
+				//to compare it to all known PERSON objects, using the overlapping search parameters that they have.
+				List<String> valuesFromResourceForSearchParam = myEmpiSearchParamSvc.getValueFromResourceForSearchParam(theResource, searchParam);
+				if (!valuesFromResourceForSearchParam.isEmpty()) {
+					criteria.add(buildResourceMatchQuery(searchParam, valuesFromResourceForSearchParam));
+				}
+			});
 		}
+
 		criteria.addAll(theFilterCriteria);
 		return Optional.of(theResourceType + "?" +  String.join("&", criteria));
 	}
--- a/hapi-fhir-jpaserver-empi/src/main/java/ca/uhn/fhir/jpa/empi/svc/candidate/EmpiCandidateSearchCriteriaBuilderSvc.java
+++ b/hapi-fhir-jpaserver-empi/src/main/java/ca/uhn/fhir/jpa/empi/svc/candidate/EmpiCandidateSearchCriteriaBuilderSvc.java
@@ -56,7 +56,6 @@ public class EmpiCandidateSearchCriteriaBuilderSvc {
 					criteria.add(buildResourceMatchQuery(searchParam, valuesFromResourceForSearchParam));
 				}
 			});
-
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Comment by fil512** [comment]
Please add a test with it's own empi-rules json with an empty [] candidateSearchParams and verify that $match returns matches

File: hapi-fhir-jpaserver-empi/src/main/java/ca/uhn/fhir/jpa/empi/svc/candidate/EmpiCandidateSearchCriteriaBuilderSvc.java [hunk 1]
```diff
@@ -43,19 +44,25 @@
...
+		if (resourceSearchParam != null) {
+			resourceSearchParam.iterator().forEachRemaining(searchParam -> {
+				//to compare it to all known PERSON objects, using the overlapping search parameters that they have.
+				List<String> valuesFromResourceForSearchParam = myEmpiSearchParamSvc.getValueFromResourceForSearchParam(theResource, searchParam);
+				if (!valuesFromResourceForSearchParam.isEmpty()) {
+					criteria.add(buildResourceMatchQuery(searchParam, valuesFromResourceForSearchParam));
+				}
+			});
+
+			if (criteria.isEmpty()) {
```
fil512, line 60: this doesn't look right to me.  if there are no criteria, then we still need to add the filter to cover the case where there is a filter but no criteria

hapi-fhir-jpaserver-empi/src/main/java/ca/uhn/fhir/jpa/empi/svc/candidate/EmpiCandidateSearchSvc.java
```diff
@@ -75,18 +75,23 @@ public EmpiCandidateSearchSvc() {
...
-
 		List<EmpiFilterSearchParamJson> filterSearchParams = myEmpiConfig.getEmpiRules().getCandidateFilterSearchParams();
-
 		List<String> filterCriteria = buildFilterQuery(filterSearchParams, theResourceType);
+		List<EmpiResourceSearchParamJson> candidateSearchParams = myEmpiConfig.getEmpiRules().getCandidateSearchParams();
 
-		for (EmpiResourceSearchParamJson resourceSearchParam : myEmpiConfig.getEmpiRules().getCandidateSearchParams()) {
+		//If there are zero EmpiResourceSearchParamJson, we end up only making a single search, otherwise we
+		//must perform one search per EmpiResourceSearchParamJson.
+		if (candidateSearchParams == null || candidateSearchParams.isEmpty()) {
```
fil512, line 84: is it possible for candidateSearchParams to ever be null?  I thought we validated this on json deserialization, but I could be wrong.

hapi-fhir-jpaserver-empi/src/test/java/ca/uhn/fhir/jpa/empi/svc/EmpiCandidateSearchCriteriaBuilderSvcTest.java
```diff
@@ -67,4 +69,21 @@ public void testIdentifier() {
 		assertTrue(result.isPresent());
 		assertEquals(result.get(), "Patient?identifier=urn:oid:1.2.36.146.595.217.0.1|12345");
 	}
+
+	@Test
+	public void testOmittingCandidateSearchParamsIsAllowed() {
```
fil512, line 74: these two tests definitely seem to cover the key issue

hapi-fhir-jpaserver-empi/src/main/java/ca/uhn/fhir/jpa/empi/svc/candidate/EmpiCandidateSearchCriteriaBuilderSvc.java
tadgh: This isn't that case. This logical block says:  If there _are_ candidate search params, but the resource has no value for _any_ of them, leave early. That's what this block is doing. The case where there _are no_ candidate search params is covered by the null check on L51

tadgh: I've removed the null check and instead just do an emptiness check

PR #2018, opened 2020-08-05
Consent bug

jvitrifork:
unit test that shows that the paging and consent service has some issues with paging and total


jamesagnew:
Closing this only because I've pulled it into a working branch.

PR #2037, opened 2020-08-11
Add test, fix code to handle empty candidate search param test

tadgh:
A use case was omitted from the original implementation of the filtering service, specifically when candidateSearchParams was empty.


tadgh:
This was already dealt with by 09d09233ff0e6e50e3875d2f307c50d0a13bd716

Issue #1940, opened 2020-06-23
HAPI FHIR does not support stateless pagination


sqshq:
**Describe the issue**

[FHIR pagination documentation states](https://www.hl7.org/fhir/http.html#paging), that server may or may not use stateful pagination:

> A server MAY add additional state tracking parameters to the links, as shown in the example above, though the server need not use a stateful paging method as shown in this example - it is at the discretion of the server how to best ensure that the continuation retains integrity in the context of ongoing changes to the resources

Currently HAPI FHIR _forces_ to use stateful pagination and hides initial search parameters in next/prev links, so there is no way to get them except saving on the server side. That complicates implementation for most common cases, when each page request can be retrieved from the database with no need for full resultset cache or continuation control. To retrieve the search metadata by request id in multi-node environment, we need to introduce additional infrastructure (e.g. cache with TTL support) or use sticky sessions, while parameters can be available in the user request.

That is exactly the reason why REST is [intended to be stateless](https://restfulapi.net/statelessness/).

It would be very beneficial to have stateless pagination option, which would not remove initial search parameters from the next/prev links

tuomoa:
I think the problem is more or less the whole search implementation. I've been working on some initial proof-of-concept of the stateless paging on the db level in !2059.

Issue #536, opened 2017-01-11
Pagination model not scalable


RJFoster:
The pagination model used by HAPI does not seem likely to scale well in real life situations. The paging model seems to require that information about all resources should be loaded when the search is run and stored somehow, then the subsequent paging requests return resources based on the original stored set. This does not work well due to unnecessary load on the server when the number of search results is large especially when Patient Privacy/consent/etc restrictions are involved. 

In the simple high volume case a search over something like Observation could easily match many thousands of resources, it seems like an unnecessary load on the server if it has to retrieve and store some details (ids) of all those resources on the off chance that the client may want to access further pages.

In the privacy restrictions case, when some resources in a search result-set should not be returned to the requesting user due to privacy restrictions, it means the privacy restrictions need to be checked on all results (expensive!) when the search is performed in order to ensure that IBundleProvider.size() returns an accurate value. If an inaccurate value is returned by size it will mean that the bundle total will be inaccurately returned to the client and the next links may be missing. This may also lead to leakage to the client that some privacy filtering has gone on as the total and actual number retrieved will not match.

Is it possible for future versions of HAPI to provide a more customizable pagination model than it currently does?

RJFoster:
Also the FHIR spec says that caching of previous pages should not be required, however HAPI's model does require this, see https://www.hl7.org/fhir/search.html#2.1.1.5.6

> 2.1.1.5.6 Paging 
> When returning paged results for a search with _include resources, all _include resources that are related to the primary resources returned for the page SHOULD also be returned as part of that same page, even if some of those resource instances have previously been returned on previous pages. This approach allows both sender and receiver to avoid caching results of other pages.

storni:
Hi @RJFoster, did you find a workaround/improvement for this? 

BTW, I've found a way to improve the search result intermediate storage (#866).

jamesagnew:
I'm not sure I understand the issue.

The [IBundleProvider](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/IBundleProvider.java) interface does not mandate that implementations store the entire search result. Certainly the default implementation does, but technically all the implementation needs to do is be able to provide a UUID that uniquely identifies the search and could be used to later fetch more results via the `IPagingProvider`.

[...]

PR #2059, opened 2020-08-31
Improve synchronous searching by providing offset & limit support

tuomoa:
**Describe the issue**
This is related to #1940, #536 

As described HAPI FHIR does not support stateless pagination(searching). Supporting stateless searching would allow using all database nodes for searches. The current stateful search forces to use master node because it needs to write and update the database. Stateless search from the database could also be more performant because it would eliminate the need to write search to the database and update the entry there (and also clean those from the db). Also the asynchronous search fetches more entries than is probably requested (batching).

Current synchronous searching does not support paging and limiting very well. This is an initial poc implementation for better paging support for the synchronous search.

**What's been done**
- Add support for offset querying which leverages paging at the query level
- Add configuration for search default page size and search maximum page size
- If using offset, always use synchronous searching to avoid extra database insert/update etc.
- When using offset, only calculate count if it's wanted
- Validate params closer to provider and handle size if search returned "all" (or last page)
Simple use cases are tested to be working. If both database backed paging and this is supported, then if there is `_offset` parameter then offset paging links are given (synchronous paging) and if not then database backed paging links are given (with page id).

**What's missing**
This is still heavily work-in-progress and therefore I'm marking this as draft. There are  issues with lots of tests that are using for example `IBundleProvider.size()` for verifications, but if we don't want to force searching to query count from the database, that field is null (if we do not fetch all or the last page of resources). I'm not really sure how this should be fixed though.. size is marked nullable, and if one resource is fetched with count 1 it's not known if there is more or not. (We could add total accurate parameter to tests). Also the bundleprovider size -> bundle total is not always correctly added if accurate count is not requested.

I haven't looked into the `$everything` operations. There might be some other issues as well.

I'm willing to put effort into this if this is something that could be merged into the HAPI FHIR library.


tuomoa:
@jamesagnew would you have time to take a look at this and let me know what do you think? If this could be something that would be considered to be merged if work would be finished? :) at the moment this is pretty much proof of concept type of change.

hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/gclient/IQuery.java
```diff
@@ -49,6 +49,14 @@
 	 */
 	IQuery<Y> count(int theCount);
 
+	/**
+	 * Specifies the <code>_offset</code> parameter, which indicates to the server the offset of the query. Use
+	 * with {@link #count(int)}.
```
jamesagnew, line 54: It should defnitely be noted in the javadoc that `_offset` is a nonstandard FHIR parameter and may not be supported on all servers.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
```diff
@@ -200,7 +210,18 @@ IBaseResource createBundleFromBundleProvider(IRestfulServer<?> theServer, Reques
 		String linkPrev = null;
 		String linkNext = null;
 
-		if (isNotBlank(theResult.getCurrentPageId())) {
+		if (theServer.getPagingProvider() == null || requestOffset != null) {
```
jamesagnew, line 213: I'm not sure I love the idea of this being the default if no paging provider exists- This feels like a significant change in behaviour from the current default.

What if we modified IPagingProvider to have a new method along the lines of

```java
default boolean supportsOffsetQueries() { return false; }
```

...and the provider could basically declare itself stateless?

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/method/MethodUtil.java
```diff
@@ -340,6 +340,8 @@ public static StringBuilder createUrl(String theResourceType, Map<String, List<S
 								outerCollectionType);
 					} else if (nextAnnotation instanceof Count) {
 						param = new CountParameter();
+					} else if (nextAnnotation instanceof Offset) {
```
jamesagnew, line 343: This new annotation should probably get a mention in [the search page](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_plain/rest_operations_search.md) and perhaps a section on stateless paging would be good on [the paging page](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_plain/paging.md)

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
tuomoa: My reasoning was that actually one wouldn't need the "extra" paging provider at all if this is used (and maybe if one wants to use database cached paging, it could be implemented later on on top of this? instead of splitting search logic to asynchronous search etc.). And if one did not define the `maximumPageSize` and/or `defaultPageSize` in the RestfulServer this wouldn't change the current behaviour (as the actual query would still fetch everything). What do you think?

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/method/MethodUtil.java
tuomoa: Yeah sure, I'll update the relevant documentation after the implementation is working.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
```diff
@@ -473,9 +474,22 @@ private IBundleProvider executeQuery(String theResourceType, SearchParameterMap
 			RequestPartitionId requestPartitionId = myRequestPartitionHelperService.determineReadPartitionForRequest(theRequestDetails, theResourceType);
 
 			Long count = 0L;
-			if (wantCount) {
+			// TODO not sure if this is good idea or not? synchronous is mostly used on internal operations, maybe count is not needed
+			//  but code has to be fixed. Most operations are looking for one resource. Those that are fetching larger could be  fixed to
+			//  handle size nullability and pages correctly.
+			if (wantCount || theParams.isLoadSynchronous()) {
```
jamesagnew, line 480: I'm not sure I get the issue being fixed here - Synchronous calls are performed when we're trying to minimize database load because we're expecting the call to be made a lot, so forcing a count goes against the whole objective.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchNoHashesTest.java
```diff
@@ -2655,6 +2655,7 @@ public void testSearchWithDate() {
 	}
 
 	@Test
+	// TODO What does this mean? The size in the result is actually wrong if we are hitting the fetchSizeDefaultMaximum?
```
jamesagnew, line 2658: Wrong how? It's incomplete sure, but that's only because you're seeing only the first 5 results as configured.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
tuomoa: That's a valid point, I'll refactor the tests to work without accurate count. There is some tests that are checking that the result bundle contains for example 1 item and size 1, but the size is not known if accurate count was not requested (if query limit wasn't greater than returned items). But I'll see what I can do about this.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchNoHashesTest.java
tuomoa: I think it's okay to just return 5 results, but saying the the search total is also 5 seems to be wrong if there is actually more in the db? If we are hitting the `fetchSizeDefaultMaximum` shouldn't we null the size in bundle provider -> requester will see that I got N entries but there might be more (or not).

When using offset paging, if we would like to actually limit every search to some value, we would need to decrease the query limit when offset is increased which seems to me would just bring more complexity without any real benefit?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
tuomoa: Actually most cases just uses `loadSynchronous` true, and do not define count and therefore the limit is the `fetchSizeDefaultMaximum` (or if it's not defined then everything. I'll refactor this.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4Test.java
```diff
@@ -2919,6 +2919,8 @@ public void testPagingOverEverythingSet() throws InterruptedException {
 	}
 
 	@Test
+	@Disabled
+	// FIXME no idea why this is returning only 6 results. stu2/3 tests are working..
```
tuomoa, line 2923: I'm not really sure why this is not working. I'm not sure but maybe it seems that the actual query returns duplicate pids and when that's combined to `LIMIT 10` we are only getting 6 pids. The same test with DSTU3 works.. Maybe there is some kind of new include param in R4 which causes query to find duplicates or something..

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchNoHashesTest.java
jamesagnew: Ok yeah I could see that. I'd say adjusting the test to check that `size()` returns null in this case is reasonable.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4Test.java
jamesagnew: I checked out this branch and played with this a bit- The issue is that in R4 there are multiple ResourceLink rows created for each Observation (one to satisfy the `subject` parameter, and another to satisfy the `patient` parameter. 

The ultimate SQL that goes to the DB is this:

```sql
select resourceta0_.RES_ID as col_0_0_ from HFJ_RESOURCE resourceta0_ left outer join HFJ_RES_LINK myresource1_ on resourceta0_.RES_ID=myresource1_.SRC_RESOURCE_ID where myresource1_.TARGET_RESOURCE_ID=1 and (resourceta0_.RES_DELETED_AT is null) limit 10
```

..but within the first 10 rows there are some duplicates because of the duplicate ResLink rows, and these get filtered out leaving you with 6 results.

I suspect adding a distinct to the query would fix this, but we have definitely heard lots of reports that distinct is a performance killer on large hapi fhir deployments.. so it would need to appear in the SQL only in cases where paging is turned off

tuomoa: Seems like I was on the right track here, I added the DISTINCT clause only if using synchronous loading (usually when doing offset paging). This way if one is using the `DatabaseBackedPagingProvider` the everything query works as it did before.

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/method/MethodUtil.java
tuomoa: I updated those docs to include mentions about count and offset parameters (and also something about the paging providers).

tuomoa:
@jamesagnew any update on this?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
jamesagnew: I'm going to merge this as is - But I do remain nervous that we're changing behaviour for existing clients. Depending on community feedback we may rework this to require an explicit setting.

jamesagnew (approved):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
tuomoa: All right that sounds good to me. 👍

PR #2064, opened 2020-09-03
Fix mssql drop index and table migrations

michelgleeson:
My idea with the `DropTableTask` change is to `execute` the `DropIndexTask` rather than just generating the sql and running it from `DropTableTask`. This change works, but I'm rather new to best practices so I might be doing something silly :-)

### DIFF
--- a/hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/BaseTask.java
+++ b/hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/BaseTask.java
@@ -60,6 +60,14 @@ public abstract class BaseTask {
 		mySchemaVersion = theSchemaVersion;
 	}
 
+	public String getMyProductVersion() {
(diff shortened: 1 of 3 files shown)


hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/BaseTask.java
```diff
@@ -60,6 +60,14 @@ protected BaseTask(String theProductVersion, String theSchemaVersion) {
 		mySchemaVersion = theSchemaVersion;
 	}
 
+	public String getMyProductVersion() {
```
jamesagnew, line 63: recommend renaming `getMyProductVersion` to `getProductVersion` here and in the method below.

There is a setting in IJ where you can tell is that the `my` in field names is a common prefix and it shouldn't add it to the generated method names like these... I'll show you next time we chat

jamesagnew (approved):
One recommended tweak, but looks awesome!

PR #2073, opened 2020-09-13
Update points.json

alanlesc1:
Update of my contact e-mail and lat./long. points


jamesagnew (approved):
Thanks for submitting. This should deploy to the live site in the next 24-48 hours.

PR #2083, opened 2020-09-15
SearchCoordinatorSvcImpl gets ExecutorService from injected ThreadPoolTaskExecutor

ttntrifork:
This enables support for adding a TaskDecorator (e.g. to add/clear ThreadLocal logging context (MDC) variables).

Here is an example of a useful TaskDecorator that adds logging context to the log statements produced by the search tasks spawned in SearchCoordinatorSvcImpl

```java
import org.apache.logging.log4j.ThreadContext;
import org.springframework.core.task.TaskDecorator;
import org.springframework.stereotype.Component;

/**
 * TaskDecorator that copies the ThreadContext from main thread to task thread.
 */
@SuppressWarnings("unused")
@Component
public class LoggingContextTaskDecorator implements TaskDecorator {
    @Override
    public Runnable decorate(Runnable runnable) {
        var contextMap = ThreadContext.getImmutableContext(); //get main thread context map
        //return new runnable that is executed on task thread
        return () -> {
            try {
                ThreadContext.putAll(contextMap); //set task thread context map
                runnable.run();
            } finally {
                ThreadContext.removeAll(contextMap.keySet());
            }
        };
    }
}

```

The decorator can be added to the ThreadPoolTaskExecutor in user code

```java
    @Bean("searchCoordinatorThreadFactory")
    @Primary
    public ThreadPoolTaskExecutor searchCoordinatorThreadFactory(TaskExecutorBuilder builder, LoggingContextTaskDecorator loggingContextTaskDecorator) {
        return builder
                .threadNamePrefix("search_coord_")
                .taskDecorator(loggingContextTaskDecorator)
                .build();
    }
```

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
@@ -264,7 +264,10 @@ public abstract class BaseConfig {
 
 	@Bean
 	public ThreadPoolTaskExecutor searchCoordinatorThreadFactory() {
-		return new ThreadPoolTaskExecutor();
+		final ThreadPoolTaskExecutor threadPoolTaskExecutor = new ThreadPoolTaskExecutor();
+		threadPoolTaskExecutor.setThreadNamePrefix("search_coord_");
+		threadPoolTaskExecutor.initialize();
+		return threadPoolTaskExecutor;
 	}
 
 	@Bean
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
@@ -263,6 +263,11 @@ public abstract class BaseConfig {
 	}
 
 	@Bean
+	public ThreadPoolTaskExecutor searchCoordinatorThreadFactory() {
+		return new ThreadPoolTaskExecutor();
+	}
+
+	@Bean
 	public TaskScheduler taskScheduler() {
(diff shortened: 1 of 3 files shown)


jamesagnew:
This is super cool - Looks like there are a couple of test failures to look at, but the approach makes perfect sense to me

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
```diff
@@ -262,6 +262,14 @@ public ISearchResultCacheSvc searchResultCacheSvc() {
 		return new DatabaseSearchResultCacheSvcImpl();
 	}
 
+	@Bean
+	public ThreadPoolTaskExecutor searchCoordinatorThreadFactory() {
+		final ThreadPoolTaskExecutor threadPoolTaskExecutor = new ThreadPoolTaskExecutor();
```
tuomoa, line 267: I think this accidentally changes the way thread pool was working before. According to javadoc this acts similar to single thread executor (because core pool size is 1 and queue is unbounded -> it won't increase the thread count). This default leads to pretty large performance degradation because all searches are ran with the same thread. Before this change SearchCoordinatorSvcImpl used unbounded thread pool that always immediately created a new thread.

https://www.baeldung.com/java-threadpooltaskexecutor-core-vs-max-poolsize (and also note the reject policy if queue is full and pool is in maximum)

ping @jamesagnew 

At least I think this should be documented that the thread pool is now acting differently and may need configuration.

ttntrifork: Oh, nicely spotted. This is unintended (not something I was aware of when making this PR). 

Please feel free to make a new PR that fixes this, restoring the original ExecutorService functionality, but still as a Spring ThreadPoolTaskExecutor (such that TaskDecorators can still be used)

PR #2087, opened 2020-09-16
 limit transaction size config

fil512:


jamesagnew (approved):

fil512:
added config param to limit transaction bundle size

PR #2112, opened 2020-10-01
xml:lang attribute prefix gets removed during serialization

oliveregger:
See also issue https://github.com/jamesagnew/hapi-fhir/issues/2111

When defining the language for the div element the lang attribute to be specified with lang and xml:lang for xhtml [reference](https://www.w3.org/International/questions/qa-html-language-declarations#basics) e.g.:

```xml
<div lang="fr" xml:lang="fr" xmlns="http://www.w3.org/1999/xhtml">
```
During serialization to xml or json the xml prefix is removed from the xml:lang attribute and we have two lang="fr" attributes which is invalid.

This PR changes the xml serialization in Xml (since its xhtml it is also used during the json serialization). A testcase is provided to check that serialization and deserialization are consistent.


jamesagnew (approved):

jamesagnew:
Thanks Oliver! Looks good, merging now.

PR #2122, opened 2020-10-07
Fixed loader so that it now sets source system version for ConceptMap…

IanMMarshall:
… resources.


dmuylwyk (approved):
Approved assuming all tests pass and any conflicts are resolved. Thanks, @IanMMarshall!

PR #2125, opened 2020-10-09
BaseClient: make invokeClient available for derived classes

oliveregger:
I'm working on a Validation Client which can execute the $validate operation with a direct POST (see http://hl7.org/fhir/resource-operation-validate.html#examples).This is not directly possible with the existing GenericClient and I would like to base the Validation Client on the Generic Client because all functionality is already implemented. The only piece missing is, that I cannot access the invoke function from the derived class, I need to do:

```java
    MethodOutcome resp = invokeClient(getFhirContext(), binding, clientInvoke, null, null, false, null, null, null, null, theExtraParams);
```
This PR changes the visibility of the invokeClient function to protected so that I can access it from a derived class which is not in the same package.

### DIFF
--- a/hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/BaseClient.java
+++ b/hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/BaseClient.java
@@ -211,15 +211,15 @@ public abstract class BaseClient implements IRestfulClient {
 		myRequestFormatParamStyle = theRequestFormatParamStyle;
 	}
 
-	<T> T invokeClient(FhirContext theContext, IClientResponseHandler<T> binding, BaseHttpClientInvocation clientInvocation) {
+	protected <T> T invokeClient(FhirContext theContext, IClientResponseHandler<T> binding, BaseHttpClientInvocation clientInvocation) {
 		return invokeClient(theContext, binding, clientInvocation, false);
 	}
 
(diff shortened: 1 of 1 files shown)


jamesagnew:
Hi @oliveregger - No concerns with the generally, but I'm curious that we're missing functionality in the client. Could your enhancements be just added directly to the generic client itself?

oliveregger:
@jamesagnew it could be added, however it is a very special case and I'm not yet sure how standard (or generic) conform it is. 

1. the $validate operation has a special syntax defined in the examples: POST /open/Patient/$validate?profile=http://hl7.org/fhir/StructureDefinition/daf-patient however according to the [spec](https://www.hl7.org/fhir/operations.html#request) of the operations this seems not to be valid to add the profile $parameter: 

```
If, when invoking the operation, there is exactly one input parameter of type Resource (irrespective of whether other possible parameters are defined), that the operation can also be executed by a POST with that resource as the body of the request (and no parameters on the url).
```

2. In addition for my Validation Client it would be very helpful to execute the $validate operation on the server level (instead of instance level), however I will need to add first also a feature request for the spec.

So to keep it short, for now, if I could access the invokeClient function directly that would be very helpful. If  I can bring up all the issues in the specs and they get resolved i will volunteer to add this functionality then also to the generic client.

jamesagnew:
@oliveregger sounds good!

jamesagnew (approved):

PR #2135, opened 2020-10-16
Im 20201015 loinc versioned resourceids

IanMMarshall:
When loading LOINC that has a code system version specified, will now first create/update a versioned CodeSystem resource for LOINC with the specified version plus an unversioned instance of the LOINC CodeSystem, also based on the upload which will become the default LOINC version (i.e. the version accessed if no version is specified).


dmuylwyk (approved):
Hi, @IanMMarshall. This looks pretty darn solid to me. Approved assuming CI passes successfully and any conflicts are resolved. Thanks!

PR #2141, opened 2020-10-21
Enable package installer to load non-Conformance resources

IanMMarshall:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -412,6 +419,22 @@ private String extractUniqueUrlFromMetadataResource(IBaseResource resource) {
 		return (String) asPrimitiveType.getValue();
 	}
 
+	private TokenParam extractIdentifierFromOtherResourceTypes(IBaseResource resource) {
+		FhirTerser terser = myFhirContext.newTerser();
+		Identifier myIdentifier = (Identifier) terser.getSingleValueOrNull(resource, "identifier");
```
jamesagnew, line 424: You should probably deal with the `null` case here (and maybe also rename that "my" variable since it's not a field) - I'm not actually sure what would happen here in the case of something that didn't have a URL or an identifier, but it's probably safe to just not upload it in that case.

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
IanMMarshall: Good point. I have changed this to throw an UnsupportedOperationException if the resource has no identifier and added a corresponding JUnit test for this case.

Issue #2148, opened 2020-10-26
EMPI - Patient matching using MPI - return mode, score, and grade


nvg-smile:
**Description**

Current EMPI implementation returns matched patient resource, but no mode, score or grade info is provided with the match. FHIR spec defines MRI $match semantics in the [section 8.1.10 of the spec](https://www.hl7.org/fhir/patient.html). More information about $match operation can be found at https://www.hl7.org/fhir/patient-operation-match.html

**Environment (please complete the following information):**
 - HAPI FHIR Version: DSTU3+
 - OS: N/A
 - Browser: N/A

**Additional context**
hapi-fhir-server-empi module contains EmpiProviderDstu3 and EmpiProviderR4 classes that return matched patient resource, but no additional information about match. These classes and potentially IEmpiMatchFinderSvc need to be updated to provide additional field defined by the spec.

PR #2149, opened 2020-10-27
#2148 Initial implementation

nvg-smile:
Updated EmpiProviderR4 match to include additional fields. Addresses #2148

### DIFF
--- a/hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/api/EmpiMatchOutcome.java
+++ b/hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/api/EmpiMatchOutcome.java
@@ -117,7 +117,12 @@ public final class EmpiMatchOutcome {
 		} else if (score > vector) {
 			return 1.0;
 		}
-		return score / vector;
+
+		double retVal = score / vector;
+		if (retVal < 0) {
+			retVal = 0.0;
+		}
+		return retVal;
 	}
 
 	@Override
--- a/hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/api/EmpiMatchOutcome.java
+++ b/hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/api/EmpiMatchOutcome.java
@@ -105,6 +105,21 @@ public final class EmpiMatchOutcome {
 		return this;
 	}
 
+	/**
+	 * Gets normalized score that is in the range from zero to one
+	 *
+	 * @return
+	 * 	Returns the normalized score
+	 */
+	public Double getNormalizedScore() {
+		if (vector == 0) {
+			return 0.0;
+		} else if (score > vector) {
+			return 1.0;
+		}
+		return score / vector;
+	}
+
 	@Override
 	public String toString() {
 		return new ToStringBuilder(this)
--- a/hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderDstu3.java
+++ b/hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderDstu3.java
@@ -26,6 +26,7 @@ import ca.uhn.fhir.empi.api.IEmpiControllerSvc;
 import ca.uhn.fhir.empi.api.IEmpiExpungeSvc;
 import ca.uhn.fhir.empi.api.IEmpiMatchFinderSvc;
 import ca.uhn.fhir.empi.api.IEmpiSubmitSvc;
+import ca.uhn.fhir.empi.api.MatchedTarget;
 import ca.uhn.fhir.empi.model.EmpiTransactionContext;
 import ca.uhn.fhir.rest.annotation.IdParam;
 import ca.uhn.fhir.rest.annotation.Operation;
@@ -37,6 +38,7 @@ import ca.uhn.fhir.rest.server.servlet.ServletRequestDetails;
 import ca.uhn.fhir.util.ParametersUtil;
 import org.apache.commons.lang3.StringUtils;
 import org.hl7.fhir.dstu3.model.Bundle;
+import org.hl7.fhir.dstu3.model.CodeType;
 import org.hl7.fhir.dstu3.model.DecimalType;
 import org.hl7.fhir.dstu3.model.InstantType;
 import org.hl7.fhir.dstu3.model.Parameters;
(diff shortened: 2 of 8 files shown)


hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderR4.java
```diff
@@ -74,25 +79,50 @@ public EmpiProviderR4(FhirContext theFhirContext, IEmpiControllerSvc theEmpiCont
...
+		if (next.isMatch()) {
+			matchGrade = MatchGrade.CERTAIN;
+		} else if (next.isPossibleMatch()) {
+			matchGrade = MatchGrade.POSSIBLE;
+		}
+		searchComponent.addExtension("http://hl7.org/fhir/StructureDefinition/match-grade",
+			new CodeType(matchGrade.toCode()));
+		return searchComponent;
+	}
+
```
tadgh, line 125: @fil512 What do you think about this implementation of converting match status -> MatchGrade? Since we are being forced to use these matchgrades here in the $match operation, how should these map to our internal assurance levels?

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderDstu3.java
```diff
@@ -77,20 +81,45 @@ public Bundle match(@OperationParam(name=ProviderConstants.EMPI_MATCH_RESOURCE,
...
+			}
+
+			Bundle.BundleEntryComponent entry = new Bundle.BundleEntryComponent();
+			entry.setResource((Resource) next.getTarget());
+			entry.setSearch(toBundleEntrySearchComponent(next));
+
+			retVal.addEntry(entry);
 		}
 
 		return retVal;
```
tadgh, line 104: Are we certain these are ordered in terms of descending score? It looks to me as though these are just in returned order. The spec says: `The response from an MPI $match operation is a set of patient records, ordered from most likely to least likely.` . Can we add a test to prove this is the case? And if it isn't the case, add a sort to the list before returning? This can easily be done via a stream comparator before return

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderDstu3.java
```diff
@@ -77,20 +81,45 @@ public Bundle match(@OperationParam(name=ProviderConstants.EMPI_MATCH_RESOURCE,
...
 	}
 
+	private Bundle.BundleEntrySearchComponent toBundleEntrySearchComponent(MatchedTarget next) {
```
tadgh, line 107: `next` is generally innapropriate unless it is directly involved in a loop. Since this is a helper function, i think a name like `theMatchedTarget` is better. Also, note the convention for naming of the hapi-fhir repository. Method params are prefixed with `the`, member variables are `my`, and static variables are `our`

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderR4.java
```diff
@@ -74,25 +79,50 @@ public EmpiProviderR4(FhirContext theFhirContext, IEmpiControllerSvc theEmpiCont
...
+		searchComponent.setMode(Bundle.SearchEntryMode.MATCH);
+		searchComponent.setScore(next.getMatchResult().getNormalizedScore());
+
+		MatchGrade matchGrade = MatchGrade.PROBABLE;
+		if (next.isMatch()) {
+			matchGrade = MatchGrade.CERTAIN;
+		} else if (next.isPossibleMatch()) {
+			matchGrade = MatchGrade.POSSIBLE;
+		}
+		searchComponent.addExtension("http://hl7.org/fhir/StructureDefinition/match-grade",
```
tadgh, line 121: This String should be ripped into a server constant and reused across both implementations

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/api/EmpiMatchOutcome.java
```diff
@@ -105,6 +105,21 @@ public EmpiMatchOutcome setEidMatch(boolean theEidMatch) {
...
+	 */
+	public Double getNormalizedScore() {
+		if (vector == 0) {
+			return 0.0;
+		} else if (score > vector) {
+			return 1.0;
+		}
+		return score / vector;
+	}
+
```
tadgh, line 122: @fil512 Would appreciate your input on this as well. I _think_ this is correct for normalization, but this needs your eyes on it.

tadgh (changes requested):
Mostly looks great! Couple of requests though, and I would like @fil512 input on a few particularly salient points.

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderR4.java
fil512: It looks like he implemented the mapping I requested

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/api/EmpiMatchOutcome.java
fil512: Nick also checked with me on this ahead of time and it looks like the best option to me

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderDstu3.java
nvg-smile: Great catch, thank you, added sorting to the matched items

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderDstu3.java
nvg-smile: Fixed, thank you for pointing out

hapi-fhir-server-empi/src/main/java/ca/uhn/fhir/empi/provider/EmpiProviderR4.java
nvg-smile: Done, thank you

tadgh (approved):

PR #2151, opened 2020-10-29
Im 20201028 fix versioned valueset expansions

IanMMarshall:
Fixed issue involving ValueSet validate-code operation where a pre-expanded ValueSet includes a versioned CodeSystem. Also fixed bug relating to ValueSet expand operation with URL and Filter parameters.

### DIFF
--- a/hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
+++ b/hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
@@ -21,6 +21,7 @@ package ca.uhn.fhir.jpa.migrate.tasks;
  */
 
 import ca.uhn.fhir.jpa.entity.EmpiLink;
+import ca.uhn.fhir.jpa.entity.TermCodeSystemVersion;
 import ca.uhn.fhir.jpa.entity.TermConceptMap;
 import ca.uhn.fhir.jpa.entity.TermValueSet;
 import ca.uhn.fhir.jpa.migrate.DriverTypeEnum;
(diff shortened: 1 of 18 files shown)


dmuylwyk:
Related ticket: https://github.com/jamesagnew/hapi-fhir/issues/2144

hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
```diff
@@ -157,6 +158,12 @@ protected void init510() {
 		trmValueSet.addColumn("20200923.2", "VER").nullable().type(ColumnTypeEnum.STRING, TermValueSet.MAX_VER_LENGTH);
 		trmValueSet.dropIndex("20200923.3", "IDX_VALUESET_URL");
 		trmValueSet.addIndex("20200923.4", "IDX_VALUESET_URL").unique(true).withColumns("URL", "VER");
+
+		//Term ValueSet Component add system version
+		Builder.BuilderWithTableName trmValueSetComp = version.onTable("TRM_VALUESET_CONCEPT");
+		trmValueSetComp.addColumn("20201028.1", "SYSTEM_VER").nullable().type(ColumnTypeEnum.STRING, TermCodeSystemVersion.MAX_VERSION_LENGTH);
```
jamesagnew, line 164: I'm going to fix this myself in another branch I have open so you don't need to address this comment but I'll comment here:
Better to use the value instead of the constant in the migration task, so that if the constant changes, the migration checker test will catch it.

Also, for the lines below, it's better to not reuse old index names, that's risky if the delete fails for any reason.

jamesagnew (approved):

hapi-fhir-jpaserver-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
IanMMarshall: Ah OK, good to know. I will keep this in mind for next time.

PR #2167, opened 2020-11-10
install-package-locally-without-download

MihawkGevko:


jamesagnew:
The PackageInstallationSpec already accepts a package binary via the `myPackageContents` field - Why not just use that?

PR #2175, opened 2020-11-16
Fixed expand filter issue

frankjtao:
Fixed ValueSet expend filter D.2 issue


IanMMarshall (changes requested):
This appears to be right to me, but I think that similar changes may be required for DSTU3 and R5 as well.

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/StringUtil.java
```diff
@@ -78,4 +79,24 @@ public static String toUtf8String(byte[] theBytes) {
...
 	}
 
+	/**
+	 * Check the input string has prefix token 
+	 * 
+	 * @param theInput       the input string
+	 * @param thePrefix      the prefix string of a token
+	 * @return Return true if a input string token separated by space start with the prefix 
+	 */
+	public static boolean isStartsWithIgnoreCase(final String theInput, final String thePrefix) {
```
dmuylwyk, line 89: Hi, @frankjtao and @IanMMarshall.

What is the issue we're trying to solve here? What was the original request?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/dstu3/FhirResourceDaoValueSetDstu3.java
```diff
@@ -238,7 +240,7 @@ private void applyFilter(IntegerType theTotalElement, List<ValueSetExpansionCont
 	private void addFilterIfPresent(String theFilter, ConceptSetComponent include) {
 		if (ElementUtil.isEmpty(include.getConcept())) {
 			if (isNotBlank(theFilter)) {
-				include.addFilter().setProperty("display").setOp(FilterOperator.EQUAL).setValue(theFilter);
+				include.addFilter().setProperty("display").setOp(FilterOperator.EXISTS).setValue(theFilter);
```
dmuylwyk, line 243: @frankjtao and/or @IanMMarshall, can you explain/justify this change from EQUAL to EXISTS?

Reviewing the FilterOperator value set [here](https://www.hl7.org/fhir/valueset-filter-operator.html), I'm not sure this makes sense. EQUAL seems like the most appropriate fit for a partial-to-complete match. EXISTS simply identifies whether or not the property (in this case the display) is present at all.

I'm calling this out only here; however, I have the same concern for all of the related resource DAOs by FHIR version.

dmuylwyk (changes requested):
I'm fairly certain the filter operator should remain EQUAL instead of EXISTS.

dmuylwyk:
These appear to be the requirements:

> Matching ngrams appropriately inside a word with usage pattern: /ValueSet/VALUE-SET-ID/$expand?filter=
> 
> For a display text of “Body height”:
> searching on "Bo" or "he" should find it
> searching on "ei" or "dy" should not find it

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/StringUtil.java
dmuylwyk: Posted the requirements here:

https://github.com/jamesagnew/hapi-fhir/pull/2175#issuecomment-730392028

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoValueSetR4.java
```diff
@@ -208,7 +210,11 @@ private void applyFilter(IntegerType theTotalElement, List<ValueSetExpansionCont
 
 		for (int idx = 0; idx < theContains.size(); idx++) {
 			ValueSetExpansionContainsComponent next = theContains.get(idx);
-			if (isBlank(next.getDisplay()) || !org.apache.commons.lang3.StringUtils.containsIgnoreCase(next.getDisplay(), theFilter)) {
+			
+			// e.g. For a display text of “Body height”:
+			// searching on "Bo" or "he" should find it
+			// searching on "ei" or "dy" should not find it
+			if (isBlank(next.getDisplay()) || !StringUtil.isStartsWithIgnoreCase(next.getDisplay(), theFilter)) {
```
dmuylwyk, line 217: This change warrants a changelog entry to indicate $expand with filter not evaluates the display with left-matching by string token, case-insensitive.

dmuylwyk (changes requested):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_3_0/2175_valueset_expension_filer_left_match.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+pr: 2175
+title: "When expanding a ValueSet using a filter, now evaluates the display with left-matching
```
dmuylwyk, line 4: Readability.

```suggestion
title: "Expanding a ValueSet using a filter now evaluates the display with left-matching 
```

dmuylwyk (changes requested):
Thanks for the changelog entry, @frankjtao.

I remain concerned about the filter operator (`exists` vs. `=`). My interpretation of the standard leads me to understand `exists` should indicate whether the display property is populated _at all_ regardless of the value of the display property. `=` is used to indicate we're evaluating the actual value of the display property. Put another way: we're saying "the filter value equals some or all of the display property value"; we are not saying "the filter value exists within the display property value".

If there's an issue with using the `EQUAL` operator _in the code_, that's a matter of revisiting the implementation and tests. I think the standard is pretty clear here.

I'd really like @IanMMarshall to weigh in on this when he has the time to investigate and consider.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/dstu3/FhirResourceDaoValueSetDstu3.java
```diff
@@ -224,7 +226,7 @@ private void applyFilter(IntegerType theTotalElement, List<ValueSetExpansionCont
 
 		for (int idx = 0; idx < theContains.size(); idx++) {
 			ValueSetExpansionContainsComponent next = theContains.get(idx);
-			if (isBlank(next.getDisplay()) || !org.apache.commons.lang3.StringUtils.containsIgnoreCase(next.getDisplay(), theFilter)) {
+			if (isBlank(next.getDisplay()) || !StringUtil.isStartsWithIgnoreCase(next.getDisplay(), theFilter)) {
```
IanMMarshall, line 229: After examining the changes in BaseTermReadSvcImpl more closely, I agree that enabling the "myDisplayNGram" in the `addDisplayFilterInexact()` method of BaseTermReadSvcImpl class is the correct solution here. That being said, I am now wondering if there is still a need for the `applyFilter` method in the FhirResourceDaoValueSet classes. I think this would only really be needed now for the case where HibernateSearch is not implemented. 

Currently the BaseTermReadSvcImpl class disallows expand with filter in cases where HibernateSearch is not enabled. As such, I would say maybe remove this for now from the FhirResourceDaoValueSetDstu3, FhirResourceDaoValueSetR4 and FhirResourceDaoValueSetR5 classes.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoValueSetR4.java
```diff
@@ -208,7 +210,8 @@ private void applyFilter(IntegerType theTotalElement, List<ValueSetExpansionCont
 
 		for (int idx = 0; idx < theContains.size(); idx++) {
 			ValueSetExpansionContainsComponent next = theContains.get(idx);
-			if (isBlank(next.getDisplay()) || !org.apache.commons.lang3.StringUtils.containsIgnoreCase(next.getDisplay(), theFilter)) {
+			
+			if (isBlank(next.getDisplay()) || !StringUtil.isStartsWithIgnoreCase(next.getDisplay(), theFilter)) {
```
IanMMarshall, line 214: See comments above in FhirResourceDaoValueSetDao3.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/r5/FhirResourceDaoValueSetR5.java
```diff
@@ -208,7 +210,7 @@ private void applyFilter(IntegerType theTotalElement, List<ValueSetExpansionCont
 
 		for (int idx = 0; idx < theContains.size(); idx++) {
 			ValueSetExpansionContainsComponent next = theContains.get(idx);
-			if (isBlank(next.getDisplay()) || !org.apache.commons.lang3.StringUtils.containsIgnoreCase(next.getDisplay(), theFilter)) {
+			if (isBlank(next.getDisplay()) || !StringUtil.isStartsWithIgnoreCase(next.getDisplay(), theFilter)) {
```
IanMMarshall, line 213: See comments above in FhirResourceDaoValueSetDstu3 and FhirResourceDaoValueSetR4.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
```diff
@@ -1084,7 +1084,7 @@ private void addDisplayFilterInexact(QueryBuilder qb, BooleanJunction<?> bool, V
 			.withSlop(2)
 			.onField("myDisplay").boostedTo(4.0f)
 			.andField("myDisplayEdgeNGram").boostedTo(2.0f)
-			// .andField("myDisplayNGram").boostedTo(1.0f)
+			.andField("myDisplayNGram").boostedTo(1.0f)
```
IanMMarshall, line 1087: I would suggest also adding a test for the expand with filter to the FhirResourceDaoR4SearchWithElasticSearchIT class as well to ensure that this change also works with the Elasticsearch mapping provider.

IanMMarshall (changes requested):
See my comments.

dmuylwyk (commented):
I concur with @IanMMarshall's review. No additional comments from me.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/dstu3/FhirResourceDaoValueSetDstu3.java
dmuylwyk: Resolved. We stuck with EQUAL.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_3_0/2175_valueset_expension_filer_left_match.yaml
dmuylwyk: Resolving. This suggestion was applied.

dmuylwyk (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/dstu3/FhirResourceDaoValueSetDstu3.java
frankjtao: @IanMMarshall No. The `applyFilter ` method can't be removed in this case, since `NGram Analyzer` will return entry e.g.  for the display value of 'Systolic blood pressure' with the filter 'lood'. The `applyFilter` will filter out the non-left matching entries.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoValueSetR4.java
frankjtao: Same as above

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/r5/FhirResourceDaoValueSetR5.java
frankjtao: Same as above

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
frankjtao: Added 3 more test cases to FhirResourceDaoR4SearchWithElasticSearchIT

IanMMarshall (approved):
Looks good. Approved.

PR #2178, opened 2020-11-17
Do not set ValueSet expansion component parameter if HibernateSearch not enabled

IanMMarshall:


dmuylwyk (approved):
Looks good. Nice and simple solution. Thanks, @IanMMarshall!

jamesagnew (approved):

PR #2180, opened 2020-11-19
Fixed NPE in package loader when package description is null

IanMMarshall:
Also renamed tests that should have caught this problem to ensure they get called during CI/CD.


dmuylwyk (approved):
Approved assuming CI completes successfully. Thanks, @IanMMarshall!

PR #2194, opened 2020-11-24
Suppress Authorization header value from the request headers

tuomoa:
Hide sensitive Authorization header value from the response in request headers.


ruoat:
@jamesagnew Do you have any plans to merge this? We have had to use create custom build because of this. Authorization header is usually sensitive information and should not be shown in ui.

jamesagnew (approved):

jamesagnew:
Thanks for the contribution! Merged.

PR #2195, opened 2020-11-24
Always apply vs filters correctly

jamesagnew:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
```diff
@@ -938,12 +946,12 @@ private Boolean expandValueSetHandleIncludeOrExcludeUsingDatabase(IValueSetConce
 		 * DM 2019-08-21 - Processing slows after any ValueSets with many codes explicitly identified. This might
 		 * be due to the dark arts that is memory management. Will monitor but not do anything about this right now.
 		 */
-		BooleanQuery.setMaxClauseCount(10000);
+		BooleanQuery.setMaxClauseCount(SearchBuilder.getMaximumPageSize());
```
tadgh, line 949: This is going to be going away in HS6 far as I'm aware. I'm pretty sure this setting is controlled on the implementing backends now. e.g. lucene has a max clause count, and ES can set a maximum bucket count. Will need to look into this

tadgh (approved):
Looks good, approved.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
jamesagnew: FWIW I don't think that's a huge deal... Affects the speed of valueset pre-expansion for loading loinc, but that is a long process no matter what you do

PR #2191, opened 2020-11-22
Resource change listener

fil512:
Added a new IResourceChangeListenerRegistry service and modified SearchParamRegistry and SubscriptionRegistry to use it.

This service contains an in-memory list of all registered {@link IResourceChangeListener} instances along
with their caches and other details needed to maintain those caches.  Register an {@link IResourceChangeListener} instance
with this service to be notified when resources you care about are changed.  This service quickly notifies listeners
of changes that happened on the local process and also eventually notifies listeners of changes that were made by
remote processes.

PR #2196, opened 2020-11-25
Fixed SubscriptionLoader NPE for a large number of subscriptions

Bugerman58:
In case you have a large number of active subscriptions (e.g. more than 10 000) SubscriptionLoader throws NullPointer Exception while trying to activate those subscriptions.
The reason is that subscriptionBundleList.size(); returns null


Bugerman58:
Closing. PR was actual before #2191

PR #2197, opened 2020-11-25
Changed the names URIs and IDs of a few LOINC concept maps in the LOINC Loader

swagers:
Hi,

We would like to have these changed to match the ids we have named for these concept maps.

Thanks,
Steven Wagers
LOINC


swagers:
Also, let us know if this is the correct process to request these type of changes or if you would prefer a different method.

Thanks again.

jamesagnew (approved):
Thanks @swagers , this is exactly the best way to request these. :) Appreciate the PR.

PR #2211, opened 2020-12-02
Split package: delete 'ca.uhn.fhir.rest.server' package from hapi-fhir-server-empi

bdenton:
The new EMPI work has created a "split package". The package 'ca.uhn.fhir.rest.server' has previously existed in the hapi-fhir-server bundle. This commit moves the class TransactionLogMessages into the 'ca.uhn.fhir.empi.model` package and deletes the split package from the hapi-fhir-server-empi bundle.


tadgh:
Hey, thanks for this PR. I am going to digest this, but at first glance, my thoughts are that the TransactionLogMessages should be held in fhir-rest-server package, as they are potentially of general interest outside of EMPI. I think this split package was a complete accident and oversight on my part. There is nothing EMPI-specific about this class, so there's no reason for it to live here. Would that work for you too? If the whole class was moved into rest-server, and kept its package?

bdenton:
Moving TransactionLogMessages to fhir-rest-server would be totally acceptable. My only concern was to fix the split package.situation. Might also have to add a test or two to fhir-rest-server to keep codecov happy. Is this change something you would have cycles to do fairly soon or would you rather I take care of it?

bdenton:
I decided it was easier to just refactor my original commit to push TransactionLogMessages to a different location... the same classes are changed in the same way.. just different target.. Seems to be passing all of the tests. If this is acceptable, could you please post a review so it can be merged?

tadgh:
Looks good to me!

tadgh (approved):

PR #2217, opened 2020-12-03
OSGi manifest maintenance

bdenton:
In addition to cleaning up some unnecessary "clutter" from the OSGi manifests, this PR also removes the Fragment-Host directives.
Adding "Fragment-Host" to the OSGi manifest is not needed and it short-circuits the normal OSGi class loading architecture. This has been creating issues for the few of us that are trying to build OSGi-based solutions for FHIR


jamesagnew (approved):

jamesagnew:
Makes sense to me!

PR #2225, opened 2020-12-10
Remove all references to Person from EMPI revisited

### LINKED ISSUE #2161: Support Arbitrary resource MDM
Current EMPI implementation is useful for Patients/Practitioners, but there is large demand for Master Data Management over arbitrary resource types, e.g. medication, organization. 

To this end we should: 
1. Drop references to person, and instead rely directly on MPI_LINK table results. 
2. Replace usages of person with an instance of the same MDM'ed class. e.g. Patient links to Patient, Organization links to Organization. Note these links are now only maintained in the MPI_LINK table. 
3. Deprecate some of the MPI_LINK columns
4. Add MDM toggles on a per-resource basis. Maybe in the empi-rules.json itself?
5. Auto-create subscriptions based on these new MDM types.

nvg-smile:
Second attempt at merging EMPI to MDM functionality.  Closes #2161.

Many large scale changes. The primary purpose of this PR is to change the EMPI solution to an MDM solution. In broad strokes, this means that the Person resource is no longer special.

The new solution (MDM) involves creating another resource where Person used to be used. For example, if MDM is occurring on a patient, we will create a new Patient for them, and tag that patient as a Golden Record. This means that several things ahve changed:

Provider methods that used to point to type of Person are now server-level operations in which you specify a resource type.
Link updating and querying no longer rely on Person IDs, but instead on arbitrary resource ids, depending on the resource type you are referring to in MDM.

Change to the EMPI config to require a list of mdmTypes

### DIFF
--- a/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/broker/MdmQueueConsumerLoader.java
+++ b/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/broker/MdmQueueConsumerLoader.java
@@ -52,9 +52,9 @@ public class MdmQueueConsumerLoader {
 		if (myMdmChannel == null) {
 			ChannelConsumerSettings config = new ChannelConsumerSettings();
 			config.setConcurrentConsumers(myMdmSettings.getConcurrentConsumers());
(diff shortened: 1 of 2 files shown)


nvg-smile:
New PR doesn't include the history, replaced with https://github.com/jamesagnew/hapi-fhir/pull/2226 to revert the revert.

PR #2226, opened 2020-12-10
Un-revert the MDM changes, now that issues have been addressed

tadgh:
Reverts jamesagnew/hapi-fhir#2223


fil512 (approved):
blindly approving on faith

PR #2228, opened 2020-12-10
Rz 20201207 remove component scan

RuslanZinovyev:


fil512 (approved):
Thanks for this @RuslanZinovyev !

PR #2231, opened 2020-12-10
Add Azure test slicing for parallel test running. 

tadgh:
Playing with Azure pipelines to see if we can parallelize the test process across modules.


jamesagnew:
Whaaaaaaaaat? This is so cool

azure-pipelines.yml
```diff
@@ -14,34 +15,151 @@ pool:
...
+
+      # Build the project fully without tests. At the end of this job, this should be auto-stored in the cache, for future retrieval by subsequent jobs
+      - task: Maven@3
+        displayName: Build Project
+        env:
+           JAVA_HOME_11_X64: /usr/local/openjdk-11
+        inputs:
+           goals: 'clean install'
+           # These are Maven CLI options (and show up in the build logs) - "-nsu"=Don't update snapshots. We can remove this when Maven OSS is more healthy
+           options: "-P ALLMODULES,JACOCO,ERRORPRONE,CI -e -B -Dmaven.repo.local=$(MAVEN_CACHE_FOLDER) -Dmaven.test.skip"
```
KevinDougan-SmileCDR, line 35: Minor nitpick: If you change "maven.test.skip" to "skipTests", the test code will be compiled but not executed. This might be useful so we can see if there are any compilation problems early, but without taking the hit of actually running the tests.

KevinDougan-SmileCDR (approved):
Approve pending review of minor changes.

PR #2232, opened 2020-12-11
fix np for empty npm package descr

oliveregger:
We have encountered fhir npm packages which do not have descriptions. According to the spec they must [have](https://github.com/jamesagnew/hapi-fhir/compare/master...ahdis:oe_jpa_np_packagecache?expand=1) a description however. 

This PR fixes the NP Exception happening if no description is set and makes hapi more liberal.


jamesagnew:
Thanks for the contribution Oliver!

This issue was actually solved in https://github.com/jamesagnew/hapi-fhir/pull/2180 though..

oliveregger:
@jamesagnew was working on a 5.2.0 branch and did not see it in the commit, sorry for the "doublette"

jamesagnew:
@oliveregger no worries at all. Thanks again :)

PR #2236, opened 2020-12-11
Add support for unregistering components in the IResourceProviderFactory

### LINKED ISSUE #2235: Add support for unregistering components in the IResourceProviderFactory
**Description**
There is currently no support for unregistering components when they are passed between modules. This manifests as bug in  MDM, when MDM configuration is passed to the persistence context, but then never refreshed, even on the module restart. There should be a way to remove a previously registered component in order to avoid this.

**To Reproduce**
1. Instantiate and start MDM module, test it to ensure it matches resources
2. Change MDM settings by updating matching rules and restart the module
3. Test it again to ensure that it uses old matching rules

**Expected behavior**
Expected to see the updated matching rules after the restart of the module.

 - HAPI FHIR Version - R4/DSTU3
 - OS: Any
 - Browser: Any - backend issue

**Additional context**
N/A

nvg-smile:
Closes #2235


hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/provider/MdmProviderLoader.java
```diff
@@ -45,17 +47,30 @@
 	@Autowired
 	private IMdmSubmitSvc myMdmSubmitSvc;
 
+	private BaseMdmProvider myMdmProvider;
+
 	public void loadProvider() {
 		switch (myFhirContext.getVersion().getVersion()) {
 			case DSTU3:
-				myResourceProviderFactory.addSupplier(() -> new MdmProviderDstu3(myFhirContext, myMdmControllerSvc, myMdmMatchFinderSvc, myMdmExpungeSvc, myMdmSubmitSvc));
+				myResourceProviderFactory.addSupplier(() -> {
```
jamesagnew, line 55: This change looks good, but I have one overall question/comment that I'm adding here for no particular reason.

You have a bunch of these MDM Providers for various versions of FHIR. This is a pattern that is used in HAPI FHIR a lot, but we're using it less these days as it's very often possible to create a generic provider that handles all versions of FHIR. See `BaseJpaSystemProviderDstu2Plus` for an example, which supplies the `$mark-all-resources-for-reindexing` operation to all versions of FHIR. 

This approach wasn't possible for a long time which is why it's uncommon in HAPI FHIR, but for anything newer I always try to at least question whether what I'm doing could be written in a cross-version way.

Can these providers work this way?

jamesagnew (approved):

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/provider/MdmProviderLoader.java
tadgh: These Providers mostly return Bundles, which appear to be STU-Specific. Do we have something that is generic that could be returned here?

jamesagnew: We have a class called [TransactionBuilder](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/TransactionBuilder.java) that can build Bundle resources in a version independent way.

It's currently focused on building transaction bundles, but I could see us renaming it to BundleBuilder and adding for working with other types of Bundles. It already looks like it does much of what would be needed to replace the version-specific stuff in the MDM providers.. it'd need `score` and a few other attributes added presumably..

PR #2257, opened 2020-12-17
Fixed the pre-expansion with filter issue

frankjtao:


jamesagnew (approved):
This is a clever solution!

IanMMarshall:
The changes look good to me. Before you merge this though, could also increase the "minGramSize" values as discussed earlier from 2 to 3 for the two "autocompleteWordEdgeAnalyzer" definitions?

PR #2262, opened 2020-12-24
Fix order by clause in search query for MySQL, MSSQL and MariaDB

IanMMarshall:
The SQL `NULLS LAST` and `NULLS FIRST` options for `ORDER BY` clauses are not supported by MySQL, MSSQL or MariaDB. Added special logic to the SearchQueryBuilder class methods to accommodate these three DBs.


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
```diff
@@ -115,6 +119,16 @@ private SearchQueryBuilder(FhirContext theFhirContext, ModelConfig theModelConfi
 		mySqlBuilderFactory = theSqlBuilderFactory;
 		myCountQuery = theCountQuery;
 		myDialect = theDialect;
+		if (myDialect instanceof org.hibernate.dialect.MySQL57Dialect){
```
jamesagnew, line 122: It's probably better to avoid the version specific dialects..

I think you can get away with just checking whether the dialect is instanceof `MySQLDialect` (which covers all versions of MySQL as well as MariaDB) or `SQLServerDialect` (which covers all versions of SQL Server).

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
```diff
@@ -550,12 +564,87 @@ public boolean haveAtLeastOnePredicate() {
...
+	}
+
+	public void addSortString(DbColumn theTheColumnValueNormalized, boolean theTheAscending, OrderObject.NullOrder theNullOrder) {
+		if ((dialectIsMariaDb || dialectIsMySql || dialectIsMsSql)) {
+			// MariaDB, MySQL and MSSQL do not support "NULLS FIRST" and "NULLS LAST" syntax.
+			// Null values are always treated as less than non-null values.
+			// As such special handling is required here.
+			String direction = theTheAscending ? " ASC" : " DESC";
+			String sortColumnName = theTheColumnValueNormalized.getTable().getAlias() + "." + theTheColumnValueNormalized.getName();
+			if ((theTheAscending && theNullOrder == OrderObject.NullOrder.LAST)
```
jamesagnew, line 589: I'm worried about the performance implications of adding this coalesce here when operating over large datasets... I think my recommendation would be to just add the sort without any null ordering on DBs that don't support it.

FYI it looks like hibernate has an interesting alternate approach for accomplishing the same thing: https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/dialect/MySQLDialect.java#L467 -- But I suspect the most performance sensitive thing is to just not use null ordering.

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
IanMMarshall: Good point. Implemented suggested changes.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
IanMMarshall: Once again, good point. I took a look at the link you provided above and coded and tested a similar approach (i.e. embedding a case statement into the order by clause) and confirmed that this could also work. However per your concerns about performance, I have opted to comment out the special order by clause for now (keeping the tested code in place in case for future reference if needed). 

The modified code will sort without any null ordering for MariaDB, MSSQL and MySQL.

PR #2271, opened 2021-01-05
Fixed package loading when partitioning is enabled

IanMMarshall:
The loader was throwing a nullpointer exception when it tried to persist the binary artifacts relating to a package. This fix addresses that problem by ensuring that the binary artifacts are persisted to the DEFAULT partition along with any conformance resources installed with the package.


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageSystemRequestDetails.java
```diff
@@ -0,0 +1,35 @@
...
+		@Override
+		public Object callHooksAndReturnObject(Pointcut thePointcut, HookParams theParams) {
+			return null;
+		}
+
+		@Override
+		public boolean hasHooks(Pointcut thePointcut) {
+			return false;
+		}
+	}
```
tadgh, line 33: Im confused as to how this requestdetails object ensures the default partition is used. Is there something downstream that normally reads partition from a SystemRequestDetails object?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/RequestPartitionHelperSvc.java
```diff
@@ -91,7 +97,7 @@ public RequestPartitionId determineReadPartitionForRequest(@Nullable RequestDeta
 
 		if (myPartitionSettings.isPartitioningEnabled()) {
 			// Handle system requests
-			if (theRequest == null && myPartitioningBlacklist.contains(theResourceType)) {
+			if ((theRequest == null && myPartitioningBlacklist.contains(theResourceType)) || theRequest instanceof SystemRequestDetails) {
```
tadgh, line 100: ah i see

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/RequestPartitionHelperSvc.java
```diff
@@ -123,7 +129,7 @@ public RequestPartitionId determineCreatePartitionForRequest(@Nullable RequestDe
...
```
tadgh, line 132: instead of SystemRequestDetails, should this not be PackageSystemRequestDetails? Or am I missing something

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/SystemRequestDetails.java
```diff
@@ -0,0 +1,134 @@
...
+		public boolean isDefaultPrettyPrint() {
+			return false;
+		}
+
+		@Override
+		public IInterceptorService getInterceptorService() {
+			return null;
+		}
+	}
+}
```
tadgh, line 134: I'm a bit confused as to why there are 2 classes at all. (PackageSystemRequestDetails and SystemRequestDetails) when it looks like this could safely be a single class

tadgh (approved):
Approved pending a few questions

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageSystemRequestDetails.java
IanMMarshall: Eliminating this class per your subsequent comments as I now realize that this is now redundant. Will use the SystemRequestDetails class instead.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/RequestPartitionHelperSvc.java
IanMMarshall: Per my comment above, I am going to instead use SystemRequestDetails class everywhere rather than having a separate extension for package registry.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/SystemRequestDetails.java
IanMMarshall: Agreed. I originally thought that I would need a specialized version of the RequestDetails class for the package registry and so I created SystemRequestDetails so as to separate the partition functionality from package registry functionality. However I see now that the package registry implementation is actually pretty generic and so it would make sense to combine these into one class.

PR #2272, opened 2021-01-06
Cleanup after deleting unneeded code

JPercival:
* Upgrade to latest cql-evaluator snapshot
* Remove unused classes
* Fix discrepancies between some R4 and DSTU3 code


hapi-fhir-jpaserver-cql/pom.xml
```diff
@@ -19,9 +19,9 @@
 	<properties>
 		<!-- FIXME KBD: Change this to a cqf-tooling released version (e.g. 1.3.1) before finalizing this work! kevin.dougan@smilecdr.com -->
 		<cqf-tooling.version>1.3.1-SNAPSHOT</cqf-tooling.version>
-		<cql-engine.version>1.5.0</cql-engine.version>
-		<cql-evaluator.version>1.0.0</cql-evaluator.version>
-		<cqframework.version>1.5.0</cqframework.version>
+		<cql-engine.version>1.5.1-SNAPSHOT</cql-engine.version>
+		<cql-evaluator.version>1.1.0-SNAPSHOT</cql-evaluator.version>
+		<cqframework.version>1.5.1-SNAPSHOT</cqframework.version>
```
KevinDougan-SmileCDR, line 24: Was this intentional to commit the changes to use -SNAPSHOT versions?
If so, then it means that I would also need to build all these components locally first, is that correct?

JPercival: No, we publish SNAPSHOTS off of our develop branches automatically. All those should exist in Maven.

KevinDougan-SmileCDR: Maybe I just need to add some other repo to my local maven settings.xml file to gain access to these SNAPSHOT releases?

JPercival: Looks like https://oss.sonatype.org/content/repositories/snapshots is already enabled in the root hapi-fhir pom. Is there a particular SNAPSHOT that you're seeing as missing?

KevinDougan-SmileCDR: No problem - my bad - I forgot to also change the artifactId down below in the POM.

PR #2273, opened 2021-01-07
Update tmpl-head.html

janol77:
wrong path in library select2 makes a 404 status code


jamesagnew (approved):

jamesagnew:
Thanks for the fix!

ruoat:
This breaks Hapi 5.3.0 version of select2. You can go to http://hapi.fhir.org/ and see select2 load failing

PR #2281, opened 2021-01-10
added IPagingProvider.canSearchByOffset()

fil512:
Also consolidated some long parameter lists

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/IRestfulServer.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/IRestfulServer.java
@@ -33,7 +33,7 @@ public interface IRestfulServer<T extends RequestDetails> extends IRestfulServer
 
 	PreferReturnEnum getDefaultPreferReturn();
 
-	default boolean canStoreSearchResults() {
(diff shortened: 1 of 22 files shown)


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/IRestfulServer.java
```diff
@@ -33,4 +33,7 @@
 
 	PreferReturnEnum getDefaultPreferReturn();
 
+	default boolean canStoreSearchResults() {
```
jamesagnew, line 36: Is this right? If a server CAN search by offset, that means we don't need to store the results and assign a UUID. So shouldn't this be 

```java
return getPagingProvider() != null && !getPagingProvider().canSearchByOffset();
```

jamesagnew (approved):
One comment but overall this is great!

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/IRestfulServer.java
fil512: good catch!  I changed the meaning and forgot to rename the helper function!

PR #2283, opened 2021-01-12
Fixed the migration scripts

frankjtao:


fil512 (approved):
Thanks for turning this around so fast Frank!

PR #2288, opened 2021-01-15
Added an extra check for an IS NOT NULL Constraint when checking an Oracle DB Column for Nullability

KevinDougan-SmileCDR:
Tested locally manually


KevinDougan-SmileCDR:
This was tested by running several migration scenarios manually.

jamesagnew (approved):

PR #2300, opened 2021-01-18
Add bean to config

### LINKED ISSUE #2299: Missing BulkExportDaoSvc Bean in application context
Attempting a bulk export without manually creating a BulkExportSvcDao bean causes the operation to fail as the steps of the job cannot fetch the bean. 

Solution is to just add this bean alongside the job config.

tadgh:
Closes #2299 

* Add the Bean to the BulkExportJobConfig.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
@@ -19,7 +19,6 @@ import ca.uhn.fhir.jpa.binstore.BinaryStorageInterceptor;
 import ca.uhn.fhir.jpa.bulk.api.IBulkDataExportSvc;
 import ca.uhn.fhir.jpa.bulk.provider.BulkDataExportProvider;
 import ca.uhn.fhir.jpa.bulk.svc.BulkDataExportSvcImpl;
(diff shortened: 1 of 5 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
```diff
@@ -19,6 +19,7 @@
 import ca.uhn.fhir.jpa.bulk.api.IBulkDataExportSvc;
 import ca.uhn.fhir.jpa.bulk.provider.BulkDataExportProvider;
 import ca.uhn.fhir.jpa.bulk.svc.BulkDataExportSvcImpl;
+import ca.uhn.fhir.jpa.bulk.svc.BulkExportDaoSvc;
```
IanMMarshall, line 22: nitpick: This import was added but does not seem to be needed now in this class. Possibly from an earlier change that was later discarded?

IanMMarshall (approved):
Other than the one nitpick, looks good. Approved.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/BaseConfig.java
tadgh: Righto, fixiing

PR #2301, opened 2021-01-19
Disable docker tests in non docker

tadgh:
Create a jupiter extension which tests if docker is available on the host, and apply this extension to all related test classes. If docker is not present, skip the tests.


IanMMarshall (approved):
Looks good. Approved.

PR #2302, opened 2021-01-19
Improved DateQuery and removed 'OR' in the SQL

frankjtao:
Removed OR for date query for gt/ge/le/lt to improve the performance.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
@@ -152,6 +152,7 @@ public class PredicateBuilderDate extends BasePredicateBuilder implements IPredi
 		return theDateParam == null || theDateParam.getPrecision().ordinal() == TemporalPrecisionEnum.DAY.ordinal();
 	}
 
+	@SuppressWarnings("unchecked")
 	private Predicate createPredicateDateFromRange(CriteriaBuilder theBuilder,
 																  From<?, ResourceIndexedSearchParamDate> theFrom,
 																  DateRangeParam theRange,
@@ -191,36 +192,75 @@ public class PredicateBuilderDate extends BasePredicateBuilder implements IPredi
 		}
 
 		if (operation == SearchFilterParser.CompareOperation.lt) {
-			if (lowerBoundInstant == null) {
-				throw new InvalidRequestException("lowerBound value not correctly specified for compare operation");
-			}
-			//im like 80% sure this should be ub and not lb, as it is an UPPER bound.
-			lb = theBuilder.lessThan(theFrom.get(lowValueField), genericLowerBound);
-		} else if (operation == SearchFilterParser.CompareOperation.le) {
-			if (upperBoundInstant == null) {
-				throw new InvalidRequestException("upperBound value not correctly specified for compare operation");
-			}
-			//im like 80% sure this should be ub and not lb, as it is an UPPER bound.
-			lb = theBuilder.lessThanOrEqualTo(theFrom.get(highValueField), genericUpperBound);
-		} else if (operation == SearchFilterParser.CompareOperation.gt) {
-			if (upperBoundInstant == null) {
-				throw new InvalidRequestException("upperBound value not correctly specified for compare operation");
-			}
-			lb = theBuilder.greaterThan(theFrom.get(highValueField), genericUpperBound);
-			} else if (operation == SearchFilterParser.CompareOperation.ge) {
-				if (lowerBoundInstant == null) {
-					throw new InvalidRequestException("lowerBound value not correctly specified for compare operation");
+			// use lower bound first
+			if (lowerBoundInstant != null) {
+				if (lowerBound.getPrefix() == ParamPrefixEnum.EQUAL)
+					lb = theBuilder.lessThan(theFrom.get(lowValueField), genericLowerBound);
+				else
+					// the value has been reduced one in this case
+					lb = theBuilder.lessThanOrEqualTo(theFrom.get(lowValueField), genericLowerBound);
+			} else {
+				if (upperBoundInstant != null) {
+					if (upperBound.getPrefix() == ParamPrefixEnum.EQUAL)
+						ub = theBuilder.lessThan(theFrom.get(lowValueField), genericUpperBound);
+					else
+						// the value has been reduced one in this case
+						ub = theBuilder.lessThanOrEqualTo(theFrom.get(lowValueField), genericUpperBound);						
+				} else {
+					throw new InvalidRequestException("lowerBound and upperBound value not correctly specified for compare theOperation");
 				}
-				lb = theBuilder.greaterThanOrEqualTo(theFrom.get(lowValueField), genericLowerBound);
-			} else if (operation == SearchFilterParser.CompareOperation.ne) {
+			}
+		} else if (operation == SearchFilterParser.CompareOperation.le) {
+				// use lower bound first
+				if (lowerBoundInstant != null) {
+					lb = theBuilder.lessThanOrEqualTo(theFrom.get(lowValueField), genericLowerBound);
+				} else {
+					if (upperBoundInstant != null) {
+						ub = theBuilder.lessThanOrEqualTo(theFrom.get(lowValueField), genericUpperBound);
+					} else {
+						throw new InvalidRequestException("lowerBound and upperBound value not correctly specified for compare theOperation");
+					}
+				}
+		} else if (operation == SearchFilterParser.CompareOperation.gt) {
+			// use upper bound first, e.g value between 6 and 10
+			// gt7 true,    10>7, gt11 false,  10>11 false, gt5 true,    10>5
+			if (upperBoundInstant != null) {
+				if (upperBound.getPrefix() == ParamPrefixEnum.EQUAL)
+					ub = theBuilder.greaterThan(theFrom.get(highValueField), genericUpperBound);
+				else
+					// the value has been added one in this case
+					ub = theBuilder.greaterThanOrEqualTo(theFrom.get(highValueField), genericUpperBound);
+			} else {
+				if (lowerBoundInstant != null) {
+					if (lowerBound.getPrefix() == ParamPrefixEnum.EQUAL)
+						lb = theBuilder.greaterThan(theFrom.get(highValueField), genericLowerBound);
+					else
+						// the value has been added one in this case
+						lb = theBuilder.greaterThanOrEqualTo(theFrom.get(highValueField), genericLowerBound);
+				} else {
+					throw new InvalidRequestException("upperBound and lowerBound value not correctly specified for compare theOperation");
+				}
+			} 
+		} else if (operation == SearchFilterParser.CompareOperation.ge) {
+			// use upper bound first, e.g value between 6 and 10
+			// gt7 true,    10>7, gt11 false,  10>11 false, gt5 true,    10>5
+			if (upperBoundInstant != null) {
+				ub = theBuilder.greaterThanOrEqualTo(theFrom.get(highValueField), genericUpperBound);;
+			} else {
+				if (lowerBoundInstant != null) {
+					lb = theBuilder.greaterThanOrEqualTo(theFrom.get(highValueField), genericLowerBound);
+				} else {
+					throw new InvalidRequestException("upperBound and lowerBound value not correctly specified for compare theOperation");
+				}
+			} 
+		} else if (operation == SearchFilterParser.CompareOperation.ne) {
 			if ((lowerBoundInstant == null) ||
 				(upperBoundInstant == null)) {
 				throw new InvalidRequestException("lowerBound and/or upperBound value not correctly specified for compare operation");
 			}
 			lt = theBuilder.lessThan(theFrom.get(lowValueField), genericLowerBound);
 			gt = theBuilder.greaterThan(theFrom.get(highValueField), genericUpperBound);
-			lb = theBuilder.or(lt,
-				gt);
+			lb = theBuilder.or(lt, gt);
 		} else if ((operation == SearchFilterParser.CompareOperation.eq) || (operation == null)) {
 			if (lowerBoundInstant != null) {
 				gt = theBuilder.greaterThanOrEqualTo(theFrom.get(lowValueField), genericLowerBound);
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
@@ -126,6 +126,9 @@ public class PredicateBuilderDate extends BasePredicateBuilder implements IPredi
 		if (theParam instanceof DateParam) {
 			DateParam date = (DateParam) theParam;
 			if (!date.isEmpty()) {
+				if (theOperation == SearchFilterParser.CompareOperation.ne) {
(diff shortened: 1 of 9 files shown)


jamesagnew (approved):
It took me a while to figure out what was going on here, but after running the affected tests and comparing the SQL I get it. This does seem like a great streamlining. Nice!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
```diff
@@ -152,6 +152,7 @@ private boolean isNullOrDayPrecision(DateParam theDateParam) {
 		return theDateParam == null || theDateParam.getPrecision().ordinal() == TemporalPrecisionEnum.DAY.ordinal();
```
jamesagnew, line 152: One comment, putting it here for no good reason - We should add a changelog for this change

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
```diff
@@ -191,36 +192,75 @@ private Predicate createPredicateDateFromRange(CriteriaBuilder theBuilder,
...
 		if (operation == SearchFilterParser.CompareOperation.lt) {
-			if (lowerBoundInstant == null) {
-				throw new InvalidRequestException("lowerBound value not correctly specified for compare operation");
+			// use lower bound first
+			if (lowerBoundInstant != null) {
+				if (lowerBound.getPrefix() == ParamPrefixEnum.EQUAL)
+					lb = theBuilder.lessThan(theFrom.get(lowValueField), genericLowerBound);
+				else
+					// the value has been reduced one in this case
+					lb = theBuilder.lessThanOrEqualTo(theFrom.get(lowValueField), genericLowerBound);
```
IanMMarshall, line 201: I am confused by this. Why is the lower bound set differently if the prefix is something other than ParamPrefixEnum.EQUAL? 

I believe the genericLowerBound is incremented by one if the prefix is GREATERTHAN or STARTS_AFTER but is unchanged if the prefix is EQUAL or GREATERTHAN_OR_EQUALS. Maybe the if statement should be checking that prefix is EQUAL or GREATERTHAN_OR_EQUALS?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
```diff
@@ -191,36 +192,75 @@ private Predicate createPredicateDateFromRange(CriteriaBuilder theBuilder,
...
+			} else {
+				if (upperBoundInstant != null) {
+					if (upperBound.getPrefix() == ParamPrefixEnum.EQUAL)
+						ub = theBuilder.lessThan(theFrom.get(lowValueField), genericUpperBound);
+					else
+						// the value has been reduced one in this case
+						ub = theBuilder.lessThanOrEqualTo(theFrom.get(lowValueField), genericUpperBound);
```
IanMMarshall, line 208: I am similarly confused by this. Why is the upper bound set differently if the prefix is something other than ParamPrefixEnum.EQUAL? 

I believe the genericUpperBound is decremented by one if the prefix is LESSTHAN or ENDS_BEFORE but is unchanged if the prefix is EQUAL or LESSTHAN_OR_EQUALS. Maybe the if statement should be checking that prefix is EQUAL or LESSTHAN_OR_EQUALS?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
```diff
@@ -191,36 +192,75 @@ private Predicate createPredicateDateFromRange(CriteriaBuilder theBuilder,
...
-				throw new InvalidRequestException("upperBound value not correctly specified for compare operation");
-			}
-			lb = theBuilder.greaterThan(theFrom.get(highValueField), genericUpperBound);
-			} else if (operation == SearchFilterParser.CompareOperation.ge) {
-				if (lowerBoundInstant == null) {
-					throw new InvalidRequestException("lowerBound value not correctly specified for compare operation");
+			// use upper bound first, e.g value between 6 and 10
+			// gt7 true,    10>7, gt11 false,  10>11 false, gt5 true,    10>5
+			if (upperBoundInstant != null) {
+				if (upperBound.getPrefix() == ParamPrefixEnum.EQUAL)
```
IanMMarshall, line 228: See my earlier comments about the EQUAL prefix..

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
```diff
@@ -191,36 +192,75 @@ private Predicate createPredicateDateFromRange(CriteriaBuilder theBuilder,
...
+					ub = theBuilder.greaterThan(theFrom.get(highValueField), genericUpperBound);
+				else
+					// the value has been added one in this case
+					ub = theBuilder.greaterThanOrEqualTo(theFrom.get(highValueField), genericUpperBound);
+			} else {
+				if (lowerBoundInstant != null) {
+					if (lowerBound.getPrefix() == ParamPrefixEnum.EQUAL)
```
IanMMarshall, line 235: See my earlier comment about the EQUAL prefix.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/DatePredicateBuilder.java
```diff
@@ -131,27 +133,68 @@ private Condition createPredicateDateFromRange(DatePredicateBuilder theFrom,
 		}
 
 		if (theOperation == SearchFilterParser.CompareOperation.lt) {
-			if (lowerBoundInstant == null) {
-				throw new InvalidRequestException("lowerBound value not correctly specified for compare theOperation");
+			// use lower bound first
+			if (lowerBoundInstant != null) {
+				if (lowerBound.getPrefix() == ParamPrefixEnum.EQUAL)
```
IanMMarshall, line 138: See my earlier comment about the EQUAL prefix.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/DatePredicateBuilder.java
```diff
@@ -131,27 +133,68 @@ private Condition createPredicateDateFromRange(DatePredicateBuilder theFrom,
...
+					lb = theFrom.createPredicate(lowValueField, ParamPrefixEnum.LESSTHAN, genericLowerBound);
+				else 
+					// the value has been reduced one in this case
+					lb = theFrom.createPredicate(lowValueField, ParamPrefixEnum.LESSTHAN_OR_EQUALS, genericLowerBound);
+			} else {
+				// use upper bound if the lower bound value is null
+				if (upperBoundInstant != null) {
+					if (upperBound.getPrefix() == ParamPrefixEnum.EQUAL)
```
IanMMarshall, line 146: See my earlier comment about the EQUAL prefix.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/DatePredicateBuilder.java
```diff
@@ -131,27 +133,68 @@ private Condition createPredicateDateFromRange(DatePredicateBuilder theFrom,
...
 			}
-			//im like 80% sure this should be ub and not lb, as it is an UPPER bound.
-			lb = theFrom.createPredicate(highValueField, ParamPrefixEnum.LESSTHAN_OR_EQUALS, genericUpperBound);
 		} else if (theOperation == SearchFilterParser.CompareOperation.gt) {
-			if (upperBoundInstant == null) {
-				throw new InvalidRequestException("upperBound value not correctly specified for compare theOperation");
+			// use upper bound first, e.g value between 6 and 10
+			// gt7 true,    10>7, gt11 false,  10>11 false, gt5 true,    10>5
+			if (upperBoundInstant != null) {
+				if (upperBound.getPrefix() == ParamPrefixEnum.EQUAL)
```
IanMMarshall, line 172: See my earlier comment about the EQUAL prefix.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/DatePredicateBuilder.java
```diff
@@ -131,27 +133,68 @@ private Condition createPredicateDateFromRange(DatePredicateBuilder theFrom,
...
+					ub = theFrom.createPredicate(highValueField, ParamPrefixEnum.GREATERTHAN, genericUpperBound);
+				else
+					// the value has been added one in this case
+					ub = theFrom.createPredicate(highValueField, ParamPrefixEnum.GREATERTHAN_OR_EQUALS, genericUpperBound);
+			} else {
+				if (lowerBoundInstant != null) {
+					if (lowerBound.getPrefix() == ParamPrefixEnum.EQUAL)
```
IanMMarshall, line 179: See my earlier comment about the EQUAL prefix.

IanMMarshall (commented):
Looks good to me, but I am a bit confused about the special logic in certain cases where the prefix is set to EQUALS.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
frankjtao: There is a test case `ParamPrefixEnum.EQUAL`, but `SearchFilterParser.CompareOperation.lt`
and the other test case `ParamPrefixEnum.LESSTHAN, and `SearchFilterParser.CompareOperation.lt`
In the first case, the lowerbound value does not change, so '<' should  be used in the comparison,
In the second case, the lowerbound value is change, so '<=' should be used in the comparison.
Please check the test cases at https://github.com/hapifhir/hapi-fhir/blob/656dd33ba893130af1e439acca9b5a10cd4015c3/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4FilterTest.java#L686 and https://github.com/hapifhir/hapi-fhir/blob/656dd33ba893130af1e439acca9b5a10cd4015c3/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4FilterTest.java#L712

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
frankjtao: Same as above, it took me several hours to figure out

frankjtao:
> Looks good to me, but I am a bit confused about the special logic in certain cases where the prefix is set to EQUALS.

Same issue. See comments above

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
jamesagnew: @frankjtao I don't think those test cases apply to the spot being discussed. I put a breakpoint on the `if (lowerBound.getPrefix() == ParamPrefixEnum.EQUAL)` and then ran `testDateComparatorGt` and `testDateComparatorEq` and it never hit.

Then I realized the line being discussed is in the method `createPredicateDateFromRange` - It applies only to ranges (eg `?date=gt20210101&date=lt20220101`). I don't actually think it's possible to have a prefix of `EQUAL` here. Is there a test that fails without this?

frankjtao: @jamesagnew just tried again. Without adding `ParamPrefixEnum.EQUAL` logic, `testDateComparatorGt `and `testDateComparatorLt` are failed.
`FhirResourceDaoR4FilterTest.testDateComparatorGt`  hits this line. 
https://github.com/hapifhir/hapi-fhir/blob/2e7872ce8e5f824ec2a1a57b5e6f0a29419608d6/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/DatePredicateBuilder.java#L172
`FhirResourceDaoR4FilterTest.testDateComparatorLt` hits this line.
https://github.com/hapifhir/hapi-fhir/blob/2e7872ce8e5f824ec2a1a57b5e6f0a29419608d6/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/DatePredicateBuilder.java#L138

IanMMarshall: I tried running all of the dao, provider, and search tests as well with breakpoints set on these changes in `PredicateBuilderDate` (the class referred to here, not `DatePredicateBuilder). I only found one test class that appeared to cover this logic, FhirResourceDaoR4LegacySearchBuilderTest. None of the tests appear to involve a case where prefix is set to EQUAL.

You might to add tests that exercise this logic in `PredicateBuilderDate`.

frankjtao: @IanMMarshall `FhirResourceDaoR4FilterTest.testDateComparatorGt `and `testDateComparatorLt`,  hit the above point.

jamesagnew: @frankjtao I'm having another look now. Thanks for pointing these out.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/DatePredicateBuilder.java
IanMMarshall: I see now that certain existing tests in `FhirResourceDaoR4FilterTest` would fail if this logic was not here in `DatePredicateBuilder` which does not appear to be the case in `PredicateDateBuilder`.  

I think maybe that part of my confusion relates to the subsequent comment which appears to be inconsistent with what I see when I trace the calculation of the genericLowerBound:

> // the value has been reduced one in this case

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/predicate/PredicateBuilderDate.java
jamesagnew: Ok @frankjtao I totally get this now. The equals check makes sense, and is correct because of the fact that the filter parser doesn't actually put the correct prefix into the DateParam when it creates it (that part was contributed ages ago- it's been wrong for a long time I think but it hasn't mattered until now).

I've made a pull request against your pull request, with a proposed cleanup of that: https://github.com/hapifhir/hapi-fhir/pull/2316

Can you have a look? If you are good with it, I'll merge it into your branch once it passes CI

PR #2311, opened 2021-01-21
bump databind version for CVE

tadgh:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-20190


jamesagnew:
That library has CVEs with a surprising frequency....

jamesagnew (approved):

PR #2316, opened 2021-01-22
Date handling cleanup

jamesagnew:


frankjtao (changes requested):
There is no issue on the SearchBuilder. But the some test cases are failed if  the Legacy flag is turned on for the the test `FhirResourceDaoR4FilterTest`

jamesagnew:
@frankjtao good call - I've fixed those issues and added `FhirResourceDaoR4FilterLegacySearchBuilderTest`, which is a copy of FhirResourceDaoR4FilterBuilderTest but with the legacy flag set to true

frankjtao (approved):

PR #2319, opened 2021-01-23
reduce _count to the maximum page size rather than throwing an error.

fil512:
Also when setting the paging provider on a restful server, automatically set the default page size and max page sized of the restful server from the paging provider.


fil512:
And fixed a couple of search coordinator bugs this uncovered

jamesagnew (approved):

PR #2323, opened 2021-01-25
Enable Auto-reference versions in JPA

jamesagnew:


hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ModelConfig.java
```diff
@@ -618,6 +625,110 @@ public void setNormalizedQuantitySearchLevel(NormalizedQuantitySearchLevel theNo
...
+	 * </p>
+	 *
+	 * @param thePaths A collection of reference paths for which the versions will be appended automatically
+	 *                 when serializing, e.g. "Patient.managingOrganization" or "AuditEvent.object.reference". Note that
+	 *                 only resource name and field names with dots separating is allowed here (no repetition
+	 *                 indicators, FluentPath expressions, etc.)
+	 * @since 5.3.0
+	 */
+	public void setAutoVersionReferenceAtPaths(String... thePaths) {
+		Set<String> paths = Collections.emptySet();
```
tadgh, line 658: would it make sense to fail to boot if we indicate that `dontStripVersions` is disabled _and_ this is enabled?

tadgh (approved):
Couple questions, overall looks good. The deep SQL stuff went a bit over my head, specifically the "r." field names, I'll leave that to somebody else to review.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseStorageDao.java
```diff
@@ -115,13 +155,21 @@ protected void preProcessResourceForStorage(IBaseResource theResource) {
...
-			Set<String> allowedBundleTypes = getConfig().getBundleTypesAllowedForStorage();
-			String bundleType = BundleUtil.getBundleType(getContext(), (IBaseBundle) theResource);
-			bundleType = defaultString(bundleType);
-			if (!allowedBundleTypes.contains(bundleType)) {
-				String message = "Unable to store a Bundle resource on this server with a Bundle.type value of: " + (isNotBlank(bundleType) ? bundleType : "(missing)");
-				throw new UnprocessableEntityException(message);
+		/*
+		 * Handle auto-populate-versions
+		 */
+		Set<IBaseReference> referencesToVersion = extractReferencesToAutoVersion(myFhirContext, myModelConfig, theResource);
```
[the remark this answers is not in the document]
jamesagnew: Good call!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseStorageDao.java
```diff
@@ -115,13 +155,21 @@ protected void preProcessResourceForStorage(IBaseResource theResource) {
...
+		if (thePerformIndexing) {
```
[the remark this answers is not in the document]
jamesagnew: Have added the following javadoc to explain:

```java
	/**
	 * Handle {@link ModelConfig#getAutoVersionReferenceAtPaths() auto-populate-versions}
	 *
	 * We only do this if thePerformIndexing is true because if it's false, that means
	 * we're in a FHIR transaction during the first phase of write operation processing,
	 * meaning that the versions of other resources may not have need updated yet. For example
	 * we're about to store an Observation with a reference to a Patient, and that Patient
	 * is also being updated in the same transaction, during the first "no index" phase,
	 * the Patient will not yet have its version number incremented, so it would be wrong
	 * to use that value. During the second phase it is correct.
	 *
	 * Also note that {@link BaseTransactionProcessor} also has code to do auto-versioning
	 * and it is the one that takes care of the placeholder IDs. Look for the other caller of
	 * {@link #extractReferencesToAutoVersion(FhirContext, ModelConfig, IBaseResource)}
	 * to find this.
	 */
```

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ModelConfig.java
jamesagnew: Will add

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseStorageDao.java
```diff
@@ -114,17 +172,38 @@ protected void preProcessResourceForStorage(IBaseResource theResource) {
...
-			bundleType = defaultString(bundleType);
-			if (!allowedBundleTypes.contains(bundleType)) {
-				String message = "Unable to store a Bundle resource on this server with a Bundle.type value of: " + (isNotBlank(bundleType) ? bundleType : "(missing)");
-				throw new UnprocessableEntityException(message);
+	/**
+	 * Handle {@link ModelConfig#getAutoVersionReferenceAtPaths() auto-populate-versions}
+	 *
+	 * We only do this if thePerformIndexing is true because if it's false, that means
+	 * we're in a FHIR transaction during the first phase of write operation processing,
+	 * meaning that the versions of other resources may not have need updated yet. For example
```
fil512, line 182:
```suggestion
	 * meaning that the versions of other resources may not have been updated yet. For example
```

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ModelConfig.java
```diff
@@ -618,6 +626,110 @@ public void setNormalizedQuantitySearchLevel(NormalizedQuantitySearchLevel theNo
...
+	 * will automatically have versions appended. The version used will be the current version of the given resource.
+	 *
+	 * @since 5.3.0
+	 */
+	public Set<String> getAutoVersionReferenceAtPaths() {
+		return myAutoVersionReferenceAtPaths;
+	}
+
+	/**
+	 * When set with resource paths (e.g. <code>"Observation.subject"</code>), any references found at the given paths
```
fil512, line 640: This is more restrictive than I expected.  Is it possible to configure the system to append version numbers to all references?

fil512 (approved):

PR #2327, opened 2021-01-27
changed $expand filter matching from left-match to contains match

fil512:


jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
```diff
@@ -644,21 +644,21 @@ private void logDesignationsExpanded(String theLogDescriptionPrefix, TermValueSe
...
+	public boolean applyFilter(final String theDisplay, final String theFilterDisplay) {
 		
 		//-- safety check only, no need to apply filter
-		if (theInput == null || thePrefixToken == null)
+		if (theDisplay == null || theFilterDisplay == null)
 			return true;
 
 		// -- sentence case
-		if (org.apache.commons.lang3.StringUtils.startsWithIgnoreCase(theInput, thePrefixToken))
+		if (org.apache.commons.lang3.StringUtils.containsIgnoreCase(theDisplay, theFilterDisplay))
```
IanMMarshall: I have a small concern with this proposed change in that it may result in additional matches that we do not want. For example if the filter was specified as `abc`, it would return  `defabcd` as well as `def abcd`. Ideally I think we only want it to return matches where the entire sentence or a phrase within the sentence starts with the filter string.

Suggestion: Maybe keep the original `startsWithIgnoreCase()` check and then add a second check for `containsIgnoreCase()`, but specifying the filter string prefixed with a space character as. the argument.

IanMMarshall (approved):
Looks good although I do recommend one small change per my comment in the code to avoid getting additional matches that we might actually want to avoid.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
fil512: Really good point.  a regression test caught the issue you predicted.  I've tightened up the matching.

PR #2331, opened 2021-01-29
Fixed Ucum Conversion NullPointerException issue

frankjtao:
`NullPointerException` when UCUM converting code `[degF]`. This is fixed.


jamesagnew:
@frankjtao I don't understand why we need to store a value in the normalized table if the value can't be normalized. That just feels like a waste of space.

Can't we just fix it to not search in that table if the client is searching using a unit that's not normalizable?

frankjtao:
> @frankjtao I don't understand why we need to store a value in the normalized table if the value can't be normalized. That just feels like a waste of space.
> 
> Can't we just fix it to not search in that table if the client is searching using a unit that's not normalizable?

No problem.  Will work on it.

jamesagnew (approved):

Issue #2350, opened 2021-02-03
Fix for CQL dependency issue in hapi-fhir-jpaserver-cql


gered:
`info.cqframework/quick` is a dependency of `info.cqframework/cql-to-elm` which hapi-fhir-jpaserver-cql depends on. But `info.cqframework/quick` is currently set with scope=runtime which is preventing it from being included in application builds that depend on hapi-fhir-jpaserver-cql.

KevinDougan-SmileCDR:
This was merged in [PR-2351](https://github.com/hapifhir/hapi-fhir/pull/2351)

PR #2351, opened 2021-02-03
Added a dependency on info.cqframework:quick since we also need this JAR

KevinDougan-SmileCDR:
Added a dependency on info.cqframework:quick since we also need this JAR


jamesagnew (approved):

KevinDougan-SmileCDR:
Fixes #2350

PR #2357, opened 2021-02-04
Update to release versions of cql components

JPercival:
* Removed cqf-tooling since it is no longer used
* Update cql-engine to 1.5.1
* Update cql-translator to 1.5.1
* Update cql-evaluator to 1.1.0


KevinDougan-SmileCDR (approved):
Approved - thanks!

PR #2365, opened 2021-02-08
Fixed bug where AND resource search-params were matched as OR params

jarimayenburg:
The `InMemoryResourceMatcher` contained a bug where AND search parameters were evaluated and matched as OR parameters.

This caused issues where for example `GET /fhir/DocumentReference?indexed=ge2018-01-01&indexed=le2019-01-01` would return DocumentReferences with an indexed data of 2020-01-01.


jarimayenburg:
@jamesagnew Let me know if you'd like me to write some tests to prove this issue

jamesagnew (approved):
This looks great, thanks for the contribution! Will merge assuming it passes CI.

jamesagnew:
Fixes #2366

jamesagnew:
Fixes #2366

PR #2381, opened 2021-02-12
Remove only provider-specific method bindings, not all base on resource name

nvg-smile:
When we are deregistering providers, there might be multiple bindings from different providers for the same resource name. Only the provider bindings should be removed.


jamesagnew (approved):
Nice!

PR #2384, opened 2021-02-13
Made unknown operation exception return an OO with issue code "not-supported"

jarimayenburg:
Made the unsupported operation exception that's thrown use an OperationOutcome with a more specific "not-supported" issue code, instead of the general "processing" error.

This is required for us to comply with certain Dutch standards, so we added it to our fork. Was wondering if this might be a valuable addition to the upstream.


jamesagnew (approved):
Thanks for the contribution! Looks good, merging.

PR #2397, opened 2021-02-16
UTF-16 characters might cause errors when calculating hash values

nvg-smile:
Attempting to create a resource with char values outside of the Basic Multilingual Plane might cause an error when calculating hash values.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/StringUtil.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/StringUtil.java
@@ -78,4 +78,37 @@ public class StringUtil {
 		return new String(bytes, StandardCharsets.UTF_8);
 	}
 
+	/**
+	 * Gets the string prefix of the specified length.
+	 *
+	 * @param theString
+	 * 	String to get the prefix from
+	 * @param theCodePointCount
+	 * 	Length of the prefix in code points
+	 * @return
+	 * 	Returns the string prefix of the specified number of codepoints.
+	 */
+	public static String left(String theString, int theCodePointCount) {
+		if (theString == null) {
+			return null;
+		}
+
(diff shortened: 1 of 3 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/util/StringUtil.java
```diff
@@ -78,4 +78,37 @@ public static String toUtf8String(byte[] theBytes) {
...
+	public static String left(String theString, int theCodePointCount) {
+		if (theString == null) {
+			return null;
+		}
+
+		if (theCodePointCount < 0) {
+			return "";
+		}
+
+		if (theString.length() <= theCodePointCount) {
```
michaelabuckley, line 100: Can we add a comment that length() is char count, while theCodePointCount isn't.  But that's ok since char count can only be bigger than code point count.

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/StringUtil.java
```diff
@@ -78,4 +78,37 @@ public static String toUtf8String(byte[] theBytes) {
...
+			return theString;
+		}
+
+		StringBuilder retVal = new StringBuilder();
```
michaelabuckley, line 104: Can we replace this loop with .substring() and offsetByCodePoints()?  I.e. return theString.substring(0, theString.offsetByCodePoints(0, theCodePointCount)).

nvg-smile: Thank you for taking a look, Michael! offsetByCodePoints still runs a for loop, but less code in our code base is always great - fixed!

jamesagnew (approved):

PR #2416, opened 2021-02-23
Skip loading a FHIRPath object if we are older than STU3

tadgh:


hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/config/MdmRuleValidator.java
```diff
@@ -61,7 +62,12 @@
 	public MdmRuleValidator(FhirContext theFhirContext, ISearchParamRetriever theSearchParamRetriever) {
 		myFhirContext = theFhirContext;
 		myTerser = myFhirContext.newTerser();
-		myFhirPath = myFhirContext.newFhirPath();
+		if (myFhirContext.getVersion().getVersion().isEqualOrNewerThan(FhirVersionEnum.DSTU3)) {
+			myFhirPath = myFhirContext.newFhirPath();
+		} else {
+			ourLog.warn("Skipping FHIRPath validation as DSTU2 does not support FHIR");
```
jamesagnew, line 68: I'm curious about 2 things:

* Is this just papering over a deeper issue? Like, why do we validate FHIRPath here unless we're gonna use it again later? Are we just delaying an inevitable fail?
* Should this be a warning, given that it seems like a routine thing to happen?

tadgh: The issue is that in CDR, the BaseAppCtxPersistence is what loads mdmSubmitterConfig. So in DSTU2 mode we load this bean, even though we never actually need it. The real solution is to add a BaseAppCtxPersistenceDSTU3Plus config, which correctly loads the Mdm Submitter beans

jamesagnew: Ah ok - Should the warn be an info then?

jamesagnew: Sorry, I see you did that nevermind

jamesagnew (approved):

PR #2410, opened 2021-02-22
Support FHIRPath in field matchers if present

tadgh:
Closes #2409

* Add field to MdmFieldMatchJson called `fhirPath`. 
* Validation will ensure one of either `resourcePath` or `fhirPath` is set, and fail otherwise.
* At match time, if fhirPath is set, evaluate the expression using the FHIRPath parser. Otherwise, use the Terser
* Update docs to reflect new reality.

PR #2418, opened 2021-02-24
Add MDM FHIRPath Changelog

tadgh:
This was overlooked in the original implementation done in #2410


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_4_0/2409-allow-fhirpath-in-mdm-field-matchers.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
michaelabuckley, line 1: The log is here.

michaelabuckley (approved):
The log is here.  We should really be able to automate this as a check.

PR #2436, opened 2021-03-03
Fix #2435: Make TransactionProcessor proxy-friendly

### LINKED ISSUE #2435: TransactionProcessor.flushSession does a proxy-unfriendly cast, causing ClassCastException in certain Spring setups
**Describe the bug**
The line [SessionImpl session = (SessionImpl) myEntityManager.unwrap(Session.class);](https://github.com/hapifhir/hapi-fhir/blob/v5.3.0/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java#L65) gives a ClassCastException in our Spring Boot setup.  
This is because [AbstractEntityManagerFactoryBean.afterPropertiesSet()](https://github.com/spring-projects/spring-framework/blob/v5.3.4/spring-orm/src/main/java/org/springframework/orm/jpa/AbstractEntityManagerFactoryBean.java#L375) sets the entityManagerFactoryInterface to `Session.class`,  
which in turn causes [SharedEntityManagerCreator's unwrap handling](https://github.com/spring-projects/spring-framework/blob/v5.3.4/spring-orm/src/main/java/org/springframework/orm/jpa/SharedEntityManagerCreator.java#L249) to return the proxy object instead of the underlying SessionImpl instance, when `Session.class` is the unwrap type requested.

**To Reproduce**
I have not found a shareable method of reproduction. 
For some reason [AbstractEntityManagerFactoryBean.afterPropertiesSet()](https://github.com/spring-projects/spring-framework/blob/v5.3.4/spring-orm/src/main/java/org/springframework/orm/jpa/AbstractEntityManagerFactoryBean.java#L375)  is not invoked in the hapi-fhir-jpaserver-starter project and therefore the logic in [SharedEntityManagerCreator's unwrap handling](https://github.com/spring-projects/spring-framework/blob/v5.3.4/spring-orm/src/main/java/org/springframework/orm/jpa/SharedEntityManagerCreator.java#L249) is different, causing the SessionImpl instance to be returned, instead of the proxy.

**Expected behavior**
I would expect the TransactionProcessor to be proxy friendly by unwrapping directly to the desired type, instead of casting the unwrap result to a different type.

The line should be replaced with the functionally equivalent line of:  
`SessionImpl session = myEntityManager.unwrap(SessionImpl.class);`

I will open a PR with the change.

This is a similar problem as reported here: https://github.com/spring-projects/spring-framework/issues/26090 where they also ended up changing the offending casts.

ttntrifork:
Fix https://github.com/hapifhir/hapi-fhir/issues/2435

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
@@ -62,7 +62,7 @@ public class TransactionProcessor extends BaseTransactionProcessor {
 	@Override
 	protected void flushSession(Map<IIdType, DaoMethodOutcome> theIdToPersistedOutcome) {
 		try {
-			SessionImpl session = (SessionImpl) myEntityManager.unwrap(Session.class);
+			SessionImpl session = myEntityManager.unwrap(SessionImpl.class);
 			int insertionCount = session.getActionQueue().numberOfInsertions();
 			int updateCount = session.getActionQueue().numberOfUpdates();
 


jamesagnew (approved):
Looks good, thanks!

PR #2441, opened 2021-03-04
Supported contained resource search

frankjtao:


hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
```diff
@@ -96,14 +100,100 @@ public void extractFromResource(RequestPartitionId theRequestPartitionId, Reques
...
+		// Do this after, because we add to strings during both string and token processing, and contained resource if any
+		populateResourceTables(theParams, theEntity);
+		
 		// Reference search parameters
 		extractResourceLinks(theRequestPartitionId, theParams, theEntity, resource, theTransactionDetails, theFailOnInvalidReference, theRequestDetails);
 
 		theParams.setUpdatedTime(theTransactionDetails.getTransactionDate());
 	}
 
+	private void extractSearchIndexParametersForContainedResources(RequestDetails theRequestDetails, ResourceIndexedSearchParams theParams, IBaseResource theResource, ResourceTable theEntity) {
```
jamesagnew, line 122: I don't think this approach is quite right - I get the idea here, where you're looking for the references in the containing resource and then inferring the search parameter name from the element name by performing a camelCase-to-dashes conversion. The issues I see:

* Not every reference in a resource is actually guaranteed to have a search parameter associated with it so this will actually index too much
* More importantly. the search parameter name is not guaranteed to follow the convention of being that tweaked version of the field name. Some are but not all, and users can create custom search parameters with any name they want.

My recommendation is to go about this in reverse: Collect the references associated with Search Parameters, then find the associated contained resource and create the index. You can ask for the references in the containing resource, along with the associated search parameter (so you can figure out the name for the index) by calling:

```java
ISearchParamExtractor.SearchParamSet<PathAndRef> refs = mySearchParamExtractor.extractResourceLinks(theResource);
```

Note that this call ignores local references (e.g. `#1`) so it won't do what you need out of the box. My recommendation would be to add another parameter to that method like `theWantLocalReferences` where `false` gives the current behaviour, but `true` returns only local refs like you need here.

frankjtao: Done.

PR #2448, opened 2021-03-05
Issues with placeholder reference targets need to be resolved.

### LINKED ISSUE #2446: Issues with placeholder reference targets need to be resolved.
Internal HubSpot reference: 322720014

**Requirements and Expectations:**

1. When [allow inline match URLs](https://smilecdr.com/docs/configuration_categories/fhir_configuration.html#property-allow-inline-match-url-references-enabled) is enabled, the resulting placeholder resource should be populated with an identifier (see 007-read-organization_expected.json attached). This is an existing feature that appears to be broken. Refer to the documentation on [Auto-Create Placeholder Reference Targets with Identifier](https://smilecdr.com/docs/fhir_repository/creating_data.html#auto-create-placeholder-reference-targets-with-identifier) for additional information.
1. When [auto-create placeholder reference targets](https://smilecdr.com/docs/configuration_categories/fhir_configuration.html#property-auto-create-placeholder-reference-targets) is enabled, all created placeholder resources should be populated with an extension identifying them as such (see 007-read-organization_expected.json attached). This needs to happen regardless of the source of the resource (i.e. type-level POST, transaction bundle, IG package, etc). The example shows the extension in the body of the resource. The current implementation has this as an extension in meta, which may not be appropriate because the extension should not persist once the resource is updated. It needs to be confirmed whether or not this occurs with an extension in meta; and moved to the body of the resource as necessary.
1. Both 1 and 2 above need to work in parallel, and we should have tests that demonstrate this.

**Future Considerations:**

- We should probably add custom SearchParameters for this extension that comes built-in with Smile CDR. To be considered for as a future enhancement.

**Attachments:**

- [001-create-patient.json.txt](https://github.com/hapifhir/hapi-fhir/files/6085788/001-create-patient.json.txt)
- [002-read-patient.json.txt](https://github.com/hapifhir/hapi-fhir/files/6085789/002-read-patient.json.txt)
- [003-read-organization.json.txt](https://github.com/hapifhir/hapi-fhir/files/6085790/003-read-organization.json.txt)
- [004-create-patient-transaction.json.txt](https://github.com/hapifhir/hapi-fhir/files/6085791/004-create-patient-transaction.json.txt)
- [005-read-patient-transaction.json.txt](https://github.com/hapifhir/hapi-fhir/files/6085792/005-read-patient-transaction.json.txt)
- [006-read-organization-transaction.json.txt](https://github.com/hapifhir/hapi-fhir/files/6085793/006-read-organization-transaction.json.txt)
- [007-read-organization_expected.json.txt](https://github.com/hapifhir/hapi-fhir/files/6085794/007-read-organization_expected.json.txt)

dmuylwyk:
Closes #2446

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/HapiExtensions.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/HapiExtensions.java
@@ -113,8 +113,9 @@ public class HapiExtensions {
 
 	/**
 	 * URL for boolean extension added to all placeholder resources
+	 *
+	 * @deprecated Deprecated in 5.4.0 in favour of {@link HapiExtensions#EXT_RESOURCE_PLACEHOLDER}
 	 */
-	// FIXME: DM 2021-03-04 - This should probably be removed, and replaced with EXT_RESOURCE_PLACEHOLDER above.
 	public static final String EXT_RESOURCE_META_PLACEHOLDER = "http://hapifhir.io/fhir/StructureDefinition/resource-meta-placeholder";
 
 
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/HapiExtensions.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/HapiExtensions.java
@@ -109,6 +109,12 @@ public class HapiExtensions {
 	/**
 	 * URL for boolean extension added to all placeholder resources
 	 */
+	public static final String EXT_RESOURCE_PLACEHOLDER = "http://hapifhir.io/fhir/StructureDefinition/resource-placeholder";
+
+	/**
+	 * URL for boolean extension added to all placeholder resources
+	 */
+	// FIXME: DM 2021-03-04 - This should probably be removed, and replaced with EXT_RESOURCE_PLACEHOLDER above
 	public static final String EXT_RESOURCE_META_PLACEHOLDER = "http://hapifhir.io/fhir/StructureDefinition/resource-meta-placeholder";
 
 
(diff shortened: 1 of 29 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/util/HapiExtensions.java
```diff
@@ -109,6 +109,13 @@
...
 	 * URL for boolean extension added to all placeholder resources
 	 */
+	public static final String EXT_RESOURCE_PLACEHOLDER = "http://hapifhir.io/fhir/StructureDefinition/resource-placeholder";
+
+	/**
+	 * URL for boolean extension added to all placeholder resources
+	 *
+	 * @deprecated Deprecated in 5.4.0 in favour of {@link HapiExtensions#EXT_RESOURCE_PLACEHOLDER}
+	 */
 	public static final String EXT_RESOURCE_META_PLACEHOLDER = "http://hapifhir.io/fhir/StructureDefinition/resource-meta-placeholder";
```
jamesagnew, line 119: Why is this changing? Isn't that going to break people's existing code?

jamesagnew (approved):
Looks good, but I'd prefer not to make a breaking change (changing the extension URL) unless there is a good reason..

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/HapiExtensions.java
dmuylwyk: This was only introduced very recently in #2423 when the extension was intended to be placed in resource meta. Since moving the extension to resource body, I thought include "meta" in the URL was misleading (nitpick).

Should I remove my newly added constant and use the existing one?

jamesagnew: Ah ok, if it's only 8 days old I withdraw my objection. I'd say just remove the old field then.

dmuylwyk: Will do. I thought @deprecated might be overkill but wanted to play it safe. : )

PR #2451, opened 2021-03-05
Avoid a crash reading an unknown ID on partitioned server

### LINKED ISSUE #1953: (partitioned server) NullPointerException on BaseHapiFhirResourceDao when querying for non existing resource
In a partitioned server, a query for a non existing resource as following:
http://localhost:8080/partID/Patient/254

Returns an error:
"Failed to call access method: java.lang.NullPointerException"

This happens becuse in BaseHapiFhirResourceDao, the block:
```
		if (entity == null) {
			throw new ResourceNotFoundException(theId);
		}

```
Shall be placed before the block just above "Verify that the resource is for the correct partition" (there's a call to the null entity).

 - HAPI FHIR Version: 5.1.0-SNAPSHOT (master)

jamesagnew:
Fix #1953

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
@@ -1148,6 +1148,10 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 		ResourcePersistentId pid = myIdHelperService.resolveResourcePersistentIds(requestPartitionId, getResourceName(), theId.getIdPart());
 		BaseHasResource entity = myEntityManager.find(ResourceTable.class, pid.getIdAsLong());
 
+		if (entity == null) {
+			throw new ResourceNotFoundException(theId);
+		}
+
 		// Verify that the resource is for the correct partition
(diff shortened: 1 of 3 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -1164,10 +1168,6 @@ public BaseHasResource readEntity(IIdType theId, boolean theCheckForForcedId, Re
 			}
 		}
 
-		if (entity == null) {
```
bbray, line 1167: This null check needs to be duplicated not moved.  

This actually creates another NPE issue if you request a resource that exists but not in the requested partition then `validateResourceType(entity);` fails (it does generate the correct debug message).  I'm working on the 5.2 branch which is a bit different but it looks like the `Validate.notNull(entity);` in this branch would fail in the same way.

bbray (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
jamesagnew: Interesting.. Will give that a shot

Issue #2445, opened 2021-03-04
Support Patient Bulk Export


tadgh:
As per FHIR Spec (https://build.fhir.org/ig/HL7/us-bulk-data/#endpoint---all-patients), implementing servers should support the Patient level export. 

This operates in much the same way as a group level export, except the group is ALL patients. 

I think I will implement this using much the same approach as Group bulk export, except instead of expanding a memberlist, I will just use 
`Observation?subject:missing=false`, which should return all results that are pertinent to any partient

PR #2463, opened 2021-03-10
Issue 2445 support patient level export

tadgh:
* Implement `/Patient/$export`, using `missing` search parameter
* Rework _typeFilter parsing so it actually respects the spec. For example, previously, `_typeFilter=Patient?identifier=foo,bar,Practitioner?identifier=zoo` would not process correctly. 
* Rework Group Bulk Export to support multiple type filter parameters. 
* We now correctly UNION multiple _typeFilter results together instead of just processing the first one. 
* Fix default behaviour when no `_type` parameter is passed to Group/Patient export.
* Add support for `Cache-Control: no-cache` header in Bulk requests, which will force a new job even if a matching one exists.


jamesagnew (approved):
Looks great!

tadgh:
This closes #2445

PR #2473, opened 2021-03-13
~ inspect instant for specific RDF datatype

ericprud:
Scalar values in RDF are types so FHIR union types like `instant` must be inspected to get one of gYear, gYearMonth, date or dateTime.

TODO: configure Jena to reject these rather than issuing a warning.


ericprud:
This one should be pretty non-controversial; the only change you could see in the test output is that it removes the Jena warnings about malformed dateTime literals.

ericprud:
pester pester pester

jamesagnew (approved):

PR #2477, opened 2021-03-15
Parameterize term valueset concept view dao

michelgleeson:
Not entirely sure if this will work - confirming with a test with Oracle

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/entity/TermValueSetConceptViewOracle.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/entity/TermValueSetConceptViewOracle.java
@@ -1,0 +1,57 @@
+package ca.uhn.fhir.jpa.entity;
+
+/*
+ * #%L
+ * HAPI FHIR JPA Server
+ * %%
+ * Copyright (C) 2014 - 2021 Smile CDR, Inc.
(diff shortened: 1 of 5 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/entity/TermValueSetConceptViewOracle.java
```diff
@@ -0,0 +1,57 @@
...
+import java.io.Serializable;
+
+@Entity
+@Immutable
+@Subselect(
+	/*
+	 * Note about the CONCAT function below- We need a primary key (an @Id) column
+	 * because hibernate won't allow the view the function without it, but
+	 */
+	"SELECT CONCAT(' ', vsc.PID, vscd.PID) AS PID, " +
```
jamesagnew, line 41: This isn't quite right. This is what we had before the change that broke things for oracle...

```
	"SELECT CONCAT(vsc.PID, CONCAT(' ', vscd.PID)) AS PID, " +
```

michelgleeson: Ah ok, thanks for the clarification!

jamesagnew (approved):
This could use a changelog but is good to go!

PR #2482, opened 2021-03-17
Update points.json for adding a point in atlas

kunjan8794:


jamesagnew (approved):
Thanks for the submission!

PR #2487, opened 2021-03-19
MDM Expansion Interceptor (poc)

### LINKED ISSUE #2520: Support MDM expansion during search queries
Create search parameter qualifier `:mdm` for reference parameters which will expand a search to include all MDM-matched resources, before the search is executed.

tadgh:
Closes #2520 

This PR adds the capability to do basic MDM Expansion during queries via a reference search parameter qualifier. 

* Move PRESEARCH_REGISTERED hook up the stack so it occurs before cache checking. 
* Add `SearchParameterMap` to the list of hook invocation parameters. This allows implementers  to manipulate the SP map before the search is registered. 
* Add an MdmSearchExpandingInterceptor, gated on two conditions
    1. Presence of an MDM module
    2. a ModelConfig/DaoConfig toggle `isAllowMdmExpansion`
    

In its current implementation, this interceptor will catch any search requests that match the following shape: 
```http_request
GET http://localhost:8000/Observation?subject:mdm=Patient/1
```
note the `:mdm` qualifier. This instructs the interceptor to reach out to the MdmLinkDao, and expand this search query to encompass not only Patient/1, but any Patient which is MDM-matched to Patient/1. 

If for example Patient/1 is MDM-linked to Patient/2 and Patient/3 via golden resource Patient/G1, then the expanding interceptor would convert the above query to:

```http_request
GET http://localhost:8000/Observation?subject:mdm=Patient/1,Patient/2,Patient/3,Patient/G1
```

**Caveats:**
* This parameter does not support chaining
* This is not FHIR spec, and strictly works only in HAPI.


patrick-vachon-smilecdr (commented):
Review done!

patrick-vachon-smilecdr (approved):
Approved!

jamesagnew (approved):

PR #2490, opened 2021-03-21
Add optional warning on unknown CS/VS instead of returning null

patrick-werner:
added boolean flags to create warnings on unknown CS/VS instead of returning null and creating an error.

### DIFF
--- a/hapi-fhir-cli/hapi-fhir-cli-api/pom.xml
+++ b/hapi-fhir-cli/hapi-fhir-cli-api/pom.xml
@@ -256,6 +256,14 @@
 
 	<build>
 		<plugins>
+			<plugin>
+				<groupId>org.basepom.maven</groupId>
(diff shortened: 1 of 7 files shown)


jamesagnew:
Hi @patrick-werner - Two comments: 

* I'm not sure the chain is an intuitive spot for this kind of logic, as it doesn't actually do any other processing. Would it make more sense to just create a new IValidationSupport implementation that hardcodes this logic, and people can add it to the end of their chain if they want this behaviour?
* Could you add a test exercizing this new functionality?

Those points aside, I'm definitely good with this being merged.

patrick-werner:
Thanks for the review and comment @jamesagnew 
Will update the PR to include the new ValidationModule including tests

jamesagnew:
FYI we had a few other requests for this over the last week so it got bumped up in priority. I have a fix being tested now here: https://github.com/hapifhir/hapi-fhir/pull/2523

PR #2507, opened 2021-03-25
fix token validation

fil512:


tadgh (approved):
Was this just found randomly, or did you bump into it in a real world scenario? I would say add a test for it, but honestly, i think its fine without

patrick-vachon-smilecdr:
Nice catch

PR #2514, opened 2021-03-29
Resolve "2513 - PointcutLatch should work with IPointcut."

dmuylwyk:
Let's use IPointcut instead of Pointcut with PointcutLatch.

Also: version bump.


fil512 (approved):
looks good

PR #2546, opened 2021-04-14
2535 ignore logical structuredefinition during package install

IanMMarshall:
Changed package installer to no longer attempt to generate a snapshot for StructureDefinition resources that have type of "logical".


tadgh (approved):
Looks good

PR #2547, opened 2021-04-14
added numeric matcher

fil512:


hapi-fhir-base/src/main/java/ca/uhn/fhir/context/phonetic/NumericEncoder.java
```diff
@@ -0,0 +1,16 @@
...
+public class NumericEncoder implements IPhoneticEncoder {
+	@Override
+	public String name() {
+		return "NUMERIC";
+	}
+
+	@Override
+	public String encode(String theString) {
+		// Remove everything but the numbers
+		return theString.replaceAll("[^0-9]", "");
```
jamesagnew, line 14: Given the performance implications of this method (I'm guessing it'll get called a -lot-), probably makes sense to optimize this. At a minimum using a precompiled regex would be better, even better something like `CharMatcher.inRange('0', '9').retainFrom(string)`

jamesagnew (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/phonetic/NumericEncoder.java
fil512: ooooh fancy!
done

PR #2561, opened 2021-04-19
Remove jetbrains annotation

tadgh:
* Remove forgotten jetbrains annotation from a previous PR 
* Add checkstyle to prevent them from ever showing up again
* Removean unrelated `@Nullable` from jetbrains


fil512 (approved):
surprised there was only one!

PR #2565, opened 2021-04-21
Enhance BundleBuilder to support Deletes

### LINKED ISSUE #2571: Add delete support to BundleBuilder
BundleBuilder.java currently supports easy adding a CREATE or UPDATE entry to the bundle, but does not support DELETE.

tadgh:
* Enhance bundlebuilder to support deletes, either by passing an IBaseResource, or the resource type + ID.
* Add test

Closes #2571


jamesagnew (approved):
nice!

Pls add changelog.

PR #2577, opened 2021-04-27
Concept properties are not being saved while using the operation apply-codesystem-delta-add

HananAwwad:
this pull request handle saving concept properties using apply-codesystem-delta-add operation


jamesagnew (approved):
Looks great, thanks for the contribution!

PR #2579, opened 2021-04-28
Offset query fixes

tuomoa:
Here is couple of fixes for synchronous offset search:

The new search builder did not correctly pass on the count limit to the offset+limit SQL query (which caused the query to default to 10 000 items (or fetch size default maximum). When using offset it should pass on the correct limit). This change is similar to the Legacy search builder.

When offset search was used the includes were dropped from the bundle response. When offset search is used, the BundleProvider will contain correct amount of resources+includes/revIncludes. There was a mistake in the implementation and it caused to "drop" included resources if count was less than resources+their includes. When offset search is used the whole bundle contents should be returned because it already contains correct resources. The DatabaseBackedPagingProvider works differently, as it fetches the includes after some slice is retrieved from the database.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
@@ -1164,7 +1164,7 @@ public class SearchBuilder implements ISearchBuilder {
 					if (myMaxResultsToFetch == null) {
 						if (myParams.getLoadSynchronousUpTo() != null) {
 							myMaxResultsToFetch = myParams.getLoadSynchronousUpTo();
-						} else if (myParams.getCount() != null) {
+						} else if (myParams.getOffset() != null && myParams.getCount() != null) {
 							myMaxResultsToFetch = myParams.getCount();
 						} else {
 							myMaxResultsToFetch = myDaoConfig.getFetchSizeDefaultMaximum();
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
@@ -1162,7 +1162,13 @@ public class SearchBuilder implements ISearchBuilder {
 				// If we don't have a query yet, create one
 				if (myResultsIterator == null) {
 					if (myMaxResultsToFetch == null) {
-						myMaxResultsToFetch = myDaoConfig.getFetchSizeDefaultMaximum();
+						if (myParams.getLoadSynchronousUpTo() != null) {
+							myMaxResultsToFetch = myParams.getLoadSynchronousUpTo();
+						} else if (myParams.getCount() != null) {
+							myMaxResultsToFetch = myParams.getCount();
+						} else {
+							myMaxResultsToFetch = myDaoConfig.getFetchSizeDefaultMaximum();
+						}
 					}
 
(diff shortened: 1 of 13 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -1162,7 +1162,13 @@ private void fetchNext() {
 				// If we don't have a query yet, create one
 				if (myResultsIterator == null) {
 					if (myMaxResultsToFetch == null) {
-						myMaxResultsToFetch = myDaoConfig.getFetchSizeDefaultMaximum();
+						if (myParams.getLoadSynchronousUpTo() != null) {
+							myMaxResultsToFetch = myParams.getLoadSynchronousUpTo();
+						} else if (myParams.getOffset() != null && myParams.getCount() != null) {
```
tuomoa, line 1167: If we would always set `myMaxResultsToFetch = myParams.getCount()` when `myParams.getCount() != null` some tests start to fail (which I presume makes assumptions that if synchronous search is used then the defaultPageSize won't matter (which is 10 when running the tests)). These tests were in `ConsentEventsDaoR4Test`.

Now when it's used like this, the assumption is still correct if it wasn't offset query. Other approach to fixing those tests could be to define explicit load synchronous up to but I wasn't sure if that would be good.

tuomoa: Hmm looks like something here broke the fetch everything when no paging provider tests. I'll take a look..

tuomoa: I think the bug I've fixed (return all of the bundle when offset is used instead of (from, to) is making the tests fail, but I think they did not work correctly before either. The second page also contains the Patient/1 that should only be on the first page. (The test actually looks only for the entry count, not the contents, I think before this change the second page contained Patient + 9 Observations instead of the second page of observations).

tuomoa: Is the issue that every everything query does this:
```
				// the one problem with this approach is that it doesn't catch Patients that have absolutely
				// nothing linked to them. So we do one additional query to make sure we catch those too.
```
?

Not really sure what to do about this issue. Should we just disable the offset paging for the everything query when no paging provider is set?

tuomoa: So it looks like the first fetch actually queries patient + 10 other items (10 is the limit). But it drops the eleventh one because it did not know it was queried similar to offset=0 query. And then the rest of the pages contain the actual Patient+count amount of other items. But if there is longer chain it seems like multiple queries are done and offset won't work. So maybe we should just drop the offset support for everything queries?

jamesagnew: @tuomoa FWIW, the `type/$everything` operation is a pretty niche operation. I think it'd be reasonable to just throw an exception if someone tries to use it in offset mode. Technically it's not even supposed to apply to "all patients", but rather the more nuanced "all patients the user has access to".. but the latter isn't really something HAPI FHIR can answer, so all patients is the best we can do.

tuomoa: That sounds good to me, I'll update this PR.

jamesagnew (approved):

PR #2581, opened 2021-04-28
Add Search Parameters for placeholder references to all resource types on boot

### LINKED ISSUE #2580: Add builtin Search Parameters for resource-placeholder extensions.
Previously, we added the following in HAPI FHIR:

**Auto-created placeholder reference targets now have an extension with the URL `http://hapifhir.io/fhir/StructureDefinition/resource-placeholder` and a value of true.** 

We'd like to come with one or more built-in custom SearchParameters to support searching for resources of all types where this extension is present with a given value (i.e. true/false).
Ideally, this will behave just like all of the other SearchParameters that come with an installation of  HAPI-FHIR (e.g. if they're missing, they'll get installed at startup).

tadgh:
Closes #2580 
* Add concept of CanonicalSearchParameter, add decanonicalization.
* Add an SP loader Bean which is only loaded when `placeholder reference targets` are set to true
* Load a new SP for every resourece type
* Test search using new Search Parameter.


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/searchparameter/PlaceholderReferenceSearchParamLoader.java
```diff
@@ -0,0 +1,51 @@
...
+		CanonicalSearchParameter canonicalSearchParameter = new CanonicalSearchParameter();
+		canonicalSearchParameter.setId(theType + "-placeholder-sp");
+		canonicalSearchParameter.setUrl("http://example.org/sp/patient/resource-placeholder");
+		canonicalSearchParameter.setName(theType + "-resource-placeholder");
+		canonicalSearchParameter.setStatus("active");
+		canonicalSearchParameter.setDescription("Search for resources of type [" + theType + "] populated with the resource-placeholder extension");
+		canonicalSearchParameter.setCode("resource-placeholder");
+		canonicalSearchParameter.setBase(Collections.singletonList(theType));
+		canonicalSearchParameter.setType("token");
+		canonicalSearchParameter.setExpression(theType + ".extension('http://hapifhir.io/fhir/StructureDefinition/resource-placeholder').value.as(boolean)");
```
jamesagnew, line 47: I'm 90% sure this won't result in an index row for resources that don't have this extension, but I think it'd be a good idea to double check or this change will cause a whole lot of new index space to be taken up in systems that don't even use placeholder references

might make sense to add a test that runs this SP and a resource without the extension through the in memory matcher to make sure no indexes are generated

might also make sense i suppose to only even run the loader if placeholder references are enabled in the DaoConfig

jamesagnew: Actually maybe an even easier way to test is to just make the expression a constant, and then just execute it directly against an empty resource with `ctx.newFhirPath()`. If it returns zero results you're good

jamesagnew (approved):

PR #2583, opened 2021-04-29
Fix reference to FHIRPath interceptor snippet

apeteri:
The snippet in section [9.6.5 Response Customizing: Evaluate FHIRPath](https://hapifhir.io/hapi-fhir/docs/interceptors/built_in_server_interceptors.html#response-customizing-evaluate-fhirpath) repeats the `ExceptionHandlingInterceptor` example from above; this PR changes it to point to the `FhirPathFilterInterceptor` snippet instead.

image


apeteri:
Not sure about the build failure &ndash; this is a small documentation change.

apeteri:
Sorry for the noise! I tried to rename my branch, which caused it to be deleted and the PR closed automatically as a result.

PR #2587, opened 2021-04-29
Issue 2585 enhance placeholder targets logic

### LINKED ISSUE #2585: Enhance Placeholder target creation identifier rules.
Identifier-based inline match URLs need to populate placeholder reference targets with an identifier.
Reference properties

Auto-Create Placeholder Reference Targets with Identifier
Property: Auto-Create Placeholder Reference Targets
Property: Allow Inline Match URL References Enabled

Presently, the inline match URL is only used to search for the reference target so that it can be resolved to a resource ID. When the Auto-Create Placeholder Reference Targets property is enabled, if such a search does not yield results than a placeholder reference target will be created. Note that a placeholder is itself a resource; we do this so we can resolve the reference to a resource ID.
In order to later update the placeholder, it must include a business identifier. This is created automatically by populating the reference with an identifier for the reference target as illustrated here:

```
"subject": {
   "reference": "Patient?identifier=http://foo|1234", <-- inline match URL instead of resource ID
   "identifier": {                                    <-- business identifier for reference target (e.g. the Patient)
      "system": "http://foo",
      "value": "1234"
   }
},
```

This works except the result is that the referencing resource (e.g. the Observation in the linked example) remains populated with a business identifier for the reference target after it's persisted in the repository. This presents at least two problems: a) we're exposing data from one resource in another; and b) changes to the reference target's business identifier won't be reflected.
We need to keep the existing functionality of using a reference identifier to populate a reference target's identifer; we don't want to remove content from resources.
We need to add new functionality as described in Solution Design below.

Acceptance Criteria

 Existing functionality is preserved.

 New functionality is added as described in Solution Design below.

 Documentation is updated accordingly; a new section before this section entitled Auto-Create Placeholder Reference Targets with Inline Match URL.

Solution Design
Note that a Reference can only have one identifier.
Note that we will never remove the reference's identifier.
Case 1:

IF the inline match URL does include an identifier
AND the reference does not include an identifier
AND a placeholder reference target is to be created
DO use the value of the inline match URL's identifier to populate an identifier in the placeholder

Case 2:

IF the inline match URL does not include an identifier
AND the reference does include an identifier
AND a placeholder reference target is to be created
DO use the value of the reference's identifier to populate an identifier in the placeholder

Case 3:

IF the inline match URL does include an identifier
AND the reference does include an identifier
AND the identifiers are the same
AND a placeholder reference target is to be created
DO use only the value of the reference's identifier to populate an identifier in the placeholder

Case 4:

IF the inline match URL does include an identifier
AND the reference does include an identifier
AND the identifiers are different
AND a placeholder reference target is to be created
DO use both the value of the inline match URL's identifier and the value of the reference's identifier to populate two identifiers in the placeholder

tadgh:
Closes #2585 
* Rework placeholder reference identifier addition to support new rules. 
* Add 4 new tests to support 4 new cases.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_4_0/2585-enhance-placeholder-reference-targets.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_4_0/2585-enhance-placeholder-reference-targets.yaml
@@ -1,0 +1,5 @@
+---
+type: add
+issue: 2585
+title: "When automatically creating a placeholder reference that is set to auto-populate identifiers, logic has been improved. If the reference does not contain an
+identifier, but the inline match URL does, the identifier found in the match URL will be added to the target. If both are populated, they will both be added to the target."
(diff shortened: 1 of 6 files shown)


jamesagnew (approved):
Nice, this is super useful

PR #2623, opened 2021-05-02
Fixed GraphQL token parameter matching bug

### LINKED ISSUE #2622: GraphQL endpoint does not parse token parameters properly
**Describe the bug**
When executing a GraphQL query that uses a token parameter, the parameter is currently being parsed as a code, without a system.

For example, the following GraphQL query is currently being executed as 'Find all Encounters with a type of code "https://example.com/example-system|"' instead of 'Find all Encounters with a type with any code in system https://example.com/example-system':
```
{
  EncounterList(type: "https://example.com/example-system|") {
    id
  }
}
```

**To Reproduce**
1. Register any resource with a Coding or CodeableConcept field with a system and a value (e.g. "https://example.com/example-system|example-code")
2. Search for that resource using either just the system (e.g. "https://example.com/example-system|") or using a system and a value (e.g. "https://example.com/example-system|example-code") using a GraphQL query.
3. Notice that the resource isn't found.

**Expected behavior**
The resource should be returned.

 - HAPI FHIR Version 5.4.0

jarimayenburg:
Fixes #2622

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/graphql/JpaStorageServices.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/graphql/JpaStorageServices.java
@@ -20,6 +20,7 @@ package ca.uhn.fhir.jpa.graphql;
  * #L%
  */
 
+import ca.uhn.fhir.context.FhirContext;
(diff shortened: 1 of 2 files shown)


jamesagnew (approved):
Looks great,thanks!

PR #2636, opened 2021-05-05
Calls with zero request details and non-partitionable resource automa…

tadgh:
…tically go to default now. This updates the test to pass in servletrequest details, simulating a user request


fil512 (approved):
Emergency fix to broken test

PR #2642, opened 2021-05-06
Resolve weakness in history operation

### LINKED ISSUE #2641: Potential Denial of Service in JPA Server via history operation
A weakness in our handling of FHIR [history](http://hl7.org/fhir/http.html#history) operations has been reported.

Specifically, on a server with a very large number of resources, if the history operation is executed by many clients (e.g. 200) concurrently, the server becomes unresponsive and ultimately consumes a large amount of disk and becomes unstable.

Our investigation has revealed that the root cause is a `SELECT COUNT(...)` query that executes at the start of all `_history` operations. Essentially, anytime a `_history` is executed, the server executes 2 SQL statements (statements here are approximate):

1. A `SELECT COUNT(pid) FROM hfj_res_ver` is performed to supply a value for `Bundle.total`
2. A `SELECT * FROM hfj_res_ver ORDER BY updated DESC LIMIT 100` is performed to supply the contents

The second query is executed against an index and is very fast. The first query by its nature requires a full index scan and is slow. Executing it 200 concurrent times quickly overwhelms the database and leads to timeouts, exceptions, and eventually instability.

The proposed fix is as follows:

* A new DaoConfig setting is added. This setting introduces a "history count mode" with 3 options:
    * Cached. This is the new default: A loading cache will be used for history counts, meaning that counts are stored in RAM for up to one minute, and the loading cache blocks all but one client thread per JVM from actually performing the count. This effectively throttles access to the database.
    * Not cached. This is the status quo and does exhibit the weakness described here, but may be appropriate in scenarios where users are trusted and accuracy is always required.
    * No count. This setting avoids the count query entirely, saving time and avoiding this weakness at the expense of not including any total in the response.

A huge thanks to **Zachary Minneker at Security Innovation** who discovered and submitted a responsible disclosure of this issue.

This issue will be resolved for the upcoming 5.4.0 release. A CVE number is forthcoming.

jamesagnew:
Fix #2641


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
```diff
@@ -0,0 +1,255 @@
...
+import static org.hamcrest.Matchers.startsWith;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class JpaHistoryR4Test extends BaseJpaR4SystemTest {
+
+	private static final Logger ourLog = LoggerFactory.getLogger(JpaHistoryR4Test.class);
+
+	@AfterEach
+	public void after() {
+		myDaoConfig.setHistoryCountMode(new DaoConfig().getHistoryCountMode());
```
michaelabuckley, line 35: I don't understand how this works.  Don't you need to save the original value in a @BeforeEach?

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
```diff
@@ -0,0 +1,255 @@
...
+		// Simulate the server requesting the Bundle.total value
+		assertEquals(null, history.size());
+
+		// Simulate the server actually loading the resources
+		history.getResources(0, 10);
+
+		assertEquals(0, myCaptureQueriesListener.countDeleteQueries());
+		assertEquals(0, myCaptureQueriesListener.countInsertQueries());
+		assertEquals(0, myCaptureQueriesListener.countUpdateQueries());
+		assertEquals(1, myCaptureQueriesListener.countSelectQueries());
```
michaelabuckley, line 59: Can you add a message to explain the 1?  Which query did you expect to have run?

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
```diff
@@ -0,0 +1,255 @@
...
+		// Simulate the server requesting the Bundle.total value
+		assertEquals(20, history.sizeOrThrowNpe());
+
+		// Simulate the server actually loading the resources
+		history.getResources(0, 10);
+
+		assertEquals(0, myCaptureQueriesListener.countDeleteQueries());
+		assertEquals(0, myCaptureQueriesListener.countInsertQueries());
+		assertEquals(0, myCaptureQueriesListener.countUpdateQueries());
+		assertEquals(2, myCaptureQueriesListener.countSelectQueries());
```
michaelabuckley, line 84: Same here.  Some day it will change, and someone-not-you might have to figure out why 2 is now 3.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
```diff
@@ -0,0 +1,255 @@
...
+		assertThat(myCaptureQueriesListener.getSelectQueries().get(0).getSql(false, false).toLowerCase(Locale.ROOT), startsWith("select count"));
```
michaelabuckley, line 105: This is nice.   Get this.

hapi-fhir-jpaserver-api/src/main/java/ca/uhn/fhir/jpa/api/config/DaoConfig.java
```diff
@@ -74,6 +75,7 @@
 	)));
 	// update setter javadoc if default changes
 	public static final int DEFAULT_MAX_EXPANSION_SIZE = 1000;
+	public static final HistoryCountModeEnum DEFAULT_HISTORY_COUNT_MODE = HistoryCountModeEnum.COUNT_CACHED;
```
michaelabuckley, line 78: I like the DEFAULT constant.  Very clear.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/PersistedJpaBundleProvider.java
```diff
@@ -241,17 +246,53 @@ public boolean ensureSearchEntityLoaded() {
...
+
+		Function<MemoryCacheService.HistoryCountKey, Integer> supplier = k -> new TransactionTemplate(myTxManager).execute(t -> {
+			HistoryBuilder historyBuilder = myHistoryBuilderFactory.newHistoryBuilder(mySearchEntity.getResourceType(), mySearchEntity.getResourceId(), mySearchEntity.getLastUpdatedLow(), mySearchEntity.getLastUpdatedHigh());
+			Long count = historyBuilder.fetchCount(getRequestPartitionId());
+			return count.intValue();
+		});
+
+		switch (myDaoConfig.getHistoryCountMode()) {
+			case COUNT_ACCURATE: {
+				int count = supplier.apply(key);
```
michaelabuckley, line 280: A bit surprised you didn't wrap up the cases inside three different suppliers.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
```diff
@@ -0,0 +1,255 @@
...
+
+		} finally {
+			threadPool.shutdown();
+		}
+
+		assertEquals(0, myCaptureQueriesListener.countDeleteQueries());
+		assertEquals(0, myCaptureQueriesListener.countInsertQueries());
+		assertEquals(0, myCaptureQueriesListener.countUpdateQueries());
+
+		// We fetch the history resources 20 times, but should only fetch the
```
michaelabuckley, line 237: This is helpful.

michaelabuckley (approved):
Approved.  Minor complaints in the tests.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
jamesagnew: This pattern is a pretty common idiom in tests that change settings in DaoConfig (or ModelConfig). It basically resets to the default value. I guess in this case we could just use the constant since there is one. I'll change it.

michaelabuckley: Oh.  I missed the `new` bit.  I read it as assigning the same value.  Ignore me.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
jamesagnew: Just the resource query, no count query. I'll add a comment.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/PersistedJpaBundleProvider.java
jamesagnew: I actually started with that, but realized that would end up forcing me to put something in the cache even if I had caching disabled, or adding extra logic to not touch the cache.. So this way ended up being simpler

PR #2650, opened 2021-05-11
Prevent consent interceptor from running on metadata

### LINKED ISSUE #2648: ConsentInterceptor is called on metadata calls
Currently, consent interceptor is invoked during calls to /metadata, and the $meta operation. This is an error, and the consent interceptor should ignore checks on those endpoints. 

Solution: 
1. if a request is a meta operation, or navigating to /metadata, the hook should just return
2. We should still call `completeOperationFailure()` and `completeOperationSuccess()`

tadgh:
Closes #2648 
* Add an allowlist for `$meta` operation and `/metadata` which skips running the consent interceptor for most checks. 
* still call `completeOperationSuccess` and `completeOperationFailure`.
* Add test
* Add some asserts to close out #2614 
* Move OPERATION_META constant from JpaConstants to ProviderConstants


fil512 (approved):
looks good

PR #2679, opened 2021-05-26
Updated identifier matcher to only match on non-empty values

### LINKED ISSUE #2678: MDM identifier matcher generates a match for an id with no value
**Describe the bug**
Identfier with just a system creates  a match

**To Reproduce**
Match two different resources that have ID with just a system - this produces a match, but should not.

**Expected behavior**
No match should be generated

nvg-smile:
Closes #2678


fil512 (approved):
perfect

PR #2690, opened 2021-05-31
Fix NPEs in DefaultProfileValidationSupport

rld4l:


jamesagnew:
Hi @rld4l - Thanks for the contribution!

I'm wondering, I can see how the code you fixed would result in an NPE if `fetchAllConformanceResources` is the first method ever called on that object. Doesn't this fix just hide the real issue though, whcih is that the resources haven't been loaded yet? I'm worried about the consequences of this method returning an empty list.

Would it make more sense to actually rework it so that if these variables are null, we just load the real contents first?

PR #2704, opened 2021-06-03
Fix DockerRequiredCondition check.

jmarchionatto:
Also add @RequiresDocker annotation to two tests.


tadgh (approved):
Looks good to me! Thanks for the fix

PR #2705, opened 2021-06-03
search by source reported incorrect size

fil512:


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SourceTest.java
```diff
@@ -218,6 +220,31 @@ public void testSourceDisabled() {
...
+			IBundleProvider result = myPatientDao.search(map);
+			assertThat(toUnqualifiedVersionlessIdValues(result), containsInAnyOrder(patientId));
+		}
+		myPatientDao.delete(new IdType(patientId));
+		{
+			myCaptureQueriesListener.clear();
+			IBundleProvider result = myPatientDao.search(map);
+			myCaptureQueriesListener.logSelectQueries();
+			assertEquals(0, result.size());
+		}
```
tadgh, line 244: I may be missing some context, but I don't see how this test tests the above implementation. I also don't see how the Source field could affect the count. Would you Mind providing a minor explanation?

tadgh (approved):
Approved pending my understanding 😄

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SourceTest.java
fil512: The test failed before the fix and passed after the fix :-)

What's going on is the pids have too much in them, we filter out (e.g. deleted resources) in the resources.  So we should really be calling the size based on the post-filtered resources rather than the pre-filtered pids.

tadgh: ah makes sense 👍🏻

PR #2717, opened 2021-06-09
Impropve transaction Performance

jamesagnew:
At the start of processing a transaction, we now gather up all of the forced IDs, and resolve them in a single SQL statement instead of letting each entry in the bundle handle the resolution one-by-one
Similarly, if there are conditional URLs (aka Match URLs) for conditional creates or conditional updates, and those URLs have a single parameter and that parameter is a token parameter (this seems hyper specific but it's the most common scenario for conditional operations if you think about it) we pre-resolve these in a single SQL statement for all of the URLs too
We then pre-fetch the index table collections for all of these pre-fetched resources in a small set of SQL statements instead of letting each inner operation handle that.

In my tests using a set of sample data, this takes a 7-resource transaction bundle down from executing about 33 SQL statements to executing 13.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
@@ -341,7 +341,11 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 		// Perform actual DB update
 		ResourceTable updatedEntity = updateEntity(theRequest, theResource, entity, null, thePerformIndexing, thePerformIndexing, theTransactionDetails, false, thePerformIndexing);
 
+		IIdType id = myFhirContext.getVersion().newIdType().setValue(updatedEntity.getIdDt().toUnqualifiedVersionless().getValue());
+		theTransactionDetails.addResolvedResourceId(id, new ResourcePersistentId(updatedEntity.getResourceId()));
+
 		theResource.setId(entity.getIdDt());
+
 		if (serverAssignedId) {
 			switch (getConfig().getResourceClientIdStrategy()) {
 				case NOT_ALLOWED:
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
@@ -257,6 +257,7 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 		entity.setResourceType(toResourceName(theResource));
 		entity.setPartitionId(theRequestPartitionId);
 		entity.setCreatedByMatchUrl(theIfNoneExist);
+		entity.setVersion(1);
 
 		if (isNotBlank(theIfNoneExist)) {
 			Set<ResourcePersistentId> match = myMatchResourceUrlService.processMatchUrl(theIfNoneExist, myResourceType, theRequest);
@@ -360,14 +361,15 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 			myMatchResourceUrlService.matchUrlResolved(theIfNoneExist, new ResourcePersistentId(entity.getResourceId()));
 		}
 
-		/*
-		 * If we aren't indexing (meaning we're probably executing a sub-operation within a transaction),
-		 * we'll manually increase the version. This is important because we want the updated version number
-		 * to be reflected in the resource shared with interceptors
-		 */
-		if (!thePerformIndexing) {
-			incrementId(theResource, entity, theResource.getIdElement());
-		}
+		// FIXME: remove?
+//		/*
+//		 * If we aren't indexing (meaning we're probably executing a sub-operation within a transaction),
+//		 * we'll manually increase the version. This is important because we want the updated version number
+//		 * to be reflected in the resource shared with interceptors
+//		 */
+//		if (!thePerformIndexing) {
+//			incrementId(theResource, entity, theResource.getIdElement());
+//		}
 
 		// Update the version/last updated in the resource so that interceptors get
 		// the correct version
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
@@ -342,7 +342,11 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 		ResourceTable updatedEntity = updateEntity(theRequest, theResource, entity, null, thePerformIndexing, thePerformIndexing, theTransactionDetails, false, thePerformIndexing);
 
 		IIdType id = myFhirContext.getVersion().newIdType().setValue(updatedEntity.getIdDt().toUnqualifiedVersionless().getValue());
-		theTransactionDetails.addResolvedResourceId(id, new ResourcePersistentId(updatedEntity.getResourceId()));
+		ResourcePersistentId persistentId = new ResourcePersistentId(updatedEntity.getResourceId());
+		theTransactionDetails.addResolvedResourceId(id, persistentId);
+		if (entity.getForcedId() != null) {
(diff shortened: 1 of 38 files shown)


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
```diff
@@ -1556,24 +1786,63 @@ public void testMassIngestionMode_TransactionWithChanges() {
 
 		myCaptureQueriesListener.clear();
 		mySystemDao.transaction(new SystemRequestDetails(), supplier.get());
-//		myCaptureQueriesListener.logSelectQueriesForCurrentThread();
-		assertEquals(3, myCaptureQueriesListener.countSelectQueriesForCurrentThread());
-		assertEquals(13, myCaptureQueriesListener.countInsertQueriesForCurrentThread());
-		assertEquals(3, myCaptureQueriesListener.countUpdateQueriesForCurrentThread());
+		assertEquals(2, myCaptureQueriesListener.countSelectQueriesForCurrentThread());
+		assertEquals(9, myCaptureQueriesListener.countInsertQueriesForCurrentThread());
```
fil512, line 1790: nice to see these numbers going down across the board!

hapi-fhir-jpaserver-base/src/test/resources/r4/transaction-perf-bundle.json
```diff
@@ -0,0 +1,904 @@
+{
+	"resourceType": "Bundle",
```
fil512, line 2: we sure all these test data are non-proprietary?

fil512 (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -338,9 +339,17 @@ private DaoMethodOutcome doCreateForPostOrPut(T theResource, String theIfNoneExi
...
 
 		// Perform actual DB update
-		ResourceTable updatedEntity = updateEntity(theRequest, theResource, entity, null, thePerformIndexing, thePerformIndexing, theTransactionDetails, false, thePerformIndexing);
+		ResourceTable updatedEntity = updateEntity(theRequest, theResource, entity, null, thePerformIndexing, false, theTransactionDetails, false, thePerformIndexing);
+
+		IIdType id = myFhirContext.getVersion().newIdType().setValue(updatedEntity.getIdDt().toUnqualifiedVersionless().getValue());
+		ResourcePersistentId persistentId = new ResourcePersistentId(updatedEntity.getResourceId());
+		theTransactionDetails.addResolvedResourceId(id, persistentId);
+		if (entity.getForcedId() != null) {
+			myIdHelperService.addResolvedPidToForcedId(persistentId, theRequestPartitionId, updatedEntity.getResourceType(), updatedEntity.getForcedId().getForcedId());
```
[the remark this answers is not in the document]
jamesagnew: I don't actually understand how that could be null (probably an artifact of lazy testing at some point). I'm removing the null guard, since it doesn't make sense to be running HAPI without that service.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
```diff
@@ -65,31 +69,61 @@
...
-			}
+	public <R extends IBaseResource> Set<ResourcePersistentId> processMatchUrl(String theMatchUrl, Class<R> theResourceType, TransactionDetails theTransactionDetails, RequestDetails theRequest) {
+		String resourceType = myContext.getResourceType(theResourceType);
+		String matchUrl = massageForStorage(resourceType, theMatchUrl);
+		ResourcePersistentId resolvedInTransaction = theTransactionDetails.getResolvedMatchUrls().get(matchUrl);
+		if (resolvedInTransaction != null) {
+			return Collections.singleton(resolvedInTransaction);
+		}
+		if (theTransactionDetails.getResolvedMatchUrls().containsKey(matchUrl)) {
+			return Collections.emptySet();
```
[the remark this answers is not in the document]
jamesagnew: I was using the value `null` (i.e. there's a key, but the value in the map is null) as a special case meaning "not found".

Agree that makes this super confusing to read though. I'm replacing the use of null with a special constant called `NOT_FOUND` instead, that has a PID of -1. That makes this much easier to follow.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
```diff
@@ -65,31 +69,61 @@
...
+		if (questionMarkIdx == 0) {
+			return theResourceType + theMatchUrl;
+		}
+		return theResourceType + "?" + theMatchUrl;
+	}
+
+	@Nullable
+	public ResourcePersistentId processMatchUrlUsingCacheOnly(String theResourceType, String theMatchUrl) {
+		ResourcePersistentId existing = null;
+		if (myDaoConfig.getMatchUrlCache()) {
```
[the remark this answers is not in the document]
jamesagnew: Good call, deprecating these methods (they were added in 5.4.0) and adding new ones that are better named

hapi-fhir-jpaserver-base/src/test/resources/r4/transaction-perf-bundle.json
jamesagnew: Yup, I gave it a good solid review before checking in

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
```diff
@@ -58,6 +108,225 @@ protected void validateDependencies() {
 		Validate.notNull(myEntityManager);
 	}
 
+	@VisibleForTesting
+	public void setFhirContextForUnitTest(FhirContext theFhirContext) {
+		myFhirContext = theFhirContext;
+	}
+
+	@Override
+	protected Map<IBase, IIdType> doTransactionWriteOperations(final RequestDetails theRequest, String theActionName, TransactionDetails theTransactionDetails, Set<IIdType> theAllIds,
```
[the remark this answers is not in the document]
jamesagnew: Agree, but I think I've introduced enough risk into this change already (and breaking it up will affect mongo as well).

Let's hold that refactor for later.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
```diff
@@ -58,6 +108,225 @@ protected void validateDependencies() {
...
+				Root<ResourceIndexedSearchParamToken> from = cq.from(ResourceIndexedSearchParamToken.class);
+				List<Predicate> orPredicates = new ArrayList<>();
+
+				for (MatchUrlToResolve next : searchParameterMapsToResolve) {
+					Collection<List<List<IQueryParameterType>>> values = next.myMatchUrlSearchMap.values();
+					if (values.size() == 1) {
+						List<List<IQueryParameterType>> andList = values.iterator().next();
+						IQueryParameterType param = andList.get(0).get(0);
+
+						if (param instanceof TokenParam) {
```
tuomoa, line 220: What about other types of params such as `UriParam`?

Because we are not handling `UriParam` here, and on line 278 the `requestUrl` is marked as not found later on in the resourceDao update search will be skipped and duplicate resource will be created for a conditional url such as `CodeSystem?url=http://example.com/something`.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
```diff
@@ -58,6 +108,225 @@ protected void validateDependencies() {
...
+							}
+
+						}
+
+					}
+
+					for (MatchUrlToResolve nextSearchParameterMap : searchParameterMapsToResolve) {
+						// No matches
+						if (!nextSearchParameterMap.myResolved) {
+							theTransactionDetails.addResolvedMatchUrl(nextSearchParameterMap.myRequestUrl, TransactionDetails.NOT_FOUND);
```
tuomoa, line 278: I think we should not mark matchUrl as not found if it was not processed (if it wasn't TokenParam).

PR #2719, opened 2021-06-09
Use pageSize variable to hold page size

jmarchionatto:
Previously used variable has other function, so not always holds right value required for previous link


hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/rest/server/PagingTest.java
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	/**
+	 * Reproduced: https://github.com/hapifhir/hapi-fhir/issues/2509
+	 *
+	 * A bundle size of 21 is used to make last page the third one to validate that the previous link of the
+	 * last page has the correct offset
+	 */
+	@Test()
+	public void test_previous_link_last_page_when_bunlde_size_equals_page_size_plus_one() throws Exception {
```
jamesagnew, line 111: Nitpick: We use camelCase for test method names in HAPI FHIR, with underscores used occasionally when mutiple tests are related but have some subtle difference in order to group them together (see FhirResourceDaoR4QueryCountTest for some examples of that)

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/rest/server/PagingTest.java
```diff
@@ -0,0 +1,224 @@
...
+	private static Server ourServer;
+	private IPagingProvider myPagingProvider;
+	private RequestDetails myRequestDetails;
+
+	@BeforeAll
+	public static void beforeClass() throws Exception {
+		ourContext = FhirContext.forR4();
+
+		ourPort = PortUtil.findFreePort();
+		ourServer = new Server(ourPort);
```
jamesagnew, line 68: There's a Junit5 extension called `RestfulServerExtension` that can be used to create a HAPI FHIR RestfulServer and a client to go with it, instead of having to manually create them. We use that on all new tests these days. We're trying to phase out using the "findFreePort" approach as fast as possible, because it causes intermittent test failures.

jamesagnew (changes requested):

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/rest/server/PagingTest.java
jmarchionatto: Adjusted the name to standard. BTW, regarding the "grouping for name prefixes", did you have a chance to see JUnit5 @Nested feature? Is awesome! It covers for that and many other niceties (as "group" before and after code, Etc.)

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/rest/server/PagingTest.java
jmarchionatto: Replaced the server by the extension. Thanks for pointing that out.

PR #2736, opened 2021-06-17
add partition support to $evaluate-measure

fil512:


hapi-fhir-jpaserver-cql/src/test/java/ca/uhn/fhir/cql/BaseCqlDstu3Test.java
```diff
@@ -80,11 +109,22 @@ protected Bundle parseBundle(String theLocation) throws IOException  {
...
 		return loadBundle(bundle);
 	}
+
+	private static class MyTestInterceptor {
+		@Hook(Pointcut.STORAGE_PARTITION_IDENTIFY_READ)
+		public void partitionIdentifyRead(RequestPartitionId theRequestPartitionId, ServletRequestDetails theRequestDetails) {
+			if (theRequestDetails == null) {
+				ourLog.info("oops");
+			}
+			assertNotNull(theRequestPartitionId);
```
michaelabuckley, line 126: What if this is never called because of misconfig?

Could you make this non-static and set a field boolean in the test, you can assert it was called in the test method.

hapi-fhir-jpaserver-cql/src/main/java/ca/uhn/fhir/cql/dstu3/helper/LibraryHelper.java
```diff
@@ -96,7 +96,7 @@ public LibraryHelper(Map<org.hl7.elm.r1.VersionedIdentifier, Model> modelCache,
 				id = id.substring(1);
 			}
 
-			org.hl7.fhir.dstu3.model.Library library = libraryResourceProvider.resolveLibraryById(id);
+			org.hl7.fhir.dstu3.model.Library library = libraryResourceProvider.resolveLibraryById(id, theRequestDetails);
```
michaelabuckley, line 99: When I see a param added this many times, I wonder if it should be a field on a request-scoped object.

michaelabuckley (approved):
I'm sure it works now, but make sure the test won't break without failing in the future.

hapi-fhir-jpaserver-cql/src/test/java/ca/uhn/fhir/cql/BaseCqlDstu3Test.java
fil512: good call

hapi-fhir-jpaserver-cql/src/main/java/ca/uhn/fhir/cql/dstu3/helper/LibraryHelper.java
fil512: no question.  we should consider this across the entire hapi-fhir dao layer

hapi-fhir-jpaserver-cql/src/test/java/ca/uhn/fhir/cql/BaseCqlDstu3Test.java
fil512: done

Issue #2589, opened 2021-04-30
On JBoss EAP 7 using RestfulServer, NPE in MethodUtil.getResourceParameters


mbw-ahc:
**Describe the bug**
On Jboss EAP7 if you obtain a IResourceProvider by CDI injection, the restful server will fail to initialize
```
10:36:51,782 ERROR [io.undertow.request] (default task-1) UT005023: Exception handling request to /hapi-fhir-eap7/DiagnosticReport: javax.servlet.ServletException: Failed to initialize FHIR Restful server
    at ca.uhn.fhir.rest.server.RestfulServer.init(RestfulServer.java:1333)
…
Caused by: java.lang.NullPointerException
        at java.lang.Class.isAssignableFrom(Native Method)
        at ca.uhn.fhir.rest.server.method.MethodUtil.getResourceParameters(MethodUtil.java:91)
        at ca.uhn.fhir.rest.server.method.BaseMethodBinding.<init>(BaseMethodBinding.java:74)
…
```
Complete stack trace is here: https://github.com/mbw-ahc/hapi-fhir-eap7#readme

**To Reproduce**
The use case is
* Servlet extending RestfulServer 
* IResourceProvider instance injected as a CDI component into the servlet
* Pass that resource provider instance to RestfulServer#setResourceProviders
* The resource provider makes use of: @IncludeParam(…) Set<Include> theIncludes
* The servlet will fail to initialize with the exception listed above

A small reproduction project can be found here: https://github.com/mbw-ahc/hapi-fhir-eap7

**Expected behavior**
The server successfully initializes

**Screenshots**
N/A

**Environment (please complete the following information):**
 - 5.1.0, and 5.3.3 (only tried those two)
 - OS Windows 10
 - Browser N/A

**Additional context**
Redhat JBoss EAP 7.3 is using WELD 3.1 for CDI. According to https://issues.redhat.com/browse/WELD-1539 and https://issues.redhat.com/browse/WELD-2423 the weld proxies will generate methods with raw types rather than generic types.  Therefore Set<Include> original class will be just a Set in the proxy method.  This causes the exception in the hapi-fhir reflection code running against the proxy instance. 

The guidance from WELD-1539 is to “Therefore, integrators are encouraged to detect client proxies and subclasses and work around the problem via reflection. “

Although found in EAP 7.3, I expect this would be an issue in all EAP 7 versions.
[...]

PR #2754, opened 2021-06-24
Add support for generic types in synthetic classes such as weld proxies (#2589)

tarekmamdouh:
I used isSynthetic method to decide whether the declaring class is a proxy one, then I get the superclass of that declaring class and resolve the method from it instead.


jamesagnew:
Hi @tarekmamdouh - Thanks for the contribution! Are you able to include a unit test demonstrating and verifying this feature?

tarekmamdouh:
@jamesagnew  I'll take a look into this. I didn't find any unit tests for MethodUtil class earlier, but I'll see if I can add some tests around it.

tarekmamdouh:
hi @jamesagnew  I've been trying to implement a unit test for this behavior but I'm facing a challenge because the behavior that exhibits this issue is explicitly observed on JBoss EAP7.x, so in order to have a unit test I need to have a mock of the method to simulate the behavior. 
However, this is not possible, because java.lang.reflect.Method is final with no public constructors, which makes it impossible for easymock to create a mock for it and therefore I am unable to implement the unit test.
Please let me know if you have other creative ideas that can help me implement this kind of tests.

tarekmamdouh:
hi @jamesagnew  Can you please provide some details about the failing check for this PR. I can't find the source code for the failing test, therefor I am unable to understand what the problem is.
I would also appreciate if you have some creative ideas to implement the unit test.

PR #2769, opened 2021-06-30
Add Cortico to HAPI FHIR Global Atlas

countable:
Cortico uses HAPI FHIR to help medical providers engage seamlessly with patients over video, email and SMS.


jamesagnew (approved):
Thanks for the contribution! This should be reflected on the website in ~24h

PR #2773, opened 2021-06-30
change out-of-order flag to switch-order flag so the default behaviour reverses

fil512:
…r reverses


jamesagnew (approved):

michaelabuckley (commented):
This is fine.  Is there another change coming to fix the docs at https://smilecdr.com/docs/smileutil/migrate_database.html#options

PR #2832, opened 2021-07-26
2829 - Add DSTU3 Support To UploadTerminologyCommand

KevinDougan-SmileCDR:


hapi-fhir-cli/hapi-fhir-cli-api/src/test/java/ca/uhn/fhir/cli/HeaderPassthroughOptionR4Tests.java
```diff
@@ -0,0 +1,46 @@
...
+	@Test
+	public void twoHeadersSameKey() throws Exception {
+		super.twoHeadersSameKey(FHIR_VERSION, myRestfulServerExtension.getPort());
+	}
+
+	@Test
+	public void twoHeadersDifferentKeys() throws Exception {
+		super.twoHeadersDifferentKeys(FHIR_VERSION, myRestfulServerExtension.getPort());
+	}
+}
```
tadgh, line 46: I gotta say, I don't see how splitting this test class is helpful. Setting headers is completely identical regardless of which version of FHIR you are using. I would undo this refactor unless you've got an amazing reason for it, because at the end of the day its just harder to read than the original, (i have to go look at a super implementation to see what it is actually doing), and the 2 different FHIR versions exercise nothing differently. If you think the version difference _is_ exercising something of value, then make this test suite parameterized, instead of using inheritance.

tadgh (changes requested):
Implementation looks good, but I think the test suite should be cleaned up a bit.

hapi-fhir-cli/hapi-fhir-cli-api/src/test/java/ca/uhn/fhir/cli/HeaderPassthroughOptionR4Tests.java
KevinDougan-SmileCDR: The HeaderPassThroughOptionTest was originally a Nested class within the UploadTerminologyCommandTest class. Once I re-factored the UploadTerminologyCommandTest class to support multiple FHIR versions, the Nested class no longer worked, so I created it as a separate class so we wouldn't lose any of those unit tests. I saw that it also specified a FHIR-VERSION when executing the App.main class, so I assumed (incorrectly, it seems) that it should also be re-factored to support multiple FHIR versions - I can revert this but I still need to create it as a separate test class to keep those tests intact.
So, we will end up with a new R4-only HeaderPassThroughOptionTest class (since you indicated that it doesn't matter which FHIR version is used), and a re-factored UploadTerminologyCommandTest class that supports DSTU3 and R4 and uses the JUnit Parameterized feature.
Is that an acceptable compromise?

hapi-fhir-cli/hapi-fhir-cli-api/src/test/java/ca/uhn/fhir/cli/UploadTerminologyCommandDSTU3Test.java
```diff
@@ -0,0 +1,94 @@
...
+	@Test
+	public void testSnapshotLargeFile() throws IOException {
+		super.testSnapshotLargeFile(FHIR_VERSION);
+	}
+
+	@Test
+	public void testAddICD10UsingCompressedFile() throws IOException {
+		super.testUploadICD10UsingCompressedFile(FHIR_VERSION);
+	}
+}
```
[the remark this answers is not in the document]
KevinDougan-SmileCDR: I'll re-factor the UploadTerminologyCommandTest class to use the JUnit Parameterized feature - see my comment above.

hapi-fhir-cli/hapi-fhir-cli-api/src/test/java/ca/uhn/fhir/cli/HeaderPassthroughOptionR4Tests.java
tadgh: That sounds perfect 👍🏻

KevinDougan-SmileCDR: I have made the suggested changes to the test classes now and this is ready for a final review.

tadgh (approved):

PR #2861, opened 2021-08-04
Fix loinc extra long designation

frankjtao:
This fix is for LOINC linguistic variants upper loader only, some fields are exceed designation max length (2000), the fix is to truncate the field after the last semicolon before max length.


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/loinc/LoincLinguisticVariantHandler.java
```diff
@@ -77,6 +79,13 @@ private void addDesignation(CSVRecord theRecord, TermConcept concept, String fie
 			return;
 		}
 		
+		// this is for loinc only, some fields are exceed the max length for the designation
+		// truncate the field after the last semicolon before max length
+		field = StringUtil.truncToMax(field, ";", TermConceptDesignation.MAX_VAL_LENGTH);
+		if (isBlank(field)) {
+			return;
```
jmarchionatto, line 86: Maybe a log line indicating a truncation problem?
Also, you may want to add a test covering line 86 to make codecov happy.

frankjtao:
Client changed their mind.

PR #2877, opened 2021-08-09
NPE in MdmResourceUtil

### LINKED ISSUE #2876: NPE caused by bad comparison in MdmResourceUtil.java
MdmResourceUtil.java contains several attempts to performing matching on a system, without considering that the tag may not have a system, causing a nullpointer like so: 

```
return theBaseResource.getMeta().getTag().stream()
           .anyMatch(tag -> tag.getSystem().equalsIgnoreCase(theSystem));
```

if `tag.getSystem() == null` then this code will produce an NPE. The simple fix here is to flip the equality as the incoming system is non-nullable.

tadgh:
Closes #2876 

* Add changelog
* Add backport tag (as i know this is also going to merge to rel_5_5_0
* Flip equality comparisons in stream operations.


nvg-smile (approved):
Looks good to me

PR #2880, opened 2021-08-10
Contained Issue

### LINKED ISSUE #2876: NPE caused by bad comparison in MdmResourceUtil.java

tadgh:
Closes #2876 

* During a Conditional Create in a bundle, set the ID of  the bundled resource  which did _not_ create (e.g. it already existed) as the resource ID for subsequent processing. (Yes, I'm also confused) 
* Minor Refactors abound
* New tests
* Add Changelog


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
```diff
@@ -1415,6 +1415,14 @@ public ResourceTable updateEntity(RequestDetails theRequest, final IBaseResource
 		return entity;
 	}
 
+	private void validateIncomingResourceTypeMatchesExisting(IBaseResource theResource, ResourceTable entity) {
```
fil512, line 1418: Yay a nice explanatory method.  THANK YOU!!  :-)

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirSystemDaoR4Test.java
```diff
@@ -1002,6 +1004,23 @@ public void testTransactionWithConditionalCreate_NoResourceTypeInUrl() {
 	}
 
 
+	@Test
+	public void testTransactionNoContainedRedux() throws IOException {
```
fil512, line 1008: I assume this test failed before you started.  (I wish we clearly had a way to express this--maybe we should add an annotation that can refer to a ticket number like @ReproducesBug("1234"))

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirSystemDaoR4Test.java
```diff
@@ -1010,12 +1029,11 @@ public void testTransactionNoContained() throws IOException {
 		mySystemDao.transaction(mySrd, request);
 
 		// Run a second time (no conditional update)
-		request = loadResourceFromClasspath(Bundle.class, "/r4/transaction-no-contained.json");
```
fil512, line 1013: this change makes me nervous.  was this test asserting behaviour we need to preserve?

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
```diff
@@ -964,6 +965,17 @@ private void addString_Address(String theResourceType, Set<ResourceIndexedSearch
 		return retVal;
 	}
 
+
+	/**
+	 * Helper function to determine if a set of SPs for a resource uses a resolve as part of its fhir path.
+	 */
+	private boolean anySearchParameterUsesResolve(Collection<RuntimeSearchParam> searchParams, RestSearchParameterTypeEnum theSearchParamType) {
```
fil512, line 972: love the fact you pulled this out into its own method.

ignore this: not anything you introduced, but the code does look like it could potentially hit a false positive....

fil512 (approved):

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirSystemDaoR4Test.java
tadgh: Ah right you are, this was a mistaken removal. Fixing

PR #2893, opened 2021-08-13
Issue 2883 unable to load open api docs

### LINKED ISSUE #2883: Unable to load OpenAPI docs

**Describe the bug**
Getting an error when trying to load OpenAPI docs on FHIR endpoint.

**Expected behavior**
OpenAPI docs should load properly.

katiesmilecdr:
Closes #2883 

Implementation URLs that end with trailing slashes cause problems with openAPI page loading as we end up adding a double slash. This ends up causing subsequent failed calls.

### DIFF
--- a/hapi-fhir-server-openapi/src/main/java/ca/uhn/fhir/rest/openapi/OpenApiInterceptor.java
+++ b/hapi-fhir-server-openapi/src/main/java/ca/uhn/fhir/rest/openapi/OpenApiInterceptor.java
@@ -283,7 +283,7 @@ public class OpenApiInterceptor {
 		context.setVariable("SERVER_VERSION", cs.getSoftware().getVersion());
 		context.setVariable("BASE_URL", cs.getImplementation().getUrl());
 		context.setVariable("BANNER_IMAGE_URL", getBannerImage());
-		context.setVariable("OPENAPI_DOCS", cs.getImplementation().getUrl() + "/api-docs");
+		context.setVariable("OPENAPI_DOCS", cs.getImplementation().getUrl() + "api-docs");
 		context.setVariable("FHIR_VERSION", cs.getFhirVersion().toCode());
 		context.setVariable("FHIR_VERSION_CODENAME", FhirVersionEnum.forVersionString(cs.getFhirVersion().toCode()).name());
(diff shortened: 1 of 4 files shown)


hapi-fhir-server-openapi/src/main/java/ca/uhn/fhir/rest/openapi/OpenApiInterceptor.java
```diff
@@ -267,10 +267,17 @@ protected boolean handleResourceRequest(HttpServletResponse theResponse, Servlet
 		return false;
 	}
 
+	public String removeTrailingSlash(String theUrl) {
+		while(theUrl.endsWith("/")) {
+			theUrl = theUrl.substring(0, theUrl.length() - 1);
+		}
+		return theUrl;
```
tadgh, line 274: Can we add a null check in here in case theUrl is null? This will NPE if it is. e.g. `while(theUrl != null && theUrl.endsWith....)`

hapi-fhir-server-openapi/src/main/java/ca/uhn/fhir/rest/openapi/OpenApiInterceptor.java
```diff
@@ -281,9 +288,9 @@ private void serveSwaggerUiHtml(ServletRequestDetails theRequestDetails, HttpSer
 		context.setVariable("DESCRIPTION", cs.getImplementation().getDescription());
 		context.setVariable("SERVER_NAME", cs.getSoftware().getName());
 		context.setVariable("SERVER_VERSION", cs.getSoftware().getVersion());
-		context.setVariable("BASE_URL", cs.getImplementation().getUrl());
+		context.setVariable("BASE_URL", baseUrl);
```
tadgh, line 291: Are you sure you also want to use the slash-removed version here? I would revert this part, as it is not part of the relevant fix, and some downstream clients may rely on a trailing slash.

tadgh (changes requested):
Overall implementation looks good! 

Minor changes requested. Also, needs a changelog associated.

tadgh (approved):
Nice!

PR #2895, opened 2021-08-15
Support batch transaction in parallel for bundle read.

frankjtao:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java
```diff
@@ -1544,5 +1525,70 @@ private static String toStatusString(int theStatusCode) {
...
+			this.myCompletedLatch = theCompletedLatch;
+			this.myRequestDetails = (ServletRequestDetails)theRequestDetails;
+			this.myNextReqEntry = theNextReqEntry;
+			this.myNextRespEntry = theNextRespEntry;			
+		}
+		
+		@Override
+		public Void call() {
+			
+			ArrayListMultimap<String, String> paramValues = ArrayListMultimap.create();
```
michaelabuckley, line 1545: Can we make this body a method back on the main object instead?  Then the worker is a just an async helper and the single threaded case looks the same.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java
```diff
@@ -464,55 +482,18 @@ private IBaseBundle processTransaction(final RequestDetails theRequestDetails, f
...
-			} catch (BaseServerResponseException e) {
-				ourLog.info("Failure processing transaction GET {}: {}", url, e.toString());
-				myVersionAdapter.setResponseStatus(nextRespEntry, toStatusString(e.getStatusCode()));
-				populateEntryWithOperationOutcome(e, nextRespEntry);
-			}
-
+			
+			// Do the read in parallel
+			BundleTask bundleTask = new BundleTask(completionLatch, theRequestDetails, nextReqEntry, nextRespEntry);
+			myExecutor.submit(bundleTask);
```
michaelabuckley, line 491: Do we maintain the same database transaction if we move this to a different thread?

michaelabuckley (changes requested):
NIH is using bundles of type "batch".  We need a similar change applied to the batch path.

I'm not sure about the transaction boundary moving these queries to a new thread.  If we aren't sure which transaction we're in, we should roll the changes to the "transaction" type bundle back.

frankjtao:
Will close this PR, and re-work for the batch

PR #2923, opened 2021-08-27
Add prefix toggle for channel settings

tadgh:
* Add isPrefixChannelName method to IChannelSettings interfaec. 
* This will allow downstream consumers to select whether the channel should use some global prefix.


tadgh:
Closing this out as we have elected to simply remove the functionality

Issue #2870, opened 2021-08-06
Server fails to start (JPA Starter - Tomcat)


ZuSe:
APPLICATION FAILED TO START
***************************

Description:

The bean 'scopedTarget.reindexWriter', defined in class path resource [ca/uhn/fhir/jpa/reindex/job/ReindexEverythingJobConfig.class], could not be registered. A bean with that name has already been defined in class path resource [ca/uhn/fhir/jpa/reindex/job/ReindexJobConfig.class] and overriding is disabled.

Action:

Consider renaming one of the beans or enabling overriding by setting spring.main.allow-bean-definition-overriding=true

06-Aug-2021 10:00:11.858 SEVERE [main] org.apache.catalina.startup.HostConfig.deployWAR Error deploying web application archive [/usr/local/tomcat/webapps/ROOT.war]

This seems to be related to: https://github.com/hapifhir/hapi-fhir/commit/f5dfc3a5ed05329b7d7c0cd89227204705cca277

@fil512  @michaelabuckley 

Version: 5.5.0 PRE-8
OS: Docker / Kubernetes

bratwurzt:
I got the same error. Will prepare PR

ZuSe:
nice

PR #2926, opened 2021-08-27
Remove essentially duplicate bean definition

### LINKED ISSUE #2925: Duplicate ReindexWriter Bean definition
See: https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/264

ReindexJob and ReindexEverythingJob both define identical beans. This should be ripped out and put into a common config file for batch jobs.

tadgh:
Closes #2925 

* Move ReindexWriter Bean definition into the common class, and autowire its usages. 

This is causing a bit of havoc without downstream implementations (see https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/264)


bratwurzt:
Closes #2870

Issue #2793, opened 2021-07-14
Expunge with expungeEverything doesn't leave database in a clean state.


theGOTOguy:
**Describe the bug**

Calling `$expunge` with
```
{
          'resourceType': 'Parameters',
          'parameter': [
              {
                  'name': 'expungeEverything',
                  'valueBoolean': True
              },
              {
                  'name': 'expungePreviousVersions',
                  'valueBoolean': True
              },
              {
                  'name': 'expungeDeletedResources',
                  'valueBoolean': True
              },
              {
                  'name': 'limit',
                  'valueInteger': 1000000000
              },
          ]}
```

should leave the database in a clean state.  In fact, it does not.

**To Reproduce**

Start a JPA server with:
```
docker run -p 8080:8080
                  -e hapi.fhir.bulk_export_enabled=true
                  -e hapi.fhir.expunge_enabled=true
                  -e hapi.fhir.client_id_strategy=ANY
                  -e hapi.fhir.auto_create_placeholder_reference_targets=true
                  hapiproject/hapi:latest
```

Import some FHIR into it.  You could use Synthea for some fake data:
```
git clone https://github.com/synthetichealth/synthea.git
cd synthea
./gradlew build check test
./run_synthea -p 100
for file in output/fhir/hospitalInformation*; do curl -X POST --header "Content-Type: application/json" -d @$file http://localhost:8080/fhir/; done
for file in output/fhir/practitionerInformation*; do curl -X POST --header "Content-Type: application/json" -d @$file http://localhost:8080/fhir/; done
for file in output/fhir/[A-Z]*.json; do curl -X POST --header "Content-Type: application/json" -d @$file http://localhost:8080/fhir/; done
```

Now export this data and save the ndjson using the `$export` endpoint.

Next, use the `$expunge` endpoint with the parameters above to purge all data from the database.   Wait for the job to complete on the server before proceeding.

Finally, take some of your ndjson and use `PUT` to restore it to its previous location at /fhir/resourceType/id.

You will get a 409 conflict error and something like this in the error logs:
```
2021-07-14 06:02:42.388 [http-nio-8080-exec-5] INFO  o.h.e.j.b.internal.AbstractBatchImpl [AbstractBatchImpl.java:208] HHH000010: On release of batch it still contained JDBC statements
[...]

PR #2930, opened 2021-08-31
Clear the memory cache whenever ExpungeEverythingService is used.

theGOTOguy:
The [ResourceExpungeService](https://github.com/hapifhir/hapi-fhir/blob/7a83c76e65befe14c83376ce8dbd2032d6f31808/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/ResourceExpungeService.java#L162) already clears out the memory cache whenever it is used to expunge specific resource.  However, the [ExpungeEverythingService](https://github.com/hapifhir/hapi-fhir/blob/6d37749be88461ce0afb7d8e7e677ce5ddb35e31/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeEverythingService.java#L178) currently does not.  This can lead to errors such as described in #2793 whenever a resource is expunged then re-created.

We solve this issue by clearing the memory cache whenever the ExpungeEverythingService is called.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeEverythingService.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeEverythingService.java
@@ -126,37 +126,37 @@ public class ExpungeEverythingService {
 			counter.addAndGet(doExpungeEverythingQuery("UPDATE " + TermCodeSystem.class.getSimpleName() + " d SET d.myCurrentVersion = null"));
 			return null;
 		});
-		counter.addAndGet(expungeEverythingByType(NpmPackageVersionResourceEntity.class, false));
-		counter.addAndGet(expungeEverythingByType(NpmPackageVersionEntity.class, false));
-		counter.addAndGet(expungeEverythingByType(NpmPackageEntity.class, false));
-		counter.addAndGet(expungeEverythingByType(SearchParamPresent.class, false));
-		counter.addAndGet(expungeEverythingByType(BulkImportJobFileEntity.class, false));
-		counter.addAndGet(expungeEverythingByType(BulkImportJobEntity.class, false));
-		counter.addAndGet(expungeEverythingByType(ForcedId.class, false));
-		counter.addAndGet(expungeEverythingByType(ResourceIndexedSearchParamDate.class, false));
-		counter.addAndGet(expungeEverythingByType(ResourceIndexedSearchParamNumber.class, false));
-		counter.addAndGet(expungeEverythingByType(ResourceIndexedSearchParamQuantity.class, false));
-		counter.addAndGet(expungeEverythingByType(ResourceIndexedSearchParamQuantityNormalized.class, false));
-		counter.addAndGet(expungeEverythingByType(ResourceIndexedSearchParamString.class, false));
-		counter.addAndGet(expungeEverythingByType(ResourceIndexedSearchParamToken.class, false));
-		counter.addAndGet(expungeEverythingByType(ResourceIndexedSearchParamUri.class, false));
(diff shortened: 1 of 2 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeEverythingService.java
```diff
@@ -122,62 +126,80 @@ public void expungeEverything(@Nullable RequestDetails theRequest) {
...
+                if (purgeMemoryCache) {
+                        myTxTemplate.execute(t -> {
+                                myMemoryCacheService.invalidateAllCaches();
+                                return null;
+                        });
+                }
+
+                return outcome;
+        }
+
```
tadgh, line 202: I like the solution, but I have a minor change requested for the approach:  this act of purging should be refactored to a new private function called purgeMemoryCaches() which is executed on line 182. Also, it is probably worth not mutating an existing element of the API of this class, and instead just adding a function which will call our purge method. e.g.
```java
expungeEverythingByTypeThenPurgeCaches(Class<?> theEntityType) {
  expungeEverythingByType(theEntityType);
  purgeAllCaches();
}
```
I like the idea of this change though, caches should definitely be purged if we are purging everything.  Let me know if you'd like a hand with the refactor, or if you are too busy I'm happy to tweak it myself and get it merged.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeHookTest.java
```diff
@@ -65,6 +68,42 @@ public void expungeEverythingHook() throws InterruptedException {
...
+                // Expunge everything with the service.
+                myEverythingLatch.setExpectedCount(1);
+                myExpungeService.expunge(null, null, null, options, mySrd);
+                myEverythingLatch.awaitExpected();
+                assertPatientGone(id);
+
+                // Create it a third time.
+                myPatientDao.update(thePatient, mySrd);
+                assertNotNull(myPatientDao.read(id));
+        }
```
tadgh, line 105: Nice test!

tadgh (approved):
Approved pending minor refactor

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeEverythingService.java
theGOTOguy: The solution you proposed isn't quite correct:  because `expungeEverythingByType` is public, other callers of `expungeEverythingByType` need the cache to be cleared after their calls as well, or alternatively all external calls to `expungeEverythingByType` need to be updated to `expungeEverythingByTypeThenPurgeCaches`.

Hopefully the refactor I implemented maintains the spirit of your request while also fixing the issue:

Instead, I added a private method `expungeEverythingByTypeWithoutPurging` which is used by `expungeEverything`.  At the end of `expungeEverything` and `expungeEverythingByType`, then, we call the private `purgeAllCaches` method.

tadgh: Just read over the changes. Looks good. I always try to squash flag arguments when I see them, and your refactor removes them, so I'm good to merge this 👍🏻

tadgh:
@theGOTOguy  is it OK if I give you the attribution for this fix in a changelog file?

theGOTOguy:
@tadgh Absolutely.  I'm proud to contribute to a project I use and love!

tadgh:
https://github.com/hapifhir/hapi-fhir/pull/2939 is for the attribution. Thanks!

PR #2932, opened 2021-08-31
Fixed $lookup with displayLanguage caching issue

frankjtao:
This is an addendum to https://github.com/hapifhir/hapi-fhir/pull/2908

### DIFF
--- a/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4CodeSystemDesignationTest.java
+++ b/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4CodeSystemDesignationTest.java
@@ -159,6 +159,171 @@ public class ResourceProviderR4CodeSystemDesignationTest extends BaseResourcePro
 		assertEquals("Systolic blood pressure 12 hour minimum", designation.getPart().get(2).getValue().toString());
 
 	}
+	
+	@Test
+	public void testLookupWithDisplayLanguageCaching() {
+		
+		//-- first call with de-AT
+		Parameters respParam = myClient
+			.operation()
+			.onType(CodeSystem.class)
+			.named("lookup")
+			.withParameter(Parameters.class, "code", new CodeType("8494-7"))
+			.andParameter("system", new UriType(CS_ACME_URL))
+			.andParameter("displayLanguage",new CodeType("de-AT"))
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4CodeSystemDesignationTest.java
```diff
@@ -159,6 +159,171 @@ public void testLookupWithoutDisplayLanguage() {
...
+		assertEquals("fr-FR:Systolic blood pressure 12 hour minimum", designation.getPart().get(2).getValue().toString());
+
+		// 3. Systolic blood pressure 12 hour minimum (no language)
+		designation = designationList.get(1);
+		assertEquals("language", designation.getPart().get(0).getName());
+		assertNull(designation.getPart().get(0).getValue());
+		assertEquals("value", designation.getPart().get(2).getName());
+		assertEquals("Systolic blood pressure 12 hour minimum", designation.getPart().get(2).getValue().toString());
+	}
+
```
tadgh, line 326: This is a 150 line test which is pretty unreadable. I'm still not sure exactly what it's doing. Can you possibly extract some of the setup? Do you need 4 separate calls to the service to replicate this bug?

tadgh (approved):
Approved pending two things: 
1. Needs a changelog. 
2. Please modify this MR and explain in plain english the scenario in which this bug would manifest. I assume its related to cache failures due to language, but there is zero description on this MR, and it doesn't look like there is an associated issue. 
3. Please try to shrink that test if at all possible, as it is >150 lines which makes it tough to maintain.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4CodeSystemDesignationTest.java
frankjtao: @tadgh 
1/2: This bug is related to the PR https://github.com/hapifhir/hapi-fhir/pull/2908 to support displayLanguage 
For the $lookup operation, the result was cached based system+code. To support displayLanguage, the the caching should be based on system+code+displayLanguage. I think this is part of displayLanguage implementation. Still need a changelog?
3. I will optimize the test case.

PR #2966, opened 2021-09-07
Fixed NPE in ResourceToFileWriter occurring during bulk exports

### LINKED ISSUE #2965: NullPointerException in ResourceToFileWriter breaks FHIR bulk export
**Describe the bug**
A NullPointerException was possible in the ResourceToFileWriter which broke the FHIR bulk export function.

 - HAPI FHIR Version: 5.4.2

jarimayenburg:
Fixes #2965

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/bulk/export/job/ResourceToFileWriter.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/bulk/export/job/ResourceToFileWriter.java
@@ -117,9 +117,11 @@ public class ResourceToFileWriter implements ItemWriter<List<IBaseResource>> {
 		int count = 0;
 		for (List<IBaseResource> resourceList : theList) {
 			for (IBaseResource nextFileResource : resourceList) {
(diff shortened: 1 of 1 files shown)


tadgh:
Out of curiosity, how did you manage to generate an NPE here? I would love to know what situation you are in that this occurred. Could you share with me: 
1. The export you attempted (specific URL request)
2. The data currently on your server (types, e.g. patients, practitioners, etc).

jarimayenburg:
Sure! I tried to export all data on my server since 2000-01-01 (so all data). There's tens of thousands of resources on that server and it has been running for 2-3 years by now. I'm not sure how some resources in this `resourceList` got to be `null`, but `null` resources can't be written to a file anyway so I'm assuming this change won't cause any issues.

tadgh:
Hmm, honestly, this is probably symptomatic of a bug higher up in the stack. There should never be a null value there. Is there any way you could write a test to replicate this NPE? A good place to write it would be in BulkDataExportSvcImplR4Test.java. It would be super valuable. If not, that's also fine, this MR won't break anything either way.

jarimayenburg:
I think you're right. Let's close this PR and move the discussion to the issue.

PR #2986, opened 2021-09-12
code cleanup

fil512:
Added a helper method to PartitionedUrl (needed for non-persistent batch jobs)
Also switched FhirContext cached calls to use helper method


hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirContext.java
```diff
@@ -225,6 +225,14 @@ private FhirContext(final FhirVersionEnum theVersion, final Collection<Class<? e
 
 	}
 
+
+	/**
+	 * @since 5.5.0
```
jamesagnew, line 230:
```suggestion
	 * @since 5.6.0
```

jamesagnew (approved):

PR #2987, opened 2021-09-13
Validate Conditional Urls are matched by stored resource before completing transaction.

tadgh:
* Once all resources have been stored during transactionWriteOperations, and references have been resolved, execute new validation. 
* New Validation ensures that all resources in a transaction that were marked by conditionals (e.g. PUT at a url, or POST with an ifNoneExist) actually match the resource that was stored. 
* Add test
* Add Changelog


jamesagnew (approved):
nice work - very easy to follow

PR #3001, opened 2021-09-16
Issue 2995 valueset.compose.include.version is not set when uploading terminology

jmarchionatto:


michaelabuckley (approved):
Looks small and thoroughly tested.

PR #3002, opened 2021-09-16
Add scope to unit test dependency

Daron-Cerner:
Add test scope to commons-lang dependency

### DIFF
--- a/hapi-fhir-structures-dstu2.1/pom.xml
+++ b/hapi-fhir-structures-dstu2.1/pom.xml
@@ -176,8 +176,9 @@
 		</dependency>
 		<dependency>
 			<artifactId>commons-lang</artifactId>
-			<groupId>commons-lang</groupId>	
-			<version>2.5</version>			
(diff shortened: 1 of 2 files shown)


jamesagnew:
Hi @Daron-Cerner - Thanks for the contribution. Looks like there's a need for `commons-lang` in the hapi-fhir-validation module's tests as well (the CI build failed because of it). But I think if you were to add this test dependency to that module as well, this should be fine

Daron-Cerner:
> Hi @Daron-Cerner - Thanks for the contribution. Looks like there's a need for `commons-lang` in the hapi-fhir-validation module's tests as well (the CI build failed because of it). But I think if you were to add this test dependency to that module as well, this should be fine

Thanks for letting me know! Added in: https://github.com/hapifhir/hapi-fhir/pull/3002/commits/528690fe6cd8db13f2813b9f18aaa27bf0120e83

tadgh (approved):

PR #3008, opened 2021-09-20
Issue 2851 upload terminology valueset parallel versioning

jmarchionatto:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
```diff
@@ -2364,10 +2421,16 @@ private CodeValidationResult codeSystemValidateCode(String theCodeSystemUrl, Str
...
 		}
 
+		// for loinc CodeSystem last version is not necessarily the current anymore, so if no version is present
+		// we need to query for the current, which is that which version is null
 		if (isNoneBlank(theCodeSystemVersion)) {
 			predicates.add(criteriaBuilder.equal(systemVersionJoin.get("myCodeSystemVersionId"), theCodeSystemVersion));
 		} else {
-			query.orderBy(criteriaBuilder.desc(root.get("myUpdated")));
+			if (theCodeSystemUrl.toLowerCase(Locale.ROOT).contains("loinc")) {
+				predicates.add(criteriaBuilder.isNull(systemVersionJoin.get("myCodeSystemVersionId")));
```
fil512, line 2430: Nothing to change here, just a comment.  Referencing field names as strings makes me super nervous.  the only safe way I know to do this is to move to something like querydsl (what App Gallery entities use), but that's out of scope for this MR.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
```diff
@@ -2489,7 +2552,26 @@ static boolean isDisplayLanguageMatch(String theReqLang, String theStoredLang) {
...
-		
+
 		return theReqLang.equalsIgnoreCase(theStoredLang);
     }
+
+
+	@Override
+	public Optional<IBaseResource> readCodeSystemByForcedId(String theForcedId) {
+		@SuppressWarnings("unchecked")
+		List<ResourceTable> resultList = (List<ResourceTable>) myEntityManager.createQuery(
```
fil512, line 2563: Can you replace this with a Spring Data dao method?

fil512: I think if you make it return Optional it should throw an exception if 2 are returned (but I'm not 100% sure)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/api/ITermReadSvc.java
```diff
@@ -119,4 +116,26 @@
...
 
+	/**
+	 * Version independent
+	 */
+	Optional<TermValueSet> findCurrentTermValueSet(String theUrl);
+
+	/**
+	 * Version independent
+	 */
+	boolean mustReturnEmptyValueSet(String theUrl);
```
fil512, line 127: I recommend moving this and the following methods to static methods in a Util class

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupportTest.java
```diff
@@ -0,0 +1,135 @@
...
+	private IValidationSupport testedClass =
+		new JpaPersistedResourceValidationSupport(theFhirContext);
+
+	private Class<? extends IBaseResource> myCodeSystemType = CodeSystem.class;
+	private Class<? extends IBaseResource> myValueSetType = ValueSet.class;
+
+
+	@BeforeEach
+	public void setup() {
+		ReflectionTestUtils.setField(testedClass, "myTermReadSvc", myTermReadSvc);
```
fil512, line 72: Use @InjectMocks instead of ReflectionTestUtils (can replace 5 lines with 1)

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupportTest.java
```diff
@@ -0,0 +1,135 @@
...
+
+	@Nested
+	public class FetchValueSetTests {
+
+		@Test
+		void fetchValueSetMustUseForcedId() {
+			when(myTermReadSvc.isLoincNotGenericUnversionedValueSet(anyString())).thenReturn(true);
+			when(myDaoRegistry.getResourceDao(ValueSet.class)).thenReturn(myValueSetResourceDao);
+
+			ResourceNotFoundException thrown = assertThrows(
```
fil512, line 115: Your way is functionally equivalent, but for consistency might be better to change it to the format we typically write tests like this:

```
try {
  testedClass.fetchValueSet("string-containing-loinc");
  fail();
  assertThat(e.getMessage(), contains("Couldn't find current version ValueSet for url"));
}
```

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/term/TermVersionAdapterSvcR4Test.java
```diff
@@ -0,0 +1,77 @@
...
+
+	private final TermVersionAdapterSvcR4 testedClass = new TermVersionAdapterSvcR4();
+
+	@Mock private IFhirResourceDao<CodeSystem> myCodeSystemResourceDao;
+	@Mock ServletRequestDetails theRequestDetails;
+	@Mock DaoMethodOutcome theDaoMethodOutcome;
+
+	@Test
+	void createOrUpdateCodeSystemMustHaveId() {
+		org.hl7.fhir.r4.model.CodeSystem codeSystem = new CodeSystem();
```
fil512, line 53:
```suggestion
		CodeSystem codeSystem = new CodeSystem();
```
and the below ones

fil512 (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupport.java
```diff
@@ -92,14 +107,55 @@ public JpaPersistedResourceValidationSupport(FhirContext theFhirContext) {
 
 	@Override
 	public IBaseResource fetchCodeSystem(String theSystem) {
+		if (myTermReadSvc.isLoincNotGenericUnversionedCodeSystem(theSystem)) {
+			Optional<IBaseResource> currentCSOpt = getCodeSystemCurrentVersion(new UriType(theSystem));
+			if (! currentCSOpt.isPresent()) {
+				ourLog.info("Couldn't find current version of CodeSystem: " + theSystem);
+			}
+			return currentCSOpt.orElse(null);
```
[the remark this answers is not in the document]
jmarchionatto: Thanks. Love to discover new ways to make code more readable. BTW, I think Java could still improve about Optional options.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
```diff
@@ -2338,6 +2340,61 @@ public CodeValidationResult codeSystemValidateCode(IIdType theCodeSystemId, Stri
...
+		if (isLoincNotGenericUnversionedValueSet(theUrl)) {
+			if (mustReturnEmptyValueSet(theUrl))   return Optional.empty();
+
+			String forcedId = theUrl.substring(LOINC_GENERIC_VALUESET_URL_PLUS_SLASH.length());
+			if (StringUtils.isBlank(forcedId))  return Optional.empty();
+
+			return myTermValueSetDao.findTermValueSetByForcedId(forcedId);
+		}
+
+		List<TermValueSet> termValueSetList = myTermValueSetDao.findTermValueSetByUrl(Pageable.ofSize(1), theUrl);
```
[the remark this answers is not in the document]
jmarchionatto: Yes. That can be done, but I was making a point of using the old construction ASIS when not in the specific new use case (which is that inside the previous condition). Replaced code is at line 456.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
```diff
@@ -2338,6 +2340,61 @@ public CodeValidationResult codeSystemValidateCode(IIdType theCodeSystemId, Stri
...
+		if (termValueSetList.isEmpty()) return Optional.empty();
+		return Optional.of(termValueSetList.get(0));
+	}
+
+
+	@Override
+	public boolean mustReturnEmptyValueSet(String theUrl) {
+		if (! theUrl.startsWith(LOINC_GENERIC_VALUESET_URL))   return true;
```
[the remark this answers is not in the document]
jmarchionatto: I also find something not completely right with them, being so trivial. What do you think about making them just static interface methods?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
jmarchionatto: I still need the conversion performed by the toResource method. I could still move the query to a dao but didn't find a place I liked for it.
Also, yes I could get a data exception if more than one is returned by using getSingleResult instead of getResultList, but I thought it was a good chance to throw a more detailed exception for that case, as I saw somewhere else in the code that that use case being considered (constraints being removed), which is pretty scary.
Where would you move it?

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupportTest.java
jmarchionatto: Those constructions are not equivalent. Please see test I sent you on slack

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/BaseTermReadSvcImpl.java
fil512: ok makes sense.  just wanted to ensure you knew that was an option

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupportTest.java
jmarchionatto: Discovered that the Juint5 construction is already heavily used in code so we left it also in this case

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupport.java
```diff
@@ -92,14 +106,55 @@ public JpaPersistedResourceValidationSupport(FhirContext theFhirContext) {
...
 	}
 
+	/**
+	 * Obtains the current version of a ValueSet using the fact that the current
+	 * version is always pointed by the ForcedId for the no-versioned VS
+	 */
+	private Optional<IBaseResource> getValueSetCurrentVersion(UriType theUrl) {
+		if (TermReadSvcUtil.mustReturnEmptyValueSet(theUrl.getValueAsString()))   return Optional.empty();
+
+		String forcedId = theUrl.getValue().substring(LOINC_GENERIC_VALUESET_URL_PLUS_SLASH.length());
```
jamesagnew, line 149: @jmarchionatto Can you clarify the intent here? This feels like it's almost certainly a bug, for 2 reasons:

* It will crash if you supply a URL shorter than the LOINC one
* It changes behaviour for any URL that happens to be the same length as the LOINC one (this seems like it will result in some very hard to diagnose bugs)

jmarchionatto: @jamesagnew conditional will return empty for both cases you mentioned which will make the method return empty.

jmarchionatto: As naming generated confusion will improve it for readability

Issue #2434, opened 2021-03-03
PostgreSQL hfj_search does not reference large objects with `oid`


tuomoa:
**Describe the bug**

The database schema for `hfj_search` table when using PostgreSQL is:
```
CREATE TABLE public.hfj_search
(
    pid bigint NOT NULL,
    created timestamp without time zone NOT NULL,
    failure_code integer,
    failure_message character varying(500) COLLATE pg_catalog."default",
    last_updated_high timestamp without time zone,
    last_updated_low timestamp without time zone,
    num_found integer NOT NULL,
    preferred_page_size integer,
    resource_id bigint,
    resource_type character varying(200) COLLATE pg_catalog."default",
    search_last_returned timestamp without time zone NOT NULL,
    search_query_string text COLLATE pg_catalog."default",
    search_query_string_hash integer,
    search_type integer NOT NULL,
    search_status character varying(10) COLLATE pg_catalog."default" NOT NULL,
    total_count integer,
    search_uuid character varying(36) COLLATE pg_catalog."default" NOT NULL,
    search_deleted boolean,
    search_param_map oid,
    optlock_version integer,
    CONSTRAINT hfj_search_pkey PRIMARY KEY (pid),
    CONSTRAINT idx_search_uuid UNIQUE (search_uuid)
)
```

The problem is that `search_query_string` column is text, even though it actually references a large object.
`ca/uhn/fhir/jpa/entity/Search.java`:
```
	@Lob()
	@Basic(fetch = FetchType.LAZY)
	@Column(name = "SEARCH_QUERY_STRING", nullable = true, updatable = false, length = MAX_SEARCH_QUERY_STRING)
	private String mySearchQueryString;
```

If we run tools such as [vacuumlo](https://www.postgresql.org/docs/11/vacuumlo.html) it will delete those large objects because there is no references from `oid` type of fields to them and leave the database in inconsistent state (and HAPI FHIR cannot resolve this as it fails to clean up the search).

**Expected behavior**

`hfj_search` table column `search_query_string` type is `oid`

**Environment (please complete the following information):**

 - HAPI FHIR Version 5.2.0

tuomoa:
@jamesagnew any insight on this?

tuomoa:
Fixed by #3016. I'm closing this.

PR #3016, opened 2021-09-22
Migrate CLOB columns to LOB

jamesagnew:
Hibernate on Postgres has just awful behaviour when you use a column annotated with `@Lob` of type `String`. In that specific case, it creates a column of type `text` which has a stringified version of the largeobject ID corresponding to the string.

As far as I can tell, Hibernate have no intention of fixing this (it would be hard for them to fix retroactively admittedly). Various apps that use Hibernate have had to introduce workarounds, e.g. https://shibboleth.atlassian.net/browse/OSJ-135

This is just awful behaviour, since this isn't a reference that Postgres actually understands - Meaning that PG sees the CLOB content as an orphan and will reclaim it if you use the `vacuumlo` command.

Fortunately hibernate does the right thing if you use a `@Lob` column with a type of `byte[]`, which is to use the SQL datatype of `oid` (a hard reference).

This PR:

* Replaces any remaining CLOBs with BLOBs to avoid this issue. In other words any existing CLOB columns are replaced with a BLOB column where the conversion from `byte[]` to `String` is handles in application code. We already do exactly this in a number of places to avoid this issue, but did not do it 100% consistently.
* Adds a new code scanner that detects CLOBs and fails the build
* Adds a migration task to migrate any existing data over


tuomoa:
Looks like this will fix #2434, right?

jamesagnew:
@tuomoa yup, it definitely should. Hibernate's default behaviour on Postgres is so weird...

SimonedeGijt:
Hey @jamesagnew, I am just bumping into this issue when upgrading hapi fhir to version 5.6.0. How can we make use of the fix described above?

martinvisser:
@SimonedeGijt Please use this as the hibernate dialect: `ca.uhn.fhir.jpa.model.dialect.HapiFhirPostgres94Dialect`

SimonedeGijt:
Thanks, this worked indeed!

Romanow88:
Hey, as far as I can see, this was already fixed in the latest Hibernate (5.6.2) -> https://hibernate.atlassian.net/browse/HHH-14211

jamesagnew:
Yup - That is just the weirdest timing. This issue has existed for ages in Hibernate, and the second we implement a workaround, they fix it. :)

PR #3022, opened 2021-09-23
Adds a NDJSON parser, in preparation for adding NDJSON $import to HAPI

theGOTOguy:
NDJSON is currently [described as a format in HL7 FHIR](https://www.hl7.org/fhir/nd-json.html).  However, HAPI FHIR does not yet have tooling to parse NDJSON into any useful structure for further processing.  Users who wish to parse NDJSON in an extended operation would either have to implement this themselves or relax the architectural vision of HAPI whereby all messages are FHIR-native in order to pass raw text into their operation for further processing.

This pull request adds an NDJSON parser that is able to take a [FHIR Bundle](https://www.hl7.org/fhir/bundle.html) and encode its entries to legal NDJSON and also take a legal NDJSON file containing FHIR objects and parse it into a [FHIR collection Bundle](https://www.hl7.org/fhir/codesystem-bundle-type.html#bundle-type-collection).  Automatic tests are included.

The motivation for this change is that the NDJSON parser would hopefully become a stepping-stone towards implementing a [`$import` endpoint](https://github.com/smart-on-fhir/bulk-import/blob/master/import.md).  In particular, our plan is to create a fork (or PR if you will accept it) that would have HAPI FHIR's `$import` accept NDJSON [in the exact same NDJSON format emitted by `$export`](https://smilecdr.com/docs/bulk/fhir_bulk_export.html) in order to facilitate data sharing between servers.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirContext.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirContext.java
@@ -15,6 +15,7 @@ import ca.uhn.fhir.parser.IParser;
 import ca.uhn.fhir.parser.IParserErrorHandler;
 import ca.uhn.fhir.parser.JsonParser;
 import ca.uhn.fhir.parser.LenientErrorHandler;
+import ca.uhn.fhir.parser.NDJsonParser;
 import ca.uhn.fhir.parser.RDFParser;
 import ca.uhn.fhir.parser.XmlParser;
 import ca.uhn.fhir.rest.api.IVersionSpecificBundleFactory;
@@ -124,6 +125,7 @@ public class FhirContext {
 	private volatile Set<String> myResourceNames;
 	private volatile Boolean myFormatXmlSupported;
 	private volatile Boolean myFormatJsonSupported;
+	private volatile Boolean myFormatNDJsonSupported;
 	private volatile Boolean myFormatRdfSupported;
 
 	/**
@@ -460,6 +462,7 @@ public class FhirContext {
 		if (retVal == null) {
 			retVal = scanResourceType(theResourceType);
 		}
+
 		return retVal;
 	}
 
@@ -735,6 +738,21 @@ public class FhirContext {
 		return retVal;
 	}
 
+        /**
+         * @return Returns <code>true</code> if the NDJSON serialization format is supported, based on the
+         * available libraries on the classpath.
+         *
+         * @since 5.4.0
+         */
+        public boolean isFormatNDJsonSupported() {
+                Boolean retVal = myFormatNDJsonSupported;
+                if (retVal == null) {
+                        retVal = tryToInitParser(() -> newNDJsonParser());
+                        myFormatNDJsonSupported = retVal;
+                }
+                return retVal;
+        }
+
 	/**
 	 * @return Returns <code>true</code> if the RDF serialization format is supported, based on the
 	 * available libraries on the classpath.
@@ -800,6 +818,22 @@ public class FhirContext {
 	public IParser newJsonParser() {
 		return new JsonParser(this, myParserErrorHandler);
 	}
+
+        /**
+         * Create and return a new NDJSON parser.
+         *
+         * <p>
+         * Thread safety: <b>Parsers are not guaranteed to be thread safe</b>. Create a new parser instance for every thread
+         * or every message being parsed/encoded.
+         * </p>
+         * <p>
+         * Performance Note: <b>This method is cheap</b> to call, and may be called once for every message being processed
+         * without incurring any performance penalty
+         * </p>
+         */
+        public IParser newNDJsonParser() {
+                return new NDJsonParser(this, myParserErrorHandler);
+        }
 
 	/**
 	 * Create and return a new RDF parser.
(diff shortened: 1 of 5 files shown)


jamesagnew:
Hi @theGOTOguy - Thanks for the contribution.

This is a neat idea. I kind of wonder whether implementing this as an implenentation of the `IParser` interface makes sense though. That interface seems like a mismatch for interacting with NDJson:

* It can only parse or encode a single resource at a time, which doesn't make much sense for NDJson
* It has a configuration options that don't make sense in NDJson, such as `setPrettyPrint(..)`

I would personally think NDJson might make more sense to be implemented as an API that would let you do streaming operations like:

```java
NDJsonWriter writer = ctx.newNDJsonWriter(new FileWriter("foo.ndjson");
writer.write(patient1);
writer.write(patient2);
```

and

```java
Iterator<IBaseResource> reader = ctx.newNDJsonReader(new FileReader("foo.ndjson"));
```

Thoughts?

theGOTOguy:
@jamesagnew 

I considered using an approach whereby we would use an iterator to allow parsing through the NDJson file, and I had a few thoughts.  I'm open to refactoring this change in order to make the most useful possible contribution.

I like the idea of using an Iterator type to allow incremental parsing through the NDJSON, since this would lead to a smaller memory footprint.  The decision not to do so was motivated first by my desired use case:  My ultimate goal is to be able to use the files emitted by HAPI FHIR's `$export` in an `$import` operation to restore a HAPI FHIR service to the same state from the command line, e.g. with a `curl` command.  This means that I can't store any additional supporting models on the server for the sake of the operation.  Looking at your [BulkImportJobFileJson](https://github.com/hapifhir/hapi-fhir/blob/060791aeb4f098c81a4beb1cd914dd24de9a8238/hapi-fhir-storage-api/src/main/java/ca/uhn/fhir/jpa/bulk/imprt/model/BulkImportJobFileJson.java#L26), it appears that it is expected that the entire contents of each file to be imported must fit in a String field in memory anyway, which means that to hold the entire NDJSON or a representation thereof in a single structure isn't out of the question as currently planned.  

I see a few routes forward:

1. We could carry on with the design philosophy that all HTTP message bodies should be parsed into FHIR-native structures, using my current NDJsonParser essentially as-is.   FHIR is a very permissive standard:  I personally think that NDJSON makes sense as a "collection" Bundle of FHIR Resources, but I can see that there's room for disagreement there.
2. We could add an additional public method to my current NDJsonParser that allows parsing through an InputStream receiving the parsed objects as an iterator.  To me, this appears to provide the best of both worlds.
3. We could decline to use the IParser interface and instead make NDJsonParser its own top-level class.  I don't love this plan as I don't see unused configuration options as a major sin, especially when it comes with the benefit of maximizing code re-use.  I might sooner change IParser and push the unused configuration options down to the derived classes so that we can avoid re-implementing parts of IParser and adding additional complexity to message parsing.

Let me know what you prefer.  If your team is willing to adopt this PR and reduce our team's maintenance overhead by that much, it will be worth refactoring this change.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirContext.java
```diff
@@ -735,6 +738,21 @@ public boolean isFormatJsonSupported() {
 		return retVal;
 	}
 
+        /**
+         * @return Returns <code>true</code> if the NDJSON serialization format is supported, based on the
+         * available libraries on the classpath.
+         *
+         * @since 5.4.0
```
jamesagnew, line 745:
```suggestion
         * @since 5.6.0
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirContext.java
```diff
@@ -801,6 +819,22 @@ public IParser newJsonParser() {
 		return new JsonParser(this, myParserErrorHandler);
 	}
 
+        /**
+         * Create and return a new NDJSON parser.
```
jamesagnew, line 823: Recommend documenting here how the parse and serialize methods expect to work with Bundles

jamesagnew:
@theGOTOguy Ok fair enough, I see your point.

My worry with this approach is not being able to stream if you want to (even if your first use case doesn't need that) but I suppose we could deal with that by creating an `INDJsonParser` interface that extends `IParser` and have NDJsonParser implement that (and make that the return type of `newNDJsonParser()`. We could add streaming methods to that interface later when they're needed.

One fun benefit I see to your approach here is that presumably the RestfulServer would easily be able to take advantage of it to return NDJson if needed. For example, I could see it being really nice to be able to ask the server `Patient?name=foo&_format=ndjson` and instead of a Bundle as the response you get the results returned to you as NDJson.

I've submitted a review with a couple of minor suggestions. Happy to merge this with those addressed.

jamesagnew (changes requested):

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirContext.java
theGOTOguy: I've added some additional description below, in terms of how the NDJsonParser expects Bundle types.

Let me know if this is sufficient, or if there's some alternate verbiage you'd prefer.

PR #3025, opened 2021-09-23
Issue 3017 remove loincxml file from application used as fallback when loading loinc terminology

jmarchionatto:


fil512 (approved):
nice work.  your code is always so easy to read.

PR #3034, opened 2021-09-24
Try commenting out clearAllStaticFieldsForUnitTest to see what happens

fil512:


jamesagnew:
Hi @fil512 

FYI the stuff that is commented out does 2 things:

* randomizeLocale() sets the timezone, platform file endings, and system locale to a randomized value in order to catch any unexpected dependencies on a particular platform/locale in the unit tests. Removing this won't break anything, but IMO it is useful behaviour.

* The "clear static fields" stuff is there to avoid a situation where each test class has its own static variable with a FhirContext in it. At one point this was causing us to run out of memory, which is why we added this. It may well not be an issue anymore given that we generally use the `FhirContext.forCached()` methods now, and we're no longer on Travis (which had much lower available resources in its runners)

fil512:
Ah good to know.  The randomizeLocale stuff is definitely useful.  If the tests pass, I'll remove the clear static stuff and rename the method to randomizeLocale()

jamesagnew (approved):

hapi-fhir-android/src/test/java/ca/uhn/fhir/android/client/GenericClientDstu3IT.java
```diff
@@ -280,7 +296,7 @@ public void testCreateWithPreferRepresentationServerReturnsResource() {
 	
 	@AfterAll
 	public static void afterClassClearContext() {
-		TestUtil.clearAllStaticFieldsForUnitTest();
+		TestUtil.randomizeLocaleAndTimezone();
```
michaelabuckley, line 299: I hate to say it, but now that all you have left is TestUtil.randomizeLocaleAndTimezone(), should these really be called in @AfterAll instead of @BeforeAll?  It's just poisoning the "next" test, but will never cause a test to fail when run standalone.

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/TestUtil.java
```diff
@@ -52,66 +44,18 @@ public static void setShouldRandomizeTimezones(boolean theShouldRandomizeTimezon
...
-
-		randomizeLocale();
-
-		/*
-		 * If we're running a CI build, set all loggers to TRACE level to ensure coverage
-		 * on trace blocks
-		 */
-		try {
-			if ("true".equals(System.getProperty("ci"))) {
-				for (Logger next : ((LoggerContext) LoggerFactory.getILoggerFactory()).getLoggerList()) {
```
michaelabuckley, line 100: I did a quick look and didn't see any conditionals around trace.  This should be fine.

michaelabuckley (approved):
Approved - but the @AfterAll should really be @BeforeAll.

hapi-fhir-android/src/test/java/ca/uhn/fhir/android/client/GenericClientDstu3IT.java
fil512: that looks like a lot of work :-).  For another day...

fil512:
pre-merge build passed.  don't want to wait for the next mr to pass again before merging

PR #3038, opened 2021-09-27
Remove parallel non gets

tadgh:
* When processing in batch, remove the capability for batch mutating operations to operate in parallel. 
* add test 
* This is a backport from nih-testing.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4BundleTest.java
+++ b/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4BundleTest.java
@@ -118,10 +118,8 @@ public class ResourceProviderR4BundleTest extends BaseResourceProviderR4Test {
 
 	@Test
 	public void testHighConcurrencyWorks() throws IOException, InterruptedException {
-		myDaoConfig.setBundleBatchPoolSize(20);
-		myDaoConfig.setBundleBatchMaxPoolSize(100);
 		List<Bundle> bundles = new ArrayList<>();
-		for (int i =0 ; i < 20; i ++) {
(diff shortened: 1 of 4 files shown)


jmarchionatto (approved):

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4BundleTest.java
```diff
@@ -110,6 +116,24 @@ public void testBundleBatch() {
 
 	}
 
+	@Test
+	public void testHighConcurrencyWorks() throws IOException, InterruptedException {
+		myDaoConfig.setBundleBatchPoolSize(20);
```
jamesagnew, line 121: Wouldn't you want to set the batch pool and max batch pool sizes to something smaller than the number of entries in the batch in order to test this? (e.g. set max to 10 since there are 20 entries). Edit: I realize now that you are sending the bundle multiple times in parallel below, so the 20 max is probably good.. but:

Actually, thinking about this more.. Don't these values only get read on startup? Meaning setting them here wouldn't actually have any effect. IMO you'd be safe to just set them to something smaller (e.g. 10) globally for the tests via the appctx. Maybe just assert that they're set to that small number in this test to prevent someone from changing it later

jamesagnew (approved):

tadgh:
Actually since the executor is lazy loaded they're checked on first
execution

jamesagnew:
@tadgh but in a full run of all of the tests, you aren't guaranteed that this method is the first time a batch will be executed..

tadgh:
> @tadgh but in a full run of all of the tests, you aren't guaranteed that this m

Ahh you're right. I was assuming that this test suite would be running its own spring context, but that isn't necessarily true eh. good point. Workin on it now.

PR #3048, opened 2021-10-04
add authorization check to inline matches

JasonRoberts-smile:


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
```diff
@@ -143,6 +160,27 @@ public ResourcePersistentId processMatchUrlUsingCacheOnly(String theResourceType
...
 		Set<ResourcePersistentId> retVal = dao.searchForIds(theParamMap, theRequest, theConditionalOperationTargetOrNull);
 
+		// Interceptor broadcast: STORAGE_PRESHOW_RESOURCES
+		if (CompositeInterceptorBroadcaster.hasHooks(Pointcut.STORAGE_PRESHOW_RESOURCES, myInterceptorBroadcaster, theRequest)) {
+			Map<IBaseResource, ResourcePersistentId> resourceToPidMap = new HashMap<>();
+			for (ResourcePersistentId pid : retVal) {
+				resourceToPidMap.put(dao.readByPid(pid), pid);
+			}
+
+			SimplePreResourceShowDetails accessDetails = new SimplePreResourceShowDetails(resourceToPidMap.keySet());
```
jamesagnew, line 170: Now that we have it working, I would like to see what can be done to optimize it (as we initially discussed). My worry about using the exact approach here is that we have to do an extra resource lookup. 

I'm wondering if we could create a new implementation of the `IPreResourceShowDetails` that lazy-loads the resource body. That way we'd only ever load it if we actually need to for a security check.

Additionally, it'd be great to add a test method to `FhirResourceDaoR4QueryCountTest` that verifies using an inline match URL with an authorization interceptor registered and verifies that we're only adding one additional select by doing this.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
```diff
@@ -99,7 +109,14 @@
...
 		paramMap.setLoadSynchronousUpTo(2);
 
-		Set<ResourcePersistentId> retVal = search(paramMap, theResourceType, theRequest, theConditionalOperationTargetOrNull);
+		Set<ResourcePersistentId> retVal;
+		try {
+			retVal = search(paramMap, theResourceType, theRequest, theConditionalOperationTargetOrNull);
+		} catch (ForbiddenOperationException e) {
+			// If the search matches a resource that the user does not have authorization for,
+			// we want to treat it the same as if the search matched no resources, in order not to leak information.
+			retVal = new HashSet<>();
```
jamesagnew, line 118: I think it probably makes sense to log a warning here - I like the idea of being careful not to leak information, but I bet some user is gonna end up getting tripped up by this. A logged warning should help the figure out what's going on.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
```diff
@@ -143,6 +160,27 @@ public ResourcePersistentId processMatchUrlUsingCacheOnly(String theResourceType
...
+				.addIfMatchesType(ServletRequestDetails.class, theRequest);
+			CompositeInterceptorBroadcaster.doCallHooks(myInterceptorBroadcaster, theRequest, Pointcut.STORAGE_PRESHOW_RESOURCES, params);
+
+			retVal = accessDetails.toList()
+				.stream()
+				.map(resourceToPidMap::get)
+				.filter(Objects::nonNull)
+				.collect(Collectors.toSet());
+		}
+
```
jamesagnew, line 183: One additional thought (not related to this line, just putting it here) - MatchResourceUrlService can also return cached results if `DaoConfig#setMatchUrlCacheEnabled` is enabled. It'd definitely be a good idea to add a test where the match results are already in the cache (e.g. by first performing the conditional operation without registering the authorization interceptor, then doing it a second time with the auth interceptor registered), and verify that we're not leaking info that way.

jamesagnew (changes requested):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
JasonRoberts-smile: I knew about the cache, but I didn't fully understand how it was being used here. In order to handle this case, I had to move the interceptor call up one level in the call stack.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
JasonRoberts-smile: Done

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
JasonRoberts-smile: Regarding the query count, calling the authorization interceptor adds two additional selects, one against the resource table and one against the history table, but the test confirms that this pair of queries is being executed only once per resource per transaction.

JasonRoberts-smile: Regarding lazy-loading, if no interceptor is listening for the `STORAGE_PRESHOW_RESOURCES` pointcut, the call to hasHooks() above short circuits the whole block, and no additional resource lookups are performed. If an interceptor is listening, it seems almost certain that it will want to access the resource, so I don't see what we are buying with the added complexity of a lazy load.

jamesagnew: I true I suppose - Well, as long as you can confirm with a test in `FhirResourceDaoR4QueryCountTest` that subsequent uses of the same match URL don't cause any additional selects (since they should come from the cache when it's enabled) I guess that's ok.

I'm trying to be really performance sensitive for this code, it gets called heavily during large data loads so small changes have a bit of a multiplier effect.. caching should avoid that though

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
jamesagnew: Maybe I mnissed it, but do you have a test verifying that this doesn't have a risk of someone getting inappropriate information due to getting cached results?

JasonRoberts-smile: Yes, I added a third case in `FhirSystemDaoR4Test` verifying that the cached data gets authenticated, and the new test case in `FhirResourceDaoR4QueryCountTest` verifies that cached data only gets authenticated once per transaction even if it's used more than once.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/MatchResourceUrlService.java
JasonRoberts-smile: There actually seem to be two caches in play, one that is local to the transaction and one that crosses transactions. The latter gets authenticated, since it may have been loaded by a different user with different privileges. The former does not get authenticated, and ensures that the cross-transaction cache only gets hit once per resource per transaction.

PR #3049, opened 2021-10-04
[3040] finish implementation

katiesmilecdr:


tadgh (changes requested):
Hey Katie, this looks great! Only a couple minor things that honestly were just requirements that weren't in the original issue, but I thought up now: 

1. create link should fail if a link already exists between a given golden resource and target resource. 
2. A source record may have precisely _one match_. Your current implementation will allow people to blindly make a single source resource MATCH to _multiple_ golden records, this is not allowed. I think to alleviate this, we need to check if the sourceResource has _any MdmLink_ that has type MATCH. if it does, and they are trying to _create_ a new MATCH link, then we should fail indiciating there is already a link (which they can go find out via $query-links)
3. Allow the user to pass in which match type they want, from the set of `MATCH, POSSIBLE_MATCH, NO_MATCH`

PR #3050, opened 2021-10-04
Allow canonical references to support treating given base URLs as local. 

### LINKED ISSUE #2843: _include does not work for QuestionnaireResponse:questionnaire
**Describe the bug**
Per FHIR spec, QuestionnaireResponse.questionnaire is a canonical. But when using a full URL for this field, a QuestionnaireResponse search request with _include=QuestionnaireResponse:questionnaire does not return the Questionnaire resource the QR refers to.
Ironically, if you use relative URL for the "questionnaire" field, the _include works, but FHIR spec requires that
 the QuestionnaireResponse.questionnaire should be a canonical.

**To Reproduce**
- Use/load the attached (see below) sample Questionnaire and QuestionnaireResource files or create/load your own,
- Make sure to use "full" URL for the QuestionnaireResponse.questionnaire field. 
- A request like:
     QuestionnaireResponse?_id=MY-QR-RES-ID&_include=QuestionnaireResponse:questionnaire
Should return Questionnaire resource as well.

Attached sample Questionnaire and QuestionaireResponse files:
[questionnaire-hapi.json.txt](https://github.com/hapifhir/hapi-fhir/files/6896612/questionnaire-hapi.json.txt)
[questionnaire-resp-hapi.json.txt](https://github.com/hapifhir/hapi-fhir/files/6895980/questionnaire-resp-hapi.json.txt)

**Expected behavior**
Should included the referenced Questionnaire in the results as requested.

 - HAPI FHIR Version: HAPI FHIR 5.5.0-PRE1-SNAPSHOT (live on hapi.fhir.org as of 2021-07-28)

tadgh:
Closes #2843 

In HAPI-FHIR, we allow people to stuff local references in Canonical type elements. If we detect such an entry, we index it as a resource link. However, we do not support fully qualified Canonicals for reference indexing, _even if_ their baseUrl appears in ModelConfig.getTreatBaseUrlsAsLocal().  This causes queries like `QuestionnaireResponse?_id=abc&_include=Questionnaire:questionnaire` to not successfully return a related Questionnaire if the QR was stored while referring to a fully qualified Questionnaire ID. 

* Modify the ingestion of canonical elements during pre-processing, such that if a canonical begins with a URL found in the `treat-as-local` list, that prefix is stripped, much as prefixed references are stripped. 
* Add new test.
* Add changelog.


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseStorageDao.java
```diff
@@ -228,6 +232,24 @@ private void replaceAbsoluteReferencesWithRelative(IBaseResource theResource) {
 					}
 				}
 			}
+	}
+
+	/**
+	 * Replace Canonical URI's with local references, if we find that the canonical should be treated as local.
+	 */
+	private void replaceAbsoluteUrisWithRelative(IBaseResource theResource, FhirTerser theTerser) {
```
michaelabuckley, line 240: I worry that we shouldn't be changing the data, but lower down in the index extraction.  Do we convert these back into absolute on the way back out?

michaelabuckley (approved):
Approved, assuming the round trip is ok.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseStorageDao.java
tadgh: Ah, we do not, but for what its worth, our current implementation (which I based mine off of), which strips references, _also_ performs this same replacement. See L224 in the same file.

PR #3063, opened 2021-10-12
Bump JENA to deal with CVE-2021-29349

tadgh:


ericprud:
inviting @afs to render his considerable wisdom

PR #3070, opened 2021-10-14
Make MDM listener transactional

jamesagnew:


jamesagnew:
This turned out to be the wrong fix.

PR #3142, opened 2021-11-04
3136 - Provide RemoteTerminologyServiceValidationSupport Implementation For Operation $validate-code

### LINKED ISSUE #3136: Provide RemoteTerminologyServiceValidationSupport Implementation For Operation $validate-code
The purpose of this Issue is to enhance the current support for the $validate-code Operation so it can be provided by an instance of RemoteTerminologyServiceValidationSupport.

Currently, the exposed operation interface goes directly to the DAO layer, which does not work when the system is configured to use a Remote Terminology Server.

KevinDougan-SmileCDR:
Closes #3136


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/ValueSetOperationProvider.java
```diff
@@ -142,24 +150,37 @@ public IBaseParameters validateCode(
...
-				valueSetIdentifier.setValue(theValueSetUrl.getValue() + "|" + theValueSetVersion);
-			} else {
-				valueSetIdentifier = theValueSetUrl;
-			}
-			IPrimitiveType<String> codeSystemIdentifier;
-			if (theSystemVersion != null) {
-				codeSystemIdentifier = (IPrimitiveType<String>) getContext().getElementDefinition("uri").newInstance();
-				codeSystemIdentifier.setValue(theSystem.getValue() + "|" + theSystemVersion);
+			// If a Remote Terminology Server has been configured, use it
+			if (myValidationSupportChain.isRemoteTerminologyServiceConfigured()) {
```
jamesagnew, line 157: IMO let's just always pass validation requests through the validation support chain, instead of having this branching logic. I don't actually see any reason we need to specifically put it through the DAO method ever, the chain has caching and that kind of nice benefit so might as well always use it - That allows us to drop the `isRemoteTerminologyServiceConfigured` method too so we don't need any special handling in the chain

jamesagnew: ..unless there's something i'm not considering of course...

jamesagnew (changes requested):
Lookin' good - But I think it could actually be simplified even more

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/ValueSetOperationProvider.java
KevinDougan-SmileCDR: Hi @jamesagnew,

I did try using your suggestion of a more simplified approach with my first implementation, but it caused 77 Unit Tests to fail!

This new approach works without negatively impacting those DAO-based Unit Tests and greatly reduces the work required to implement a solution.

Let me know if this new approach is still acceptable, or if you want me to go back and try to address it using the single validation support chain approach, and working on those failing DAO-based Unit Tests...

Thanks!
Kevin

KevinDougan-SmileCDR: Actually, let me try that simplified approach one more time. I might've ran into problems earlier with my Unit Tests not properly cleaning up their artifacts in the @AfterEach method...

jdar8 (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/ValueSetOperationProvider.java
KevinDougan-SmileCDR: The DAO Unit Test failures were not a "quick fix" and therefore, I'll go ahead with this approach, and look for an opportunity to re-factor the implementation to make it simpler in the future.

tadgh:
I am going to merge this for now, with sights towards a refactor defined here: https://github.com/hapifhir/hapi-fhir/issues/3184

PR #3171, opened 2021-11-15
3170 language portion of language code is case insensitive

### LINKED ISSUE #3170: Allow LanguageCodeValidation to be case insensitive on language code portion.
When trying to ingest language code only, uppercase language codes cannot be found.

Description
When we try to ingest a practitioner resource which has communication, with Preferred Binding to Common languages and Max binding to All languages(same mapping as base R4 Practitioner) , it fails with Validation error Validation failed for 'urn:ietf:bcp:47#SGN'. This is failing for lot of values in BCP47

Steps to Reproduce

Submit Any Practitioner resource or Patient resource which has communication element that contains a language code in uppercase only (eg here: SGN)
Receives Validation error for 'urn:ietf:bcp:47#SGN'

TipzCM:
closes #3170 

- Updated CommonCodeSystemsTerminologyService to allow language portion of language-region code to be case insensitive
- updated changelog

### DIFF
--- a/hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/CommonCodeSystemsTerminologyServiceTest.java
+++ b/hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/CommonCodeSystemsTerminologyServiceTest.java
@@ -12,6 +12,7 @@ import org.junit.jupiter.api.Test;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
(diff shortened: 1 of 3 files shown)


hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/CommonCodeSystemsTerminologyServiceTest.java
```diff
@@ -54,6 +55,26 @@ public void testUcum_LookupCode_UnknownSystem() {
...
+		assertNotNull(outcomeUpper);
+		assertNotNull(outcomeLower);
+		assertTrue(outcomeLower.isFound());
+		assertTrue(outcomeUpper.isFound());
+	}
+
+	@Test
+	public void lookupCode_languageAndRegionLookup_languageIsCaseInsensitive() {
+		IValidationSupport.LookupCodeResult outcomeUpper = mySvc.lookupCode(newSupport(), "urn:ietf:bcp:47", "EN-US", "English");
+		IValidationSupport.LookupCodeResult outcomeLower = mySvc.lookupCode(newSupport(), "urn:ietf:bcp:47", "en-US", "English");
```
tadgh, line 71: while we're in here, we might as well support `en-us` too. Can you add that test case?

tadgh:
Approved pending two things: 
1. A changelog. 
2. one minor addition to support lower case region. We might as well do it now, since we're probably going to get asked for it anyhow.

tadgh (approved):

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/CommonCodeSystemsTerminologyServiceTest.java
TipzCM: done

PR #3175, opened 2021-11-18
Updated cql docs

JPercival:
* Additional documentation for $evaluate-measure


KevinDougan-SmileCDR (approved):
Thanks for the detailed documentation and insight into how this feature works!

PR #3183, opened 2021-11-23
Add executeRawSql() method to Migrator

tadgh:
Currently executing driver-based raw sql is limited to Tasks in which you are adding a table. This pull request:

 *   Adds the ability to create a single migration which accepts a map of Driver -> SQL String for its SQL argument.
  *  Adds a test

Changelog is omitted here as there is nothing public facing that I'm aware of.


hapi-fhir-sql-migrate/src/test/java/ca/uhn/fhir/jpa/migrate/taskdef/ExecuteRawSqlTaskTest.java
```diff
@@ -81,5 +82,40 @@ public void testOnlyAppliesToPlatforms(Supplier<TestDatabaseDetails> theTestData
 		}
 	}
 
+	@ParameterizedTest(name = "{index}: {0}")
+	@MethodSource("data")
+	public void testDriverTypeBasedRawSqlExecution(Supplier<TestDatabaseDetails> theTestDatabaseDetails) {
```
fil512, line 87: Wow that's a really cool test!

fil512 (approved):

PR #3192, opened 2021-11-25
Support https with elastic

ZuSe:
This is providing support for https protocol when using Elastic or OpenSearch

Many cloud providers such as Aiven, OVH or AWS only provide their services with SSL in front (what makes basically sense).

It fixes https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/275

MR to support proper configuration in jpa-starter is coming soon.


tadgh (approved):
Looks good. Thanks for the contribution!

PR #3193, opened 2021-11-25
test fix

TipzCM:


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/match/registry/SubscriptionCanonicalizer.java
```diff
@@ -105,12 +106,26 @@ private CanonicalSubscription canonicalizeDstu2(IBaseResource theSubscription) {
 	 * @return A map of tags System:Code
 	 */
 	private Map<String, String> extractTags(IBaseResource theSubscription) {
-		return theSubscription.getMeta().getTag()
-			.stream()
-			.collect(Collectors.toMap(
-				IBaseCoding::getSystem,
-				IBaseCoding::getCode
-			));
+		List<? extends  IBaseCoding> tags = theSubscription.getMeta().getTag();
```
fil512, line 109: I don't think we should log these.  I recommend just adding a filter to the original stream code to filter out the null ones.

fil512 (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/match/registry/SubscriptionCanonicalizer.java
```diff
@@ -107,6 +108,10 @@ private CanonicalSubscription canonicalizeDstu2(IBaseResource theSubscription) {
 	private Map<String, String> extractTags(IBaseResource theSubscription) {
 		return theSubscription.getMeta().getTag()
 			.stream()
+			.filter(t -> {
+				// ignore the tags with null system or code
+				return !(t.getSystem() == null || t.getCode() == null);
+			})
```
fil512, line 114:
```suggestion
			.filter(t -> t.getSystem() != null && t.getCode() != null)
```

fil512 (approved):

PR #3203, opened 2021-12-01
Add executeRawSqlStub() to Migrator

joanne-m:
For skipping failing migrations

### DIFF
--- a/hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/api/Builder.java
+++ b/hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/api/Builder.java
@@ -76,9 +76,19 @@ public class Builder {
 	}
 
 	public BuilderCompleteTask executeRawSql(String theVersion, @Language("SQL") String theSql) {
-		ExecuteRawSqlTask task = new ExecuteRawSqlTask(myRelease, theVersion).addSql(theSql);
-		mySink.addTask(task);
+		ExecuteRawSqlTask task = executeRawSqlOptional(false, theVersion, theSql);
 		return new BuilderCompleteTask(task);
+	}
+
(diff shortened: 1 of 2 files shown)


hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/api/Builder.java
```diff
@@ -76,9 +76,19 @@ public BuilderAddTableRawSql addTableRawSql(String theVersion, String theTableNa
...
+
+	public void executeRawSqlStub(String theVersion, @Language("SQL") String theSql) {
+		executeRawSqlOptional(true, theVersion, theSql);
+	}
+
+	private ExecuteRawSqlTask executeRawSqlOptional(boolean theDoNothing, String theVersion, @Language("SQL") String theSql) {
 		ExecuteRawSqlTask task = new ExecuteRawSqlTask(myRelease, theVersion).addSql(theSql);
+		task.setDoNothing(theDoNothing);
 		mySink.addTask(task);
-		return new BuilderCompleteTask(task);
```
fil512, line 81: I recommend keeping new BuilderCompleteTask(task).  It has helper fluid methods we can use in the future.

fil512 (approved):

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/api/Builder.java
joanne-m: It's on line 80. The stub isn't returning anything so I kept it on the original executeRawSql. Should the private method return BuilderCompleteTask instead?

fil512: Oh I see.  Good point, I missed that.  Disregard my comment.

PR #3230, opened 2021-12-09
Improve mail service

VK-SMILECDR:


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/mail/MailSvc.java
```diff
@@ -36,41 +36,44 @@
...
+							@Nonnull ExceptionConsumer theErrorHandler) {
 		Validate.notNull(theEmail);
 		Validate.notNull(theOnSuccess);
 		Validate.notNull(theErrorHandler);
 		try {
-			final AsyncResponse asyncResponse = theMailer.sendMail(theEmail, true);
+			final AsyncResponse asyncResponse = myMailer.sendMail(theEmail, true);
 			if (asyncResponse != null) {
 				asyncResponse.onSuccess(theOnSuccess);
 				asyncResponse.onException(theErrorHandler);
```
tadgh, line 79: This feels a bit odd. When the call returns, we call _both_ the onSuccess and errorHandler? Is this  AsyncReponse  repsonsible for determining whether it actually  errorred out?

tadgh:
Looks good, but I think this would benefit from a changelog, as this affects the API surface for extenders of HAPI-FHIR

tadgh (approved):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/mail/MailSvc.java
VK-SMILECDR: Since it is async, the method does not actual send email. AsyncResponse mimics promise behaviour, that why need to set both callbacks (error and success), when actual call will be executed.

[Reference docs](https://www.simplejavamail.org/features.html#section-handling-async-result)

PR #3233, opened 2021-12-10
Loop over new list to avoid ConcurrentModificationException in case a…

jmarchionatto:
… new hook is added while looping


tadgh:
Out of curiosity, how did you ever run into this?

tadgh (approved):
approved pending a question

jmarchionatto:
> 

ConcurrentModificationException was raised because while BaseSubscription is looping over the hooks, the compliance handling code adds another hook to validate the compliance request's response.

PR #3245, opened 2021-12-13
Allow Subscription to be configured to send delete messages

jmarchionatto:


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/model/CanonicalSubscription.java
```diff
@@ -73,6 +74,8 @@
 	private String myPayloadSearchCriteria;
 	@JsonProperty("partitionId")
 	private Integer myPartitionId;
+	@JsonProperty("mustSendDeleteMessages")
+	private boolean myMustSendDeleteMessages;
```
michaelabuckley, line 78: This name seems overly complicated.  Maybe something like 
```suggestion
	private boolean myIncludeDeleteMessages;
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/HapiExtensions.java
```diff
@@ -130,6 +130,11 @@
 	 */
 	public static final String EX_RETRY_COUNT = "http://hapifhir.io/fhir/StructureDefinition/subscription-delivery-retry-count";
 
+	/**
```
fil512, line 133: Adding comment in random location.

Please add a test to `RestHookTestR4Test` to validate that rest-hook subscription deliverer successfully delivers deleted resources.

I suspect when you test this you may discover that it only works if channel.payload is omitted in the subscription

PR #3281, opened 2022-01-06
2681 member match fix method

jmarchionatto:
Adding changelog


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_7_0/2681-member-match-operation-must-be-invoked-by-post.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
fil512, line 1: is the fix in a different MR?

fil512 (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_7_0/2681-member-match-operation-must-be-invoked-by-post.yaml
jmarchionatto: Yes. commit 21066317183ed333e2077b9aa66b320dd1f30ea6

PR #3295, opened 2022-01-13
Fix batching of mdm-clear tasks

IanMMarshall:


IanMMarshall:
Found an issue in the $mdm-clear batch operation that appears to be similar to one recently fixed for reindexing as part of https://github.com/hapifhir/hapi-fhir/pull/3211.

fil512 (approved):

PR #3297, opened 2022-01-13
3156 multiple tag race condition

TipzCM:
* Fixed race condition issue in BaseHapiFhirDao that could result in multiple tags of the same kind being created
* Added change log


katiesmilecdr (approved):
Looks good to me overall!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
```diff
@@ -449,11 +446,15 @@ protected TagDefinition getTagOrNull(TransactionDetails theTransactionDetails, T
...
 
 				TypedQuery<TagDefinition> q = myEntityManager.createQuery(cq);
-				try {
-					retVal = q.getSingleResult();
-				} catch (NoResultException e) {
-					retVal = new TagDefinition(theTagType, theScheme, theTerm, theLabel);
-					myEntityManager.persist(retVal);
+
+				// synchronize to ensure no duplicates
+				synchronized (this) {
```
tadgh, line 451: This _feels_ like it should work, but  may wildly impact performance when processing transaction bundles. We have a caffeine-backed tag cache that we use, which is supposed to prevent issues like this. See BaseHapiFhirDao.java L425 for an example of reaching into the tag cache. This implementation effectively allows one thread at a time be allowed to create a tag. If a bundle comes in with 1k transactions, with 1k unique tags, this essentially single-threads bundle processing. @jamesagnew what are your thoughts on this? 99.9% of the time this should be a cache hit in this scenario, so maybe the overhead is not that painful?

Maybe it's simpler to add a unique constraint in the DB and catch-and-retry failures that result in UniqueConstraintViolation exceptions. That will catch the edges while not impacting performance as much.

Also, @TipzCM were you able to actually replicate the original failure in test or in practice? I see the mocked dao test, but that doesn't necessarily replicate the real world implications. Any luck getting the failure to occur?

TipzCM: FYI, there is a unique constraint already. It's the error that ended up causing this bug to begin with.

And i did replicate it in my test. Not their exact issue (cause i couldn't replicate via calls through postman), but my test does cause exceptions (and you can try by removing the synchronized blocks and testing if you like).

jamesagnew: Honestly I'm not sure this is a workable solution as developed, for a couple of reasons:

* It won't work in a cluster, since the synchronized block doesn't span JVMs obviously
* It would also fail if two concurrent requests both tried to create resources with a new tag at the same time, if those resources were different types (since in that case there would be 2 different DAOs, so each request would synchronize of a different lock)

You could solve the second point by having some better lock object.. And that would almost certainly be better than the status quo and therefore worth pursuing. I don't love global locks in our request processing since that's a recipe for bad bugs. At a minimum we should do a double-check kind of algorithm. I.e. "check if the tag exists, if it doesn't then synchronize and check again, and if it still doesn't exist, create it"

I think even better than this though would be something like:

* Check if the tag exists
* If it doesn't exist, start a new sub-transaction (using a TransactionTemplate with propagation `REQUIRES_NEW`)
* Inside that sub-transaction, check if the tag exists
* If it still doesn't exist, create it and commit the sub-transaction
* If that fails, repeat this whole process
* If repeating this whole process fails after X times (say, 10?), then throw an InternalErrorException

TipzCM: @tadgh @jamesagnew @katiesmilecdr 

I've updated the code to do exactly this.

Can you guys all recheck please?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
```diff
@@ -466,6 +449,98 @@ protected TagDefinition getTagOrNull(TransactionDetails theTransactionDetails, T
...
+			cq.where(
+				builder.and(
+					builder.equal(from.get("myTagType"), theTagType),
+					builder.isNull(from.get("mySystem")),
+					builder.equal(from.get("myCode"), theTerm)));
+		}
+
+		TypedQuery<TagDefinition> q = myEntityManager.createQuery(cq);
+
+		TransactionTemplate template = new TransactionTemplate(myTransactionManager);
```
jamesagnew, line 479: This looks overall good! One issue though: You need to do the loop outside of the new transaction. Once you're in the transaction, if the write fails you won't be able to try again because the active transaction is already marked for rollback.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDaoTest.java
```diff
@@ -1,11 +1,353 @@
...
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.lenient;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
+@ExtendWith(MockitoExtension.class)
 public class BaseHapiFhirDaoTest {
 
+	private static class TestDao extends BaseHapiFhirDao<Patient> {
```
jamesagnew, line 63: This is a great test - Definitely not suggesting to change it here, but I'd recommend adding an additional test that is more of a traditional integration test, perhaps in `FhirResourceDaoR4CreateTest`. In that test you would concurrently create 10 patients all of which have an identical tag to make sure you they all save correctly.

See `FhirResourceDaoR4Test#testConflictingUpdates()` for a test that is very similar, you could probably copy-paste it more or less verbatim but add a tag to it and make sure that there are no failures.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
TipzCM: updated

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDaoTest.java
TipzCM: Added test in FhirResourceDaoR4CreateTest

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4CreateTest.java
```diff
@@ -968,6 +975,57 @@ public void testCreateWithNormalizedQuantityStorageSupported_SmallerThanCanonica
...
+	public void testResourceWithTagCreationNoFailures() throws ExecutionException, InterruptedException {
+		ExecutorService pool = Executors.newFixedThreadPool(5);
+		try {
+			Coding tag = new Coding();
+			tag.setCode("code123");
+			tag.setDisplay("Display Name");
+			tag.setSystem("System123");
+
+			Patient p = new Patient();
+			IIdType id = myPatientDao.create(p).getId();
```
jamesagnew, line 988: You should get rid of this line, as it pre-creates the tag so that the concurrent requests will never fail even without the fixes you've made in this ticket.

jamesagnew: You know what, nevermind... I was clearly reading with my nose, this line doesn't create any tags at all. Marking this as resolved.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDaoTest.java
```diff
@@ -1,11 +1,353 @@
...
+
+		@Override
+		protected TagDefinition getTagOrNull(TransactionDetails theDetails,
+														 TagTypeEnum theEnum,
+														 String theScheme,
+														 String theTerm,
+														 String theLabel) {
+			// we need to init synchronization due to what
+			// the underlying class is doing
+			TransactionSynchronizationManager.initSynchronization();
```
jamesagnew, line 79: It'd probably be worth putting the following line in a try-finally with `clearSynchronization()` in the finally block

Presumably threads will get reused by later tests, so having a threadlocal set by a previous test could cause unexpected outcomes

TipzCM: added

jamesagnew (approved):
This looks awesome. Thanks for all the revisions on it, feels like a really solid improvement that will help a lot of people. :)

PR #3300, opened 2022-01-14
JUnit jupiter fix for @Disabled tag order.

markiantorno:
`testEverythingWithLargeSet()` had the tags in the wrong order

It appears the latest version of JUnit Jupiter was not respecting the `@Disabled` tag if the `@Test` tag was first. This resulted in this test (which was previously ignored) to run. The test was disabled due to previously failing, so when the tag was no longer recognized, it failed the build.


tadgh:
what a wild bug. nice find!

tadgh (approved):

PR #3309, opened 2022-01-18
Add support for of-type token modifier

### LINKED ISSUE #1892: Support for :of-type Token Modifier
**Describe the issue**
Currently HAPI FHIR does not support `ofType` Token Modifier - there is no such enum value in the [TokenParamModifier](https://github.com/jamesagnew/hapi-fhir/blob/master/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/param/TokenParamModifier.java) and therefore ofType requests are not parsed correctly (contain empty modifier).

**Environment:**
 - HAPI FHIR Version: 4.2.0

Is there a reason why `:of-type` is not supported, or the enum value just needs to be added?

jamesagnew:
Fixes #1892

This PR completes the support for the :of-type modifier. Previously the client and plain server supported the modifier, but the JPA server didn't actually implement the search logic. Now we do!

I'm also dropping the :text modifier's indexing of `Identifier.type.text` because the spec doesn't actually say to index that.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_7_0/1892-drop-text-identifer-search.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_7_0/1892-drop-text-identifer-search.yaml
@@ -1,0 +1,7 @@
+---
+type: remove
+issue: 1892
+title: "In the JPA server, searching using the token `:text` modifier will no longer index and match
+  values in `Identifier.type.text`. Previously we indexed this element, but the FHIR specification does not
+  actually show this field as being indexed for this modifier, and intuitively it doesn't make sense to
+  do so."
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
@@ -1259,6 +1259,11 @@ public abstract class BaseSearchParamExtractor implements ISearchParamExtractor 
 
 	}
 
+	/**
+	 * Note that this should only be called for R4+ servers. Prior to
+	 * R4 the paths could be separated by the word "or" or by a "|"
+	 * character, so we used a slower splitting mechanism.
+	 */
 	@Nonnull
 	public static String[] splitPathsR4(@Nonnull String thePaths) {
 		StringTokenizer tok = new StringTokenizer(thePaths, " |");
@@ -1292,7 +1297,7 @@ public abstract class BaseSearchParamExtractor implements ISearchParamExtractor 
 		}
 	}
 
-	private static String extractValueAsString(BaseRuntimeChildDefinition theChildDefinition, IBase theElement) {
+	protected static String extractValueAsString(BaseRuntimeChildDefinition theChildDefinition, IBase theElement) {
 		return theChildDefinition
 			.getAccessor()
 			.<IPrimitiveType<?>>getFirstValueOrNull(theElement)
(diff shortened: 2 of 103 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_7_0/1892-drop-text-identifer-search.yaml
```diff
@@ -0,0 +1,7 @@
+---
+type: remove
+issue: 1892
+title: "In the JPA server, searching using the token `:text` modifier will no longer index and match
+  values in `Identifier.type.text`. Previously we indexed this element, but the FHIR specification does not
+  actually show this field as being indexed for this modifier, and intuitively it doesn't make sense to
+  do so."
```
tadgh, line 7: https://xkcd.com/1172/

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
```diff
@@ -1239,6 +1259,88 @@ public void start() {
...
+	}
+
+	private static void addIgnoredType(FhirContext theCtx, String theType, Set<String> theIgnoredTypes) {
+		BaseRuntimeElementDefinition<?> elementDefinition = theCtx.getElementDefinition(theType);
+		if (elementDefinition != null) {
+			theIgnoredTypes.add(elementDefinition.getName());
+		}
+	}
+
+	private static String extractValueAsString(BaseRuntimeChildDefinition theChildDefinition, IBase theElement) {
```
tadgh, line 1295: Should these potentially be protected? Feels like they should be useful to subclasses.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
```diff
@@ -1239,6 +1259,88 @@ public void start() {
 
 	}
 
+	@Nonnull
+	public static String[] splitPathsR4(@Nonnull String thePaths) {
```
tadgh, line 1263: is this different for non-r4 versions? Feels generic enough to me.

tadgh (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_7_0/1892-drop-text-identifer-search.yaml
jamesagnew: I thought about this for sure..... Decided that the perf hit wasn't worth the risk, we'll see if anyone complains

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
jamesagnew: Yeah I guess they could be. Will change.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
jamesagnew: I'll add a javadoc.. We don't use this for DSTU3 and below because prior to R4 you could specify multiple paths either as `A or B` or as `A | B`. From R4 onward you can only use the latter, so the splitter can be much more efficient with a stringtokenizer.

PR #3319, opened 2022-01-20
Issue 3299 conformance validation is happening once per client endpoint thread instead as just once per client endpoint

jmarchionatto:


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/client/ClientThreadedCapabilitiesTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+
+	@BeforeEach
+	public void beforeEach() throws Exception {
+		ourServer.getFhirClient().registerInterceptor(myCountingMetaClientInterceptor);
+	}
+
+
+	@Test
+	public void capabilityRequestSentOnlyOncePerClient() {
```
fil512, line 72: Nice test!

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/RestfulClientFactory.java
```diff
@@ -257,13 +259,21 @@ public void validateServerBaseIfConfiguredToDoSo(String theServerBase, IHttpClie
...
+				break;
+
+			case ONCE:
+				if (myValidatedServerBaseUrls.contains(serverBase)) {
+					break;
+				}
+
+				synchronized (myValidatedServerBaseUrls) {
+					if (!myValidatedServerBaseUrls.contains(serverBase)) {
+						myValidatedServerBaseUrls.add(normalizeBaseUrlForMap(theServerBase));
```
fil512, line 272: It's not obvious that this line will cause the above conditional to be true on the next pass.  They look different.

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/RestfulClientFactory.java
```diff
@@ -257,13 +259,21 @@ public void validateServerBaseIfConfiguredToDoSo(String theServerBase, IHttpClie
...
-		case ONCE:
-			if (!myValidatedServerBaseUrls.contains(serverBase)) {
-				validateServerBase(serverBase, theHttpClient, theClient);
-			}
-			break;
+			case NEVER:
+				break;
+
+			case ONCE:
+				if (myValidatedServerBaseUrls.contains(serverBase)) {
```
fil512, line 266: Is this a thread-safe map?  I guess we want the single-write multi-read one.

fil512 (approved):

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/RestfulClientFactory.java
jmarchionatto: Thanks

jmarchionatto: It is the same, because serverBase was normalized at method entry, but the normalization inside the if is redundant. Changing it. Thanks

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/RestfulClientFactory.java
jmarchionatto: It is a concurrent set, which locking is a more strict than a ReadWriteLock, because we not only need to guarantee exclusion, but also that the capabilities request goes first, which was not sure was going to be achieved using a RW lock. Note that the double checks are to avoid the synchronization when possible.

PR #3343, opened 2022-01-28
add needed dependendencies for mysql and mssql

tadgh:
A recent upgrade in flyway versions caused a regression which meant that we can no longer connect to mysql or mssql databases, as the dependencies from flyway were pulled out into additional dependencies. This MR adds those dependencies, and writes a small test to catch if other dependencies are pulled out.


dmuylwyk (approved):
LGTM

The only thing I'm not sure of are some of the values in the new tests (e.g. CockroachDB). Approved assuming this passes CI.

tadgh:
Yup this is just the internal driver that is used based on the jdbc url provided. I'm essentially just checking "Hey is this non-null and have some sane value

PR #3376, opened 2022-02-10
Can't specify specific resource type permissions for bulk export

oliviayou:
Closes [#3394](https://github.com/hapifhir/hapi-fhir/issues/3394)
* When exporting data that should only belong to a specific group of resources the user is permitted to export, data from other groups is returned. This has been corrected.
* New tests
* Add changelog

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_7_0/3394-bulk-export-permissions-are-over-permissive.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/5_7_0/3394-bulk-export-permissions-are-over-permissive.yaml
@@ -1,0 +1,4 @@
+---
+type: fix
+issue: 3394
+title: "Bulk export permissions were over-permissive in RuleBulkExportImpl. This has been corrected."


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImpl.java
```diff
@@ -66,7 +66,7 @@
 			}
 			for (String next : options.getResourceTypes()) {
 				if (!myResourceTypes.contains(next)) {
-					return null;
+					return new AuthorizationInterceptor.Verdict(PolicyEnum.DENY,this);
```
fil512, line 69: please add a unit test for this fix

fil512 (changes requested):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImpl.java
oliviayou: added

fil512 (approved):

PR #3377, opened 2022-02-10
Jr 20220210 handle common search params in contained searches

JasonRoberts-smile:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -731,18 +731,35 @@ public Condition createPredicateReference(@Nullable DbColumn theSourceJoinColumn
...
+
+		private String extractPath(String theResourceType, RuntimeSearchParam theSearchParam) {
+			List<String> pathsForType = theSearchParam.getPathsSplit().stream()
+				.map(String::trim)
+				.filter(t -> t.startsWith(theResourceType))
+				.collect(Collectors.toList());
+			if (pathsForType.isEmpty()) {
+				ourLog.warn("Search parameter {} does not have a path for resource type {}.", theSearchParam.getName(), theResourceType);
+				return "";
+			} else if (pathsForType.size() > 1) {
```
michaelabuckley, line 758: I can't decide if this should be a warning, an error, a documented limitation, or something nobody should do.

There are definitely token params that have multiple paths for the same type, but for a reference?  That does seem unlikely.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -731,18 +731,35 @@ public Condition createPredicateReference(@Nullable DbColumn theSourceJoinColumn
...
-		public RuntimeSearchParam getSearchParam() {
-			return mySearchParam;
+		public String getPath() { return myPath; }
+		
+		public String getSearchParameterName() { return mySearchParam.getName(); }
+
+		private String extractPath(String theResourceType, RuntimeSearchParam theSearchParam) {
+			List<String> pathsForType = theSearchParam.getPathsSplit().stream()
+				.map(String::trim)
+				.filter(t -> t.startsWith(theResourceType))
```
michaelabuckley, line 753: I'm not sure this is safe.  The resource type at the front is optional. E.g. `​encounter` instead of `Observation.​encounter`. I think you can assume that a resource name is capitalized.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
JasonRoberts-smile: It does happen, but it's rare. Is it worth holding back the release while I try to come up with a solution, or would it be acceptable for now to just create a follow-up ticket and throw it in the backlog?

michaelabuckley (approved):
The test look good.  Approved

PR #3386, opened 2022-02-12
Mb new token search index

michaelabuckley:
New indexes for the token search table hfj_spidx_token.
We satisfy both merge join and nested loops as both the first and second tables.  Also support sort.


jaison-b (approved):

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4StandardQueriesNoFT.java
```diff
@@ -0,0 +1,252 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.hasItems;
+import static org.hamcrest.Matchers.not;
+
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = {TestR4Config.class, TestHibernateSearchAddInConfig.NoFT.class})
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
+public class FhirResourceDaoR4StandardQueriesNoFT extends BaseJpaTest {
```
jamesagnew, line 54: I don't think Surefire will actually treat this class as a test unless it's named `*Test`

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceIndexedSearchParamToken.java
```diff
@@ -59,16 +59,12 @@
...
-	@Index(name = "IDX_SP_TOKEN_HASH", columnList = "HASH_IDENTITY"),
-	@Index(name = "IDX_SP_TOKEN_HASH_S", columnList = "HASH_SYS"),
-	@Index(name = "IDX_SP_TOKEN_HASH_SV", columnList = "HASH_SYS_AND_VALUE"),
-	// TODO PERF change this to:
-	//	@Index(name = "IDX_SP_TOKEN_HASH_V", columnList = "HASH_VALUE,RES_ID"),
-	@Index(name = "IDX_SP_TOKEN_HASH_V", columnList = "HASH_VALUE"),
-
-	@Index(name = "IDX_SP_TOKEN_UPDATED", columnList = "SP_UPDATED"),
-	@Index(name = "IDX_SP_TOKEN_RESID", columnList = "RES_ID")
+	@Index(name = "IDX_SP_TOKEN_HASH_V2", columnList = "HASH_IDENTITY,SP_SYSTEM,SP_VALUE,RES_ID,PARTITION_ID"),
```
jamesagnew, line 62: I'm pondering this order (not for this specific index, but in general): Does it really make sense for the partition ID to be last?

Like, I can see two approaches here:

1. Partition ID comes last, and we include it in the SQL only if we are querying a server with partitioning enabled
2. Partition ID comes first, and we always include it in the SQL

I'm sure you've done way more thinking about this than I have so I'm not at all asking you to change this, just spitballing... Option 1 has lots of advantages: It doesn't require changing any SQL for non-partitioned servers which is great for people who have already hand tuned their indexes, and it allows for efficient queries across all paritions.

Option 2 would clearly be better performing for partitioned servers as long as you're querying a specific partition, and presumably wouldn't be any worse than the status quo for non-partitioned servers.

Even in typing this up I think I'm landing on option 1 being the right one, but maybe we should consider a future enhancement to have a way of auto creating indexes for option 2 instead, in cases where that is the better option?

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
```diff
@@ -87,70 +87,138 @@ public HapiFhirJpaMigrationTasks(Set<String> theFlags) {
 	private void init600() {
 		Builder version = forVersion(VersionEnum.V6_0_0);
 
-		/*
+		/**
```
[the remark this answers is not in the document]
michaelabuckley: It tells IntelliJ to parse the syntax and turn those into links.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceIndexedSearchParamToken.java
michaelabuckley: This was a big lesson for me from Markus.  Before that course, I would have ordered them in a hierarchical way, like a taxonomy.  So partition_id would be first.  But partition (usually) has terrible selectivity!  The hash is much more selective, since it usually includes partition.

If `INCLUDE` were better supported, I'd write this as `(HASH_IDENTITY,SP_SYSTEM,SP_VALUE) INCLUDE (RES_ID,PARTITION_ID)`.

michaelabuckley: And TBH - I was hoping to do this with as few code changes as possible.  There's a _lot_ of code that knows about these tables, I have very little confidence that I could find _all_ of it.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4StandardQueriesNoFT.java
michaelabuckley: Thanks.  Changed.

PR #3403, opened 2022-02-17
3400 bulk export rules incorrectly applied to group and patient exports

oliviayou:


hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImplTest.java
```diff
@@ -0,0 +1,45 @@
...
+	private Pointcut myPointcut = Pointcut.STORAGE_INITIATE_BULK_EXPORT;
+	@Mock
+	private RequestDetails myRequestDetails;
+	@Mock
+	private IRuleApplier myRuleApplier;
+	@Mock
+	private Set<AuthorizationFlagsEnum> myFlags;
+
+	@Test
+	public void testDenyBulkRequestWithInvalidGroupId() {
```
JasonRoberts-smile, line 31: Please consider adding test coverage for the allowed case as well. As it stands, the test suite would pass if there were an overly restrictive rule that unconditionally denies every request, which is clearly not what is needed.

JasonRoberts-smile (approved):

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImplTest.java
oliviayou: Good idea, a test has been added to cover both cases :)

PR #3413, opened 2022-02-23
Run online index operations non-transactionally on Postgres

michaelabuckley:
Oracle just ignores the transaction, but Postgres actually fails.  So we add a `transactional` boolean property to tasks, and set it for add/drop index when online.


tadgh (approved):

tadgh:
lgtm

PR #3436, opened 2022-02-28
Avoid error when unknown code system behaviour set to warning

jamesagnew:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/validation/JpaValidationSupportChain.java
```diff
@@ -85,6 +84,7 @@ public void postConstruct() {
 		addValidationSupport(myNpmJpaValidationSupport);
 		addValidationSupport(new CommonCodeSystemsTerminologyService(myFhirContext));
 		addValidationSupport(myConceptMappingSvc);
+		addValidationSupport(myUnknownCodeSystemWarningValidationSupport);
```
KevinDougan-SmileCDR, line 87: Just a thought - Should we add a comment here stating that the UnknownCodeSystemWarning class needs to always be last in this List? I'll Approve either way, but I had added a comment myself, in the hopes that we would avoid the same problem again in the future where this got moved to the top of this List.

KevinDougan-SmileCDR (approved):
I added 1 comment but it's not required - just a thought.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/validation/JpaValidationSupportChain.java
jamesagnew: Good call, will do

Issue #3441, opened 2022-03-02
query being ignored during $reindex operation 


tadgh:
**Problem Statement**

Currently, there is a bug when attempting to reindex Patient resources within the database through a filtered basis primarily regarding the _lastUpdated search parameter. When defining the reindex job to use lastUpdated to limit the number of patients that were reindexed, it shows that it has that URL, however, it is reindexing all patients again.

**Steps to Reproduce**

1. Create a Patient resource and POST to http://localhost:8000/Patient
2. POST a reindex request using the _lastUpdated search parameter where the previously submitted Patient resource would not meet the requirement and POST to localhost:8000/$reindex with the following body: 
```json
{
	"resourceType": "Parameters",
	"parameter": [
		{
			"name": "url",
			"valueString": "Patient?_lastUpdated=ge2022-02-25T13:03:00-05:00&_lastUpdated=lt2022-02-25T13:04:00-05:00"
		},
		{
			"name": "batchSize",
			"valueDecimal": 100
		}
	]
}
```

**Observed Results**

It is not filtering the Patient resource based off of the _lastUpdated search parameter and instead it prompts all the patient resources to be reindexed even if they do not meet the _lastUpdated search parameter criteria.

PR #3451, opened 2022-03-04
Issue 3441 reindex query being ignored

markiantorno:
This fixes [3441](https://github.com/hapifhir/hapi-fhir/issues/3441).


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/delete/job/ReindexJobTest.java
```diff
@@ -81,6 +85,71 @@ public void testReindexJob() throws Exception {
 		assertEquals(obsFinalId.getIdPart(), alleleObservationIds.get(0));
 	}
 
+	@Test
+	public void testReindexJobLastUpdatedFilter() throws Exception {
+		// Given
+		DaoMethodOutcome T1_Patient = myReindexTestHelper.createEyeColourPatient(true);
+		Thread.sleep(3000);
```
jamesagnew, line 92: Nitpick - Instead of sleeping for 3 seconds and using that interval to calculate T2_date and T4_date below, I'd recommend using `sleepUntilTimeChanges()` and grabbing the actual timestamps before and after. This reduces the pause time, and also guarantees that you don't get any weird side effects from clock inaccuracies (calling `Thread.sleep(3000)` can actually result in a delay that is higher or lower than that number)

markiantorno: Makes sense. I updated the code. Thanks.

PR #3464, opened 2022-03-10
New indexing for number and quantity search parameters

michaelabuckley:
New indexing for number and quantity SearchParams.

Some new helpers for test data creation.


JasonRoberts-smile (approved):
Looks great. I guess now we'll only ever touch the heap during create and delete operations, and all searches can be index only.

PR #3471, opened 2022-03-16
3470 ne date search

oliviayou:
* add support for the `NOT_EQUAL` prefix on dates
* refactored the `addParam(...)` method to group together operations when `theParsed.getPrefix()` is null, EQUAL, or NOT_EQUALS
* add changelog
* add test

### DIFF
--- a/hapi-fhir-base/src/test/java/ca/uhn/fhir/rest/param/DateRangeParamTest.java
+++ b/hapi-fhir-base/src/test/java/ca/uhn/fhir/rest/param/DateRangeParamTest.java
@@ -26,7 +26,7 @@ public class DateRangeParamTest {
 	 * Can happen e.g. when the query parameter for {@code _lastUpdated} is left empty.
 	 */
 	@Test
-	public void testParamWithoutPrefixAndWithoutValue() {
+	public void testParamWithoutPrefixAndWithoutValue_dateRangeParamRemainsEmpty() {
 		QualifiedParamList qualifiedParamList = new QualifiedParamList(1);
 		qualifiedParamList.add("");
 
@@ -42,7 +42,7 @@ public class DateRangeParamTest {
 	 * Can happen e.g. when the query parameter for {@code _lastUpdated} is given as {@code lt} without any value.
 	 */
 	@Test
-	public void testUpperBoundWithPrefixWithoutValue() {
+	public void testUpperBoundWithPrefixWithoutValue_throwsDateFormatException() {
 		QualifiedParamList qualifiedParamList = new QualifiedParamList(1);
 		qualifiedParamList.add("lt");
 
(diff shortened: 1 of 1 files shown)


jdar8 (commented):
I like the refactor, great work!

hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/param/DateRangeParam.java
```diff
@@ -161,41 +162,42 @@ public DateRangeParam(String theLowerBound, String theUpperBound) {
...
+			theParsed.setPrefix(EQUAL);
+		}
 
-			switch (theParsed.getPrefix()) {
-				case GREATERTHAN:
-				case GREATERTHAN_OR_EQUALS:
-					if (myLowerBound != null) {
-						throw new InvalidRequestException(Msg.code(1923) + "Can not have multiple date range parameters for the same param that specify a lower bound");
-					}
+		switch (theParsed.getPrefix()) {
```
TipzCM, line 169: Can we at least add some unit tests for all your various DateRange options (including error states)?

There's a DateRangeParamTest class. I suggest adding them there.

`setValueAsQueryTokens_conditionUnderTest_expectedResult()`


Might help with understanding the lower down logic for me, too. Cause it seems unusual to me that you check lowerbound is not null. Then set it to something anyways (for instance)

TipzCM (changes requested):

hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/param/DateRangeParam.java
oliviayou: Sure, I added some tests for the `ne` DateRange option inside `DateRangeParamTest`.
The null check for the lower bound was written back in 2015 (it shows up because i refactored this method) so I am not sure if the error would be invoked unless we introduce a bug that sets the lower or upper bound more than once.

hapi-fhir-base/src/test/java/ca/uhn/fhir/rest/param/DateRangeParamTest.java
```diff
@@ -74,4 +75,52 @@ public void testLowerBoundWithPrefixWithoutValue() {
...
+
+		dateRangeParam.setValuesAsQueryTokens(fhirContext, "_lastUpdated", params);
+		assertEquals("1965-11", dateRangeParam.getLowerBound().getValueAsString());
+		assertEquals("1965-11", dateRangeParam.getUpperBound().getValueAsString());
+		assertEquals(ParamPrefixEnum.NOT_EQUAL, dateRangeParam.getLowerBound().getPrefix());
+		assertEquals(ParamPrefixEnum.NOT_EQUAL, dateRangeParam.getUpperBound().getPrefix());
+	}
+
+	@Test
+	public void testSetValueAsQueryTokens_neDay() {
```
TipzCM, line 112: Looking good!

But as a minor point (i want to try and adapt a new standard here), can you rename all your unit tests as follows?

methodUnderTest_testConditions_expectedResult()

eg:
setValuesAsQueryTokens_withQualifiedParamListNEToGivenDate_setsLowerBoundValue

or something like this

TipzCM (approved):
Just minor changes please :)

PR #3478, opened 2022-03-17
Skip database query when hibernate search fully satisfies search

michaelabuckley:
Some queries can be completely satisfied by Hibernate Search.
Before, every query was sent to the database, even when the query reduced to
```
select res_id from hfj_resource where res_id in (:res_id)
```

This change skips that redundant query when possible.
The database is still used for sorting, or when some query clauses can not be satisfied by Hibernate Search.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
@@ -23,7 +23,6 @@ package ca.uhn.fhir.jpa.dao;
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.jpa.api.config.DaoConfig;
-import ca.uhn.fhir.jpa.dao.data.IForcedIdDao;
 import ca.uhn.fhir.jpa.dao.search.ExtendedLuceneClauseBuilder;
 import ca.uhn.fhir.jpa.dao.search.ExtendedLuceneIndexExtractor;
 import ca.uhn.fhir.jpa.dao.search.ExtendedLuceneSearchBuilder;
@@ -36,6 +35,7 @@ import ca.uhn.fhir.jpa.searchparam.SearchParameterMap;
 import ca.uhn.fhir.jpa.searchparam.extractor.ISearchParamExtractor;
 import ca.uhn.fhir.jpa.searchparam.extractor.ResourceIndexedSearchParams;
 import ca.uhn.fhir.model.api.IQueryParameterType;
+import ca.uhn.fhir.model.primitive.IdDt;
 import ca.uhn.fhir.parser.IParser;
 import ca.uhn.fhir.rest.api.Constants;
 import ca.uhn.fhir.rest.api.server.RequestDetails;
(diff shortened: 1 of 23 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
```diff
@@ -271,4 +271,23 @@ private void ensureElastic() {
 		return convertLongsToResourcePersistentIds(pidList);
 	}
 
+	@Override
+	public List<IBaseResource> getResources(Collection<Long> thePids) {
+		SearchSession session = getSearchSession();
+		List<ExtendedLuceneResourceProjection> rawResourceDataList = session.search(ResourceTable.class)
+			.select(
+				f -> f.composite(
+					(pid, forcedId, resource)-> new ExtendedLuceneResourceProjection(pid, forcedId, resource),
```
jaison-b, line 280: This is interesting. So you are basically selecting the fields and projecting it into another to do more work on it?

jaison-b, line 280: This is interesting. So you are basically selecting the fields and projecting it into another to do more work on it?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -865,23 +892,55 @@ public void loadResourcesByPid(Collection<ResourcePersistentId> thePids, Collect
...
+	 * @return can we fetch from Hibernate Search?
+	 */
+	private boolean isLoadingFromElasticSearchSupported(Collection<ResourcePersistentId> theIncludedPids) {
+		// todo mb we can be smarter here.
+		// todo check if theIncludedPids has any with version not null.
+
+		// is storage enabled?
+		return myDaoConfig.isStoreResourceInLuceneIndex() &&
+			// only support lastN for now.
+			myParams.isLastN() &&
```
jaison-b, line 919: we only support loading from ES and skipping DB on Observation lastN operation? I was under the impression that we will support regular resource search as well(as far as we can satisfy the search params in ES).

jaison-b, line 919: we only support loading from ES and skipping DB on Observation lastN operation? I was under the impression that we will support regular resource search as well(as far as we can satisfy the search params in ES).

jaison-b (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
michaelabuckley: When you select a single field, you can provide the simple type (e.g. we were using Long before for myId).  But for multiple fields, that doesn't work.  The default return type would be `List<Object>`.  This is a bit nicer, and let's me pull out some code.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
michaelabuckley: I'd like to, but this I was getting lost in all the difficulties here.  I wanted to fix the 2-second bug first.  We can come back to this later.

PR #3501, opened 2022-03-29
Fixing ValueSet expansion not supporting canonical url.

### LINKED ISSUE #3515: Supporting expansion of ValueSet include/exclude system URI expressed in canonical format during validation
**Describe the bug**
Validation of a code against a ValueSet with an including where the system uri is expressed in canonical format fails.

**To Reproduce**
1. Create a ValueSet including a CodeSystem
2. Create a CodeableConcept.code with uri expressed in a canonical format
3. Validated the code against the ValueSet.
4. Validation will fail.

**Expected behavior**
Validation succeeds.

epeartree:
Fixing issue where `code` validation would fail when the` system` identifying the code is represented in canonical format.

- Adding constructions to detect and parse canonical codeSystem url.

closes #3515

### DIFF
--- a/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupport.java
+++ b/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupport.java
@@ -36,6 +36,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import static org.apache.commons.lang3.StringUtils.contains;
 import static org.apache.commons.lang3.StringUtils.defaultString;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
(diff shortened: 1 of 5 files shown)


hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupportTest.java
```diff
@@ -423,8 +423,45 @@ public void testExpandValueSet_VsUsesVersionedSystem_CsIsCompleteWithCode() {
...
+
+		 String theCodeToValidateCodeSystemUrl = pipedCodeSystemUrl;
+		 String theCodeToValidate = codeMD;
+
+		 IValidationSupport.CodeValidationResult codeValidationResult = mySvc.validateCodeInValueSet(
+			 valCtx,
+			 options,
+			 theCodeToValidateCodeSystemUrl,
+			 theCodeToValidate,
+			 "dontCare",
```
jdar8, line 457:
```suggestion
			 null,
```
i noticed the other tests used `null`, would `null` be better?

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupport.java
```diff
@@ -554,15 +559,24 @@ private boolean expandValueSetR5IncludeOrExclude(ValidationSupportContext theVal
...
 		CodeSystem includeOrExcludeSystemResource = null;
+
 		if (isNotBlank(includeOrExcludeConceptSystemUrl)) {
 
+			if( includeOrExcludeConceptSystemUrl.contains(PIPE_CHARACTER) ){
+				if(isBlank(includeOrExcludeConceptSystemVersion)){
+					includeOrExcludeConceptSystemVersion = substringAfter(includeOrExcludeConceptSystemUrl, PIPE_CHARACTER);
+				}
+				includeOrExcludeConceptSystemUrl = substringBefore(includeOrExcludeConceptSystemUrl, PIPE_CHARACTER);
+			}
```
tadgh, line 578: Suggestion: move this nested arrow code into a helper method called `optionallPopulateVersionFromUrl()`

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupport.java
```diff
@@ -542,10 +544,13 @@ private void expandValueSetR5IncludeOrExcludes(ValidationSupportContext theValid
 	 * Returns <code>true</code> if at least one code was addded
```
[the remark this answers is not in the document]
epeartree: i can change that for sure.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupportTest.java
epeartree: In this case, it would not matter and i can change it for consistency.  But as a general rule, passing a NULL value implies that the implementer of the functionality assumes that the parameter can be NULL and will code accordingly.  In such case, I would expect an @Nullable annotation on the parameter or have it wrapped in an Optional<T>.  That is not the case here so as a rule of thumb, i always provide a filler value since i assume that the inner working of the method will dereference the point without asserting that it is non-null.

Issue #3462, opened 2022-03-10
In hapi 5.7.0 setting setResourceServerIdStrategy(DaoConfig.IdStrategyEnum.UUID) in DaoConfig 


tyfoni-systematic:
**Describe the bug**
If the dao is configured to use resourceServerIdStrategy = DaoConfig.IdStrategyEnum.UUID, then the translateForcedIdToPids in IdHelperService fails because the findAndResolveByForcedIdWithNoType returns an empty collection.

It seems that the lookup may be done before the HFJ_FORCED_ID is populated. 

**To Reproduce**
1. Configure the dao: daoConfig().setResourceServerIdStrategy(DaoConfig.IdStrategyEnum.UUID)
2. Try creating a resource, eg. a CodeSystem
3. See exception:
`ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException: HAPI-2001: Resource CodeSystem/1 is not known
        at ca.uhn.fhir.jpa.dao.index.IdHelperService.resolveResourceIdentity(IdHelperService.java:143)
        at ca.uhn.fhir.jpa.dao.index.IdHelperService.resolveResourcePersistentIds(IdHelperService.java:184)
        at ca.uhn.fhir.jpa.dao.index.IdHelperService.resolveResourcePersistentIds(IdHelperService.java:215)
        at ca.uhn.fhir.jpa.term.TermCodeSystemStorageSvcImpl.getCodeSystemResourcePid(TermCodeSystemStorageSvcImpl.java:607)
`

**Expected behavior**
That the resource is created and has an id with UUID 

**Screenshots**
N/A

**Environment (please complete the following information):**
 - HAPI FHIR Version: 5.7.0
 - OS: [e.g. iOS]: Kubernetes/Linux
 - Browser [e.g. chrome, safari]: N/A

**Additional context**

jkiddo:
FYI - Bug is also triggered when installing IG's with `DaoConfig.IdStrategyEnum.UUID` enabled

rogtrifork:
This seems to be caused due to the following:
---
Starting with **5.7.0**, the flow of creating resources changed which seems to have broken **CodeSystem** objects specifically:
[...]

PR #3505, opened 2022-03-29
Allowing any client assigned ID throws ResourceNotFoundException when creating a CodeSystem

oliviayou:
There was an error creating CodeSystem resources because the the PID lookup uses the forced ID which hasn't yet been populated. The solution was to directly get the PID directly in storeNewCodeSystemVersionIfNeeded.


jkiddo:
I believe this relates to https://github.com/hapifhir/hapi-fhir/issues/3462

jamesagnew (changes requested):
I'm having trouble understanding this fix TBH - There is a new parameter `isPid` added to `resolveResourcePersistentIds(...)` but it doesn't have any javadoc and I'm not clear on what it actually signifies.

You seem to be calling it as `false` for everything except the codesystem lookup, but it's not clear to me why that specific lookup satisfies `isPid`. Perhaps that parameter needs a more descriptive name and a javadoc explaining what it does what when you might set it to true/false?

oliviayou:
I added some javadocs for this new parameter, now named, `theIdIsPid`. I felt this parameter was required because we don't have a way to track whether an id passed into `resolveResourcePersistentIds(...)` is a pid or a forced id when `ClientIdStrategyEnum=ANY` because both can be fully numeric.

I might be wrong, but it looks like the only instance where `resolveResourcePersistentIds(...)` is called and the id is a pid is after the CodeSystem lookup, hence being set to true.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcImpl.java
```diff
@@ -320,7 +320,7 @@ public IIdType storeNewCodeSystemVersion(CodeSystem theCodeSystemResource, TermC
 		// Note that this creates the TermCodeSystem and TermCodeSystemVersion entities if needed
 		IIdType csId = myTerminologyVersionAdapterSvc.createOrUpdateCodeSystem(theCodeSystemResource, theRequest);
 
-		ResourcePersistentId codeSystemResourcePid = myIdHelperService.resolveResourcePersistentIds(RequestPartitionId.allPartitions(), csId.getResourceType(), csId.getIdPart());
+		ResourcePersistentId codeSystemResourcePid = myIdHelperService.resolveResourcePersistentIds(RequestPartitionId.allPartitions(), csId.getResourceType(), csId.getIdPart(), true);
```
jamesagnew, line 323: Ok, now that I see your addition.. I get what you're doing but this fix makes me uneasy. Technically an `IIdType` should never have a PID in it. So I think you're fixing the wrong problem here.

Based on the unit test you added, your server ID strategy is UUID, meaning that `csId` should have that UUID. IMO the right thing to do is to investigate why the UUID that is assigned to the new resource doesn't end up being populated into that `IIdType`

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -208,12 +209,13 @@ public Map<String, ResourcePersistentId> resolveResourcePersistentIds(@Nonnull R
 	 */
 	@Override
 	@Nonnull
-	public ResourcePersistentId resolveResourcePersistentIds(@Nonnull RequestPartitionId theRequestPartitionId, String theResourceType, String theId) {
+	public ResourcePersistentId resolveResourcePersistentIds(@Nonnull RequestPartitionId theRequestPartitionId, String theResourceType, String theId, boolean theIdIsPid) {
```
fil512, line 212: Adding a boolean flag to a method is an anti-pattern.   See https://martinfowler.com/bliki/FlagArgument.html

Instead, create a new method name.

fil512 (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
oliviayou: Thanks! the plan is to take out this parameter and fix the underlying problem with the IIdType having a PID or resource path in it

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcImpl.java
```diff
@@ -277,7 +278,7 @@ public void storeNewCodeSystemVersionIfNeeded(CodeSystem theCodeSystem, Resource
 			if (theCodeSystem.getContent() == CodeSystem.CodeSystemContentMode.COMPLETE || theCodeSystem.getContent() == null || theCodeSystem.getContent() == CodeSystem.CodeSystemContentMode.NOTPRESENT) {
 				ourLog.info("CodeSystem {} has a status of {}, going to store concepts in terminology tables", theResourceEntity.getIdDt().getValue(), theCodeSystem.getContentElement().getValueAsString());
 
-				ResourcePersistentId codeSystemResourcePid = getCodeSystemResourcePid(theCodeSystem.getIdElement());
+				ResourcePersistentId codeSystemResourcePid = new ResourcePersistentId(theCodeSystem.getUserData(RESOURCE_PID_KEY));
```
jamesagnew, line 281: Tiny nitpick:

Recommend replacing:

```java
ResourcePersistentId codeSystemResourcePid = new ResourcePersistentId(theCodeSystem.getUserData(RESOURCE_PID_KEY));
```

with

```java
Long pid = (Long)theCodeSystem.getUserData(RESOURCE_PID_KEY);
assert pid != null;
ResourcePersistentId codeSystemResourcePid = new ResourcePersistentId(pid);
```

Just in case someone ever makes a change that causes the PID to not end up in that map, this assert should avoid any hard-to-figure out bugs, since that would probably cause an NPE much later

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcImpl.java
oliviayou: That makes perfect sense! Replaced 👍🏻

oliviayou: The assert statement helped me realize that it's breaking stuff in r5 . This is because in the hl7 fhir convertor does not convert the userData field (which contains the pid) from the r5 to the r4 model in `convertCodeSystem(org.hl7.fhir.r5.model.CodeSystem src)` 

Our options are 
1. add the userData field to the convertor, or 
2. find another way to get the pid (`theCodeSystem.getIdElement().myUnqualifiedId`)

Thoughts?

jamesagnew: @oliviayou could we copy it over wherever it gets converted? ie. not modify the converter, but modify whatever thing in the term service classes is calling it?

jamesagnew (approved):

PR #3529, opened 2022-04-11
move expunge core classes and interfaces to storage project

JasonRoberts-smile:
This MR teases apart the portions of the expunge business logic that are JPA-specific from those that are DB technology agnostic, and moves the generic pieces into the storage package. This is a preliminary step towards the goal of implementing a version of the expunge process that supports MongoDB.


fil512 (approved):
Looks great.  Nice idea to split this out as a separate MR!

PR #3535, opened 2022-04-13
Make Hibernate Search Path asynchronous to speed up first bundle

michaelabuckley:
Two changes:
- stream the results from Hibernate Search to return results faster
- restrict the number of results returned to avoid saving too many hits in the SavedSearch


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
```diff
@@ -170,9 +188,9 @@ private List<ResourcePersistentId> doSearch(String theResourceType, SearchParame
 					//DROP EARLY HERE IF BOOL IS EMPTY?
 
 				})
-			).fetchAllHits();
+			).scroll(scrollSize);
```
jaison-b, line 191: So we are still sticking to scroll with scrollSize? fetch implementation wouldn't still work here with count passed in?

jaison-b (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
michaelabuckley: It did, but I went back.  Two reasons:
- we'll want the longer scroll when we fix the prefetch
- for the case where we still need to join into the database, we may need more than maxResults from elastic to produce maxResults from the db.

PR #3536, opened 2022-04-13
Add New Index On HFJ_RESOURCE For $reindex Operation

### LINKED ISSUE #3534: Add New Index On HFJ_RESOURCE For $reindex Operation
The `$reindex` Operation takes a very long time to gather the list of Resources to be used in that Operation.

An example request would look like:
```
POST http://myfhirserver.com:8000/$reindex
{
  "resourceType": "Parameters",
  "parameter": [ {
    "name": "url",
    "valueString": "Organization?"
  } ]
}
```

The SQL Query it generates is:
```
SELECT t0.RES_ID 
FROM HFJ_RESOURCE t0 
WHERE (((t0.RES_TYPE = 'Organization') 
  AND (t0.RES_DELETED_AT IS NULL)) 
  AND (t0.RES_UPDATED <= '2021-11-30 10:17:11.119')) 
ORDER BY t0.RES_UPDATED DESC NULLS LAST limit 800;
```

On a sample Postgres database loaded with data, this query takes almost 10 minutes to complete. See the internal Wiki for all the Query Plan details.

When I add the following Index on the HFJ_RESOURCE Table:
```
CREATE INDEX CONCURRENTLY idx_res_type_del_updated ON hfj_resource USING btree (res_type, res_deleted_at, res_updated, res_id);
```
the same query can use the Index to gather all the required data and it completes in only 16 seconds!

**Rationale for adding this Index to HFJ_RESOURCE**

First of all, I understand that changing anything related to this table is potentially very impactful!

The example that lead me to propose adding this new Index is the `$reindex` operation scenario.

The first thing that a `$reindex` operation does is query the HFJ_RESOURCE table to build a list of resources to be re-indexed. In a "real-world scenario", which means a database with many resources such as the sample database that I am working with, this initial query can take more than 10 minutes to complete, and can fail due to database query timeout errors.

Many sites will perform a `$reindex` operation on a specific resource type, rather than the entire database. That implies that several of these _Full Table Scan_ queries will be needed in order to finish a full re-index, which means a lot more long-running queries.

Additionally, this proposed Index is not only good for this one scenario, even though this one is fairly generic and common, but it also satisfies the Index Demand of a different scenario that has been raised as a problem. In that other scenario, the query in question requires 2 of the 3 columns being indexed, so it is satisfied with using this Index in that query to return the `res_id` values. This aligns with the SQL Performance initiative which encourages you to create the minimal amount of indexes to satisfy the entire set of queries in a system.

KevinDougan-SmileCDR:
Add columns to hfj_resource indices to avoid random row reads.  Also change migration tasks to have no timeout since adding indices can be very slow.

Closes #3534.

### DIFF
--- a/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
+++ b/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
@@ -74,7 +74,7 @@ import java.util.stream.Collectors;
 @Entity
 @Table(name = "HFJ_RESOURCE", uniqueConstraints = {}, indexes = {
 	// Do not reuse previously used index name: IDX_INDEXSTATUS, IDX_RES_TYPE
-	@Index(name = "IDX_RES_UPDATED_V2", columnList = "RES_UPDATED,RES_ID,PARTITION_ID"),
+	@Index(name = "IDX_RES_DATE", columnList = "RES_UPDATED"),
 	@Index(name = "IDX_RES_TYPE_DEL_UPDATED", columnList = "RES_TYPE,RES_DELETED_AT,RES_UPDATED,PARTITION_ID,RES_ID"),
 })
 @NamedEntityGraph(name = "Resource.noJoins")
--- a/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
+++ b/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
@@ -74,7 +74,7 @@ import java.util.stream.Collectors;
 @Entity
 @Table(name = "HFJ_RESOURCE", uniqueConstraints = {}, indexes = {
 	// Do not reuse previously used index name: IDX_INDEXSTATUS, IDX_RES_TYPE
-	@Index(name = "IDX_RES_DATE", columnList = "RES_UPDATED"),
+	@Index(name = "IDX_RES_UPDATED_V2", columnList = "RES_UPDATED,RES_ID,PARTITION_ID"),
 	@Index(name = "IDX_RES_TYPE_DEL_UPDATED", columnList = "RES_TYPE,RES_DELETED_AT,RES_UPDATED,PARTITION_ID,RES_ID"),
 })
 @NamedEntityGraph(name = "Resource.noJoins")
--- a/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
+++ b/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
@@ -75,7 +75,7 @@ import java.util.stream.Collectors;
 @Table(name = "HFJ_RESOURCE", uniqueConstraints = {}, indexes = {
 	// Do not reuse previously used index name: IDX_INDEXSTATUS, IDX_RES_TYPE
 	@Index(name = "IDX_RES_DATE", columnList = "RES_UPDATED"),
-	@Index(name = "IDX_RES_TYPE_DEL_UPDATED", columnList = "RES_TYPE,RES_DELETED_AT,RES_UPDATED,RES_ID"),
+	@Index(name = "IDX_RES_TYPE_DEL_UPDATED", columnList = "RES_TYPE,RES_DELETED_AT,RES_UPDATED,PARTITION_ID,RES_ID"),
 })
 @NamedEntityGraph(name = "Resource.noJoins")
 public class ResourceTable extends BaseHasResource implements Serializable, IBasePersistedResource, IResourceLookup {
--- a/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
+++ b/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
@@ -73,9 +73,9 @@ import java.util.stream.Collectors;
 @Indexed(routingBinder= @RoutingBinderRef(type = ResourceTableRoutingBinder.class))
 @Entity
 @Table(name = "HFJ_RESOURCE", uniqueConstraints = {}, indexes = {
-	// Do not reuse previously used index name: IDX_INDEXSTATUS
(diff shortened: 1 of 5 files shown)


KevinDougan-SmileCDR:
Referenced from https://gitlab.com/simpatico.ai/cdr/-/issues/2793

KevinDougan-SmileCDR:
I still have lots of DB-specific manual testing to do before this is ready for merging.

kailiu-smile:
Tested that changes took effect on a fresh install of smilecdr. Also tested that changes took effect on an upgraded version of smilecdr.  
Verified that other functions(post and get) were not affected.

KevinDougan-SmileCDR:
@michaelabuckley 
This has been verified to work as expected on a fresh DB as well as on an upgraded DB, per Kai's comment above.
I will leave it to you to decide when this gets merged to master.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
```diff
@@ -73,9 +73,9 @@
 @Indexed(routingBinder= @RoutingBinderRef(type = ResourceTableRoutingBinder.class))
 @Entity
 @Table(name = "HFJ_RESOURCE", uniqueConstraints = {}, indexes = {
-	// Do not reuse previously used index name: IDX_INDEXSTATUS
-	@Index(name = "IDX_RES_DATE", columnList = "RES_UPDATED"),
-	@Index(name = "IDX_RES_TYPE", columnList = "RES_TYPE"),
+	// Do not reuse previously used index name: IDX_INDEXSTATUS, IDX_RES_TYPE
+	@Index(name = "IDX_RES_UPDATED_V2", columnList = "RES_UPDATED,RES_ID,PARTITION_ID"),
```
fil512, line 77: shouldn't RES_UPDATED be last on this list?

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
```diff
@@ -73,9 +73,9 @@
...
+	@Index(name = "IDX_RES_TYPE_DEL_UPDATED", columnList = "RES_TYPE,RES_DELETED_AT,RES_UPDATED,PARTITION_ID,RES_ID"),
```
fil512, line 78: similarly, should RES_UPDATED be last?

fil512 (approved):

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
michaelabuckley: Good question.  This is for all the `_lastUpdated=le[TIMESTAMP]&_sort=-_lastUpdated&_count=500` that you use for the batch jobs.  If the main clause returns zillions of rows, you don't want to fetch them all to sort by _lastUpdated after.  Instead, the planner uses the index to start fetching res_ids, then checks the rest of the clauses one by one.  There's lots of waste, but less waste.

Otherwise, we should change how the we do the batch jobs.

michaelabuckley: Thanks for poking.  I had them backwards.  We usually order descending, so I reversed them.

This query: `/Observation?code=151700&_lastUpdated=lt2022&_sort=-_lastUpdated`

Produces this sql
```
SELECT t1.RES_ID FROM HFJ_RESOURCE t1 
LEFT OUTER JOIN HFJ_SPIDX_TOKEN t0 ON (t1.RES_ID = t0.RES_ID) 
LEFT OUTER JOIN HFJ_RESOURCE t2 ON (t1.RES_ID = t2.RES_ID) 
WHERE ((t0.HASH_VALUE = -5787047884846015926) 
AND (t2.RES_UPDATED <= '2022-01-01 06:59:59.999')) 
ORDER BY t1.RES_UPDATED DESC NULLS LAST
limit 500
```

Which has this plan - note two index scans which are intersected:
```
Limit  (cost=6126.18..6127.43 rows=500 width=16) (actual time=11.506..11.589 rows=500 loops=1)
  Buffers: shared hit=5694
  ->  Sort  (cost=6126.18..6127.80 rows=645 width=16) (actual time=11.505..11.536 rows=500 loops=1)
        Sort Key: t1.res_updated DESC NULLS LAST
        Sort Method: quicksort  Memory: 58kB
        Buffers: shared hit=5694
        ->  Nested Loop  (cost=1.29..6096.08 rows=645 width=16) (actual time=0.086..11.246 rows=710 loops=1)
              Join Filter: (t0.res_id = t1.res_id)
              Buffers: shared hit=5694
              ->  Nested Loop  (cost=0.86..5711.89 rows=645 width=16) (actual time=0.075..9.629 rows=710 loops=1)
                    Buffers: shared hit=2854
                    ->  Index Only Scan using idx_sp_token_hash_v_v2 on hfj_spidx_token t0  (cost=0.43..32.45 rows=687 width=8) (actual time=0.049..0.210 rows=710 loops=1)
                          Index Cond: (hash_value = '-5787047884846015926'::bigint)
                          Heap Fetches: 0
                          Buffers: shared hit=14
                    ->  Index Scan using hfj_resource_pkey on hfj_resource t2  (cost=0.43..8.27 rows=1 width=8) (actual time=0.013..0.013 rows=1 loops=710)
                          Index Cond: (res_id = t0.res_id)
                          Filter: (res_updated <= '2022-01-01 06:59:59.999'::timestamp without time zone)
                          Buffers: shared hit=2840
              ->  Index Scan using hfj_resource_pkey on hfj_resource t1  (cost=0.43..0.58 rows=1 width=16) (actual time=0.002..0.002 rows=1 loops=710)
                    Index Cond: (res_id = t2.res_id)
                    Buffers: shared hit=2840```

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
michaelabuckley: This is for `_lastUpdated=lt[TIMESTAMP]` as the first table.  res_id is an include column here.

Anyway - rolling this one back.  My testing says it doesn't help much.  HFJ_RESOURCE is already basically sorted by RES_UPDATED, so including columns in the index doesn't buy much.

michaelabuckley (approved):
Thanks for the help!

PR #3539, opened 2022-04-14
2822 mdm log enhancement

alackerbauer:

PR #3543, opened 2022-04-15
3506 mdm log enhancement

alackerbauer:
this pull request is instead of https://github.com/hapifhir/hapi-fhir/pull/3539 to be consistent with ticket numbers and naming.

Matcher fields now contain scores in TRACE level:
```
DEBUG Matcher patient-given is valid for resource type: Patient. Evaluating match.
TRACE No match: Matcher patient-given did not match (score: 0.0).
DEBUG Matcher patient-last is valid for resource type: Patient. Evaluating match.
TRACE Match: Successfully matched matcher patient-last with score 1.0.
```
The summary now contains information on success for each matcher field:
```
DEBUG NO_MATCH Patient/3503: MdmMatchOutcome[vector=1,score=1.4642857015132904,myCreatedNewResource=false,myEidMatch=false,myMatchResultEnum=NO_MATCH]
TRACE Field matcher results: 
birthday: YES
phone: NO
firstname-meta: NO
lastname-meta: NO
firstname-jaro: NO
lastname-jaro: NO
```

### DIFF
--- a/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/dao/MdmLinkDaoSvc.java
+++ b/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/dao/MdmLinkDaoSvc.java
@@ -100,6 +100,7 @@ public class MdmLinkDaoSvc {
 			mdmLink.setPartitionId(new PartitionablePartitionId(partitionId.getFirstPartitionIdOrNull(), partitionId.getPartitionDate()));
 		}
 
+		// FIXME Anna: NULL properties in theMatchOutcome is the issue here
 		String message = String.format("Creating MdmLink from %s to %s -> %s", theGoldenResource.getIdElement().toUnqualifiedVersionless(), theSourceResource.getIdElement().toUnqualifiedVersionless(), theMatchOutcome);
 		theMdmTransactionContext.addTransactionLogMessage(message);
 		ourLog.debug(message);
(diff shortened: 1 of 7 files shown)


hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/dao/MdmLinkDaoSvc.java
```diff
@@ -100,7 +100,7 @@ public MdmLink createOrUpdateLinkEntity(IBaseResource theGoldenResource, IBaseRe
 			mdmLink.setPartitionId(new PartitionablePartitionId(partitionId.getFirstPartitionIdOrNull(), partitionId.getPartitionDate()));
 		}
 
-		String message = String.format("Creating MdmLink from %s to %s -> %s", theGoldenResource.getIdElement().toUnqualifiedVersionless(), theSourceResource.getIdElement().toUnqualifiedVersionless(), theMatchOutcome);
+		String message = String.format("Creating MdmLink from %s to %s.", theGoldenResource.getIdElement().toUnqualifiedVersionless(), theSourceResource.getIdElement().toUnqualifiedVersionless());
 		theMdmTransactionContext.addTransactionLogMessage(message);
 		ourLog.debug(message);
```
jamesagnew, line 105: Nitpick (non blocking): SLF4j already has a variable substitution mechanism, and using it is more performant then String.format, especially when debug level logging is disabled (since in that case the string isn't substitututed at all).

In other words, better to just call:

```
ourLog.debug("Creating MdmLink from {} to {}.", theGoldenResource.getIdElement().toUnqualifiedVersionless(), theSourceResource.getIdElement().toUnqualifiedVersionless());
```

jamesagnew (approved):

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/dao/MdmLinkDaoSvc.java
alackerbauer: Cool, I didn't know about that performance difference! But in this very specific case, I think I would leave it as it is, given the message is being used twice (line 104 and 105).

alackerbauer: Cool, I didn't know about that performance difference! But in this very specific case, I think I would leave it as it is, given the message is being used twice (line 104 and 105).

PR #3549, opened 2022-04-18
Supported search

### LINKED ISSUE #3548: _has parameter query failures on `Resource` search parameters.
While checking for active search params on a resource, SPs on `Resource` should also be checked during `_has` query parameter resolution.

tadgh:
Closes #3548 

* When resolving a _has parameter, use more than just the SPs from the resourcetype itself. Also use the `Resource` search params.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
@@ -546,11 +546,11 @@ public class QueryStack {
 				//Ensure that the name of the search param
 				// (e.g. the `code` in Patient?_has:Observation:subject:code=sys|val)
 				// exists on the target resource type.
-				RuntimeSearchParam owningParameterDef = getRuntimeSearchParam(targetResourceType, paramName, 1209);
+				RuntimeSearchParam owningParameterDef = mySearchParamRegistry.getRuntimeSearchParam(targetResourceType, paramName);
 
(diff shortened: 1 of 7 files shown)


michaelabuckley (approved):
Super tidy

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -595,6 +590,25 @@ private Condition createPredicateHas(@Nullable DbColumn theSourceJoinColumn, Str
...
+	 * Find a search param for a resource. First, check the resource itself, then check the top-level `Resource` resource.
+	 * @param theResourceType
+	 * @param theParamName
+	 * @param theFailureCodeIfAbsent
+	 *
+	 * @return the {@link RuntimeSearchParam} that is found.
+	 */
+	private RuntimeSearchParam getRuntimeSearchParam(String theResourceType, String theParamName, int theFailureCodeIfAbsent) {
+		RuntimeSearchParam availableSearchParamDef = mySearchParamRegistry.getActiveSearchParam(theResourceType, theParamName);
+		if (availableSearchParamDef == null) {
```
jamesagnew, line 603: Would it be possible to push this logic up one level? Might as well push it up into `SearchParamRegistryImpl` so that all callers of this method would get the same benefit.

One other point - This PR should probably have a test

tadgh: Method has been pushed down, test is on the CDR side as it was unreplicable here.

PR #3554, opened 2022-04-19
Add implementation and changelog, and test.

### LINKED ISSUE #3553: Feature Request: Prevent deexternalization of binaries
Currently, if you are requesting a Binary/DocumentReference with an attachment, the BinaryStorageINterceptor will automatically deexternalize up to 10MB of binary data into the response. This can cause super slow responses.

tadgh:
Closes #3553 

* Add 2 new DaoConfig options: 
  * `setAllowAutoInflateBinaries`
  * `setAutoInflateBinariesMaximumBytes`
* In BinaryStorageInterceptor, if we detect that it is disabled, or the binary size is set to 0, just skip iteration completely. This should improve processing speed for servers who never need to automatically de-externalize binaries.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
@@ -263,8 +263,8 @@ public class JpaConfig {
 	@Lazy
 	public BinaryStorageInterceptor binaryStorageInterceptor(DaoConfig theDaoConfig) {
 		BinaryStorageInterceptor interceptor = new BinaryStorageInterceptor();
-		interceptor.setAllowAutoDeExternalizingBinaries(theDaoConfig.isAllowAutoDeExternalizingBinaries());
-		interceptor.setAutoDeExternalizeMaximumBytes(theDaoConfig.getAutoDeExternalizeMaximumBytes());
+		interceptor.setAllowAutoInflateBinaries(theDaoConfig.isAllowAutoInflateBinaries());
+		interceptor.setAutoInflateBinariesMaximumSize(theDaoConfig.getAutoInflateBinariesMaximumBytes());
 		return interceptor;
 	}
 
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
@@ -261,8 +261,11 @@ public class JpaConfig {
 
 	@Bean(name = "myBinaryStorageInterceptor")
 	@Lazy
-	public BinaryStorageInterceptor binaryStorageInterceptor() {
-		return new BinaryStorageInterceptor();
+	public BinaryStorageInterceptor binaryStorageInterceptor(DaoConfig theDaoConfig) {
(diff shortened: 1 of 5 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
```diff
@@ -261,8 +261,11 @@ public BinaryAccessProvider binaryAccessProvider() {
 
 	@Bean(name = "myBinaryStorageInterceptor")
 	@Lazy
-	public BinaryStorageInterceptor binaryStorageInterceptor() {
-		return new BinaryStorageInterceptor();
+	public BinaryStorageInterceptor binaryStorageInterceptor(DaoConfig theDaoConfig) {
+		BinaryStorageInterceptor interceptor = new BinaryStorageInterceptor();
+		interceptor.setAllowAutoDeExternalizingBinaries(theDaoConfig.isAllowAutoDeExternalizingBinaries());
```
fil512, line 266: suggestion: instead call these:
setAllowAutoInflateOutboundBinaries
 setAutoInflateOutboundBinaryMaximumBytes

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/provider/r4/BinaryStorageInterceptorR4Test.java
```diff
@@ -109,6 +113,29 @@ public void testCreateAndRetrieveBinary_ServerAssignedId_ExternalizedBinary_NoSe
...
+		String encoded = myFhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(outcome.getResource());
+		ourLog.info("Encoded: {}", encoded);
+		assertThat(encoded, containsString(HapiExtensions.EXT_EXTERNALIZED_BINARY_ID));
+		assertThat(encoded, not(containsString("\"data\"")));
+
+		// Now read it back and make sure it was not  de-externalized
+		Binary output = myBinaryDao.read(id, mySrd);
+		assertEquals("application/octet-stream", output.getContentType());
+		assertArrayEquals(null, output.getData());
+		assertThat(output.getDataElement().getExtensionByUrl(HapiExtensions.EXT_EXTERNALIZED_BINARY_ID), is(notNullValue()));
```
fil512, line 137: question: Shouldn't we use `attachment.url` for this and not an extension?

fil512 (approved):

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/provider/r4/BinaryStorageInterceptorR4Test.java
tadgh: So, the answer is "maybe?". It's worth a discussion to have the interceptor inject the url into this field, but I don't want to roll that into this PR. Currently, it replaces whatever data element it finds with an extension on the data element. Note that DocuRef.attachment is not the _only_ source of binaries, this logic also applies to actual Binary resources ( which don't have such a URL field)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
tadgh: Renamed

PR #3556, opened 2022-04-20
Freetext search query builder consolidation

jmarchionatto:


michaelabuckley (approved):
These tests cover things

PR #3559, opened 2022-04-21
3547 Interceptor hook for client-assigned IDs

### LINKED ISSUE #3547: Interceptor hook for client-assigned IDs
**Describe the issue**
We have client-assigned IDs disabled on our server because we want to prevent customers from creating resources that conflict with each other. However, there have been a variety of occasions where it would be desirable to set a client-assigned ID.
Right now it is possible to allow client-assigned IDs on some resource and not others by using multiple persistence modules. However, we would like more flexibility so that we can decide whether not a certain user is authorized to create client-assigned IDs.

**Solution design**
In this case we think the most flexible option would be to have an interceptor that decides whether creation can proceed when the client provides resource ID. We want to be able to decide this based on the user sending the request.
- Add new void Pointcut just before client-assigned id is created so can allow/deny based on `ResourceDetails` and `RequestDetails`

jdar8:
closes #3547 

- Added new pointcut that takes in `ResourceDetails` and `RequestDetails`
- Added test
- Added changelog

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
@@ -1346,6 +1346,14 @@ public enum Pointcut implements IPointcut {
 	),
 
 	/**
+	 *
+	 */
+	STORAGE_ACCEPT_CLIENT_ASSIGNED_ID(void.class,
+		"org.hl7.fhir.instance.model.api.IBaseResource",
+		"ca.uhn.fhir.rest.api.server.RequestDetails"
+	),
+
+	/**
 	 * <b>Storage Hook:</b>
 	 * Invoked before a resource will be updated, immediately before the resource
 	 * is persisted to the database.
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
@@ -1368,7 +1368,7 @@ public enum Pointcut implements IPointcut {
 	 * Hooks should return <code>void</code>.
 	 * </p>
 	 */
-	STORAGE_ACCEPT_CLIENT_ASSIGNED_ID(void.class,
+	STORAGE_PRESTORAGE_ACCEPT_CLIENT_ASSIGNED_ID(void.class,
 		"org.hl7.fhir.instance.model.api.IBaseResource",
 		"ca.uhn.fhir.rest.api.server.RequestDetails"
 	),
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
@@ -1346,7 +1346,27 @@ public enum Pointcut implements IPointcut {
 	),
 
 	/**
-	 *
+	 * <b>Storage Hook:</b>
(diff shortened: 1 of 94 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
```diff
@@ -1345,6 +1345,34 @@ public enum Pointcut implements IPointcut {
...
+	 * pulled out of the servlet request. Note that the bean
+	 * properties are not all guaranteed to be populated, depending on how early during processing the
+	 * exception occurred.
+	 * </li>
+	 * </ul>
+	 * <p>
+	 * Hooks should return <code>void</code>.
+	 * </p>
+	 */
+	STORAGE_PRESTORAGE_ACCEPT_CLIENT_ASSIGNED_ID(void.class,
```
tadgh, line 1371: This name should probably be `STORAGE_PRESTORAGE_CLIENT_ASSIGNED_ID`. The implementers may or may not actually perform any acceptance checking.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_0_0/3547-interceptor-hook-for-client-assigned-ids.yaml
```diff
@@ -0,0 +1,5 @@
+type: add
+issue: 3547
+jira: SMILE-4141
+title: "Added a pointcut just before the client assigned id is created. This allows interceptors to allow or deny the
```
tadgh, line 4:
```suggestion
title: "Added a new pointcut: `STORAGE_PRESTORAGE_CLIENT_ASSIGNED_ID`  which is invoked when a user attempts to create a resource with a client-assigned ID. 
```

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4InterceptorTest.java
```diff
@@ -511,13 +514,33 @@ public void resourcePreCreate(IBaseResource theResource) {
...
+		mySrdInterceptorService.registerInterceptor(interceptor);
+
+		Patient p = new Patient();
+		try {
+			myPatientDao.create(p, mySrd).getId().toUnqualifiedVersionless();
+			fail();
+		} catch (ForbiddenOperationException e) {
+			assertEquals("Client assigned id rejected.", e.getMessage());
+		}
+	}
```
tadgh, line 542: Issue: This seems wrong. Calling `.create()` on the dao will _always_ use a server assigned ID. I would expect this call to _not_ invoke your interceptor. I would expect your interceptor to be invoked on `.update()` calls to the dao _only_.

tadgh: When you call .create(), does it actually _use_ your client assigned ID?

tadgh (changes requested):

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4InterceptorTest.java
jdar8: yea it is wrong. create sets it to an empty string

fixed

tadgh (approved):

PR #3562, opened 2022-04-23
3558 bulk export premature deletion

### LINKED ISSUE #3561: BulkExport files are deleted prematurely.
On Bulk Export job instantiation, we set an expiry on the BulkExportJobEntity (see EXP_TIME in BulkExportJobEntity.java) which is equal to 2 hours from instantiation. This is unsuitable, as many jobs take over two hours to complete, causing the cleanup task to clean up jobs _that are currently running_. See BulkDataExportJobSchedulingHelperImpl#purgeExpiredFiles L203 for this. Thi s is a hard coded limit which is unsuitable for real production workloads. 

Recommendation: 
1. Making this setting configurable via a DaoConfig setting. 
2. Increasing the default to 24 hours instead of 2. Setting the value to <=0 sets bulk files to never expire.

tadgh:
Closes #3561 

* Make retention period configurable, and also use hours instead of millis. 
* Add DaoConfig property to expose this option. 
* Parameterize test for multiple expiry scenarios 
* Allow never expiring files by usage of negative or 0 value in new setting.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IBulkExportJobDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IBulkExportJobDao.java
@@ -43,6 +43,9 @@ public interface IBulkExportJobDao extends JpaRepository<BulkExportJobEntity, Lo
 	@Query("SELECT j FROM BulkExportJobEntity j WHERE j.myExpiry < :cutoff")
 	Slice<BulkExportJobEntity> findByExpiry(Pageable thePage, @Param("cutoff") Date theCutoff);
 
+	@Query("SELECT j FROM BulkExportJobEntity j WHERE j.myExpiry < :cutoff AND j.myStatus <> 'BUILDING'")
+	Slice<BulkExportJobEntity> findNotRunningByExpiry(Pageable thePage, @Param("cutoff") Date theCutoff);
+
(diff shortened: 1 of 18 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IBulkExportJobDao.java
```diff
@@ -43,6 +43,9 @@ public interface IBulkExportJobDao extends JpaRepository<BulkExportJobEntity, Lo
 	@Query("SELECT j FROM BulkExportJobEntity j WHERE j.myExpiry < :cutoff")
 	Slice<BulkExportJobEntity> findByExpiry(Pageable thePage, @Param("cutoff") Date theCutoff);
 
+	@Query("SELECT j FROM BulkExportJobEntity j WHERE j.myExpiry < :cutoff AND j.myStatus <> 'BUILDING'")
```
fil512, line 46: Will this match null in some databases?
Also, is BUILDING the only state that should block expiry?
Wondering if maybe there should be further filtering in Java

fil512 (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IBulkExportJobDao.java
tadgh: like will null match for expiry? Man...I hope not? I suppose I can guard with j.myExpiry IS NOT NULL. Updating.

PR #3569, opened 2022-04-28
Add res_id, and partition_id to string norm and exact indexes

michaelabuckley:
The current string search indexes require row reads for each match, and these will be random blocks.
We add res_id, and partition_id so the queries can be satisfied directly by an index scan.


JasonRoberts-smile (approved):
Looks good to me.

PR #3585, opened 2022-05-04
add graphql test

fil512:


jmarchionatto (approved):
Soooo much readable using text blocks. Thanks!

PR #3593, opened 2022-05-09
Bump fhir.core to latest version

patrick-werner:
- Updated fhir.core version to 5.6.43
- hapi now populates OO.issue.details with the messageID from the RI java validator
- fixed failing tests and test resources (introduced by newer fhir.core version being stricter on some data type setters)
- removed duplicated commons-codec dependency from [hapi-fhir-structures-dstu3/pom.xml](https://github.com/hapifhir/hapi-fhir/pull/3593/files#diff-95c71976d9099af7b8855f398d28c9ea2bb53b1b42da9692eade9e78b0f65be7)
- modified message filtering in [ValidatorWrapper](https://github.com/hapifhir/hapi-fhir/blob/60034008c36b1a7fd20bab6091ffed283e7008e5/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/ValidatorWrapper.java), replaced Stringmatchers with messageId based Issue filtering


patrick-werner:
**This PR is ready to be merged.**

I had to remove the 1.1 commons-codec dependency which was added to also check the android build.
fhir.core now also uses Base64.isBase64() (introduced in commons 1.5) from common-codecs - which breaks the android build during runtime.

fyi @tadgh

jamesagnew (approved):
Looks good! Merging now.

patrick-werner:
> FYI @patrick-werner this failed with a duplicate message code, but I resolved it and pushed a fix to your branch.

thanks @jamesagnew 
I'm back from vacation, merged latest master & fixing duplicated Msg.codes now

jamesagnew:
FYI I suspect this is going to have to be bumped back to post 6.1.0 - We're currently prevented from upgrading core because the R5 structures are missing the search parameter annotations. Grahame is investigating that, but I think it's unlikely that will happen before 6.1.0 hits code freeze.

I have a separate branch also going that fixes the unexpected strictness with date/time parsing. I really want to avoid merging in the "watering down" of the unit tests, since they represent cases that may be invalid but also represent things we know people do in the real world.

patrick-werner:
> FYI I suspect this is going to have to be bumped back to post 6.1.0 - We're currently prevented from upgrading core because the R5 structures are missing the search parameter annotations. Grahame is investigating that, but I think it's unlikely that will happen before 6.1.0 hits code freeze.
> 
> I have a separate branch also going that fixes the unexpected strictness with date/time parsing. I really want to avoid merging in the "watering down" of the unit tests, since they represent cases that may be invalid but also represent things we know people do in the real world.

@jamesagnew that's understandable.
I had to adapt another test because of the parser strictness. Build should work now.

I just disabled the auto-merge you enabled to not merge the new core into master.
Is there anything i can help to resolve the parser issues in core @grahamegrieve ?

jamesagnew:
A fix to the parser issues in core just merged as well yesterday. I think this should all be good to go as soon as the next core build is pushed to central.

@dotasek can you comment here when the next core release is pushed to central? This can merge once that's happened.

patrick-werner:
> A [fix to the parser issues in core](https://github.com/hapifhir/org.hl7.fhir.core/pull/874) just merged as well yesterday. I think this should all be good to go as soon as the next core build is pushed to central.
> 
> @dotasek can you comment here when the next core release is pushed to central? This can merge once that's happened.
> 
> This may make the 6.1.0 release after all!

Great News! 
I'm currently reverting the changed Tests/ adapting latest R5 changes into this PR. Will rename this PR.

patrick-werner:
@jamesagnew i tried the latest core snapshot as dependency. Now some tests which expected parsing errors are failing:
`dt.setValueAsString("2013-02-03T11:22Z");` was expected to fail: `"Invalid date/time string (datatype DateTimeType does not support MINUTE precision): 2013-02-03T11:22Z"`

Do you want to remove/disable these tests? Or was the recent core change too lenient in this use-case?

patrick-werner:
upgraded to latest R5 structures, identified a bug in fhir.core (can't handle "5.0.0-snapshot2"), created a fixing PR: https://github.com/hapifhir/org.hl7.fhir.core/pull/885

PR #3603, opened 2022-05-10
Handle batch2 job cancellation

jmarchionatto:


hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/impl/JobMaintenanceServiceImpl.java
```diff
@@ -136,6 +138,21 @@ public void runMaintenancePass() {
...
+	private void handleCancellation(JobInstance theInstance) {
+		if (! theInstance.isCancelled()) { return; }
+
+		if (theInstance.getStatus() == StatusEnum.QUEUED || theInstance.getStatus() == StatusEnum.IN_PROGRESS) {
+			String msg = "Job instance cancelled";
+			if (theInstance.getCurrentGatedStepId() != null) {
+				msg += " while running step " + theInstance.getCurrentGatedStepId();
+			}
+			theInstance.setErrorMessage(msg);
+			theInstance.setStatus(StatusEnum.ERRORED);
```
fil512, line 150: This should be CANCELLED not ERRORED.

jmarchionatto: I would prefer that, but there is no CANCELLED status and I understood James wanted not to be one
Let me check with him, as he joined yesterday's thread about this
Or you want me to just add it?

jamesagnew: We could definitely add that status to the enum. The reason I made it a boolean flag as opposed to an entry in the status enum was just to avoid any race conditions (e.g. person clicks cancel at the same time as the maintenance service is also updating the status).

Your fix would avoid that issue since it's the maintanance loop that's setting it to CANCELLED.

StevenXLi (approved):

PR #3616, opened 2022-05-15
Batch 2 documentation and SRP cleanup

fil512:


hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/impl/JobStepExecutor.java
```diff
@@ -0,0 +1,114 @@
...
+			outcome = worker.run(stepExecutionDetails, theDataSink);
+			Validate.notNull(outcome, "Step theWorker returned null: %s", worker.getClass());
+		} catch (JobExecutionFailedException e) {
+			ourLog.error("Unrecoverable failure executing job {} step {}", theJobDefinitionId, targetStepId, e);
+			myJobPersistence.markWorkChunkAsFailed(chunkId, e.toString());
+			return false;
+		} catch (Exception e) {
+			ourLog.error("Failure executing job {} step {}", theJobDefinitionId, targetStepId, e);
+			myJobPersistence.markWorkChunkAsErroredAndIncrementErrorCount(chunkId, e.toString());
+			throw new JobExecutionFailedException(Msg.code(2041) + e.getMessage(), e);
```
JasonRoberts-smile, line 97: question (non-blocking): I don't often see a catch block throw an exception of a type that is caught by a different branch of the same try..catch structure. It seems that the responsibility for handling `JobExecutionFailedException`s is diffused across several layers of the call hierarchy. Is that intentional?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/JobInstance.java
```diff
@@ -113,6 +113,17 @@ public JobInstance(JobInstance theJobInstance) {
 		setCurrentGatedStepId(theJobInstance.getCurrentGatedStepId());
 	}
 
+	public static JobInstance fromJobDefinition(JobDefinition<?> theJobDefinition) {
+		JobInstance instance = new JobInstance();
+
+		instance.setJobDefinitionId(theJobDefinition.getJobDefinitionId());
+		instance.setJobDefinitionVersion(theJobDefinition.getJobDefinitionVersion());
+		if (theJobDefinition.isGatedExecution()) {
+			instance.setCurrentGatedStepId(theJobDefinition.getSteps().get(0).getStepId());
```
JasonRoberts-smile, line 122: suggestion (non-blocking): I think I've seen this `definition.getSteps().get(0)` pattern in a few places now. Might `definition.getFirstStep()` be more readable?

JasonRoberts-smile (approved):
nice decomposition of that monster class. I found the documentation very helpful for understanding what I was seeing in the code

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/impl/JobStepExecutor.java
fil512: Good question.  This is a feature of having a callback api.  In the framework, step workers (that are written by "other" people) can indicate to the batch framework that an unrecoverable error occurred by throwing a `JobExecutionFailedException` which is the polite way to fail unrecoverably.  However, a step may fail in a way the author of the step writer didn't anticipate and in those cases we rethrow their failure as a JobExecutionFailedException so the channel pub-sub mechanism tries to run that worker again later.  I guess we could have just rethrown e again here, but the advantage of wrapping the exception in a known exception type is if we chose to we could add special handling for it in the channel retry framework later.  I agree it is confusing to overload `JobExecutionFailedException` for this though.  I will speak with @jamesagnew about it this morning.

fil512: Excellent catch.  After consulting with @jamesagnew changed the throw to JobStepFailedException

PR #3618, opened 2022-05-16
Update files.properties

jkiddo:
Updated doc of years


tadgh:
Heya! I appreciate the patch, I just checked out our release branch, and looks like it has already been resolved in the rel_6_0 branch: https://github.com/hapifhir/hapi-fhir/blob/rel_6_0/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/files.properties#L4 That should merge back into master on thursday. Thanks for noticing this though, it was also bothering me that the dates were wrong.

PR #3628, opened 2022-05-20
improve nickname matcher

fil512:
traverse the nickname tree for stronger nickname matching

### DIFF
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/nickname/NicknameSvc.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/nickname/NicknameSvc.java
@@ -28,8 +28,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.Reader;
-import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 public class NicknameSvc {
 	private final NicknameMap myNicknameMap = new NicknameMap();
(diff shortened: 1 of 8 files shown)


hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/nickname/NicknameSvc.java
```diff
@@ -47,27 +48,26 @@ public int size() {
...
-			retval.addAll(expansions);
-		} else {
-			expansions = getFormalNamesFromNicknameOrNull(theName);
-			if (expansions != null) {
-				retval.addAll(expansions);
+		if (retval.isEmpty()) {
+			List<String> formalNames = getFormalNamesFromNickname(theName);
+			retval.addAll(formalNames);
+			for (String formalName : formalNames) {
+				retval.addAll(getNicknamesFromFormalName(formalName));
```
JasonRoberts-smile, line 57: Question: All these formal names have at least one nickname in common, so it seems probable that there may be additional overlap. Do we care about uniqueness in the resulting list?

suggestion: add all the formal names and nicknames to a Set, and add the contents of the Set to retval after the loop.

JasonRoberts-smile (approved):
Approved with one suggestion, which is blocking only in the sense that there seems to be an associated test failure

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/nickname/NicknameSvc.java
fil512: Man you're so good.  Very good point here. A different test caught this same issue.

PR #3643, opened 2022-05-24
Group Bulk Export rules exiting too early. 

### LINKED ISSUE #3642: Multiple Group Bulk Export rules conflict
The current implementation of the RuleBulkExportImpl returns a DENY if the group ID does not match, but it is possible for _another_ rule to match. E.g. if we were to write the following rules: 

```
				return new RuleBuilder()
					.allow().bulkExport().groupExportOnGroup(new IdType("Group/123")).andThen()
					.allow().bulkExport().groupExportOnGroup(new IdType("Group/789")).andThen()
					.build();
```

This would cause a failure for somebody attempting to export Group/789, as first the Group/123 rule is hit, and it returns a deny. If instead it returns null, it can delegate to further rules.

tadgh:
Closes #3642 

* Make RuleBulkExportImpl return null instead of DENY if group ID does not match. Default implementation will DENY if no rules match. 
* This will allow a failed bulk export rule to delegate to further ones
* Fix tests, add test component.
* Add changelog.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3642-group-bulk-export-failure.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3642-group-bulk-export-failure.yaml
@@ -1,0 +1,5 @@
+---
+type: fix
+issue: 3642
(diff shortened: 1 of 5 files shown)


fil512 (approved):
nice.  you made quick work of this!!

PR #3646, opened 2022-05-26
hapi-fhir-checkstyle now generates javadocs

markiantorno:
Adding a dummy javadoc package so we can publish checkstyle without sonatype complaining.


markiantorno:
doing as backport

PR #3647, opened 2022-05-26
pom changes to publish checkstyle to sonatype

markiantorno:
Sonatype would not allow for checkstyle to upload as there are no javadocs or published source packages. This change fixes that.


hapi-fhir-checkstyle/pom.xml
```diff
@@ -8,6 +8,7 @@
 		<version>6.0.1</version>
 		<relativePath>../pom.xml</relativePath>
 	</parent>
+	<version>0.0.1-TEST</version>
```
fil512, line 11: you sure this is what you want on this MR?

fil512 (approved):

PR #3652, opened 2022-05-27
Avoid unsupported features during migrations on MSSQL

### LINKED ISSUE #3651: Migration failure on MSSQL Standard Edition
#### Problem Statement

Migration to 2022.05 fails on MSSQL Standard Edition

#### Steps to Reproduce

- Install 2022.02 on MSSQL Standard Edition.
- Upgrade to 2022.05
 
#### Observed Results

Migration error with message "Online index operations can only be performed in Enterprise edition of SQL Server."

#### Expected Results

No errors.

Hapi 6.0 upgrade.
MSSQL Server 2017 Standard

Sql Server Standard doesn't support ONLINE=ON.

michaelabuckley:
Closes #3651 
Check for Standard Edition, and don't use ONLINE

### DIFF
--- a/hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/MetadataSource.java
+++ b/hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/MetadataSource.java
@@ -1,5 +1,6 @@
 package ca.uhn.fhir.jpa.migrate.taskdef;
 
+import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.jpa.migrate.DriverTypeEnum;
 
 /**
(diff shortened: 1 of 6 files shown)


hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/MetadataSource.java
```diff
@@ -0,0 +1,30 @@
...
+				return false;
+		}
+	}
+
+	private String getEdition(DriverTypeEnum.ConnectionProperties theConnectionProperties) {
+		final String result;
+		if (theConnectionProperties.getDriverType() == DriverTypeEnum.MSSQL_2012) {
+			result = theConnectionProperties.newJdbcTemplate().queryForObject("SELECT SERVERPROPERTY ('edition')", String.class);
+		} else {
+			 result = "unimplemented";
```
JasonRoberts-smile, line 26: question: is this "unimplmented" because the concept of edition doesn't apply to other databases, or just because you didn't need the others for this story? If the latter, might it be safer to throw an exception instead, to force future developers using this method for other purposes to extend the implementation to support the databases they care about?

JasonRoberts-smile (approved):
Approved with a question

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/MetadataSource.java
michaelabuckley: Thanks.  That's better.  It also prompted me to write some javadoc.

PR #3663, opened 2022-05-31
Rel 5 7 mergeback

tadgh:
* Bring in the Postgres LOB reindex fix. 
* New version.


tadgh:
Closing because we elected to back this change out.

PR #3665, opened 2022-06-01
replace search by example.

fil512:
it caused a cryptic hibernate session error


JasonRoberts-smile (approved):
I still don't understand why this fixed it, but it clearly did

PR #3670, opened 2022-06-03
added fix for subscription in default partition when default partitio…

longma1:
Fix issue when partition is in default partition when the partition id is null being matched against partitions from all partitions


tadgh (approved):
LGTM. You need a changelog though

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/matcher/subscriber/SubscriptionMatchingSubscriber.java
```diff
@@ -127,7 +127,7 @@ private void doMatchActiveSubscriptionsAndDeliver(ResourceModifiedMessage theMsg
 		for (ActiveSubscription nextActiveSubscription : subscriptions) {
 			// skip if the partitions don't match
 			CanonicalSubscription subscription = nextActiveSubscription.getSubscription();
-			if (subscription != null && subscription.getRequestPartitionId() != null && theMsg.getPartitionId() != null &&
+			if (subscription != null && theMsg.getPartitionId() != null &&
```
jaison-b, line 130: Suggestion: for a long list of conditions to check pull them into a private method like 'checkPartitionSupport' or something and put these assertions there. It will read better.

jaison-b (approved):

PR #3685, opened 2022-06-09
3684 subscription for delivering delete events formatted wrong

StevenXLi:
- Fixed issue with placement of extension Delivering Delete Events
- Fixed all the tests

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/i18n/Msg.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/i18n/Msg.java
@@ -25,7 +25,7 @@ public final class Msg {
 
 	/**
 	 * IMPORTANT: Please update the following comment after you add a new code
-	 * Last code value: 2084
+	 * Last code value: 2086
 	 */
 
 	private Msg() {}
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3684-subscription-for-delivering-delete-events-formatted-wrong.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3684-subscription-for-delivering-delete-events-formatted-wrong.yaml
@@ -1,4 +1,6 @@
 ---
 type: fix
+issue: 3684
+jira: SMILE-4398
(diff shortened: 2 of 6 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/i18n/Msg.java
```diff
@@ -25,7 +25,7 @@ public final class Msg {
 
 	/**
 	 * IMPORTANT: Please update the following comment after you add a new code
-	 * Last code value: 2084
+	 * Last code value: 2086
```
jdar8, line 28: Question: why was this value updated? I don't think a new code was added.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3684-subscription-for-delivering-delete-events-formatted-wrong.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+title: "Previously, the working format for the Delivering Delete Events had the extension and extension value outside the
+the channel, that has been fixed to now be inside of the channel."
```
jdar8, line 4:
```suggestion
---
type: fix
issue: 3684
jira: SMILE-4398
title: "Previously, the working format for the Delivering Delete Events had the extension and extension value outside the
the channel, that has been fixed to now be inside of the channel."
```

jdar8 (commented):
Looks good to me! Suggested a few changes to changelog

tadgh:
I'm confused why this had to be done at all. What is the problem with the extension being on the subscription instead of the channel?

tadgh:
Approved pending changes to the changelog for readability.

StevenXLi:
> I'm confused why this had to be done at all. What is the problem with the extension being on the subscription instead of the channel?

@tadgh Since the subscription is converted to `CanonicalSubscription` on cdr side, the extension would be loss on conversion, but it will be preserved if it's in the channel. 

It's just to lessen the amount of machinery required to get the ticket working on the cdr side while making this feature consistent to what the doc has.

tadgh (approved):

Issue #3702, opened 2022-06-14
MPI_LINK Query Performance Improvement Needed


kateryna-mironova:
While importing Patient data with MDM enabled, we noticed that queries on the MPI_LINK table were performing poorly and being raised in the AWS Console as a "Top 2 Most Active Query".

We added the following Index and it improved our performance:

```
CREATE INDEX CONCURRENTLY idx_empi_match_tgt_ver ON mpi_link USING btree (match_result, target_pid, version);
```

tadgh:
Super interesting. By any chance can you share the size of that table in rows?  I would be happy to add this index if it provides a nice improvement.

PR #3708, opened 2022-06-15
3702 mpi link add index

kateryna-mironova:
Closes https://github.com/hapifhir/hapi-fhir/issues/3702

Added index for table MPI_LINK on attributes MATCH_RESULT, TARGET_PID and VERSION


fil512 (approved):
Looks good to me.  Only thing I'm not sure about is the online flag.

PR #3711, opened 2022-06-15
Unknown resource type error message contains plain providers

### LINKED ISSUE #3710: Unknown resource type error message contains plain providers
When the RestfulServer encounters an unknown resource type, the error message includes plain providers

jdar8:
- Modify error message to include only known resource providers
- Modify a test to ensure plain providers are not included in error message
- Modify expected value of `testSearchWithUnknownResourceType` to reflect this change. Reasoning is: this test expected the error message to contain `server knows how to handle ... OperationDefinition` but it shouldn't since the server doesn't actually know how to handle `OperationDefinition` for CRUD operations. `OperationDefinition` is (in that case) set to be a plain provider and can only be used for system-level operations, but `testSearchWithUnknownResourceType` is testing a `GET` operation which is not a system-level operation.

Closes #3710

### DIFF
--- a/hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderOnlySomeResourcesProvidedR4Test.java
+++ b/hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderOnlySomeResourcesProvidedR4Test.java
@@ -1,11 +1,20 @@
 package ca.uhn.fhir.jpa.provider.r4;
 
 import ca.uhn.fhir.jpa.api.dao.DaoRegistry;
+import ca.uhn.fhir.jpa.provider.BaseJpaProvider;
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderOnlySomeResourcesProvidedR4Test.java
```diff
@@ -32,7 +48,20 @@ public void testCreateUnsupportedType() {
 		try {
 			myClient.create().resource(pract).execute();
 		} catch (ResourceNotFoundException e) {
-			assertThat(e.getMessage(), containsString("Unknown resource type 'Practitioner' - Server knows how to handle:"));
+			String errorMessage = e.getMessage();
+			assertThat(errorMessage, CoreMatchers.allOf(
```
fil512, line 52: ooh this is a cool matcher.  never seen that one before!  :-)

fil512 (approved):

PR #3727, opened 2022-06-22
Implement direct HSearch search path

michaelabuckley:
Provide a search strategy to query resource json directly from Hibernate Search without fetching pids first.


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
```diff
@@ -81,6 +86,9 @@
 
 public class FulltextSearchSvcImpl implements IFulltextSearchSvc {
 	private static final org.slf4j.Logger ourLog = org.slf4j.LoggerFactory.getLogger(FulltextSearchSvcImpl.class);
+
+	public static final int RSRC_SEARCH_LIMIT_DEFAULT = 50;
```
michaelabuckley, line 90: I think this constant already exists somewhere else in the Search Coordinator

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
```diff
@@ -103,8 +111,8 @@ public class FulltextSearchSvcImpl implements IFulltextSearchSvc {
 	private IHSearchSortHelper myExtendedFulltextSortHelper;
 
 	final private ExtendedLuceneSearchBuilder myAdvancedIndexQueryBuilder = new ExtendedLuceneSearchBuilder();
-
 	private Boolean ourDisabled;
+	private int mySearchCount = 0;
```
michaelabuckley, line 115: What is this for?  I don't see it used.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
```diff
@@ -382,4 +409,42 @@ public long count(String theResourceName, SearchParameterMap theParams) {
...
+		}
+
+		mySearchCount++;
+
+		var query = getSearchSession().search(ResourceTable.class)
+				.select(this::buildResourceSelectClause)
+				.where(f -> buildWhereClause(f, theResourceType, theParams, null));
+
+		if (theParams.getSort() != null && offset == 0) {
+			query.sort(
```
michaelabuckley, line 430: Shouldn't this be 
```suggestion
			query = query.sort(
```
Is this tested?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/IFulltextSearchSvc.java
```diff
@@ -95,4 +95,15 @@ public interface IFulltextSearchSvc {
...
 	long count(String theResourceName, SearchParameterMap theParams);
+
+	List<IBaseResource> searchForResources(String theResourceType, SearchParameterMap theParams);
+
+	boolean supportsAllOf(SearchParameterMap theParams);
+
+	/**
+	 * Required for testing
+	 */
+	public int getSearchCount();
```
michaelabuckley, line 106: I'd rather not have this in the service interface if we can.  Can we test this with a concrete class somewhere?

michaelabuckley: Or use the Logging Extension on the elasticsearch client and count the queries that way?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/search/ExtendedLuceneSearchBuilder.java
```diff
@@ -67,6 +67,42 @@ public boolean isSupportsSomeOf(SearchParameterMap myParams) {
...
+			! myParams.isDeleteExpunge() && // ???
+
+			// this works. doesn't it?
+//			! myParams.isLastN() && // ???
+
+//			// how about this?
+//			myParams.getLastNMax() != 0  && // ???
+
+			! myParams.isLoadSynchronous()  && // ???
+			myParams.getLoadSynchronousUpTo() == null && // ???
```
michaelabuckley, line 90: I believe the new path is the equivalent of the synchronous path, so you can drop these.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/search/ExtendedLuceneSearchBuilder.java
```diff
@@ -67,6 +67,42 @@ public boolean isSupportsSomeOf(SearchParameterMap myParams) {
...
+	/**
+	 * Are all the queries supported by our indexing?
+	 */
+	public boolean isSupportsAllOf(SearchParameterMap myParams) {
+		return
+//			fixme jm: confirm
+			myParams.getSummaryMode() == null &&
+			myParams.getSearchTotalMode() == null &&  // ???
+			myParams.getRevIncludes() == null && // ???
+			myParams.getIncludes() == null && // ???
```
michaelabuckley, line 79: We could implement include and rev-include.  It would require a second query, but should be fast.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/search/ExtendedLuceneSearchBuilder.java
```diff
@@ -67,6 +67,42 @@ public boolean isSupportsSomeOf(SearchParameterMap myParams) {
...
 	}
 
+	/**
+	 * Are all the queries supported by our indexing?
+	 */
+	public boolean isSupportsAllOf(SearchParameterMap myParams) {
+		return
+//			fixme jm: confirm
+			myParams.getSummaryMode() == null &&
+			myParams.getSearchTotalMode() == null &&  // ???
```
michaelabuckley, line 77: These two are supported when using HSearch, but not in this new direct path.  So move these to the direct strategy instead.

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/TestDaoSearch.java
```diff
@@ -110,6 +115,25 @@ public List<IBaseResource> searchForResources(String theQueryUrl) {
 		return result.getAllResources();
 	}
 
+
+	/**
+	 * Search for resources in the first query, instead of searching for IDs first
+	 */
+	@Transactional
+	public List<IBaseResource> searchForFastResources(String theQueryUrl) {
```
michaelabuckley, line 123: Why add this?  How is this different from using the existing `searchForResources()` when the direct path is supported?

I think this is too specialized for this test helper.  Maybe keep in the test class.

michaelabuckley (commented):
Looks good.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
jmarchionatto: Found it in BasePagingProvider

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/search/ExtendedLuceneSearchBuilder.java
jmarchionatto: OK

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/search/ExtendedLuceneSearchBuilder.java
jmarchionatto: Sure

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/TestDaoSearch.java
jmarchionatto: It uses that, but also used TestDaoSearch facility to get params from url.
Moved it to test.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
jmarchionatto: For counting searches, but you are asking a different way, so it will be removed once I can find that new way.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
jmarchionatto: No. Works as is. And it is tested now

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/search/ExtendedLuceneSearchBuilder.java
jmarchionatto: Not sure about that. For synchronous queries includes and revIncluded are all loaded before the query, but that doesn't consider offset and count, which we do here.
Let's discuss this further.

michaelabuckley: This will be fine for now.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
```diff
@@ -382,4 +410,37 @@ public long count(String theResourceName, SearchParameterMap theParams) {
...
+		myHSearchEventListener.hsearchEvent(IHSearchEventListener.HSearchEventType.SEARCH);
+
+		var query = getSearchSession().search(ResourceTable.class)
+				.select(this::buildResourceSelectClause)
+				.where(f -> buildWhereClause(f, theResourceType, theParams, null));
+
+		if (theParams.getSort() != null && offset == 0) {
+			query.sort(
+				f -> myExtendedFulltextSortHelper.getSortClauses(f, theParams.getSort(), theResourceType) );
+		}
```
JasonRoberts-smile, line 433: I don't understand why offset == 0 is here. Doesn't that mean that the sort rule will only apply to the first page, and subsequent pages would be unsorted?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/IHSearchEventListener.java
```diff
@@ -0,0 +1,10 @@
+package ca.uhn.fhir.jpa.dao;
+
+public interface IHSearchEventListener {
+
+	enum HSearchEventType {
+		SEARCH
```
JasonRoberts-smile, line 6: What's the value of an enum with only one entry?

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchWithElasticSearchIT.java
```diff
@@ -198,6 +207,22 @@ public class FhirResourceDaoR4SearchWithElasticSearchIT extends BaseJpaTest {
...
+
+	@Autowired
+	private MatchUrlService myMatchUrlService;
+
+	private int hSearchCount;
+
+
+	@BeforeEach
+	public void registerAsSearchEventListener() {
+		myHSearchEventDispatcher.register(this);
```
JasonRoberts-smile, line 224: nitpick: I probably would have chosen to register a mock listener and then verify how many times it gets called, rather than create an implementation of the interface. This test seems unusual enough to me that I would find it a speed bump if I were called upon to maintain or extend the feature

JasonRoberts-smile (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
jmarchionatto: Yes. We can't sort paginated request, so they shouldn't be paginated in the first place. 
See comment in SynchronousSearchSvcImpl 142:
			 * For synchronous queries, we load all the includes right away
			 * since we're returning a static bundle with all the results
			 * pre-loaded. This is ok because synchronous requests are not
			 * expected to be paged

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/IHSearchEventListener.java
jmarchionatto: Future extension

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
jmarchionatto: It is a synchronous requests and as such, not expected to be paged, as per the (old) comment for the service.

PR #3735, opened 2022-06-23
3719 enable map response terminology for bulk export

StevenXLi:
- Refactored out the logic for mapping codes in `ResponseTerminologyTranslationInterceptor` into a service called `ResponseTerminologyTranslationSvc`
- added logic in the write to file step of bulk export to call the `ResponseTerminologyTranslationSvc` to map the codes before export, in the file `BaseResourceToFileWriter`


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
```diff
@@ -227,6 +229,12 @@ public ResponseTerminologyTranslationInterceptor responseTerminologyTranslationI
 		return new ResponseTerminologyTranslationInterceptor(theValidationSupport);
 	}
 
+	@Lazy
+	@Bean
+	public ResponseTerminologyTranslationSvc responseTerminologyTranslationSvc(@Autowired ResponseTerminologyTranslationInterceptor theResponseTerminologyTranslationInterceptor) {
+		return new ResponseTerminologyTranslationSvc(theResponseTerminologyTranslationInterceptor);
+	}
```
tadgh, line 236: Qustion: Have you tried to build this locally and ensured CDR does not break? Or at least checked the repo for instances of the interceptor to ensure it is only instantiated here in HAPI? Should be safe, but should also check.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ResponseTerminologyTranslationInterceptor.java
```diff
@@ -62,17 +46,8 @@
...
-	private final BaseRuntimeElementDefinition<IPrimitiveType<?>> myCodeDefinition;
-	private final Class<? extends IBase> myCodeableConceptType;
-	private final Class<? extends IBase> myCodingType;
-	private final BaseRuntimeChildDefinition myCodeableConceptCodingChild;
-	private final BaseRuntimeElementCompositeDefinition<?> myCodingDefinitition;
-	private final RuntimePrimitiveDatatypeDefinition myStringDefinition;
-	private final BaseRuntimeChildDefinition myCodingDisplayChild;
-	private Map<String, String> myMappingSpecifications = new HashMap<>();
+	private final ResponseTerminologyTranslationSvc myResponseTerminologyTranslationSvc = new ResponseTerminologyTranslationSvc(this);
+	private final Map<String, String> myMappingSpecifications = new HashMap<>();
```
tadgh, line 50: design issue: It feels very strange to me that the interceptor still retains these mappings. Why does the interceptor need to know about them?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ResponseTerminologyTranslationSvc.java
```diff
@@ -0,0 +1,129 @@
...
+	private final BaseRuntimeElementDefinition<IPrimitiveType<?>> myUriDefinition;
+	private final BaseRuntimeElementDefinition<IPrimitiveType<?>> myCodeDefinition;
+	private final Class<? extends IBase> myCodeableConceptType;
+	private final Class<? extends IBase> myCodingType;
+	private final BaseRuntimeChildDefinition myCodeableConceptCodingChild;
+	private final BaseRuntimeElementCompositeDefinition<?> myCodingDefinition;
+	private final RuntimePrimitiveDatatypeDefinition myStringDefinition;
+	private final BaseRuntimeChildDefinition myCodingDisplayChild;
+
+	public ResponseTerminologyTranslationSvc(ResponseTerminologyTranslationInterceptor theMyResponseTerminologyTranslationInterceptor) {
```
tadgh, line 37: Issue: This service should need to know precisely nothing about the Interceptor. I see that you are using it internally to yank out context and mappings. I think a better design is to have this service accept the mappings and the FhirContext directly. This class ideally knows nothing about its related interceptors.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ResponseTerminologyTranslationSvc.java
```diff
@@ -0,0 +1,129 @@
...
+		myCodingSystemChild = myCodingDefinition.getChildByName("system");
+		myCodingCodeChild = myCodingDefinition.getChildByName("code");
+		myCodingDisplayChild = myCodingDefinition.getChildByName("display");
+
+		myUriDefinition = (RuntimePrimitiveDatatypeDefinition) myResponseTerminologyTranslationInterceptor.getContext().getElementDefinition("uri");
+		myCodeDefinition = (RuntimePrimitiveDatatypeDefinition) myResponseTerminologyTranslationInterceptor.getContext().getElementDefinition("code");
+		myStringDefinition = (RuntimePrimitiveDatatypeDefinition) myResponseTerminologyTranslationInterceptor.getContext().getElementDefinition("string");
+	}
+
+	public void processResourcesForTerminologyTranslation(List<IBaseResource> resources) {
```
tadgh, line 55: Design: This method should read: `processResourcesForTerminilogyTranslationWithMappings(List<IBaseREsource> theResources, Map theMappings){}` And you can pass the mappings down to the model visitor. This will remove the need for you to know about the interceptor in this code.

tadgh: Alternatively, change the `addMapping()` so that it lives in the translator SVC instead, and that way it will easily apply to both locations.

tadgh (changes requested):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ResponseTerminologyTranslationInterceptor.java
StevenXLi: I've tested the version with the mappings fully moved to the service. It didn't work as expected in the bulk export service call.
I think the issue is that after the service and the interceptor is created and autowired in the config, when trying to set the mapping with `addMappingSpecification` call in the interceptor somewhere in the code, the same service doesn't get updated with the new mappings

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ResponseTerminologyTranslationSvc.java
StevenXLi: see: https://github.com/hapifhir/hapi-fhir/pull/3735#discussion_r907566323

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ResponseTerminologyTranslationSvc.java
StevenXLi: see: https://github.com/hapifhir/hapi-fhir/pull/3735#discussion_r907566323

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ResponseTerminologyTranslationSvc.java
StevenXLi: I've also tried without passing in the interceptor, passing in only the `IValidationSupport`, the same issue as the other comment happens

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
StevenXLi: I have built this locally, but haven't ran all the tests.
The only place outside of HAPI where this interceptor is used is when this setting: https://smilecdr.com/docs/configuration_categories/fhir_endpoint_terminology.html#property-response-mapping
is enabled

StevenXLi:
The mapping specification will be removed from `ResponseTerminologyTranslationInterceptor` once the cdr side has been changed to use the new service as a way to set the mappings in terminology translation.

This way cdr will not break during this code change

tadgh (approved):
Approved pending changelog change

PR #3740, opened 2022-06-26
only register mdm clear job definition if it is not registered yet

fil512:


fil512:
hapi and cdr are out of sync until this is resolved

PR #3746, opened 2022-06-29
3745 - fixed NPE for bundle with duplicate conditional create resourc…

### LINKED ISSUE #3745: Transaction filing error - "theResource must not be null"
**Description**
During the input the bundle that has transaction type and contains several Conditional Create entries and an operation (for example, DELETE), an error "HAPI-0389: Failed to call access method: java.lang.NullPointerException: theResource must not be null" is displayed.

**To Reproduce**
1. Insert the attached Bundle via Postman. 
2. See error:
```
{
    "issue": [
        {
            "severity": "error",
            "code": "processing",
            "diagnostics": "HAPI-0389: Failed to call access method: java.lang.NullPointerException: theResource must not be null"
        }
    ]
}
```

**Expected behaviour**
Objects are successfully created and operation performed without errors.

**Environment:**
 - HAPI FHIR Version: 6.1.0

**Additional context**
Bundle example to reproduce the error:
[TransactionConditionalCreate.txt](https://github.com/hapifhir/hapi-fhir/files/9004788/TransactionConditionalCreate.txt)

kateryna-mironova:
Closes #3745

Added check for resource not null before replacing references in entries with consolidated UUID. Now the system will not try to change references in requests that do not contain a resource (for example DELETE requests).


tadgh (changes requested):
Nice work! Implementation looks like it will work,  but this will require a test before it can be merged.

tadgh (approved):

PR #3752, opened 2022-07-02
Add error-handler callback and ensure batch2 completion handler can't be called twice

fil512:


JasonRoberts-smile (approved):
Looks good to me

PR #3787, opened 2022-07-13
Fix sort in hsearch - use full millis

michaelabuckley:
We were sorting on the date ordinal, not the timestamp.

### DIFF
--- a/hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchWithElasticSearchIT.java
+++ b/hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchWithElasticSearchIT.java
@@ -145,7 +145,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 	DependencyInjectionTestExecutionListener.class
 	,FhirResourceDaoR4SearchWithElasticSearchIT.TestDirtiesContextTestExecutionListener.class
 	})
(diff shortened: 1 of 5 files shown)


hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchWithElasticSearchIT.java
```diff
@@ -1944,19 +1954,19 @@ public void byTokenCode() {
...
 					myCaptureQueriesListener.clear();
-					IBundleProvider result = myTestDaoSearch.searchForBundleProvider("/Observation?_sort=-date");
+					List<String> result = myTestDaoSearch.searchForIds("/Observation?_sort=-date");
 
 					assertEquals(0, myCaptureQueriesListener.getSelectQueriesForCurrentThread().size(), "we build the bundle with no sql");
-					// requesteddate descending so order should be id2, id1
-					assertThat(getResultIds(result), contains(id2, id1));
+					ourLog.info("byDate sort {}", result);
+					// date descending - order should be id2, id1
+					assertThat(result, contains(id2, id3, id4, id1));
```
jmarchionatto, line 1969: I don't see how ID's make this sort is deterministic given id2 and id4 have exactly the same date, given that IDs are not part of the query

michaelabuckley: The ids are for debugging the test.
The fix is changing the property we are sorting: from `dt.lower-ord` to `dt.lower`

PR #3833, opened 2022-07-21
Make sure that direct resource load path is only executed for synchro…

jmarchionatto:
…nous searches


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
```diff
@@ -361,13 +361,13 @@ public IBundleProvider registerSearch(final IFhirResourceDao<?> theCallingDao, f
 //		SearchStrategyFactory.ISearchStrategy searchStrategy = mySearchStrategyFactory.pickStrategy(theResourceType, theParams, theRequestDetails);
 //		return searchStrategy.get();
 
-		if (mySearchStrategyFactory.isSupportsHSearchDirect(theResourceType, theParams, theRequestDetails)) {
-			ourLog.info("Search {} is using direct load strategy", searchUuid);
-			SearchStrategyFactory.ISearchStrategy direct =  mySearchStrategyFactory.makeDirectStrategy(searchUuid, theResourceType, theParams, theRequestDetails);
-			return direct.get();
-		}
-
 		if (theParams.isLoadSynchronous() || loadSynchronousUpTo != null || isOffsetQuery) {
```
michaelabuckley, line 364: Please move this line inside the strategy calculation.  That's the whole point of the strategy factory -- avoiding the boolean checks being distributed over the entire trace.

michaelabuckley (changes requested):
Looks ok, but I'd like to consolidate the logic.

Have your verified that we produce next links on the non-offset path?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
jmarchionatto: I don't understand your ask, as we still want to have a path for not-direct synchronous query, so the check would be duplicated unless we left it outside of the strategy.

michaelabuckley (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
jmarchionatto: We agreed on look into this change by deepening the scope of strategy sometime later

PR #3836, opened 2022-07-22
Add support for deletes to stu3/dstu2 subscriptions. 

tadgh:
* Previously only r4 was supported. 
* Add tests
* Add changelog


JasonRoberts-smile (approved):
Looks good to me

PR #3842, opened 2022-07-23
everything operation with type param

jdar8:
- Adds support for `_type` parameter for the `$everything` operation on resource types and instances (all FHIR versions)
- Add tests 
- Add changelog


fil512 (approved):
approved assuming there is a test for this on the cdr side

TipzCM (changes requested):

TipzCM (approved):

Issue #3837, opened 2022-07-22
Query to include substances does not work


Qingyixia:
**Describe the bug**
Query to include substances does not work after upgrading. Search to include operation works fine in 2022.02 and returns the list of medications including substances in this case. After upgrading to 2022.05 it returns medications only.

**To Reproduce**
1. Create a Substance resource and a Medication resource that references the substance as an ingredient.
2. Send a GET request with the condition "Medication?\_include=Medication:ingredient".
3. The response contains only the Medication resource.

**Expected behavior**
The response contains both the Medication resource and the Substance resource.

PR #3845, opened 2022-07-25
Fixed the issue

Qingyixia:
#3837

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3837-fix-query-to-include-substances.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3837-fix-query-to-include-substances.yaml
@@ -2,4 +2,6 @@
 type: fix
 issue: 3837
 jira: SMILE-4467
-title: "Fix the query to include substances."
+title: "Method for finding the path of the included substances is fixed. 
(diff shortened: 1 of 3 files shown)


jmarchionatto (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3837-fix-query-to-include-substances.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 3837
+jira: SMILE-4467
+title: "Fix the query to include substances."
```
tadgh, line 5: Issue: Changelog requires much more description. Somebody reading this changelog has no idea what was fixed. Which query, which class of queries, what was fixed? It looks like the meat of the MR is a correction to _includes behaviour, so maybe talk about that.

tadgh (changes requested):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_1_0/3837-fix-query-to-include-substances.yaml
Qingyixia: Fixed

tadgh (approved):

PR #3847, opened 2022-07-25
Bring ES tests back in

michaelabuckley:
Reduce elasticsearch heap from 4GB to 512MB.  Revert the earlier change to remove all the elastic tests.


michaelabuckley:
Abandoned - see #3850 instead

PR #3852, opened 2022-07-27
3840 modify fetching batch jobs to not need any parameters

dzijianc:
Removed parameters of fetch jobs API and changed return type


hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
```diff
@@ -106,12 +105,11 @@ public interface IJobPersistence {
...
 	/**
-	 * Fetches all job instances based on the JobFetchRequest
-	 * @param theRequest - the job fetch request
-	 * @return - a page of job instances
+	 * Fetches all job instances with specified sort
+	 * @return - a list of job instances
 	 */
-	default Page<JobInstance> fetchJobInstances(JobInstanceFetchRequest theRequest) {
-		return Page.empty();
+	default List<JobInstance> fetchJobInstances() {
```
TipzCM, line 111: What's the need for this when you have various fetchInstances above (with page sizes, etc)?

The reason we added the Page<JobInstance> was because we required a paging API, no?

What's more, no sort is specified - so the documentation doesn't even match.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -136,15 +136,10 @@ public List<JobInstance> fetchInstancesByJobDefinitionId(String theJobDefinition
...
 	@Override
-	public Page<JobInstance> fetchJobInstances(JobInstanceFetchRequest theRequest) {
-		PageRequest pageRequest = PageRequest.of(
-			theRequest.getPageStart(),
-			theRequest.getBatchSize()
-		);
-
-		Page<Batch2JobInstanceEntity> pageOfEntities = myJobInstanceRepository.findAll(pageRequest);
+	public List<JobInstance> fetchJobInstances() {
+		List<Batch2JobInstanceEntity> pageOfEntities = myJobInstanceRepository.findAll();
```
TipzCM, line 140: I'm concerned about fetching literally all jobs.

Do we know that this won't be an unreasonably large amount, or that fetching all from the DB won't be problematic?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobCoordinator.java
```diff
@@ -68,8 +69,7 @@ public interface IJobCoordinator {
 
 	/**
 	 * Fetches all job instances tht meet the FetchRequest criteria
-	 * @param theFetchRequest - fetch request
-	 * @return - page of job instances
+	 * @return - list of job instances
 	 */
-	Page<JobInstance> fetchAllJobInstances(JobInstanceFetchRequest theFetchRequest);
+	List<JobInstance> fetchAllJobInstances();
```
TipzCM, line 74: Documentation does not match (no criteria is specified).

Moreover, i'd question the need for this when there are getInstances() above which can be used to fetch all (with some paging on your own).

hapi-fhir-storage/src/main/java/ca/uhn/fhir/batch2/models/JobInstanceFetchRequest.java
```diff
@@ -1,5 +1,7 @@
 package ca.uhn.fhir.batch2.models;
 
+import org.springframework.data.domain.Sort;
+
 public class JobInstanceFetchRequest {
```
TipzCM, line 5: Is this class even used anymore? It looks like you've excised it.

TipzCM (changes requested):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/batch2/models/JobInstanceFetchRequest.java
dzijianc: Yeah it has no uses anymore, I have deleted it.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
dzijianc: Updated documentation. Originally the paging works because it went through one database and retrieved the jobs inside, but since now it requires all jobs List can combine results easier.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
dzijianc: I can see the concern, but since the use case has no filters it has to retrieve all jobs. If there was a single database for all jobs then it would be possible to limit how many are retrieved.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobCoordinator.java
dzijianc: Updated documentation. I cannot use getInstances() since jobs can exist over multiple databases. As a result, if I need to page the result I can only do so with the whole list since I don't know how many results come from each database, and getInstances() only refers to one database.

PR #3858, opened 2022-07-28
Issue 3857 elastic lucene search returns only 50 results when offset0 and count100

jmarchionatto:


michaelabuckley (approved):
Thanks for the quick fix.

PR #3882, opened 2022-08-05
Fix illegalstateexception when an exception is thrown during stream response

### LINKED ISSUE #3881: Error during stream handling
Transcribed:

It appears that when an exception occurs in HAPI while a response to a FHIR request is being streamed back to the client application, the handler for the exception ends up throwing another unhandled exception. The result is that rather than getting an Operation Outcome, the client application receives a message saying {{Problem: 500: STREAM}}.

**Steps to Reproduce**
The triggers for this issue vary in the real production environment, but have been able to replicate similar behaviour by using an interceptor to throw an
 
1. Create and compile an interceptor that throws an exception the first time Pointcut.SERVER_OUTGOING_WRITER_CREATED occurs for a given request. A sample interceptor class has been attached
2. Load the interceptor
3. Using a REST client (e.g. Postman or Insomnia) submit a generic GET request, e.g. GET /Patient, with a request header "Accept-Encoding" set to "gzip" (including this header will cause the response to be streamed to the client).

tadgh:
* Previously, if an error was thrown after the output stream had been opened, the ExceptionHandlingInterceptor's attempt to open the stream was causing an IllegalStateException that was bubbling up to clients. 
* Now, the stream is reset, and an OperationOutcome is returned as intended. 
* Add changelog
* Add test

Closes #3881

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptor.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptor.java
@@ -108,7 +108,6 @@ public class ExceptionHandlingInterceptor {
 			resetOutputStreamIfPossible(response);
 		} catch (Throwable t) {
 			ourLog.error("HAPI-FHIR was unable to reset the output stream during exception handling. The root causes follows:", t);
-			return null;
 		}
 
 		return response.streamResponseAsResource(oo, true, Collections.singleton(SummaryEnum.FALSE), statusCode, statusMessage, false, false);
@@ -121,8 +120,9 @@ public class ExceptionHandlingInterceptor {
 	 * Also, it strips the content-encoding header if present, as the method outcome will negotiate its own.
 	 */
 	private void resetOutputStreamIfPossible(IRestfulResponse response) {
-		if (response instanceof HttpServletResponse) {
-			HttpServletResponse servletResponse = ((ServletRestfulResponse) response).getRequestDetails().getServletResponse();
+		if (response.getClass().isAssignableFrom(ServletRestfulResponse.class)) {
+			ServletRestfulResponse servletRestfulResponse = (ServletRestfulResponse) response;
+			HttpServletResponse servletResponse = servletRestfulResponse.getRequestDetails().getServletResponse();
 			Collection<String> headerNames = servletResponse.getHeaderNames();
 			Map<String, Collection<String>> oldHeaders = new HashedMap<>();
 			for (String headerName : headerNames) {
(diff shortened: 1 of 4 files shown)


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptor.java
```diff
@@ -101,10 +103,21 @@ public Object handleException(RequestDetails theRequestDetails, BaseServerRespon
 		}
 
 		BaseResourceReturningMethodBinding.callOutgoingFailureOperationOutcomeHook(theRequestDetails, oo);
+		resetOutputStreamIfPossible(response);
```
jmarchionatto, line 106: thought: I would have liked it better `resetOutputStreamIfNecessary `

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptorTest.java
```diff
@@ -77,6 +81,30 @@ public void testInternalError() throws Exception {
...
+		//Then: This should still return an OperationOutcome, and not explode with an HTML IllegalState response.
+		String responseContent = IOUtils.toString(status.getEntity().getContent(), Charsets.UTF_8);
+		IOUtils.closeQuietly(status.getEntity().getContent());
+		ourLog.info(responseContent);
+		assertEquals(500, status.getStatusLine().getStatusCode());
+		OperationOutcome oo = (OperationOutcome) servlet.getFhirContext().newXmlParser().parseResource(responseContent);
+		ourLog.info(servlet.getFhirContext().newXmlParser().encodeResourceToString(oo));
+		assertThat(oo.getIssueFirstRep().getDiagnosticsElement().getValue(), StringContains.containsString("Simulated IOException"));
+	}
+
```
jmarchionatto, line 107: Brilliant!

jmarchionatto (approved):

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptorTest.java
tadgh: 🥳

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ExceptionHandlingInterceptor.java
tadgh: Ah that's a fair thought. I personally prefer `ifPossible` since it feels to me as though `ifNecessary` implies some sort of scenario where we would _choose_ not to due to some business reason

Issue #3764, opened 2022-07-07
HAPI FHIR validator 5.2.0 issue related to meta.extension


yiupeter:
NOTE: Before filing a ticket, please see the following URL:
https://github.com/hapifhir/hapi-fhir/wiki/Getting-Help

**Describe the bug**
 HAPI FHIR validator 5.2.0 issue related to meta.extension

**To Reproduce**
The payload is enclosed here –
```
{
  "resourceType" : "MedicationRequest",
  "meta" : {
	"versionId" : "1",
	"tag" : [
  	{
    	"system" : "https://smarthealthit.org/tags",
    	"code" : "synthea-8-2017"
  	}
	],
	"lastUpdated" : "2021-04-17T02:39:06.885-04:00",
	"extension" : [
  	{
    	"valueUri" : "https://r2.smarthealthit.org/",
    	"url" : "http://hl7.org/fhir/4.0/StructureDefinition/extension-Meta.source"
  	}
	]
  },
  "encounter" : {
	"reference" : "Encounter/a83e5350-6462-4331-adba-dc72552cecbd"
  },
  "subject" : {
	"reference" : "Patient/ab606cad-b848-4a24-b3b6-61075563fadf"
  },
  "authoredOn" : "2018-02-19",
  "status" : "stopped",
  "medicationCodeableConcept" : {
	"coding" : [
  	{
    	"display" : "Atorvastatin",
    	"system" : "http://www.nlm.nih.gov/research/umls/rxnorm",
    	"code" : "259255"
  	}
	],
	"text" : "Atorvastatin"
  },
  "intent" : "order"
```

	In our dependency, we are having
           	hapi-fhir-base: 5.2.0
           	hapi-fhir-structures-r4: 5.2.0
           	hapi-fhir-validation: 5.2.0
           	hapi-fhir-validation-resources-r4: 5.2.0
            	We are initializing the validator through the following code and use validateWithResult on the validator:
final FhirContext ctx = FhirContext.forR4();
final FhirInstanceValidator instanceValidator = new FhirInstanceValidator(ctx);
instanceValidator.setNoTerminologyChecks(true);
validator = ctx.newValidator();
validator.registerValidatorModule(instanceValidator);
    	validator.validateWithResult(…);
* What result is being received
            	Result was not generated, validator throws a null pointer, stack trace like this:
java.lang.NullPointerException
[...]

PR #3896, opened 2022-08-09
3764 WIP Fix for #3764

dotasek:
WORK IN PROGRESS 

This is the bare minimum of code required for our test case to pass, and to allow a review of the solution itself.


hapi-fhir-docs/src/main/java/ca/uhn/hapi/fhir/docs/ValidateDirectory.java
```diff
@@ -94,7 +94,7 @@ public static void main(String[] args) throws Exception {
 
       ValidationSupportChain validationSupportChain = new ValidationSupportChain();
       validationSupportChain.addValidationSupport((ca.uhn.fhir.context.support.IValidationSupport) new DefaultProfileValidationSupport(ctx));
-      validationSupportChain.addValidationSupport((ca.uhn.fhir.context.support.IValidationSupport) new PrePopulatedValidationSupport(ctx, structureDefinitions, valueSets, codeSystems));
```
jamesagnew, line 97: Recommend leaving the old constructor in place (the one without the final parameter) but also creating the new constructor, and having the old one just chain to the new one. That way we don't break any existing users' code.

hapi-fhir-validation/src/test/java/org/hl7/fhir/r4/validation/ParserWithValidationR4Test.java
```diff
@@ -0,0 +1,48 @@
...
+import org.hl7.fhir.common.hapi.validation.validator.FhirInstanceValidator;
+import org.hl7.fhir.r4.model.MedicationRequest;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+
+import static ca.uhn.fhir.util.ClasspathUtil.loadResource;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+public class ParserWithValidationR4Test {
```
jamesagnew, line 21: A couple of comments on this test:

* The naming for this class and the test method don't seem to align with what's actually being tested. Maybe `XverValidationR4Test`?
* We should include a negative test as well - E.g. same resource being validated but instead of `"valueUri" : "https://r2.smarthealthit.org"` it ias `"valueInteger": 123` and presumably we'd get a validation error because the value is of the wrong datatype?
* It'd be nice to include an example on [this page](https://hapifhir.io/hapi-fhir/docs/validation/examples.html) that shows how to validate with xver refs. I think the code snippet would basically by the same logic as this test.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
```diff
@@ -258,7 +258,8 @@ public List<StructureDefinition> getStructures() {
 
 	@Override
 	public void cacheResource(Resource res) {
-		throw new UnsupportedOperationException(Msg.code(660));
+		//FIXME Should we implement anything here?
```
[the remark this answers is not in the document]
jamesagnew: @dotasek definitely no concerns with making that method a no-op.

PR #3905, opened 2022-08-11
3904 during $delete expunge job hibernate search indexed documents are left orphaned

### LINKED ISSUE #3904: During $delete-expunge job, hibernate search indexed documents are left orphaned.
1. Enable hibernate search (either lucene/elasticsearch). 
2. Create some data
3. Query Elasticsearch to see that `resourcetable-000001` index contains documents. \
4. Run a $delete-expunge job. 
5. Wait for it to complete

Expected Behaviour: 
- resources expunged from database should also be expunged from the hibernate search index. 

Actual Behaviour: 
- Documents representing expunged resources remain visible in Elasticsearch. 

Expunging should also clean out the HS indexes.

tadgh:
* Perform a manual index clean post-job
* Add test
* Add changelog

Closes #3904

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
@@ -444,4 +444,14 @@ public class FulltextSearchSvcImpl implements IFulltextSearchSvc {
 	}
 
 
+	@Override
+	public void deleteIndexedDocumentsByTypeAndId(Class theClazz, List<Object> theGivenIds) {
+		SearchSession session = Search.session(myEntityManager);
+		SearchIndexingPlan indexingPlan = session.indexingPlan();
+		for (Object givenId : theGivenIds) {
+			indexingPlan.purge(theClazz, givenId, null);
+		}
+		indexingPlan.process();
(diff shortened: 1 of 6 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
```diff
@@ -444,4 +444,14 @@ private void dispatchEvent(IHSearchEventListener.HSearchEventType theEventType)
 	}
 
 
+	@Override
+	public void deleteIndexedDocumentsByTypeAndId(Class theClazz, List<Object> theGivenIds) {
```
StevenXLi, line 448: Question: I only see the Class parameter used for ResourceTable.class, does this class not have access to that import? or are we anticipating other class types to be passed in?

also, nice name, does using `theClass` break something?

tadgh: 1. Yup we need to support arbitrary indexed types, it's just our first implementation happens to be on resourcetable. 

2. It does not break anything, but `clazz` is an old java convention. Since `class` is a reserved keyword in java, developers would use `clazz` instead to indicate that the argument was a class. Now this is completely rendered moot by our `the` prefix, but I kept the convention because I'm an old man.

PR #3938, opened 2022-08-19
3937 searching for resources with a target resource parameter in  include fail

epeartree:
Root cause of the issue:
- Use of contiguous ternary operators that would both add sql conditions each with named parameters but subsequently providing only one named parameter value.  As a result, we'd get an exception from Hibernate because one of the named parameter was not provided.

What was done:
- Replacing string concatenation to a StringBuilder in native query building.  Making use of the StringBuilder simplifies the code and provides an easy way to replace the contiguous ternary operators.
- Adding test case;
- Adding changeLog.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
@@ -1171,17 +1171,24 @@ public class SearchBuilder implements ISearchBuilder {
 							}
 						}
 						//@formatter:off
-						String resourceUrlBasedQuery = "SELECT " + fieldsToLoadFromSpidxUriTable +
+						StringBuilder resourceUrlBasedQuery = new StringBuilder("SELECT " + fieldsToLoadFromSpidxUriTable +
 							" FROM hfj_res_link r " +
 							" JOIN hfj_spidx_uri rUri ON ( " +
 							"   r.target_resource_url = rUri.sp_uri AND " +
-							"   rUri.sp_name = 'url' " +
-							    (targetResourceType != null ? " AND rUri.res_type = :target_resource_type " : "") +
-							    (haveTargetTypesDefinedByParam ? " AND rUri.res_type IN (:target_resource_types) " : "") +
-							" ) " +
+							"   rUri.sp_name = 'url' ");
+
+						if(targetResourceType != null) {
+							resourceUrlBasedQuery.append(" AND rUri.res_type = :target_resource_type ");
+
+						} else if(haveTargetTypesDefinedByParam) {
+							resourceUrlBasedQuery.append(" AND rUri.res_type IN (:target_resource_types) ");
+						}
+
+						resourceUrlBasedQuery.append(" ) ");
+						resourceUrlBasedQuery.append(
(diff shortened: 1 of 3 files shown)


hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4Test.java
```diff
@@ -1819,6 +1820,27 @@ public void testElements() throws IOException {
 		}
 	}
 
+	@Test
+	public void testSearchWithIncludeAndTargetResourceParameterWillSucceed() {
```
nathandoef, line 1824: Are both conditions covered by tests?
- `if(targetResourceType != null)` 
- `else if(haveTargetTypesDefinedByParam)`

nathandoef (commented):
Left one question

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -1171,17 +1171,24 @@ public Set<ResourcePersistentId> loadIncludes(FhirContext theContext, EntityMana
 							}
 						}
 						//@formatter:off
-						String resourceUrlBasedQuery = "SELECT " + fieldsToLoadFromSpidxUriTable +
+						StringBuilder resourceUrlBasedQuery = new StringBuilder("SELECT " + fieldsToLoadFromSpidxUriTable +
```
jamesagnew, line 1174: I may be missing something but I don't understand how this change (the whole block, not specifically this line) is actually any different in the end.

The ternary version or the stringbuilder version look functionally identical to me. What is the difference?

epeartree: The first problem is with the way that the ternary operators are being used. If both of their condition evaluate to true, both sql strings will be added when we should be adding only one.

The second problem with the previous construction is that if we add both sql strings, it implies that we have to provide values for named parameter `target_resource_type` AND `target_resource_type**s**`.  We end up with Hibernate throwing an exception because we don't provide a value for `target_resource_type**s**` due to the `if() else if()` construction starting at line 1201.

Using the StringBuilder allows for an easy appending within the `if() else if()` construction starting at line 1180.

jamesagnew (approved):

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4Test.java
epeartree: The second condition is asserted in test 'testElements' where an '_include' directive is provided but without a targetResoureType.  The targetResourceType is the third and optional parameter of an _include directive where each parameters is separated by a colon.  As per [the specification](https://www.hl7.org/fhir/search.html#include):

1. The name of the source resource from which the join comes

2. The name of the search parameter which must be of type reference

3. (Optional) A specific of type of target resource (for when the search parameter refers to multiple possible target types)

vadi2:
Could the public sandbox be updated to include this fix?

PR #3952, opened 2022-08-24
3951 reimplement missing search support

TipzCM:


TipzCM:
Refactored Searching to use a new method of searching for missing fields.

This method is better for some cases (searching for non-missing fields on resources) than the old method. But slower on others (searching for missing-fields on similar resources when there are many similarly typed resources).

Both will be supported and "toggled" by setting/unsetting the IndexMissingFields value.

Included a changelog

Issue #3956, opened 2022-08-25
Bulk export returning the wrong path if "Fixed Value for Endpoint Base URL" is set


Qingyixia:
NOTE: Before filing a ticket, please see the following URL:
https://github.com/hapifhir/hapi-fhir/wiki/Getting-Help

**Describe the bug**
If the “Fixed Value for Endpoint Base URL” value is set to something different from the default value, the URL that is export-poll-status URL that is returned is incorrect. Likewise, after correcting the export-poll-status URL, the URL that is returned is also incorrect.

**To Reproduce**
1. Enable Bulk Export.
2. Set a fixed value for endpoint base URL. (eg. http://example/fixedvalue)
3. Create a partition and post a patient to the partition.
4. Run a bulk export against the partition
5. Do a GET on the URL from step 4, which will give you a 404 Not Found error.
6. In the URL, remove the extra part (e.g. /fixedvalue), and run the GET again. You will now get a Binary file URL similar to the following: [](http://example/fixedvalue/DEFAULT/Binary/1111)
7. Do a GET on the URL from step 6, and you will get this error: Unknown resource type 'DEFAULT'.
8. Remove the "/fixedvalue" part from the URL, and do a GET. You will now be able to retrieve the binary file.

**Expected behavior**
The URLs that are returned from $export and $export-poll-status should not contain the extra path from "Fixed Value for Endpoint Base URL".

PR #3958, opened 2022-08-25
Fixed the issue of bulk export returning the wrong path if "Fixed Value for Endpoint Base URL" is set

Qingyixia:
#3956

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/3956-bulk-export-returning-the-wrong-path-if-fixed-value-for-endpoint-base-url-is-set.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/3956-bulk-export-returning-the-wrong-path-if-fixed-value-for-endpoint-base-url-is-set.yaml
@@ -1,0 +1,6 @@
+---
+type: fix
+issue: 3956
+title: "Previously, if the Endpoint Base URL is set to something different from the default value, the URL that export-poll-status returned is incorrect.
+And after correcting the export-poll-status URL, the binary file URL returned is also incorrect. Now, this error has been fixed and the URLs that are returned 
+from $export and $export-poll-status will not contain the extra path from 'Fixed Value for Endpoint Base URL'."
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/bulk/export/provider/BulkDataExportProvider.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/bulk/export/provider/BulkDataExportProvider.java
@@ -45,6 +45,7 @@ import ca.uhn.fhir.rest.api.PreferHeader;
 import ca.uhn.fhir.rest.api.server.RequestDetails;
 import ca.uhn.fhir.rest.api.server.bulk.BulkDataExportOptions;
 import ca.uhn.fhir.rest.server.RestfulServerUtils;
+import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
 import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import ca.uhn.fhir.rest.server.servlet.ServletRequestDetails;
 import ca.uhn.fhir.rest.server.util.CompositeInterceptorBroadcaster;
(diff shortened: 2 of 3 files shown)


jdar8:
Nice work!

hapi-fhir-base/src/main/java/ca/uhn/fhir/i18n/Msg.java
```diff
@@ -25,7 +25,7 @@ public final class Msg {
 
 	/**
 	 * IMPORTANT: Please update the following comment after you add a new code
-	 * Last code value: 2130
+	 * Last code value: 2132
```
tadgh, line 28: pretty sure this will conflict. Pull master into your branch and update accordingly. We are already at 2132

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/3956-bulk-export-returning-the-wrong-path-if-fixed-value-for-endpoint-base-url-is-set.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 3956
+title: "Previously, if the Endpoint Base URL is set to something different from the default value, the URL that export-poll-status returned is incorrect.
+And after correcting the export-poll-status URL, the binary file URL returned is also incorrect. Now, this error has been fixed and the URLs that are returned
```
tadgh, line 5:
```suggestion
After correcting the export-poll-status URL, the binary file URL returned is also incorrect. This error has also been fixed and the URLs that are returned 
```

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/bulk/export/provider/BulkDataExportProvider.java
```diff
@@ -143,7 +144,17 @@ private boolean shouldUseCache(ServletRequestDetails theRequestDetails) {
...
+		if (theRequestDetails.getCompleteUrl().contains(theRequestDetails.getServerBaseForRequest())) {
+			// Base URL not Fixed
+			return StringUtils.removeEnd(theRequestDetails.getServerBaseForRequest(), "/");
+		} else {
+			// Base URL Fixed
+			int index = StringUtils.indexOf(theRequestDetails.getCompleteUrl(), theRequestDetails.getOperation());
+			if (index == -1) {
+				return null;
+			}
+			return theRequestDetails.getCompleteUrl().substring(0, index - 1);
```
tadgh, line 156: issue: This logic is potentially dangerous. Is there no way to actually fetch the current server base from config and simply add it here, instead of trying to do indexing math? If you have tested this thoroughly, I am ok with it, but feels a bit dangerous. What is actually going on here? This is how it reads to me: 

1. If the full url contains the server's base for the request, then we use the old logic. 
2. If the full url does not contain the base of the request, we use the base of the complete URL? 

Why is the complete url different base from the serverBaseForRequest?

@jamesagnew @fil512 can you think of anyway to do this that isn't parsing out of the completeUrl?

tadgh (approved):
Approved pending resolution of comments.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/bulk/export/provider/BulkDataExportProvider.java
Qingyixia: Yeah, the logic you stated in the bullet points is exactly the same as what I meant. 

1. If the base URL is not fixed which means it's the default value, then we just simply get the server base. 
2. If the base URL is set to a fixed value other than the default value, then the server base is now becoming the fixed value. For example, if the fixed value for the endpoint base URL is set to be "**http://example.com/fixedvalue**", the fhir server base will now become "**http://example.com/fixedvalue**" and only the complete URL will contain the server base of "**http://localhost:PORT**" (which is the one we aimed to get). So in this case, I choose to get the server base using the indexing method. 

Currently, this method can successfully deal with all the cases that I can think of. But I totally agree that this logic is kind of dangerous, I'll try to find a better way to get the server base if possible!

(To answer the question of "Why is the complete URL different base from the serverBaseForRequest?", 
On the web admin page, it stated that If "Fixed Value for Endpoint Base URL" is specified, the server will always use this base URL for FHIR responses.)

Issue #2444, opened 2021-03-04
Creating HapiWorkerContext crashes: dependency missing: com.github.benmanes.caffeine.cache.Caffeine


jingtang10:
**Describe the bug**

Caused by: java.lang.ClassNotFoundException: com.github.benmanes.caffeine.cache.Caffeine

**To Reproduce**

Create a new java project using gradle.

Add dependency to hapi structures lib

```
    implementation 'ca.uhn.hapi.fhir:hapi-fhir-structures-r4:5.3.0'
```

Create a main class that does the following:

```
package com.example;

import ca.uhn.fhir.context.FhirContext;
import ca.uhn.fhir.context.support.DefaultProfileValidationSupport;
import org.hl7.fhir.r4.utils.FHIRPathEngine;
import org.hl7.fhir.r4.hapi.ctx.HapiWorkerContext;

public class Main {
  public static void main (String[] args) {
    FhirContext context = FhirContext.forR4();
    DefaultProfileValidationSupport validationSupport = new DefaultProfileValidationSupport(context);
    HapiWorkerContext workerContext = new HapiWorkerContext(context, validationSupport);
    FHIRPathEngine fhirPathEngine = new FHIRPathEngine(workerContext);
  }
}
```

**Expected behavior**
It should run.

**Actual behavior**
```
Caused by: java.lang.ClassNotFoundException: com.github.benmanes.caffeine.cache.Caffeine
```

**Environment (please complete the following information):**
 - HAPI structures version: ca.uhn.hapi.fhir:hapi-fhir-structures-r4:5.3.0
 - IDE:
IntelliJ IDEA 2020.2.3 (Community Edition)
Build #IC-202.7660.26, built on October 6, 2020
Runtime version: 11.0.8+10-b944.34 x86_64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
macOS 10.16
GC: ParNew, ConcurrentMarkSweep
Memory: 1981M
Cores: 8
 - java: project sdk: 11.0.4

**Additional info**
This can be resolved by adding the following explicit dependency to my project:

```
implementation 'com.github.ben-manes.caffeine:caffeine:2.9.0'
```

but I shouldn't need to do this. Is this a problem with the pom.xml file?

vitorpamplona:
[...]

Issue #3690, opened 2022-06-10
Make sure HAPI structures are Android friendly


jingtang10:
NOTE: Before filing a ticket, please see the following URL:
https://github.com/hapifhir/hapi-fhir/wiki/Getting-Help

**Describe the issue**
A clear and concise description of what the feature request is. Please include details about what problem you are ultimately trying to solve (i.e. what are you building with HAPI FHIR) and how this feature would help.

HAPI structures is currently used by the [Android FHIR SDK](https://github.com/google/android-fhir), but it's possible (and perhaps should be encouraged?) that other Android projects use the HAPI structures as FHIR data models. 

However, not all Java libraries can be used on Android automatically, certain dependencies will cause issues. For example JAXB as we discovered recently https://github.com/cqframework/clinical_quality_language/pull/752, https://github.com/google/android-fhir/issues/1423. Using Java9+ features can also cause issues on Android.

It would be useful to have a strategy wrt HAPI structures and Android. Is there a set of minimum tests we can add to ensure things work on Android? Is there some GitHub plugin that can help us with this? What documentation should we add to the HAPI docs to be explicit about this with developers?

@jamesagnew @tadgh @yigit @vitorpamplona @stevenckngaa @kevinmost @aditya-07 @fredhersch

**Environment (please complete the following information):**
NA

**Additional context**
https://developer.android.com/studio/write/java8-support
https://jakewharton.com/androids-java-9-10-11-and-12-support/

tadgh:
I had a gander around for maven plugins that could check android compatibility but came up empty. Happy to take suggestions on maintaining android support via tests/github actions if any exist.

jingtang10:
Discussed with Vitor (vitorpamplona@ on GitHub).

I think one of the reasons it's so critical to address this issue on HAPI's side is that it's much more difficult to reverse new dependencies after a release. For example, if HAPI introduces a dependency on some java api or system api that is not available on Android, we will only discover that after the release has been made and when we're testing the new HAPI structures lib version. Once we discover the issue, we'll then need to report to HAPI and see if there's a way to override the dependency and/or reverse the usage in the next version. This would be hugely time consuming.

And if something happens that is irreversible, that means HAPI structures will not be usable on Android from that point on. This is pretty limiting for HAPI's usefulness.

[...]

PR #3977, opened 2022-08-31
Extracts Caching Libraries into Service Loaders (Caffeine -> Java, Guava -> Android)

vitorpamplona:
### Background

[Caffeine 3.0+ does not plan to support Android](https://github.com/ben-manes/caffeine/issues/763). The 3.0 version requires a `java.lang.System.getLogger` method from the [JEP-264: Platform Logging API](https://openjdk.java.net/jeps/264) to redirect logs to `slf4j`. Unfortunately, Android has not provided a `System.getLogger` interface yet. This means a HAPI migration to Caffeine 3.0+ will simply crash in runtime on Android. 

### Solution

Abstract the Caching function, remove direct dependencies on Caffeine, and build an Android-friendly alternative. 

1. This PR establishes a new Caching structure via ServiceLoaders in 4 new maven modules: 
   - `hapi-fhir-cache`: Cache service loader interfaces and factory methods. 
   - `hapi-fhir-cache-guava`: Delegator classes to Guava (ideal for Android)
   - `hapi-fhir-cache-caffeine`: Delegator classes to Caffeine (current implementation) 
   - `hapi-fhir-cache-testing`: Utility module to quickly alternate between `guava` and `caffeine` and run the project's test cases on.

2. All modules dependent on `caffeine` now depend on `hapi-fhir-cache` directly. Users must choose a caching implementation by also importing `hapi-fhir-cache-caffeine`  or `hapi-fhir-cache-guava`. If none is present, the following error message will be displayed: `No Cache Service Providers found. Choose between hapi-fhir-cache-caffeine (Default) and hapi-fhir-cache-guava (Android)`

3. `hapi-fhir-cache-guava` contains brand new Cache and LoadingCache implementations that mimic Caffeine's behavior completely and it's ideal for the use of HAPI on Android. In my tests, it fully replaces Caffeine for the needs of this project. 

Closes: #2444, Helps: #3690


ben-manes:
Just want to make sure this wasn't forgotten in the discussion. If the [cache-api](https://github.com/Xanthic/cache-api) project fit your needs, it might be a good way for you and others to collaborate. It may not and I am not familiar with it or the developers, but an fyi in case that avoids duplication of efforts.

vitorpamplona:
Thank you, Ben! I didn't know the project existed. Maybe we can collaborate with them in the future. 

At a first glance, it does a similar job. However, the cache-api does not standardize the interface among all implementations and it doesn't offer a provider for the `LoadingCache` class. This makes testing for all implementations really difficult. There are behavioral differences between Caffeine and Guava that need to be sorted out by the Delegator. You can see them in the `Cache.get` method in the Guava module that it: 

1. Has to wrap ExecutionExeptions as RuntimeExeptions 
2. Has to unwrap UncheckedExecutionException to expose the original Exception 
3. Has to not throw exceptions when the CacheLoader returns null

These are needed to completely match Caffeine's API. 

The CacheDelegator ends up looking like this: 

```java
try {
	return cache.get(key, () -> mappingFunction.apply(key));
} catch (ExecutionException e) {
	e.printStackTrace();
	throw new RuntimeException(e);
} catch (UncheckedExecutionException e) {
	if (e.getCause() instanceof RuntimeException) {
		// Unwrap exception to match Caffeine
		throw (RuntimeException)e.getCause();
	}
	throw e;
} catch (CacheLoader.InvalidCacheLoadException e) {
	// If the entry is not found or load as null, returns null instead of an exception
	// This matches the behaviour of Caffeine
	return null;
}
```

ben-manes:
It might be unsatisfactory, but would using Caffeine's Guava Cache adapters fit your goals? That attempts to make Caffeine conform to Guava, with similar workarounds to your snippet. Just trying to make sure you have all the options to consider, and whatever the answer will be the best choice for your project.

vitorpamplona:
I did see those components (thank you for making them) but I discarded that approach because it adds significant complexity to the exception handling on the HAPI side. We could go there, but it wouldn't be a simple 3hr PR by then. 

HAPI was clearly coded with Caffeine's API in mind. And I do believe that, after looking at both, Caffeine's API is superior for users.

jamesagnew:
@vitorpamplona - I want to get this merged in time for the 6.2.0 release. I've pulled these changes into a fresh PR that I'll be submitting to our internal review process tomorrow: #4196

vitorpamplona:
Nice! Yes, let's go! :)

vitorpamplona:
Thank you, James!

vitorpamplona:
Hi @jamesagnew 

Is there any chance we can avoid using the caffeine library as default? Today every Android developer hits a roadblock when trying to use HAPI because they need to figure out that they need to manually remove caffeine and add guava in their build files. 

The current dependency stack doesn't provide a great dev experience for Android. :( 

CC @jingtang10 @brynrhodes @JPercival

jamesagnew:
@vitorpamplona I thought we had? What is pulling it in?

jamesagnew:
@vitorpamplona pushing through a fix in https://github.com/hapifhir/hapi-fhir/pull/5472

vitorpamplona:
Thank you, James!

PR #3978, opened 2022-08-31
Permit uplift

tadgh:
* Permit Uplifted search parameters to be created with identical base + code. 
* Add test cases 
* perform internal comparison of uplift extensions to ensure we retain no duplicates.


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/searchparam/submit/interceptor/SearchParamValidatingInterceptor.java
```diff
@@ -53,6 +57,7 @@
 public class SearchParamValidatingInterceptor {
 
 	public static final String SEARCH_PARAM = "SearchParameter";
+	public static final String UPLIFT_EXTENSION_URL = "https://smilecdr.com/fhir/ns/StructureDefinition/searchparameter-uplift-refchain";
```
fil512, line 60: suggestion: replace this hard-coded URL with a `List<String> myUpliftExtensions` and add a public `addUpliftExtension` method for adding to it.

fil512 (approved):

PR #3983, opened 2022-09-01
SMILE-3459 Implemented -l parameter in validator. Added SnapshotGener…

### LINKED ISSUE #4003: CLI Validation against FHIR profile not working
**Describe the bug**

KGJ-software:
Implemented support for -l parameter. Added test to ensure local profile is used.

Closes #4003


dotasek:
I would need to see some manner of test for this to verify that it's working.

I'd imagine that it would follow a similar format to: ca.uhn.fhir.cli.ValidateCommandTest testValidateLocalProfileR4() for example.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/FhirInstanceValidator.java
```diff
@@ -229,7 +229,6 @@ protected List<ValidationMessage> validate(IValidationContext<?> theValidationCt
 			.setBestPracticeWarningLevel(getBestPracticeWarningLevel())
 			.setErrorForUnknownProfiles(isErrorForUnknownProfiles())
 			.setExtensionDomains(getExtensionDomains())
-			.setValidatorResourceFetcher(validatorResourceFetcher)
```
dotasek, line 232: This looks like it will cause breakage; ValidatorWrapper would have no way to have its ValidatorResourceFetcher set here.

hapi-fhir-cli/hapi-fhir-cli-api/src/main/java/ca/uhn/fhir/cli/ValidateCommand.java
```diff
@@ -164,7 +163,11 @@ public void run(CommandLine theCommandLine) throws ParseException {
 				case R4: {
 					FhirInstanceValidator instanceValidator = new FhirInstanceValidator(ctx);
 					val.registerValidatorModule(instanceValidator);
-					ValidationSupportChain validationSupport = new ValidationSupportChain(new DefaultProfileValidationSupport(ctx), new InMemoryTerminologyServerValidationSupport(ctx));
+					ValidationSupportChain validationSupport = new ValidationSupportChain(
+						new DefaultProfileValidationSupport(ctx),
+						new InMemoryTerminologyServerValidationSupport(ctx),
+						new LocalFileValidationSupport(ctx),
```
dotasek, line 169: I think, based on what I see below, that LocalFileValidationSupport and SnapshotGeneratingValidationSupport can be added optionally. Like it is when theCommandLine.hasOption("r") adds support with addValidationSupport.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/FhirInstanceValidator.java
KGJ-software: setValidatorResourceFetcher is called below, I think the one I removed was merely a duplicate.

KGJ-software:
@dotasek @nathandoef 

I think this PR is in a good place now.

dotasek (changes requested):
See comments in code for details:

1) The ValidateCommandTest is perhaps a little too brittle. There's a less verbose check that might be sufficient.
2) LocalFileValidationSupport doesn't discriminate between types of resource, which could cause errors, and could be updated to be more robust. 
3) LocalFileValidationSupport should have its own unit test.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/FhirInstanceValidator.java
markiantorno: Might be a good opportunity to change this implementation of Validator wrapper to take a parameter argument, vs just chaining in 10+ vars here.

markiantorno (commented):
So, if there are tests that already exist that test that validation works when you provide a local profile, I wouldn't test that again. 

What I would suggest is going even more granular and tests the changes specifically, test the 'l' param is detected and causes the load to occure, test the struct def is loaded correctly, and test that the validator is called with the appropriate structure def used.

If we add another validation here, it will add more time to the overall tests. We want to keep them as lean as possible.

dotasek (approved):
This looks good to me. If @markiantorno approves as well, we're good to go!

markiantorno (approved):

PR #4049, opened 2022-09-20
changed the comment in Msg class, and added changelog

StevenXLi:


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4048-update-comment-for-msg-code-to-use-wiki.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: add
+issue: 4048
+title: "Changed the comment in Msg class to point to the wiki when new code are added, instead of updating in the comment itself."
```
TipzCM, line 4: I don't think we need a changelog for this.

It's internal info only. And just a comment. It would be confusing to the users otherwise.

hapi-fhir-base/src/main/java/ca/uhn/fhir/i18n/Msg.java
```diff
@@ -24,8 +24,8 @@ public final class Msg {
 	private static final String ERROR_CODE_PREFIX = "HAPI";
 
 	/**
-	 * IMPORTANT: Please update the following comment after you add a new code
-	 * Last used code value: 2140
+	 * IMPORTANT: Please update the table in the following link after you add a new code
+	 * https://gitlab.com/simpatico.ai/cdr/-/wikis/Bumping-Error-Code
```
TipzCM, line 28: Can we also include the number from before? or is that too complex?

TipzCM (changes requested):

hapi-fhir-base/src/main/java/ca/uhn/fhir/i18n/Msg.java
StevenXLi: That's the whole point of the new process, it doesn't make sense to keep it in 2 places. Also, updating codes in 2 different places can introduce desync

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4048-update-comment-for-msg-code-to-use-wiki.yaml
StevenXLi: I thought all MR's need a change log, is there documentation on the cases like this, where it's an exception

StevenXLi: removed

TipzCM (approved):

Issue #4051, opened 2022-09-21
Bulk Export Inconsistencies


tadgh:
Two issues need rectifying: 

1. After a bulk export, attempting to read the resulting binary while using an `Accept` header of `application/fhir+ndjson` should stream the data back to the client as NDJSON. Currently, it returns an XML binary resource. If the Accept header is omitted, currently the results are streamed back as JSON, not NDJSON. 

2. `$export-poll-status` is not currently sufficient. According to the HL7 [specification](http://hl7.org/fhir/uv/bulkdata/2021May/export.html#response---complete-status%7Chttp://hl7.org/fhir/uv/bulkdata/2021May/export.html%23response---complete-status%7Csmart-link) for bulk export,  the poll response for a completed Bulk Export job is required to include four elements:
       
      {{transactionTime}} 
      {{request}}
      {{requiresAccessToken}}
      {{output}}

Currently HAPI-FHIR only returns {{transactionTime}} and {{output}}.

PR #4053, opened 2022-09-21
Fix content negotiation on binary streaming

tadgh:
Works on #4051 

* Fixes content-type negotiation when using `Accept` header.
* Add test
* Add changelog

Now, when using an `Accept` header that resolves to the same content-type as the binary, this will cause the binary contents to be streamed out, instead of an XML representation of the binary being returned.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4051-content-type-negotiation-binaries.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4051-content-type-negotiation-binaries.yaml
@@ -2,6 +2,6 @@
 type: fix
 issue: 4051
 jira: SMILE-5170
-title: "There was previously a bug in content-type negotiation when reading Binary resources. Previously, if a client requested a Binary and used an `Accept` 
-header that matched the Binary resource's `contentType`, the server would return an XML representation of the Binary resource. This has been fixed, and using a matching `Accept` header will cause
-the Binary's data to be returned directly in the requested content type."
+title: "There was a bug in content-type negotiation when reading Binary resources. Previously, when a client requested a Binary resource and with an `Accept` 
+header that matched the `contentType` of the stored resource, the server would return an XML representation of the Binary resource. This has been fixed, and a request with a matching `Accept` header will receive
+the stored binary data directly as the requested content type."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4051-content-type-negotiation-binaries.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4051-content-type-negotiation-binaries.yaml
@@ -1,0 +1,7 @@
+---
+type: fix
+issue: 4051
+jira: SMILE-5170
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -52,7 +61,86 @@ public class BulkExportUseCaseTest extends BaseResourceProviderR4Test {
 
 	@Nested
 	public class SystemBulkExportTests {
-		//TODO add use case tests here
+		@Test
+		public void testBinariesAreStreamedWithRespectToAcceptHeader() throws IOException {
```
michaelabuckley, line 65: Good test.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4051-content-type-negotiation-binaries.yaml
```diff
@@ -0,0 +1,7 @@
+---
+type: fix
+issue: 4051
+jira: SMILE-5170
+title: "There was previously a bug in content-type negotiation when reading Binary resources. Previously, if a client requested a Binary and used an `Accept` 
+header that matched the Binary resource's `contentType`, the server would return an XML representation of the Binary resource. This has been fixed, and using a matching `Accept` header will cause
+the Binary's data to be returned directly in the requested content type."
```
michaelabuckley, line 7:
```suggestion
title: "There was a bug in content-type negotiation when reading Binary resources. Previously, when a client requested a Binary resource and with an `Accept` 
header that matched the `contentType` of the stored resource, the server would return an XML representation of the Binary resource. This has been fixed, and a request with a matching `Accept` header will receive
the stored binary data directly as the requested content type."
```

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/RestfulServerUtils.java
```diff
@@ -1037,6 +1038,29 @@ public static Object streamResponseAsResource(IRestfulServerDefaults theServer,
...
+	 * - If they request octet-stream, return true;
+	 * - If the content-type happens to be a match, return true.
+	 * - Construct an EncodingEnum out of the contentType. If this matches the responseEncoding, return true.
+	 * - Otherwise, return false.
+	 *
+	 * @param theResponseEncoding the requested {@link EncodingEnum} determined by the incoming Content-Type header.
+	 * @param theBinary  the {@link IBaseBinary} resource to be streamed out.
+	 * @return True if response can be streamed as the requested encoding type, false otherwise.
+	 */
+	private static boolean matchesBinaryContentType(ResponseEncoding theResponseEncoding, IBaseBinary theBinary) {
```
michaelabuckley, line 1052: Any chance theResponseEncoding is null?

michaelabuckley (approved):
Approved with minor wording suggestion and question.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/RestfulServerUtils.java
tadgh: yup, but that condition is handled by the conditional calling it. 

On second thought, i'll collapse that functionality into this method.

PR #4055, opened 2022-09-21
Improve `$poll-export-status` to include all required attributes. 

### LINKED ISSUE #4051: Bulk Export Inconsistencies

tadgh:
Closes #4051

* Add all needed attributes to the bulk export polling response when it is completed. 
* Set `requiresAccessToken` to be true. 
* Set `request` to be the originating request URL which caused this bulk export to occur. This necessitated modifying the contents of the BulkExportJobParameters so it could be propagated all the way to the Report Writer step. 
* Add test.

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
@@ -4,6 +4,8 @@ import ca.uhn.fhir.jpa.api.config.DaoConfig;
 import ca.uhn.fhir.jpa.api.model.BulkExportJobResults;
 import ca.uhn.fhir.jpa.api.svc.IBatch2JobRunner;
 import ca.uhn.fhir.jpa.batch.models.Batch2JobStartResponse;
+import ca.uhn.fhir.jpa.bulk.export.model.BulkExportResponseJson;
+import ca.uhn.fhir.jpa.model.util.JpaConstants;
 import ca.uhn.fhir.jpa.provider.r4.BaseResourceProviderR4Test;
 import ca.uhn.fhir.jpa.util.BulkExportUtils;
 import ca.uhn.fhir.parser.IParser;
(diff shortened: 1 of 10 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -59,6 +63,42 @@ public class BulkExportUseCaseTest extends BaseResourceProviderR4Test {
...
+			myBatch2JobHelper.awaitJobCompletion(jobId);
+
+			//Then: When the poll shows as complete, all attributes should be filled.
+			HttpGet statusGet = new HttpGet(pollingLocation);
+			try (CloseableHttpResponse status = ourHttpClient.execute(statusGet)) {
+				String responseContent = IOUtils.toString(status.getEntity().getContent(), StandardCharsets.UTF_8);
+				ourLog.info(responseContent);
+
+				BulkExportResponseJson result = JsonUtil.deserialize(responseContent, BulkExportResponseJson.class);
+				assertThat(result.getRequest(), is(equalTo(myClient.getServerBase() + "/$export?_type=Patient")));
```
fil512, line 94: nitpick: when a test asserts a response contains a string from the request, best practice is to put it in a variable

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/bulk/export/provider/BulkDataExportProvider.java
```diff
@@ -282,16 +287,18 @@ public void exportPollStatus(
 					BulkExportResponseJson bulkResponseDocument = new BulkExportResponseJson();
 					bulkResponseDocument.setTransactionTime(info.getEndTime()); // completed
 
+					bulkResponseDocument.setRequiresAccessToken(true);
```
fil512, line 290: I see now why you were asking about this.  If we're not enforcing this, how does it affect anything setting this to true?  Or is this just informative for audit purposes?  Anyway, I assume it's for compliance for now, but post-merge I would like to understand what this means.

fil512 (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
tadgh: Fixed, good call.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/bulk/export/provider/BulkDataExportProvider.java
tadgh: ok cool 👍🏻

PR #4063, opened 2022-09-23
Fix transactional issue in everything operation

tadgh:
* Add transactional annotation to patient instance and type everything. 
* Remove runInTransaction from test to allow internal transactional semantics to work.


jmarchionatto (approved):
Looks good

PR #4066, opened 2022-09-23
Cleanup and utilities from composite work

michaelabuckley:
Small changes from #3839 to reduce the patch size.


hapi-fhir-base/src/main/java/ca/uhn/fhir/util/ObjectUtil.java
```diff
@@ -25,14 +28,12 @@
 
 public class ObjectUtil {
 
+	/**
+	 * Just use Objects.equals() instead;
```
jamesagnew, line 32:
```suggestion
	 * @deprecated Just use Objects.equals() instead;
```

I think errorprone will complain if you use the `@Deprecated` annotation without a corresponding `@deprecated` javadoc (and I think that's a good idea anyhow)

hapi-fhir-jpaserver-test-utilities/src/main/resources/logback-test.xml
```diff
@@ -5,71 +5,42 @@
...
+	</root>
 
-	<logger name="ca.uhn.fhir.rest.client" additivity="false" level="info">
-		<appender-ref ref="STDOUT" />
-	</logger>
-
-	<logger name="ca.uhn.fhir.jpa.dao" additivity="false" level="info">
-		<appender-ref ref="STDOUT" />
-	</logger>
+	<logger name="ca.uhn.fhir.jpa.subscription.match.matcher.subscriber.SubscriptionMatchingSubscriber" level="info"/>
```
jamesagnew, line 13: What does this do? I'd assume with no appenders it means to suppress these messages? Is that the desired intent? If so, recommend documenting that in a comment here.. but maybe i've misunderstood.

jamesagnew (approved):

hapi-fhir-jpaserver-test-utilities/src/main/resources/logback-test.xml
michaelabuckley: No.  I'm fixing our misuse of Logback.  By moving the root to the top, they all inherit the appender.  We needed all this noise because the root wasn't declared until the end.

PR #4070, opened 2022-09-23
Android compatibility checks

vitorpamplona:
Adding a maven plugin to check the compatibility of given modules with Android APIs: 
- base
- structures
- validation
- converters

The idea is to declare compliance with Android 8 and up. The plugin runs through the source code to check if all classes and all methods used are available in that version of Android. Android 8 uses SDK 26 and was released in 2017. This should give us good coverage without blocking the use of newer Java APIs. 

Run with:
- `./mvnw animal-sniffer:check`


jamesagnew:
@vitorpamplona does this work for you?

I just checked it out, and when I run it I see the following error:

> [ERROR] Failed to execute goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.22:check (default-cli) on project hapi-fhir-server: Execution default-cli of goal org.codehaus.mojo:animal-sniffer-maven-plugin:1.22:check failed: PermittedSubclasses requires ASM9 -> [Help 1]

I thought maybe it was an issue because I'm running on JDK17 but I see the same error if I downgrade to JDK11. Bumping the ASM dependency to the animal sniffer plugin didn't help unfortunately.

jamesagnew:
Weird! What does `mvn -v` give you in terms of an environment? I'm on:

```
❯ mvn -v
Apache Maven 3.8.6 (84538c9988a25aec085021c365c560670ad80f63)
Maven home: /opt/homebrew/Cellar/maven/3.8.6/libexec
Java version: 11.0.17, vendor: Amazon.com Inc., runtime: /Users/james/.sdkman/candidates/java/11.0.17-amzn
Default locale: en_CA, platform encoding: UTF-8
OS name: "mac os x", version: "13.0", arch: "aarch64", family: "mac"
```

vitorpamplona:
```
➜  ./mvnw -v                  
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /Users/vitor/.m2/wrapper/dists/apache-maven-3.6.3-bin/1iopthnavndlasol9gbrbg6bf2/apache-maven-3.6.3
Java version: 17.0.5, vendor: Homebrew, runtime: /usr/local/Cellar/openjdk@17/17.0.5/libexec/openjdk.jdk/Contents/Home
Default locale: en_US, platform encoding: UTF-8
OS name: "mac os x", version: "11.6.7", arch: "x86_64", family: "mac"
```

vitorpamplona:
Maybe this is an M1/M2 issue? I am still on intel.

jamesagnew:
I wonder... I just got this new M2 MacBook. It's crazy fast, but definitely some things don't work well on it.

I'll try on my home linux/x86_64 machine tonight.

vitorpamplona:
Good to know... I am about to buy one :)

jamesagnew:
Well, sure enough, this works just fine on my Linux box.

What a weird side effect.

I'm definitely good to merge this, although it should probably be moved to the `CI` profile in the POM so that it doesn't run by default (most of us on the project are on Macs). We'd also actually need to tackle the issues causing the failure before merging it. Would you be up for moving it into the CI profile, then doing any one of the following?

* Set the plugin to warn but not fail and we can fix the API issues in a separate PR
* Make me a contributor to your fork and I'll fix the API issues
* You take a stab at fixing the API issues

PS - Despite this wacky behaviour, I highly recommend the M2! It cuts my build times in half compared to my 4 year old MBP, and the battery life is... well, i have trouble believing it. I worked away in IJ and watched a webinar yesterday for an hour and only used about 10%.

JPercival:
@vitorpamplona - Can you merge the latest from `master` and revalidate?

vitorpamplona:
Good to go. But I can never get CI to pass :(

vitorpamplona:
Now we are a go! :) @JPercival

JPercival (approved):

tadgh (approved):

PR #4085, opened 2022-09-26
4084 error segment of bulk export job poll status is missing

tadgh:
* Change serialization rules for error field, ensure it is always populated. 
* Add Changelog
* Add Test


IanMMarshall (approved):
Looks good.

PR #4086, opened 2022-09-27
Fix hashCode() to reflect the equals() contract by passing in getValu…

lukedegruchy:
Closes [4087](https://github.com/hapifhir/hapi-fhir/issues/4087)

* Fix hashCode() and equals() in ResourceIndexedSearchParamNumber to have the same contract, namely, for both to include value and hashNumber
* Fix bug in DaoSearchParamSynchronizer where the removed count is set as the added count in AddRemoveCount
* Add unit tests
* Fix changelog from previous MR that referred to the wrong issue number


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoSearchParamSynchronizer.java
```diff
@@ -91,7 +91,8 @@ private <T extends BaseResourceIndex> void synchronize(ResourceTable theEntity,
 			myEntityManager.merge(next);
 		}
 
-		theAddRemoveCount.addToAddCount(paramsToRemove.size());
+		// TODO:  are there any unintended consequences to fixing this bug?
```
jamesagnew, line 94: This only gets used for log messages, so fixing this should not cause any issues. Well spotted. :)

jamesagnew (approved):

Issue #4080, opened 2022-09-26
Support ONC requirements for Group Bulk Export


StevenXLi:
**Describe the issue**
ONC requirements mentioned [here](https://www.healthit.gov/test-method/standardized-api-patient-and-population-services#ccg), require that HAPI-FHIR be able to include Practitioner, Organization, Device and Location resources as part of the responses for requests for multiple patient Group Bulk Export. The resources are referenced within the resources that is a part of the patient compartment.

**Environment (please complete the following information):**
 - HAPI FHIR Version 6.2.0
 - OS: MacOS
 - Browser: chrome

sjayswal:
https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/446
its not resolved yet in 6.2.0 pre18.

PR #4093, opened 2022-09-28
Make sure to not mix resource types in bulk export

jamesagnew:
The fix for #4080 resulted in bulk export files with mixed resource types. This isn't allowed.


dmuylwyk (approved):
LGTM, thanks!

Issue #2979, opened 2021-09-09
If-Match does not specify the weak ETag attribute "W/"


Romanow88:
When using the generic client for an update, assuming communicating to a versioned server, the "If-Match" header is set to the version of the respource that is being updated, but the required "W/" indicating a weak ETag is omitted.  Relevant part of the standard: https://hl7.org/fhir/r4/http.html#versioning

**To Reproduce**
1. Setup a versioned server that has strict "If-Match" handling
2. Query for a resource using the generic client
3. Change any part of the resource and try to do an update using the generic client.
4. Observe server rejecting the update 

**Expected behavior**
The client should send a weak ETag (prepended with "W/") as per the standard

**Environment (please complete the following information):**
 - HAPI FHIR Version: 5.4.2
 - OS: Windows
 - JRE: 11

Relevant code: https://github.com/hapifhir/hapi-fhir/blob/d1c2d839d187b0820e3c38658a2fef889a908dd3/hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/method/MethodUtil.java#L208

berkant-k:
I create PR #4129 to solve this issue.

ogg:
This is still an issue. I currently have to use an `IClientInterceptor` to rewrite the header.

cspielmann:
Same issue here. HAPI client version 6.6.0

jamesagnew:
#4129 has merged.

PR #4129, opened 2022-10-10
Version aware update should use weak etag

berkant-k:
The value from .meta.versionId should be  "weak" ETag, prefixed with W/ and enclosed in quotes (e.g., W/"3141").
(Documented here: https://www.rfc-editor.org/rfc/rfc7232#section-2.3 and here https://build.fhir.org/http.html

This PR closes issue #2979


jamesagnew (approved):
Thanks for the contribution, and sorry it took so long to address. Merging this shortly.

PR #4134, opened 2022-10-12
fix generate-multi-file and generate-single-file goals with R5

bdenton:
The tinder-plugin does not work for R5 and the generic generate-multi-file and generate-single-file Mojo plugins. This PR fixes that and adds more tests to insure the generic plugins continue to work for all supported FHIR versions.


jamesagnew (approved):
Thanks for the PR! Merging now.

PR #4135, opened 2022-10-13
Issue 4052 addition of properties to the loinc terminology uploading process

jmarchionatto:


hapi-fhir-jpaserver-base/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
jamesagnew, line 1: Out of curiosity, what is this?

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
jmarchionatto: It is the configuration for Mockito to be able to mock final classes, statics, etc. 
https://stackoverflow.com/questions/14292863/how-to-mock-a-final-class-with-mockito

PR #4144, opened 2022-10-14
4140-upload-terminology-does-not-allow-snomed-ct-canada-edition-zip-file-to-load-due-to-a-different-file-name

karneet1212:


jmarchionatto (approved):

jmarchionatto:
Looks good. Approved pending a small detail.

PR #4155, opened 2022-10-18
allow disabling of batch job reuse

JasonRoberts-smile:


tadgh (approved):
Looks good overall if its needed.  This is the intent of the no-cache in the first place. Can the caller not just use that param?

PR #4157, opened 2022-10-19
wip migration lock attempt 1

fil512:
Lock the migration table when performing migrations to support migrating in a cluster


fil512:
Likely abandoning this branch

PR #4174, opened 2022-10-20
Change to a minimal first pre-fetch to avoid latency on the async path.

### LINKED ISSUE #3664: JPA async path is slower now that is synchronous
**Describe the bug**
The JPA persistence provider became slower when the async cached query threadpool was removed.

## To Reproduce
1. Install 5.3
2. Load some synthea data
3. Make a query that returns more that 500 results. E.g. /Observation?status=active
4. Upgrade to 6.0
5. Compare results
6. The new server is much slower.  E.g. 200ms vs 30ms.

## Expected behaviour
Sub 50ms response

 - HAPI FHIR Version 6.0 vs 5.3
 - Postgres

michaelabuckley:
Fixes #3664 

The "async" path is now synchronous, and was causing latency for first page by loading 500 results. We avoid pre-fetch on the first page, and only pre-fetch after users fetch the second page.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/3664-shorten-first-prefetch.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/3664-shorten-first-prefetch.yaml
@@ -1,0 +1,4 @@
+---
+type: fix
+issue: 3664
+title: "The first JPA prefetch window was shortened to avoid slow first bundles."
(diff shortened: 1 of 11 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/3664-shorten-first-prefetch.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 3664
+title: "The first JPA prefetch window was shortened to avoid slow first bundles."
```
jamesagnew, line 4: Suggested slightly more descriptive changelog

```suggestion
title: "Initial page loading has been optimized to reduce the number of prefetched resources. This should improve the speed of initial search queries in many cases."
```

jamesagnew (approved):

PR #4177, opened 2022-10-21
Add support for clustered database migrations

fil512:
Lock the database so only one server migrates at once


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4177-migration-lock.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 4177
+title: "Support has been added for clustered database upgrades.  Only one database migration is permitted at once."
```
jamesagnew, line 4:
```suggestion
title: "Support has been added for clustered database upgrades in the new (non-FlyWay) migrator codebase.  Only one database migration is permitted at once."
```

TipzCM (approved):

PR #4195, opened 2022-10-24
Fix ForcedId criteria query

jmarchionatto:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -351,7 +351,7 @@ private void doResolvePersistentIds(RequestPartitionId theRequestPartitionId, Li
 				andPredicates.add(typeCriteria);
 			}
 
-			Predicate idCriteria = cb.equal(from.get("myForcedId").as(String.class), next.getIdPart());
+			Predicate idCriteria = cb.equal(from.get("myResourcePid").as(String.class), next.getIdPart());
```
jamesagnew, line 354: This is the wrong fix - That logic is definitely wrong. Looking up a forced ID by its PID is the opposite of how this should work.

I just ran the test locally and put a breakpoint on this line. This method is being called do resolve `Patient/1` but the patient resource's ID isn't `Patient/1`, it's `Patient/PING1`. 1 is the PID and should not be placed into `theIds`.

You need to address why the wrong ID type is being passed in.

jamesagnew (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
jamesagnew: My guess is that `getMembersFromGroupWithFilter` is the culprit. It returns PIDs, so you don't need to be resolving the IDs from that method, since they are already PIDs.

tadgh: Think I figured this out.... the _old_ implementation just read out from the member list of the group resource, which would occasionally require us to convert from forcedid to PID. However on a subsequent rework we did a query for group membership via a _has query, and the searchbuilder used for this would return the resourcepersistentid objects. It appears as though in ALPHANUMERIC mode, the IdHelperService simply realizes that the PIDs are already PIDs,and silently succeeds. I think we can safely remove the IdHelperService call instead. I'll tackle this.

tadgh: @jamesagnew can you have a look now? I've reworked it to primarily use ResourcePersistentId across the board, and only resolve that into a fhir resource id when actually needed for a query

jamesagnew (approved):

PR #4209, opened 2022-10-27
4207-getpagesoffset-set-to-total-number-of-resources-results-in-inconsistent-amount-of-entries-when-requests-are-sent-consecutively

### LINKED ISSUE #4207: _getpagesoffset set to total number of resources results in inconsistent amount of entries when requests are sent consecutively
When _getpagesoffset is set to the total number of resources in a GET request, the first time the request is sent, no entries are visible in the response (the expected response). If another request is sent not long after the first one is complete, it returns one entry, which should be the response for _getpagesoffset = {number of resources - 1}.

Steps to Reproduce
1. Post 10 resources
2. Send a GET request with _getpagesoffset = 10

http://localhost:8000/<Resource>?_getpagesoffset=10&_count=50&_pretty=true

3. Observe that the response shows no resource entries
4. Send the same GET request not long after the first one
5. Observe that there is one resource entry

Expected Results
Regardless of the amount of consecutive requests sent, the response should always show 0 resource entries.

karneet1212:
Fixed the way the offset is calculated when offset is more than or equal to the total number of resources available

Closes #4207

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4207-getpagesoffset-set-to-total-number-of-resources-results-in-inconsistent-amount-of-entries-when-requests-are-sent-consecutively.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4207-getpagesoffset-set-to-total-number-of-resources-results-in-inconsistent-amount-of-entries-when-requests-are-sent-consecutively.yaml
@@ -1,6 +1,6 @@
 ---
 type: fix
 issue: 4207
-title: "Previously to improve performance, if the total number of resources was less than the _getpagesoffset given, it would default to 
-         the offset for the last resource available. This would result in one entry being displayed in some requests, especially when they are sent consecutively,
-          even when there should be none. This has been fixed."
+title:  "Previously to improve performance, if the total number of resources was less than the _getpageoffset, 
+      the results would default to last resource offset. This is especially evident when requests are consecutive 
+      resulting in one entry being displayed in some requests. This issue is now fixed."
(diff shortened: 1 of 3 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4207-getpagesoffset-set-to-total-number-of-resources-results-in-inconsistent-amount-of-entries-when-requests-are-sent-consecutively.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 4207
+title: "Previously to improve performance, if the total number of resources was less than the _getpagesoffset given, it would default to
```
jaison-b, line 4: Minor nitpick to fix: Should read "Previously to improve performance, if the total number of resources was less than the _getpageoffset, the results would default to last resource offset. This is especially evident when requests are consecutive resulting in one entry being displayed in some requests. This issue is now fixed.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
```diff
@@ -375,11 +375,9 @@ public IBaseResource doInvokeServer(IRestfulServer<?> theServer, RequestDetails
 					}
 
 					Integer resultSize = result.size();
-					int start;
+					int start = offsetI;
```
jaison-b, line 378: Nitpick: why the `offsetI` what is the 'I' at the end? does the suffix of 'I' have any meaning? It confuses me when I read it. Not an issue just a very minor nitpick.

jaison-b (approved):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/BaseResourceReturningMethodBinding.java
karneet1212: I believe the 'I' meant index. However, I agree that it can be confusing, so I have switched it to being just offset.

PR #4213, opened 2022-10-28
jm wrong bundle entry url

jmarchionatto:


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4210-wrong-bundle-entry-uri-in-transaction-bundle.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 4210
+jira: SMILE-4685
+title: "Generating Bundle with resources was setting `entry.request.url` as absolute url when it should be relative. This has been fixed"
```
tadgh, line 5: suggestion: indicate that this is specifically for `withResources` calls to the client in DSTU3.

tadgh (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_2_0/4210-wrong-bundle-entry-uri-in-transaction-bundle.yaml
jmarchionatto: But it is not. Had to do the same fix for R4, R4b, and R5 also

PR #4223, opened 2022-10-31
Fix some issues with 'mvn test' on windows

bdenton:
This PR addresses three issues with running Maven test on Windows workstations:
1. Different line ends cause some multi-line string compares to fail
2. Different path separator character causes one test to fail
3. Many tests don't wait long enough for things to finish. Introduced a little more patience into the tests


jamesagnew (approved):

jamesagnew:
Thanks Bill!

Issue #3942, opened 2022-08-22
OutOfMemoryException due to endless loop in SnapshotGeneratingValidationSupport when havinig circular dependency


joergreichert:
**Describe the bug**
When there is a circular dependency within a structure definition the structure definitions are traversed again and again finally resulting in an OutOfMemoryException. 

**To Reproduce**
See minimal example JUnit test in Gist and comments for project setup and output: https://gist.github.com/joergreichert/8d0c714cfb2d3c5a7eef1bdc0e9be097

**Expected behavior**
The validator should stop traversing resources already loaded.

**Environment (please complete the following information):**
 - HAPI FHIR Version: 6.0.1

ABDA-FHIR:
Proposed pull request to fix this [https://github.com/hapifhir/hapi-fhir/pull/4228]

patrick-werner:
i couldn't reproduce the issue with hapi 8.0.0. @joergreichert

PR #4228, opened 2022-11-02
Fix circular dependency in SnapshotGeneratingValidationSupport

ABDA-FHIR:
Correct scoping for the finally frame that undoes the addition of the input URL to the 'currently generating snapshots' hash set.

### DIFF
--- a/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/SnapshotGeneratingValidationSupport.java
+++ b/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/SnapshotGeneratingValidationSupport.java
@@ -50,7 +50,6 @@ public class SnapshotGeneratingValidationSupport implements IValidationSupport {
 	@Override
 	public IBaseResource generateSnapshot(ValidationSupportContext theValidationSupportContext, IBaseResource theInput, String theUrl, String theWebUrl, String theProfileName) {
 
-		String inputUrl = null;
 		try {
 			FhirVersionEnum version = theInput.getStructureFhirVersionEnum();
 			assert version == myCtx.getVersion().getVersion();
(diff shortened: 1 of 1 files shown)


hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/SnapshotGeneratingValidationSupport.java
```diff
@@ -60,70 +59,69 @@ public IBaseResource generateSnapshot(ValidationSupportContext theValidationSupp
...
 
-			inputUrl = inputCanonical.getUrl();
+			final String inputUrl = inputCanonical.getUrl();
 			if (theValidationSupportContext.getCurrentlyGeneratingSnapshots().contains(inputUrl)) {
 				ourLog.warn("Detected circular dependency, already generating snapshot for: {}", inputUrl);
 				return theInput;
 			}
 			theValidationSupportContext.getCurrentlyGeneratingSnapshots().add(inputUrl);
 
-			String baseDefinition = inputCanonical.getBaseDefinition();
```
jamesagnew, line 70: It's really hard to tell what is changed here. Could you you please tweak this to not reformat the whole file?

jamesagnew (changes requested):

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/SnapshotGeneratingValidationSupport.java
```diff
@@ -50,7 +50,6 @@ public SnapshotGeneratingValidationSupport(FhirContext theCtx) {
 	@Override
 	public IBaseResource generateSnapshot(ValidationSupportContext theValidationSupportContext, IBaseResource theInput, String theUrl, String theWebUrl, String theProfileName) {
```
jamesagnew, line 51: Can you please add a unit test that exercizes this fix?

ABDA-FHIR: coming soon....
...until then https://github.com/hapifhir/hapi-fhir/issues/3942 -> https://gist.github.com/joergreichert/8d0c714cfb2d3c5a7eef1bdc0e9be097

PR #4249, opened 2022-11-04
bulk export permanently reusing cached results

### LINKED ISSUE #4247: Bulk Export permanently reusing cached results.
Steps to reproduce: 
1. Submit a Bulk Export request
2. Wait for it to complete. 
3. Submit a new identical Bulk Export Request

Observed results: 
- the already completed job is returned. 

Expected results: 
- a new job is started. 

Existing job should only be returned if it is not yet complete.

tadgh:
* Fix bug where we were checking completed jobs as cache candidates 
* Add changelog
* Add test. 

Closes #4247


fil512 (approved):
I assume there is already some kind of test that validates in-progress / queued jobs are reused?

tadgh:
> 

You make a good point, there is not. That is the current behaviour, will roll this into a subsequent PR

PR #4252, opened 2022-11-04
Consolidate all System Properties into a single class

fil512:
Actually ones only read by tests are in a separate class


hapi-fhir-cli/hapi-fhir-cli-api/src/main/java/ca/uhn/fhir/cli/RunServerCommand.java
```diff
@@ -94,11 +97,6 @@ public void run(CommandLine theCommandLine) throws ParseException {
 
 		myPort = parseOptionInteger(theCommandLine, OPTION_P, DEFAULT_PORT);
 		
-		if (theCommandLine.hasOption(OPTION_LOWMEM)) {
-			ourLog.info("Running in low memory mode, some features disabled");
-			System.setProperty(OPTION_LOWMEM, OPTION_LOWMEM);
-		}
-
```
StevenXLi, line 101: is this feature no longer required?

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/system/HapiTestSystemProperties.java
```diff
@@ -0,0 +1,54 @@
+package ca.uhn.fhir.system;
```
StevenXLi, line 1: Is there a reason to separate out the operations to two different classes? I feel since all the operation in both has some relation to tests, keeping them in two places makes it harder for other devs to find them

hapi-fhir-cli/hapi-fhir-cli-api/src/main/java/ca/uhn/fhir/cli/RunServerCommand.java
fil512: correct.  James responded to my question about this on slack.

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/system/HapiTestSystemProperties.java
fil512: That's a really good question.  I went back and forth on it.  The prod ones actually affect prod behaviour and it felt important to separate those from the ones that only touch test code.

hapi-fhir-base/src/main/java/ca/uhn/fhir/system/HapiSystemProperties.java
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	/**
+	 * This property is used by unit tests - do not rely on it in production code
+	 * as it may change at any time. If you want to capture responses in a reliable
+	 * way in your own code, just use client interceptors
+	 */
+	public static void enableHapiClientKeepResponses() {
+		System.setProperty(HAPI_CLIENT_KEEPRESPONSES, "true");
+	}
```
tadgh, line 29: suggestion: if you have an enabler there should probably be an equivalent disabler.

tadgh (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/system/HapiSystemProperties.java
fil512: ty

PR #4270, opened 2022-11-10
4269 implement a toString for OkHttpRestfulRequest to have better log…

Roel-Scholten:
…gin…

Implement a toString for OkHttpRestfulRequest logging to have better logging. Currently this still differs from the ApacheHttpRequest.java as we do not have the used protocol (http/1.1/alpn) on request creation. This is added to the Response object.


jamesagnew (approved):
Appreciate the contribution! Looks great, will merge when it passes CI.

PR #4275, opened 2022-11-11
Adding okio dependency for r4b structures for running DIST profile

markiantorno:
DIST profile fails with error cannot resolve dependency okio.


hapi-fhir-structures-r4b/pom.xml
```diff
@@ -15,6 +15,10 @@
 	<name>HAPI FHIR Structures - FHIR r4b</name>
 
 	<dependencies>
+		<dependency>
+			<groupId>com.squareup.okhttp3</groupId>
```
jamesagnew, line 19: This should be marked as `<optional>true</optional>` - Most users of the structure don't need okhttp

jamesagnew (approved):

PR #4283, opened 2022-11-15
bundle not throwing an error when multiple matches were found

karneet1212:
Adding a resolved match url now checks if multiple persistent ids are found for a conditional URL

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
@@ -20,9 +20,11 @@ package ca.uhn.fhir.rest.api.server.storage;
  * #L%
  */
 
+import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.interceptor.api.HookParams;
 import ca.uhn.fhir.interceptor.api.Pointcut;
 import ca.uhn.fhir.rest.api.InterceptorInvocationTimingEnum;
+import ca.uhn.fhir.rest.server.exceptions.PreconditionFailedException;
 import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.ListMultimap;
 import org.apache.commons.lang3.Validate;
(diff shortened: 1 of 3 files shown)


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
```diff
@@ -162,10 +164,22 @@ public void addResolvedMatchUrl(String theConditionalUrl, @Nonnull ResourcePersi
...
 		myResolvedMatchUrls.put(theConditionalUrl, thePersistentId);
 	}
 
+	private boolean matchUrlWitDiffIdExists(String theConditionalUrl, @Nonnull ResourcePersistentId thePersistentId) {
+		if (myResolvedMatchUrls.containsKey(theConditionalUrl) && myResolvedMatchUrls.get(theConditionalUrl) != NOT_FOUND) {
+			if (myResolvedMatchUrls.get(theConditionalUrl).getId() != thePersistentId.getId()) {
+				return true;
+			}
+		}
+		return false;
```
JasonRoberts-smile, line 180: nitpick: Stylistically, this is inelegant.

In general,
`
if(x) {
  if(y) {
    //do stuff
  }
}
`
can be simplified to
`
if (x && y) {
  // do stuff
}
`
and
`
if (x) {
  return true;
}
return false;
`
can be simplified to
`
return x;
`

I find the extra syntax makes the code harder to read and maintain.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
```diff
@@ -162,10 +164,22 @@ public void addResolvedMatchUrl(String theConditionalUrl, @Nonnull ResourcePersi
...
 		if (myResolvedMatchUrls.isEmpty()) {
 			myResolvedMatchUrls = new HashMap<>();
+		} else if (matchUrlWitDiffIdExists(theConditionalUrl, thePersistentId)) {
+			String msg = "Invalid match URL " + theConditionalUrl + " - Multiple resources match this search";
+			throw new PreconditionFailedException(Msg.code(2207) + msg);
 		}
 		myResolvedMatchUrls.put(theConditionalUrl, thePersistentId);
 	}
 
+	private boolean matchUrlWitDiffIdExists(String theConditionalUrl, @Nonnull ResourcePersistentId thePersistentId) {
```
JasonRoberts-smile, line 174: nitpick: There is a type in the method name - Wit instead of With.

JasonRoberts-smile (commented):
You should probably update the documentation to describe this new constraint on the use of conditional creates.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
JasonRoberts-smile: And there is a typo in my comment, so what goes around comes around

PR #4296, opened 2022-11-22
Points deprecated retry constants to new values

nathandoef:


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/UserRequestRetryVersionConflictsInterceptor.java
```diff
@@ -49,15 +49,15 @@ public class UserRequestRetryVersionConflictsInterceptor {
 
 	/** Deprecated and moved to {@link ca.uhn.fhir.rest.api.Constants#HEADER_RETRY_ON_VERSION_CONFLICT} */
 	@Deprecated
-	public static final String HEADER_NAME = "X-Retry-On-Version-Conflict";
+	public static final String HEADER_NAME = Constants.HEADER_RETRY_ON_VERSION_CONFLICT;
```
fil512, line 52: I think these should be in ProviderConstants

fil512: But if they've been in Constants for a while now, we should leave a deprecated breadcrumb there (which is unfortunate because that makes 2 breadcrumbs now :-()

fil512 (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/UserRequestRetryVersionConflictsInterceptor.java
nathandoef: Just tried to update `Constants` to point to `ProviderConstants`, but it’s resulting in a circular dependency when I import `ProviderConstants`

Issue #3885, opened 2022-08-05
HIGH Security Vulnerability in hapi-fhir-test-utilities


joepaquette:
hapi-fhir-test-utilities has a dependency on `net.sourceforge.htmlunit` which has a dependency on `xalan:xalan`.

`xalan:xalan:2.7.2` has a HIGH security vulnerability and was last updated on July 24, **2014**. [see](https://mvnrepository.com/artifact/xalan/xalan/2.7.2)

Results from a WHITESOURCE security scan:
- Artifact ID: xalan
- Group ID: xalan
- Library Version: 2.7.2
- Library Path: /root/.gradle/caches/modules-2/files-2.1/xalan/xalan/2.7.2/d55d3f02a56ec4c25695fe67e1334ff8c2ecea23/xalan-2.7.2.jar
- Type: MAVEN_ARTIFACT
- Description: The Apache Xalan Java XSLT library is vulnerable to an integer truncation issue when processing malicious XSLT stylesheets. This can be used to corrupt Java class files generated by the internal XSLTC compiler and execute arbitrary Java bytecode. The Apache Xalan Java project is dormant and in the process of being retired. No future releases of Apache Xalan Java to address this issue are expected. Note: Java runtimes (such as OpenJDK) include repackaged copies of Xalan.
- Suggested Fix: No suggested fix

tadgh:
Heya! Greatly appreciate the report.  The test utilities are test-scoped, and not shipped in production code. Since there seems to be no remediation immediately available, do you have any recommendations? Do you think it would be safe to just remove the dependency?

joepaquette:
Someone should look at [hapi-fhir-server-openapi](https://github.com/hapifhir/hapi-fhir/tree/master/hapi-fhir-server-openapi) because the test utilities are not scoped to test only. That's how we stumbled onto this issue. I'll add a new issue to resolve this.

The scanning tool we use apparently does not make distinctions between "test" and "production" dependencies, so we cannot make use of the test utilities package in our project. 

Since the test utilities has a dependency on HtmlUnit and HtmlUnit has the dependency on Xalan, the only thing I can think of is to find an alternative to HtmlUnit that does not contain a dependency on Xalan. I'm not familiar with these tools, but a quick Google search (htmlunit alternative java) identified https://github.com/dhamaniasad/HeadlessBrowsers with numerous suggestions.

jamesagnew:
[...]

Issue #2716, opened 2021-06-08
Hapi-fhir-server-openapi: Do not rely on hapi-fhir-test-utilities for javax.servlet:javax.servlet-api.


Ro3Ls:
**Describe the issue**
The hapi-fhir-server-openapi jar could be smaller to change the scope from hapi-fhir-test-utilities to test and add the javax.servlet:javax.servlet-api as a dependency.

**Environment (please complete the following information):**
 - HAPI FHIR Version
 - OS: Windows 10
 - Browser: Firefox

**Additional context**
Decreases the size of the hapi-fhir-server that uses hapi-fhir-server-openapi package. This way test-containers-${version} will not be added to final deployment (Size: 11.454kB) and possible other not needed dependencies.

If needed i can provide a PR

PR #4306, opened 2022-11-29
Clean up dependencies

jbonzohln:
Cleans up dependencies so multiple versions of a dependency aren't brought in unnecessarily.
* `swagger-models` `2.1.7` was specified, but `2.1.12` is brought in with `swagger-core`.
* `hapi-fhir-test-utilities` did not have scope=test to allow for the servlet-api classes to be referenced, but depending on `javax.servlet-api` is cleaner and does not bring in all of the test dependencies.
 * `okhttp` was a dependency but wasn't being used

Fixes https://github.com/hapifhir/hapi-fhir/issues/3885
Fixes https://github.com/hapifhir/hapi-fhir/issues/2716

### DIFF
--- a/hapi-fhir-server-openapi/pom.xml
+++ b/hapi-fhir-server-openapi/pom.xml
@@ -35,7 +35,6 @@
 		<dependency>
 			<groupId>io.swagger.core.v3</groupId>
 			<artifactId>swagger-models</artifactId>
-			<version>2.1.7</version>
 		</dependency>
 		<dependency>
 			<groupId>io.swagger.core.v3</groupId>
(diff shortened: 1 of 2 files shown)


tadgh:
Heya, I'll have to test this against our upstream dependencies, and if all looks good, I'll get this merged :)

tadgh (approved):

tadgh:
So, this actually causes problems in javadoc generation (specifically the okhttp dep) . I'm going to revert the okhttp removal, and see if that fixes it.

tadgh:
Could you do me a favour and see if you can successfully run 
```
mvn clean install -Dmaven.test.skip -P DIST
```

? I get an error thrown due to the missing OKhttp dep.

jbonzohln:
@tadgh you're right, I had to add okhttp and apache httpcomponents to hapi-fhir-structures-r4 to get the javadoc step to work.

PR #4314, opened 2022-12-01
4312 add support for member identifier parameter return from $member match operation

StevenXLi:
- added member identifier as a part of the return parameters resource
- changed identifier update on the consent to member identifier value


michaelabuckley (approved):
Nice work.  The one must-fix is the identifier type.

PR #4328, opened 2022-12-05
Enhance OperationOutcomeUtil

JasonRoberts-smile:
Add a method to OperationOutcomeUtil to check for the presence of issues with a given severity


jmarchionatto (approved):
Looks good.

PR #4361, opened 2022-12-12
4360 bulk export questionnaireresponses should get picked when author is not empty

StevenXLi:
- added new logic to make a second query to check for resources that only reference to the patient through author or performer

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
@@ -59,7 +59,7 @@ public class SearchParameterUtil {
 	 * Given the resource type, fetch its patient-based search parameter name
 	 * 1. Attempt to find one called 'patient'
 	 * 2. If that fails, find one called 'subject'
-	 * 3. If that fails, find find by Patient Compartment.
+	 * 3. If that fails, find one by Patient Compartment.
 	 * 3.1 If that returns >1 result, throw an error
 	 * 3.2 If that returns 1 result, return it
 	 */
@@ -72,6 +72,22 @@ public class SearchParameterUtil {
 			if (myPatientSearchParam == null) {
 				myPatientSearchParam = getOnlyPatientCompartmentRuntimeSearchParam(runtimeResourceDefinition);
 			}
+		}
+		return Optional.ofNullable(myPatientSearchParam);
+	}
+
+	/**
+	 * Given the resource type, fetch its secondary patient-based search parameter name that's different from what's found in getOnlyPatientSearchParamForResourceType()
+	 * 1. Attempt to find one called 'author'
+	 * 2. If that fails, find one called 'performer'
+	 * This can support more search param, but for now these are the only 2 supported
+	 */
+	public static Optional<RuntimeSearchParam> getOtherPatientSearchParamForResourceType(FhirContext theFhirContext, String theResourceType) {
+		RuntimeSearchParam myPatientSearchParam;
+		RuntimeResourceDefinition runtimeResourceDefinition = theFhirContext.getResourceDefinition(theResourceType);
+		myPatientSearchParam = runtimeResourceDefinition.getSearchParam("author");
+		if (myPatientSearchParam == null) {
+			myPatientSearchParam = runtimeResourceDefinition.getSearchParam("performer");
 		}
 		return Optional.ofNullable(myPatientSearchParam);
 	}
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
@@ -82,7 +82,16 @@ public class SearchParameterUtil {
 	public static Set<String> getPatientSearchParamsForResourceType(FhirContext theFhirContext, String theResourceType) {
 		RuntimeResourceDefinition runtimeResourceDefinition = theFhirContext.getResourceDefinition(theResourceType);
 
-		List<RuntimeSearchParam> searchParams = runtimeResourceDefinition.getSearchParamsForCompartmentName("Patient");
+		List<RuntimeSearchParam> searchParams = new ArrayList<>(runtimeResourceDefinition.getSearchParamsForCompartmentName("Patient"));
+		// add patient search parameter for resources that's not in the compartment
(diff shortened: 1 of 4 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
```diff
@@ -76,6 +76,22 @@ public static Optional<RuntimeSearchParam> getOnlyPatientSearchParamForResourceT
 		return Optional.ofNullable(myPatientSearchParam);
 	}
 
+	/**
+	 * Given the resource type, fetch its secondary patient-based search parameter name that's different from what's found in getOnlyPatientSearchParamForResourceType()
+	 * 1. Attempt to find one called 'author'
+	 * 2. If that fails, find one called 'performer'
+	 * This can support more search param, but for now these are the only 2 supported
+	 */
+	public static Optional<RuntimeSearchParam> getOtherPatientSearchParamForResourceType(FhirContext theFhirContext, String theResourceType) {
```
tadgh, line 85: Issue: I believe this shim is a code smell.

I See you're trying to design around the existing API `getOnlyPatientSearchParam`, but the intent here is non-obvious. Look for other usages of this method. Where is it used? the only location is JpaBulkExportProcessor.java.

What was the actual intent of the original design? What you've done here is just adding a shim to a functionality that needs to be changed fundamentally. The more of these magic shims we add for special edge cases, the less clear the intent of the code is, and the more difficult it is for other developers to work on it in the future. 

Based on all these recent changes to Bulk Export, there is clearly the desire to export more rather than less. The purpose of the original code was to just get the narrowest patient SP possible.

Our new knowledge tells us that many resources have more than one patient SP, so why not just make that the default? 

Now the requirement appears to be: "Get all SPs that could possibly point to patient and check em all".

I would argue that the bulk export processor needs to be changed to do this new behaviour for all resource types. Maintaining a hardcoded list of special resources is a code smell in this case. Also, what if other people add search parameters on resources you haven't considered that point to patient? I recommend refactoring JpaBulKExportProcessor to simply always expect a set of search parameters, and have the util method fetch all SPs that could possibly point to patients.

tadgh (changes requested):

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
StevenXLi: This is a valid concern, I have implemented it now to get all the patient based search params that's available to the resource type being processed.
I did run into one thing, it seems to me that, for some resources, when getting all the patient based search params, some of the ones returned are not listed in the [compartment definition](https://www.hl7.org/fhir/compartmentdefinition-patient.html). I don't know if it warrants special case handling for those exceptions

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_4_0/4360-bulk-export-questionnaireresponses-should-get-picked-when-author-is-not-empty.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 4360
+title: "Previously, Patient Bulk Export on certain resources with reference to patient only in author or performer did not get exported.
```
tadgh, line 4: I imagine there should be a `jira` block here, as this is based on a JIRA issue yeah?

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
```diff
@@ -76,6 +76,20 @@ public static Optional<RuntimeSearchParam> getOnlyPatientSearchParamForResourceT
...
+		RuntimeResourceDefinition runtimeResourceDefinition = theFhirContext.getResourceDefinition(theResourceType);
+
+		List<RuntimeSearchParam> searchParams = runtimeResourceDefinition.getSearchParamsForCompartmentName("Patient");
+		if (searchParams == null || searchParams.size() == 0) {
+			String errorMessage = String.format("Resource type [%s] is not eligible for this type of export, as it contains no Patient compartment, and no `patient` or `subject` search parameter", runtimeResourceDefinition.getId());
+			throw new IllegalArgumentException(Msg.code(2222) + errorMessage);
+		}
+		// deduplicate list of searchParams and get their names
+		return searchParams.stream().map(RuntimeSearchParam::getName).collect(Collectors.toSet());
+	}
```
tadgh, line 92: Question: What happened to the previous shims that were there? The ones that attempted to lookup by `patient` or `subject`. Are those no longer needed, because they're covered by the compartment? Those shims covered resources _not_ in the Patient Compartment, e.g. `device.patient` in Device R4. 

I like that you are returning the whole set of SPs in the compartment, but this is not the universe of things we care about. This is why that lookup by `patient` and `subject` was being done. Would probably be good to amalgamate the SPs of: 
1. Patient Compartment
2. SP that has `patient` code. 
3. SP that has `subject` code. 

This will involve cannibalizing the `getOnlyPatientSearchParamForResourceType` code, and instead of returning on the first one found, just returning em all as a set. 

EDIT: A way to see if you broke the functionality is to write a test as follows: 
1. Create a patient. 
2. Create a Device that refers to the patient from 1. 
3. Call a bulk export on the patient. 
4. Ensure the device is in it .

tadgh (approved):
Approved pending another change.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_4_0/4360-bulk-export-questionnaireresponses-should-get-picked-when-author-is-not-empty.yaml
StevenXLi: Nop, no jira linked in the original ticket, it is for TrustSphere

StevenXLi: https://gitlab.com/simpatico.ai/cdr/-/issues/3881

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
StevenXLi: test added, and fixed implementation

PR #4365, opened 2022-12-13
fix index migration error

nathandoef:


tadgh (approved):

KevinDougan-SmileCDR:
@nathandoef   
This change is confusing to me.

I just added this new INDEX yesterday and you changed the name of it to have a `FK_` prefix, but it is NOT a Foreign Key at all, is is a brand new DB Index.

nathandoef:
> @nathandoef This change is confusing to me.
> 
> I just added this new INDEX yesterday and you changed the name of it to have a `FK_` prefix, but it is NOT a Foreign Key at all, is is a brand new DB Index.

Hey @KevinDougan-SmileCDR , we have a test that compares database schemas before and after a migration that was failing.

This is because in some databases (H2), an index is created on Foreign Key columns. So in this case H2 already has an index called `FK_SEARCHINC_SEARCH` on the `SEARCH_PID` field of the `HFJ_SEARCH_INCLUDE` table.

When the new index `IDX_HFJ_SEARCH_INCLUDE_SRCHPID` was added there were essentially two different index names on the `SEARCH_PID` column in H2.

This fix puts those index names in sync. So now if the database platform is H2, nothing needs to be done. If the database platform does not automatically create a foreign key index, a new index is created named `FK_SEARCHINC_SEARCH` that matches the H2 name.

KevinDougan-SmileCDR:
>
> Hey @KevinDougan-SmileCDR , we have a test that compares database schemas before and after a migration that was failing.
> 
> This is because in some databases (H2), an index is created on Foreign Key columns. So in this case H2 already has an index called `FK_SEARCHINC_SEARCH` on the `SEARCH_PID` field of the `HFJ_SEARCH_INCLUDE` table.
> 
> When the new index `IDX_HFJ_SEARCH_INCLUDE_SRCHPID` was added there were essentially two different index names on the `SEARCH_PID` column in H2.
> 
> This fix puts those index names in sync. So now if the database platform is H2, nothing needs to be done. If the database platform does not automatically create a foreign key index, a new index is created named `FK_SEARCHINC_SEARCH` that matches the H2 name.

Hi @nathandoef ,

Thanks for the explanation! It still strikes me that the name `FK_SEARCHINC_SEARCH` is invalid for an Index though, because our standard naming convention requires them to start with `IDX_` and we typically try and squeeze in the Table Name followed by Column(s).

To me, this implies that the H2 DB Index was created incorrectly, so I wonder how that happened and whether we should be fixing that incorrect name in H2?

Thanks again!
Kevin

nathandoef:
> > Hey @KevinDougan-SmileCDR , we have a test that compares database schemas before and after a migration that was failing.
> > This is because in some databases (H2), an index is created on Foreign Key columns. So in this case H2 already has an index called `FK_SEARCHINC_SEARCH` on the `SEARCH_PID` field of the `HFJ_SEARCH_INCLUDE` table.
> > When the new index `IDX_HFJ_SEARCH_INCLUDE_SRCHPID` was added there were essentially two different index names on the `SEARCH_PID` column in H2.
> > This fix puts those index names in sync. So now if the database platform is H2, nothing needs to be done. If the database platform does not automatically create a foreign key index, a new index is created named `FK_SEARCHINC_SEARCH` that matches the H2 name.
> 
> Hi @nathandoef ,
> 
> Thanks for the explanation! It still strikes me that the name `FK_SEARCHINC_SEARCH` is invalid for an Index though, because our standard naming convention requires them to start with `IDX_` and we typically try and squeeze in the Table Name followed by Column(s).
> 
> To me, this implies that the H2 DB Index was created incorrectly, so I wonder how that happened and whether we should be fixing that incorrect name in H2?
> 
> Thanks again! Kevin

Hi @KevinDougan-SmileCDR, you make a good point about the index names. There are a couple existing indexes that follow the same naming convention to address the H2 issue (`TestUtil#duplicateNameValidationExceptionList`).

Do you want to create a new ticket for this one, or should this MR fix all the previous issues too?

PR #4366, opened 2022-12-13
Fix Schematron issue when running in OSGi container

bdenton:
This fix allows Schematron validation to work in an OSGi container. Currently, Schematron is not able to load the .sch files from the "validation-resources" bundles. This was caused by the fact that, in OSGi, the "validation-resource" bundles are not visible to the classloader for the 'ph-schematron*' bundle.

The fix is to crate the SCH Resource using the classloader for hapi-fhir-base which *does* have visibility to the 'validation-resource' bundles


bdenton:
Any chance of merging this PR into an upcoming release??

tadgh (approved):
This doesnt break any tests, approved and merging.

PR #4367, opened 2022-12-13
Split-package conflict with org.hl7.fhir.* model projects

bdenton:
when the Fhir resource models were moved over to `org.hl7.fhir.core`, the `fhirversion.properties` files in `hapi-fhir` remained in the `*.model` Java packages (in src/main/resources directory). This created a "split package" situation that cannot be resolved with any OSGi tricks like `Fragment-Host`

This PR corrects this split-package situation.


bdenton:
This problem still exists but I will resubmit this PR ported to current master rather than trying to compensate for the merge conflicts

Issue #4379, opened 2022-12-19
Allow Interceptor annotation on methods


doumdoum:
NOTE: Before filing a ticket, please see the following URL:
https://github.com/hapifhir/hapi-fhir/wiki/Getting-Help

**Describe the issue**
So far Interceptor annotation is only allowed on types and is just used as a marker.
It can only be used as a bean marker for interceptor classes but not when the bean is provided by a configuration method.
To be able to use this annotation in such case the interceptor annotation should allow METHOD as location

**Environment (please complete the following information):**
 - HAPI FHIR Version: ALL
 - OS: ALL
 - Browser N/A

doumdoum:
Hello there, wish you the best for 2023.
Any feedback on that ticket ?
Thx

jamesagnew:
PR was merged

PR #4380, opened 2022-12-19
4379: allow interceptor annotation on methods

doumdoum:
Hi,
Fixing https://github.com/hapifhir/hapi-fhir/issues/4379

Regards,

/DV


doumdoum:
Hello, kind reminder

doumdoum:
Hello, any news regarding this PR ?

doumdoum:
Hello, any news regarding this PR ?

doumdoum:
Hi, kind reminder...

doumdoum:
Hi, any news regarding this PR ?

jamesagnew:
Hi @doumdoum -

Thanks for the contribution. I'm concerned though that this could be confusing to users if we allow this at the method level, since HAPI FHIR doesn't actually look at it when used at the method level. E.g. a user might do something like:

```java
@Interceptor(order=5)
@Hook(Pointcut.FOO)
public void blah() { }
```

..and then expect the order to be respected, when it won't be. At a minimum the javadoc should probably be updated to point out that the annotation is purely a marker when used at method level, and the properties will be ignored.

doumdoum:
> Hi @doumdoum -
> 
> Thanks for the contribution. I'm concerned though that this could be confusing to users if we allow this at the method level, since HAPI FHIR doesn't actually look at it when used at the method level. E.g. a user might do something like:
> 
> ```java
> @Interceptor(order=5)
> @Hook(Pointcut.FOO)
> public void blah() { }
> ```
> 
> ..and then expect the order to be respected, when it won't be. At a minimum the javadoc should probably be updated to point out that the annotation is purely a marker when used at method level, and the properties will be ignored.

Hi @jamesagnew, sorry for this late response. 
If the interceptor is registered through the registerInterceptor method then order is respected no ?
Should it be sufficient to mention that order is respected as long as interceptor is registered ?

PR #4384, opened 2022-12-20
Add Device to the list of resource types considered for group bulk export async

lukedegruchy:
- In the case of a group export async with no provided resource types, add Device to the implicitly built list of the resource types.

Closes [4385](https://github.com/hapifhir/hapi-fhir/issues/4385)


jmarchionatto (approved):
Looks good!

PR #4388, opened 2022-12-21
Add test and implementation to fix potential NPE in pre-show resources

tadgh:
* Prevent hook invocation if the resolved resources would be null. 
* Prevent NPE in downstream interceptors. 
* Add test and changelog


patrick-vachon-smilecdr:
praise: That looks like the easiest way to fix that! 

Review done.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/binary/interceptor/BinaryStorageInterceptor.java
```diff
@@ -256,30 +256,37 @@ public void preShow(IPreResourceShowDetails theDetails) throws IOException {
...
+				ourLog.warn("Received a null resource during STORAGE_PRESHOW_RESOURCES. This is a bug and should be reported. Skipping resource.");
+				continue;
+			}
+			unmarshalledByteCount = inflateBinariesInResource(unmarshalledByteCount, nextResource);
+		}
+	}
 
-			IIdType resourceId = nextResource.getIdElement();
-			List<IBinaryTarget> attachments = recursivelyScanResourceForBinaryData(nextResource);
+	private long inflateBinariesInResource(long theUnmarshalledByteCount, IBaseResource theResource) throws IOException {
```
jamesagnew, line 267: Nitpick: I think it'd be a bit easier to follow if you didn't pass theUnmarshalledByteCount in, and just returned the total for this rsources and added it to the running total outside this method

jamesagnew (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/binary/interceptor/BinaryStorageInterceptor.java
tadgh: ah yeah makes sense. Laziness on my part, trusting the auto-refactor's choices. Fixing.

tadgh: Ah actually... we need to know the current total _in_ the loop for the purpose of stopping if we are over-inflating.

PR #4400, opened 2023-01-04
Dont fail on batch2 double delivery

jamesagnew:
If we receive the same work chunk notification twice in Batch2, we can fail the second time with a blocking error because the data was already purged.


tadgh (approved):
Looks good to me

PR #4414, opened 2023-01-06
Schedule jobs via IJobScheduler instead of in @PostConstruct

fil512:
Previously, jobs were scheduled via a @PostConstruct annotated method.  This has been changed to an implementation
of IJobScheduler interface so the job scheduler can schedule all the jobs after it has been started.


fil512:
bad branch

PR #4415, opened 2023-01-07
Schedule jobs via IJobScheduler instead of in @PostConstruct

fil512:


hapi-fhir-jpa/src/main/java/ca/uhn/fhir/jpa/model/sched/IJobScheduler.java
```diff
@@ -0,0 +1,5 @@
+package ca.uhn.fhir.jpa.model.sched;
+
+public interface IJobScheduler {
```
jamesagnew, line 3: I find the name of this interface a bit confusing - It kinda suggests that implementors _are_ a job scheduler, but implementors just have scheduled jobs.

How about `IHasScheduledJobs`?

One other thought, feel free to ignore: Would it make sense for the `scheduleJobs` method to accept a parameter of type `ISchedulerService`? That way all of the various beans with jobs wouldn't need to Autowire that dependency and there would be slighly less tigher coupling?

jamesagnew (approved):

hapi-fhir-jpa/src/main/java/ca/uhn/fhir/jpa/model/sched/IJobScheduler.java
fil512: ooooh very nice!

PR #4424, opened 2023-01-11
Improve query count accuracy

jamesagnew:


hapi-fhir-jpa/src/main/java/ca/uhn/fhir/jpa/util/BaseCaptureQueriesListener.java
```diff
@@ -63,7 +63,7 @@ public void execute(ExecutionInfo theExecutionInfo, List<QueryInfo> theQueryInfo
 		for (QueryInfo next : theQueryInfoList) {
 			String sql = trim(next.getQuery());
 			List<String> params;
-			int size = 0;
+			int size;
 			if (next.getParametersList().size() > 0 && next.getParametersList().get(0).size() > 0) {
 				size = next.getParametersList().size();
 				List<ParameterSetOperation> values = next
```
jmarchionatto, line 69: question: can't the list have more than 1 element? Looks only first element is being processed

jmarchionatto (approved):
Looks good! Approved

hapi-fhir-jpa/src/main/java/ca/uhn/fhir/jpa/util/BaseCaptureQueriesListener.java
jamesagnew: This is deliberate but I agree it's not obvious. Adding the following comment:
```java
		/*
		 * Note on how this works:
		 *
		 * Hibernate batches queries where the SQL is identical other than parameter values.
		 *
		 * So for example, if we call "select PID from SOMETABLE where NAME = ?" twice with
		 * two different name values, this method will be called only once with 2 parameter
		 * lists. In this case, we say size=2 and we capture the first parameters in case
		 * we want to log them. We don't capture subsequent parameters currently because
		 * this is only used for troubleshooting logging anyhow so the values are used
		 * only as a representative example only.
		 */
```

PR #4425, opened 2023-01-11
Clean up ResourceMetadataKeyEnum

jamesagnew:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -466,8 +466,7 @@ private List<JpaPid> queryHibernateSearchForEverythingPids() {
 		if (myParams.get(IAnyResource.SP_RES_ID) != null) {
 			String idParamValue;
 			IQueryParameterType idParam = myParams.get(IAnyResource.SP_RES_ID).get(0).get(0);
-			if (idParam instanceof TokenParam) {
-				TokenParam idParm = (TokenParam) idParam;
+			if (idParam instanceof TokenParam idParm) {
```
jmarchionatto: question: can't understand how this construction passed the build. Isn't the target still 11 for prod?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/dao/MetadataKeyResourcePid.java
```diff
@@ -20,34 +20,13 @@
...
 
-import ca.uhn.fhir.model.api.IResource;
-import ca.uhn.fhir.model.api.ResourceMetadataKeyEnum.ResourceMetadataKeySupportingAnyResource;
-import org.hl7.fhir.instance.model.api.IAnyResource;
+import ca.uhn.fhir.model.api.ResourceMetadataKeyEnum;
 
-public final class MetadataKeyResourcePid extends ResourceMetadataKeySupportingAnyResource<Long, Long> {
-	private static final long serialVersionUID = 1L;
+public final class MetadataKeyResourcePid extends ResourceMetadataKeyEnum<Long> {
+	private static final long serialVersionUID = 2L;
```
jmarchionatto, line 26: question: won't this cause backwards compatibility problems for saved resources?

jmarchionatto (approved):
Approved with a couple of questions.    
(BTW, "tiny" is in the eye of the beholder :-) Took me a while to understand that class)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
jamesagnew: It actually didn't. :) I accepted an IJ cleanup suggestion (which is the source of this change) and failed torealize that it was introducing a JDK17 feature. Fortunately our CI build caught it and it's already reverted in this PR.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/dao/MetadataKeyResourcePid.java
jamesagnew: It will - but we make no guarantees about serialization capabilities across HAPI FHIR versions. Given that our model has a natural wire format that is hapi-version-agnostic I'm not worried about this. The only use real case for serialization is for very temporary storage (e.g. as a part of user sessions and that kind of thing).

PR #4430, opened 2023-01-11
Demote PHI logging to debug or lower

michaelabuckley:
Clean up logging of resource bodies and other possible PHI to debug or lower.

The important items:
- LoggingInterceptor now logs request/response bodies at `debug` rather than `info`.
- Any troubleshooting or other log actions of resource bodies have been demoted to `debug` as well.


jmarchionatto (approved):
Looks good, however there are still some tests with info logging which could display some PHI. Doing a search for "ourlog.info" on files: *Test.java,*IT.java shows hits as in FhirResourceDaoR4SearchWithElasticSearchIT line 758 and more. 
I am approving change, but you may want to do that search (and maybe also with ourLog.error) for further validation

Issue #3466, opened 2022-03-13
OpenAPI generates incomplete and incorrect specs


lawley:
NOTE: Before filing a ticket, please see the following URL:
https://github.com/hapifhir/hapi-fhir/wiki/Getting-Help

**Describe the bug**

There are (at least) three issues:

1. Non-primitive parameters are still listed as query parameters in the GET version of an operation. For example, the valueSet parameter to `ValueSet/$expand`.

2. POST should always be supported and thus should always be included in the OpenAPI specification, rather than omitted when GET is indicated via `affectsState=true` (from `idempotent=true`).

3. GET is always indicated when `affectsState=false`, even when non-primitive required parameters are specified.

Specifically, the FHIR documentation for OperationDefinition.affectsState says:

> What http methods can be used for the operation depends on the `affectsState` value and whether the input parameters are primitive or complex:
> 
> - Servers SHALL support POST method for all operations.
> - Servers SHALL support GET method if all the parameters for the operation are primitive or there are no parameters and the operation has affectsState = false.

But `OpenApiInterceptor.addFhirOperation` always indicates support for GET when the value of `affectsState` is false.
It should also check that there are no non-primitive required parameters (since there's no way to pass these in a GET).

**To Reproduce**
1. Define an operation like:
```
    @Operation(name="$convert", idempotent=true, type=IBaseResource.class)
    public IBaseResource convert(
            RequestDetails requestDetails,
            HttpServletRequest request,
            @ResourceParam IBaseResource resource
            )
```
3. Enable the OpenApiInterceptor
4. Go to `[base]/swagger-ui/?page=System Level Operations`
5. Only an entry for the GET form is present and input is shown as a query parameter

**Expected behavior**

`$convert` should be shown as a POST operation with an example value of `{}` rather than
```
{
  "resourceType": "Parameters"
}
```

**Environment (please complete the following information):**
 - HAPI FHIR Version 5.7.0
 - OS: N/A
 - Browser N/A

**Additional context**

I understand there may be a desire to omit POST requests where GET requests are indicated in order to simplify a generated Swagger UI for human consumption, but this affects other use-cases such as providing an API definition for an API Manager where completeness is required.

PR #4436, opened 2023-01-17
#3466 fix and improve openapi generation

lawley:
fixes hapifhir/hapi-fhir#3466
    
* Includes a POST version for every GET endpoint.
* Excludes non-primitive parameters from GET endpoints.
* For an idempotent (affectsState=false) endpoint with at least one required non-primitive parameter, only support POST

* Also includes */_search endpoints for the Search operation.


jamesagnew (approved):
Looks great! Thanks, merging now

PR #4445, opened 2023-01-18
Reduce postconstruct use

fil512:
Step 2 in reducing use of PostConstruct in the codebase.  Most were replaced with constructor params.  Some depended on ContextRefreshed event in an ordered way.  Replaced "magic" numbers with a config file as an interim measure.  In the future, the plan is to replace these with named events.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
@@ -1,0 +1,13 @@
+package ca.uhn.fhir;
+
+/**
+ * Spring Boot order constants.
+ */
+public interface IHapiBootOrder {
+	int ADD_JOB_DEFINITIONS = 100;
+	int REGISTER_INTERCEPTORS = 200;
+
+	int SUBSCRIPTION_MATCHING_CHANNEL_HANDLER = 300;
+	int AFTER_SUBSCRIPTION_INITIALIZED = 310;
+
+}
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
@@ -21,7 +21,7 @@ package ca.uhn.fhir;
  */
 
 /**
- * Spring startup order constants.
+ * Spring bean initialization constants.
  */
 public interface IHapiBootOrder {
 	int ADD_JOB_DEFINITIONS = 100;
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
@@ -21,7 +21,7 @@ package ca.uhn.fhir;
  */
 
 /**
- * Spring Boot order constants.
+ * Spring startup order constants.
  */
 public interface IHapiBootOrder {
 	int ADD_JOB_DEFINITIONS = 100;
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
@@ -1,5 +1,25 @@
 package ca.uhn.fhir;
 
+/*-
+ * #%L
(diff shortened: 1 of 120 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+/**
+ * Spring Boot order constants.
+ */
+public interface IHapiBootOrder {
```
jamesagnew, line 26: I love this

hapi-fhir-base/src/main/java/ca/uhn/fhir/IHapiBootOrder.java
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+
+/**
+ * Spring Boot order constants.
```
jamesagnew, line 24:
```suggestion
 * Spring startup order constants.
```

Just to avoid any confusion about this being for Spring Boot

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
```diff
@@ -74,18 +73,13 @@ public class HapiTransactionService implements IHapiTransactionService {
 	protected PlatformTransactionManager myTransactionManager;
 	@Autowired
 	protected IRequestPartitionHelperSvc myRequestPartitionHelperSvc;
-	private boolean myCustomIsolationSupported;
+	private Boolean myCustomIsolationSupported;
```
jamesagnew, line 76:
```suggestion
	private volatile Boolean myCustomIsolationSupported;
```

If you're potentially setting this field from multiple threads, it needs to be volatile.

jamesagnew (approved):

jmarchionatto (approved):
Approving with a few comments

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
fil512: nice catch

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/cache/ResourceChangeListenerRegistryInterceptor.java
```diff
@@ -46,7 +47,7 @@ public class ResourceChangeListenerRegistryInterceptor {
 
 
 	@EventListener(classes = {ContextRefreshedEvent.class})
-	@Order
+	@Order(IHapiBootOrder.REGISTER_INTERCEPTORS)
```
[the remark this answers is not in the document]
fil512: Before, register interceptors was ContextRefreshed Lowest precedence whereas subscription matching was run on ContextRefreshed with no order which is equivalent to lowest precedence.  So before the two beans would have equal precendence whereas now subscriptions are explicitly started after interceptors are registered which is correct since message handlers have pointcuts.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
```diff
@@ -162,11 +156,15 @@ public <T> T execute(@Nullable RequestDetails theRequestDetails, @Nullable Trans
 	}
 
 	public boolean isCustomIsolationSupported() {
```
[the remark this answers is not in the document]
fil512: That feels risky to me.  This used to be set in a @PostConstruct, and now it is lazy-initialized when first requested which will happen WAAAAY later and certainly after the jpadialect has been set.

PR #4449, opened 2023-01-19
fix ValueSet$expand?includeHierarchy=true with fts disabled

ivagulin:
Unable to $expand icd10cm ValueSet without this fix. Fails with NumberFormatException. Is it possible to apply it?

Looks like broken by TermConcept changes in e9cb5180124. Not sure why its not triggered in a wild. Starts to work after fix for me.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_4_0/4449-fix-expand-with-hierarchy.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_4_0/4449-fix-expand-with-hierarchy.yaml
@@ -2,4 +2,5 @@
 type: fix
 issue: 4449
 title: "A bug prevented valueset expansion with includeHierarchy=true when lucene/elasticsearch 
-   was not enabled. This has been corrected. Thanks to GitHub user @ivagulin for the contribution!"
+   was not enabled. This has been corrected. Thanks to GitHub user @ivagulin for the 
+   contribution!"


jamesagnew:
Hi @ivagulin - Thanks for the contribution. Ideally this should include a test that verifies the fix, would you be able to add one?

ivagulin:
I will try to add test. Not sure if I will be able to, project is pretty big :)

ivagulin:
Hi @jamesagnew. I've added test, doesn't work without my fix. Need testordering to don't affect other tests with disabled fts, so added to FhirResourceDaoR4QueryCountTest.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
```diff
@@ -2700,6 +2701,49 @@ public void testValueSetExpand_PreExpanded_UseHibernateSearch() {
 		assertEquals(0, myCaptureQueriesListener.countRollbacks());
 	}
 
+	@Test
+	public void testExpandValueSet_PreExpandedWithHierarchyNoHibernateSearch() {
```
jamesagnew, line 2705: Test method looks good, thanks!

I'd propose that `FhirResourceDaoR4ValueSetTest` is a better home for this test since this test is about ValueSet expansion and not query counting (and that class already resets the `setForceDisableHibernateSearchForUnitTest` value so the test won't have side effects). 

My recommendation:
* Move this test method to `FhirResourceDaoR4ValueSetTest`
* Add a javadoc comment to the method linking to this PR (Just `/** See #4449 */` is good enough)

With that done, we're good to merge!

ivagulin:
Hi @jamesagnew. I've applied your suggestions. Please merge if everything ok or give me more guidance.

BTW if there are any chances this can hit rel_6_2 branch anytime soon? Can I help with PR? We are evaluating possibility to use hapi-fhir for our small hospital project and this issue struck as in a head.

jamesagnew (approved):

jamesagnew:
Hi @ivagulin - Thanks for the contribution!

It's unlikely we'll be doing any more 6.2.x releases, but this will be included in the 6.4.0 release in February.

ivagulin:
Hi, that is great news thanks. We will switch to 6.4 releases as soon as they will be available. Thank you for collaboration!

PR #4452, opened 2023-01-19
Add utils for testing Optional, bulk-export polling, and http requests.

michaelabuckley:
Some minor test utilities.


hapi-fhir-base/src/main/java/ca/uhn/fhir/util/MethodOutcomeUtil.java
```diff
@@ -0,0 +1,23 @@
...
+import ca.uhn.fhir.rest.api.MethodOutcome;
+
+import javax.annotation.Nonnull;
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Extra methods for {@link MethodOutcome}.
+ */
+public class MethodOutcomeUtil {
```
jamesagnew, line 12: Nitpick: This class should have a private constructor

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/MethodOutcomeUtil.java
```diff
@@ -0,0 +1,23 @@
...
+
+	@Nonnull
+	public static Optional<String> getFirstResponseHeader(MethodOutcome theMethodOutcome, String theHeader) {
```
jamesagnew, line 15: Actually, thinking even more, why not just put this method on MethodOutcome as `getFirstResponseHeader(String)`? That might be more discoverable.

jamesagnew (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/MethodOutcomeUtil.java
michaelabuckley: Good suggestion.  No reason besides cowardice.  This sort of looked like a core lib, and I don't touch those.  I'll do that, and defend getResponseHeaders() against null.

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/MethodOutcomeUtil.java
michaelabuckley: It's dead, Jim.

PR #4453, opened 2023-01-20
Batch2 fastracking configurable

fil512:
A new DaoConfig configuration setting has been added called JobFastTrackingEnabled, default false.
If this setting is enabled, then gated batch jobs that produce only one chunk will immediately trigger a batch
maintenance job.  This may be useful for testing, but is not recommended for production use.  Prior to this change,
fasttracking was always enabled which meant if the server was not busy, small batch jobs would be processed quickly.
However this lead do instability on high-volume servers, so this feature is now disabled by default.

### DIFF
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/config/DaoConfig.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/config/DaoConfig.java
@@ -359,6 +359,10 @@ public class DaoConfig {
 			ourLog.info("Status based reindexing is DISABLED");
 			setStatusBasedReindexingDisabled(true);
 		}
+		if (HapiSystemProperties.isUnitTestModeEnabled()) {
+			setJobFastTrackingEnabled(true);
+		}
+
 	}
 
 	/**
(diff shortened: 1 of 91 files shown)


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/config/DaoConfig.java
```diff
@@ -355,6 +359,10 @@ public DaoConfig() {
 			ourLog.info("Status based reindexing is DISABLED");
 			setStatusBasedReindexingDisabled(true);
 		}
+		if (HapiSystemProperties.isUnitTestModeEnabled()) {
```
jamesagnew, line 362: The one thing that makes me nervous about this - Do we have any tests that ensure that the non-unit test mode will actually work? Ie are we sure that with fast tracking disabled, the non-fast-track path actually triggers the job?

jamesagnew (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/api/config/DaoConfig.java
fil512: Most of our ITs disable the scheduler.  I'll add a test that gets at your concern.

fil512: done

PR #4470, opened 2023-01-26
cleaning up checkstyle files

markiantorno:
- moved all checkstyle files to checkstyle module
- changed CI profile to run checkstyle across _all_ modules to catch cross module error code duplication
- combined base-pom checkstyle and module checkstyle into one checkstyle config
- set up to delete checkstyle cache on every run, regardless of if `clean` is included in maven command
- local module checkstyle now runs on every build, but is excluded from CI profile (this runs the checkstyle across the entire project)


michaelabuckley (approved):
Nice to see all those code dupes fixed.

PR #4476, opened 2023-01-27
Enabling massIngestionMode causes incomplete resource deletion

epeartree:
#### Background:
Enabling MassIngestionMode will have negative effects on the resource deletion process. When a resource is deleted with the mode enabled:

- entries related to the deleted resource in the various index tables are not removed;
-  the entry marking the resource deletion is not added to the HFJ_RES_VER table;

#### what was done:
- modification of the deletion process to keep on marking the resource being deleted has  'changed' to trigger updates of the index tables and adding an entry to the version table;
- adding tests;
- adding changelog;

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
@@ -143,7 +143,7 @@ import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
```diff
@@ -601,7 +601,11 @@ protected EncodedResource populateResourceIntoEntity(TransactionDetails theTrans
 
 		} else {
 
-			theEntity.setHashSha256(null);
+			if(nonNull(theEntity.getHashSha256())){
```
jamesagnew, line 604: Wow...... Of all the things to have caused this issue....

Nice catch

jamesagnew (approved):

PR #4479, opened 2023-01-30
Added information on how referential integrity works in HAPI. 

JorisHeadease:
Only indexed fields are validated. Devs might expect foreign-key constraint logic when enabling this option. Also added examples on how to enable this setting in both JPA and JPA starter.


jamesagnew (approved):
Great docs, thanks for the contribution!

jamesagnew:
BTW - This should appear on the live site in about 24 hours/

JorisHeadease:
Great, thanks a lot @jamesagnew!

JorisHeadease:
Hi @jamesagnew, just a quick head up as I don't see the changes applied to the live docs.

PR #4488, opened 2023-01-31
Delete methods deprecated in 4485.

lukedegruchy:


jmarchionatto (approved):
Good move! Maybe you want to expose the method in a tech talk?

PR #4493, opened 2023-02-01
care-gaps integration

chalmarm:


pom.xml
```diff
@@ -956,7 +956,7 @@
 		 <!-- CQL Support -->
 		 <cqframework.version>2.4.0</cqframework.version>
 		 <cql-engine.version>2.3.0</cql-engine.version>
-		 <cql-evaluator.version>2.3.0</cql-evaluator.version>
+		 <cql-evaluator.version>2.6.0-SNAPSHOT</cql-evaluator.version>
```
JPercival, line 959: The engine/translator versions should also be bumped.

hapi-fhir-storage-cr/src/main/java/ca/uhn/fhir/cr/r4/measure/CareGapsProvider.java
```diff
@@ -0,0 +1,143 @@
...
+						organization,
+						status,
+						measureId,
+						measureIdentifier,
+						measureUrl,
+						program
+					);
+	}
+
+	OperationOutcome generateIssue(String severity, String issue) {
```
JPercival, line 122: There is likely a utility class for generating OperationOutcomes.

hapi-fhir-storage-cr/src/test/java/ca/uhn/fhir/cr/r4/CareGapsProviderIT.java
```diff
@@ -0,0 +1,727 @@
...
+				, null
+				, null
+				, status
+				, measures
+				, null
+				, null
+				, null
+			);
+		} catch (DataFormatException de){
+			//do nothing.
```
JPercival, line 219: We should use the `expectedException` attribute of the `@Test` annotation rather than catching and swallowing.

hapi-fhir-storage-cr/src/test/java/ca/uhn/fhir/cr/r4/CareGapsProviderIT.java
```diff
@@ -0,0 +1,727 @@
...
+			, null
+			, null
+			, status
+			, measures
+			, null
+			, measureUrls
+			, null
+		);
+
+		assertNotNull(result);
```
JPercival, line 694: This doesn't test that the output is correct, only that it's "not null".

hapi-fhir-storage-cr/src/test/java/ca/uhn/fhir/cr/r4/CareGapsProviderIT.java
```diff
@@ -0,0 +1,727 @@
...
+				stringPart("subject", subjectPatientValid),
+				stringPart("subject", subjectGroupValid),
+				stringPart("status", statusValid),
+				stringPart("measureId", measureIdValid));
+
+		SystemRequestDetails requestDetails = new SystemRequestDetails();
+		requestDetails.setRequestType(RequestTypeEnum.POST);
+		requestDetails.setFhirContext(getFhirContext());
+		requestDetails.setResource(params);
+		Parameters result = theCareGapsProvider.careGapsReport(requestDetails
```
JPercival, line 509: This test is apparently trying to test that the RequestDetails can pass input parameters. That's all handled by HAPI infrastructure, so the scenario that this test is covering is not possible. The input parameters would be populated from the RequestDetails as part of REST processing.

JPercival: We should probably strip out the Operation Parameter validation stuff. That should be enforced at a different level.

JPercival (changes requested):

hapi-fhir-storage-cr/src/main/java/ca/uhn/fhir/cr/r4/measure/CareGapsProvider.java
JPercival: Also, this should throw a meaningful exception, it shouldn't return a successful response in the form of a Parameters resource with an "invalid parameters" part.

JPercival: The HAPI REST processing layer will take care of converting that to an OperationOutcome.

https://build.fhir.org/operations.html#response

PR #4499, opened 2023-02-02
fix batch2 reduction step

### LINKED ISSUE #4491: Batch2 Jobs in FINALIZE state cannot be cancelled
Batch 2 reduction steps can be fired multiple times for long running jobs.

Once fired a second time, jobs can:
* fail
* overwrite existing data
* stay in FINALIZE status

Once a job sticks in FINALIZE status, it cannot be cancelled

TipzCM:
Closes #4491

* moved reduction step to message handling to prevent multiple triggerings of the same job during maintenance run
* will allow jobs in FINALIZE state to be cancelled

### DIFF
--- a/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/services/Batch2JobRunnerImpl.java
+++ b/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/services/Batch2JobRunnerImpl.java
@@ -109,15 +109,17 @@ public class Batch2JobRunnerImpl implements IBatch2JobRunner {
 	public static BulkExportJobStatusEnum fromBatchStatus(StatusEnum status) {
 		switch (status) {
 			case QUEUED:
+			case FINALIZE:
 				return BulkExportJobStatusEnum.SUBMITTED;
 			case COMPLETED :
 				return BulkExportJobStatusEnum.COMPLETE;
 			case IN_PROGRESS:
 				return BulkExportJobStatusEnum.BUILDING;
+			default:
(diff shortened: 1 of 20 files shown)


hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/services/Batch2JobRunnerImpl.java
```diff
@@ -109,15 +109,17 @@ private Batch2JobInfo fromJobInstanceToBatch2JobInfo(@Nonnull JobInstance theIns
 	public static BulkExportJobStatusEnum fromBatchStatus(StatusEnum status) {
 		switch (status) {
 			case QUEUED:
+			case FINALIZE:
```
jamesagnew, line 112: Is this right? I would have expected finalize to map to IN_PROGRESS

Also I guess as a nitpick aside, I find this to be an odd place for this method given that it's BulkExport specific but it's in the generic batch2 runner. I might suggest to refactor->move the method over to being a static method right on BulkExportJobStatusEnum itself

jamesagnew (approved):

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/services/Batch2JobRunnerImpl.java
TipzCM: Will move it to bulk export area.

might be a tech debt item to either change this enum to a generic batch2 job enum (which it seems like it is now) or use the statusenum only (or merge the 2)

For now, i've just moved it to the bulk export area.

PR #4514, opened 2023-02-06
changed what score is set for mdmlinks that created new golden resource

longma1:
Changed behaviour when an incoming resource has to create a new golden resource.
Currently the score for the link is null, this changes it to 1.0


fil512 (approved):
looks sensible

Issue #4523, opened 2023-02-08
meta.source field in subscription messages is inconsistently populated


Qingyixia:
**Describe the bug**
We have FHIR subscriptions that send all resource updates to Kafka. We have noticed that the {{meta.source}} field in the messages is sometimes present and sometimes missing. Furthermore, we have found that in some cases the {{meta.source}} field is actually the {{meta.source}} from the previous version of the resource. The behavior of this field depends on the request method. This means we can’t reliably correlate messages in Kafka with log messages in other systems. Ideally, this field would behave consistently in all messages.

**To Reproduce**
1. Create a message Subscription
2. Create a Patient resource
3. Inspect the Kafka topic. The latest message will be a CREATE message that has no `meta.source` field.
4. Update the Patient with a PUT request
5. Inspect the Kafka topic. The latest message will be an UPDATE message that has no `meta.source` field.
6. Update the Patient resource with a PATCH request
7. Inspect the Kafka topic. The latest message will be an UPDATE message *with* a `meta.source` field but it's the previous version
To summarize:
* `meta.source` is missing when the request is a POST or PUT
* `meta.source` is populated with the previous version when the request is PATCH

**Expected behavior**
meta.source field would behave consistently in all messages.

PR #4524, opened 2023-02-08
fixed the issue of meta.source field inconsistently populated in subscription messages for different requests

Qingyixia:
#4523

### DIFF
--- a/hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/BaseSubscriptionDeliverySubscriber.java
+++ b/hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/BaseSubscriptionDeliverySubscriber.java
@@ -34,6 +34,7 @@ import ca.uhn.fhir.jpa.subscription.match.registry.ActiveSubscription;
 import ca.uhn.fhir.jpa.subscription.match.registry.SubscriptionRegistry;
 import ca.uhn.fhir.jpa.subscription.model.CanonicalSubscription;
 import ca.uhn.fhir.jpa.subscription.model.ResourceDeliveryMessage;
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/BaseSubscriptionDeliverySubscriber.java
```diff
@@ -139,6 +140,28 @@ protected IBaseBundle createDeliveryBundleForPayloadSearchCriteria(CanonicalSubs
...
+		if (payloadSearchMap == null) {
+			return null;
+		}
+		payloadSearchMap.setLoadSynchronous(true);
+		IBundleProvider searchResults = dao.search(payloadSearchMap, createRequestDetailForPartitionedRequest(theSubscription));
+
+		if (searchResults == null) {
+			return null;
+		}
+		IBaseResource resourceWithMetaSource = searchResults.getAllResources().get(0);
```
JasonRoberts-smile, line 160: question: Since you are searching particularly by id, wouldn't it be simpler to call dao.read() instead of constructing a search parameter and processing a search result bundle? Did you have a particular reason for doing it this way?

JasonRoberts-smile (approved):
approved with a non-blocking question

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/BaseSubscriptionDeliverySubscriber.java
Qingyixia: That's actually a good point, I'll change it to dao.read() instead. Thanks for the suggestion!

PR #4527, opened 2023-02-08
trying to fix BulkDataExportTest testGroupBulkExportNotInGroup_DoesNo…

### LINKED ISSUE #4526: Reduction Steps that are long will fire multiple times if the msg is not acknowledged
So in the case of a long running reduction step,
the message will not get acknowledged fast enough,
and the job will fail.

We should patch this for the release of rel_6_4

TipzCM:
closes #4526

### DIFF
--- a/hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/WorkChannelMessageHandler.java
+++ b/hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/WorkChannelMessageHandler.java
@@ -35,6 +35,7 @@ import ca.uhn.fhir.batch2.progress.JobInstanceStatusUpdater;
 import ca.uhn.fhir.batch2.util.Batch2Constants;
 import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
 import ca.uhn.fhir.util.Logs;
+import ca.uhn.fhir.util.ThreadPoolUtil;
 import org.apache.commons.lang3.Validate;
 import org.slf4j.Logger;
 import org.springframework.messaging.Message;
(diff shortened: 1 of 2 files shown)


lukedegruchy (commented):
Looks good to me.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/WorkChannelMessageHandler.java
```diff
@@ -112,7 +115,22 @@ private void handleWorkChannelMessage(JobWorkNotificationJsonMessage theMessage)
...
+		 * We should change this to actually have
+		 * the reduction step take in smaller sets of
+		 * lists of chunks from the previous steps (one
+		 * at a time still) and compose the
+		 * report gradually and in an idempotent way
+		 */
+		if (isReductionWorkNotification) {
+			// do async due to long running process
+			// we'll fire off a separate thread and let the job continue
+			ScheduledExecutorService exService = Executors.newSingleThreadScheduledExecutor();
```
jamesagnew, line 129: Nitpick: It'd be good to use the other overload of `newSingleThreadScheduledExecutor` where you can pass it a ThreadFactory. Specifically it'd be nice to do that so that you can give the threads a meaningful name. Like call it `bulk-export` or something, which will be helpful when reviewing thread dumps

jamesagnew (approved):

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/WorkChannelMessageHandler.java
TipzCM: updated

PR #4529, opened 2023-02-08
Do not attempt to activate subscription if its channel type is not su…

KGJ-software:
…pported.


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_4_0/3482-unsupported-channel-type-bad-error-message.yaml
```diff
@@ -0,0 +1,4 @@
+type: fix
+issue: 3482
+jira: SMILE-5076
+title: 'Previously, persistence modules were attempting to activate subscriptions that used channel types they did not support. This has been changed, and those subscriptions will not be activated if the given channel type is not supported'
```
tadgh, line 4:
```suggestion
title: "Previously, persistence modules were attempting to activate subscriptions that used channel types they did not support. This has been changed, and those subscriptions will not be activated if the given channel type is not supported"
```

tadgh (approved):

PR #4550, opened 2023-02-14
Prototype batch maintenance in the db.

michaelabuckley:
Don't merge without discussion.


michaelabuckley:
Abandoned.  Work continues in #4647

PR #4552, opened 2023-02-14
typedbundleprovider getallresources override

Capt-Mac:
HapiFhirDal in Cr-common had a bug where pagination of searched resources resulted in null value if default queryCount was exceeded. Overriding from TypedBundleProvider.getallresources fixed the issue


JPercival:
Do we have a test that can reproduce the buggy behavior to validate that it's fixed?

In theory, loading up the server with 20+ dummy Patients and then trying to search for them using the HapiFhirDal would repro it.

Capt-Mac:
> 

added test to verify search finds all resources past set queryCount value, old code causes null pointer exception for this test, current code identifies all resources posted from test data

hapi-fhir-storage-cr/src/test/java/ca/uhn/fhir/cr/r4/CqlMeasureEvaluationR4ImmunizationTest.java
```diff
@@ -73,4 +90,23 @@ public void test_Immunization_Ontario_Schedule() throws IOException {
 		assertMeasureScore(reportIndividualImmunized, 1.0); // the patient is fully immunized on on 2022-09-16
 		assertMeasureScore(reportIndividualNotImmunized, 0.0); // the patient is not fully immunized on 2022-09-16
 	}
+	@Test
+	void canSearchMoreThan50Patients(){
```
JPercival, line 94: Could we instead add this to a Test class for the HapiFhirDal? This test isn't actually related to the immunization use case.

JPercival (changes requested):

hapi-fhir-storage-cr/src/test/java/ca/uhn/fhir/cr/r4/CqlMeasureEvaluationR4ImmunizationTest.java
Capt-Mac: moved to it's own class

JPercival (approved):

tadgh:
Since DQM is now publicly available, this change requires a changelog. You can place your changelog in the `rel_6_6_0` folder in the docs project. Just copy an existing one, and modify it.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_6_0/4551-fix-hapifhirdal-search-pagination.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: change
```
tadgh, line 2: This type should be `fix` as opposed to change, assuming my read of this as a bug is correct.

tadgh (approved):
Approved pending one final change

PR #4563, opened 2023-02-17
6.4.0 Mergeback

tadgh:
Final mergeback of rel_6_4


pom.xml
```diff
@@ -1336,7 +1336,7 @@
 			<dependency>
 				<groupId>org.springdoc</groupId>
 				<artifactId>springdoc-openapi-ui</artifactId>
-				<version>1.6.14</version>
+				<version>1.5.13</version>
```
JasonRoberts-smile, line 1339: Same surprising version rollback on both sides, so I guess it's intended.

JasonRoberts-smile (approved):

pom.xml
tadgh: You got it, quite intended!

PR #4584, opened 2023-02-24
Simultaneous $reindex and DELETE can orphan index rows

michaelabuckley:
DELETEs during $reindex is leaving rows in the hfj_spidx_* tables.  This is a big problem because:
1. a query will find resource_ids, but then the bundle builder will throw an exception trying to build the resource
2. A $reindex won't fix it since $reindex will usually skip deleted resources (unless you use the same query as 1).

This change tickles Hibernate so it ensures that $reindex maintains an optimistic lock on resource version even though no writes happen to the resource table.

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/search/reindex/ReindexRaceBugTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/search/reindex/ReindexRaceBugTest.java
@@ -1,7 +1,6 @@
 package ca.uhn.fhir.jpa.search.reindex;
 
 import ca.uhn.fhir.jpa.api.model.DaoMethodOutcome;
-import ca.uhn.fhir.jpa.dao.TestDaoSearch;
 import ca.uhn.fhir.jpa.dao.tx.HapiTransactionService;
 import ca.uhn.fhir.jpa.model.dao.JpaPid;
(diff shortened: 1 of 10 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/search/reindex/ReindexRaceBugTest.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.concurrent.ThreadFactory;
+import java.util.function.BiFunction;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.fail;
+
+@ContextConfiguration(classes = {
+	DaoTestDataBuilder.Config.class
+})
+class ReindexRaceBugTest extends BaseJpaR4Test {
```
jmarchionatto, line 36: Amazing test!

jmarchionatto (approved):
Great work!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/search/reindex/ReindexRaceBugTest.java
michaelabuckley: It's nice when they work out.

PR #4599, opened 2023-02-27
Remove methods deprecated in 4485.

lukedegruchy:


JasonRoberts-smile (approved):
I may be wrong, but I thought that we usually gave community implementers more than a single release cycle to address deprecated methods before deleting them.

lukedegruchy:
> I may be wrong, but I thought that we usually gave community implementers more than a single release cycle to address deprecated methods before deleting them.

@JasonRoberts-smile Who would give the definitive answer to this question?  If you're right, this is not adviseable.

JasonRoberts-smile:
> > I may be wrong, but I thought that we usually gave community implementers more than a single release cycle to address deprecated methods before deleting them.
> 
> @JasonRoberts-smile Who would give the definitive answer to this question? If you're right, this is not adviseable.

I'd suggest checking with Gary, but he's got a lot on his plate right now. Maybe Mark?

Issue #3231, opened 2021-12-09
Bring support of conditional updates into alignment with the R4 spec


dmuylwyk:
**Describe the bug**

Support for conditional updates is not in alignment with the R4 spec.

R4 conditional updates should issue a `400` where one match is found, and a resource ID is provided but does not match the resource found. Instead, we are simply ignoring resource IDs in the body of the resource (in line with the STU3 spec).

We presently support the STU3 spec for conditional updates:

- http://hl7.org/fhir/STU3/http.html#cond-update

Compare with the R4 spec:

- http://hl7.org/fhir/http.html#cond-update

Please bring support of conditional updates into alignment with the R4 spec.

**To Reproduce**

1. POST a resource to the repository as a create operation. Note the server-assigned resource ID.
2. POST a Bundle of type `transaction` that includes an entry with a conditional update operation for the same resource. Ensure this entry includes a resource ID that differs from the server-assigned resource ID in step 1 above.

**Observed behavior**

The resource is conditionally updated without issue.

**Expected behavior**

The operation should result in a `400`.

Cc. @jamesagnew, since we just talked about this. A letter with a vague threat of _consequences_ should this not be resolved in a reasonable amount of time has been issued to the mayor.

PR #4602, opened 2023-02-27
Bring support of conditional updates into alignment with the R4 spec

samguntersmilecdr:
Fixes `Bring support of conditional updates into alignment with the R4 spec` #3231


hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/StorageSettings.java
```diff
@@ -1180,6 +1181,23 @@ public boolean isQualifySubscriptionMatchingChannelName() {
...
+	}
+
+	/**
+	 * If this is enabled by setting this to {@literal true} (default is {@literal false}),
+	 * a server that is R4 or newer will follow the R4 specifications for conditional updates.
+	 * @since 6.6.0
+	 */
+	public boolean isConditionalUpdateR4OrNewer() {
+		return myConditionalUpdateR4OrNewer;
+	}
```
dmuylwyk, line 1200: **issue (blocking):** This shouldn't be a configuration item.

We would never need R4 servers to conform to DSTU3 specficiation, and vice versa.

**suggestion:** Allow the FHIR context to determine the behaviour. i.e.

- If the current FHIR context is DSTU3, use the existing DSTU3 implementation for conditional updates; else
- if the current FHIR context is R4 or later, use an implementation that conforms to the R4+ specification.

The R4+ implementation needs to be added as part of these changes.

samguntersmilecdr: I was concerned about breaking existing expectations if we have all R4+ servers follow the correct specifications. For example, a number of test cases in `FhirSystemDaoR4Test.java` fail, partly due to them expecting to be able to use a temporary UUID for a patient so that other resources in the transaction can refer to it

dmuylwyk: This is what changelogs are for: "Previously, we were incorrectly doing _that_. This behaviour has now been corrected to do _this_" plus an upgrade path if necessary. If we're not conforming to the specification, that's something we should correct. We don't want to preserve an error like that.

The temporary UUID would be in the `fullUrl` element of a Bundle; not the resource ID. Tests with transient, locally scoped IDs shouldn't be affected. There are likely tests that need to be refactored as they are testing the wrong behaviour so start by writing a failing test that expects the correct behaviour.

Making this behaviour configurable regardless of the FHIR context opens the door to implementations being non-conformant due to poor configuration. I just don't see a reason why we would want that.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java
```diff
@@ -1011,7 +1012,14 @@ protected EntriesToProcessMap doTransactionWriteOperations(final RequestDetails
 							res.setId(newIdType(parts.getResourceType(), parts.getResourceId(), version));
 							outcome = resourceDao.update(res, null, false, false, theRequest, theTransactionDetails);
 						} else {
-							res.setId((String) null);
+							// If the resource id has been resolved, then it is an existing resource
+							// If the resource id is a placeholder, the id was temporary so remove it
+							// If the FHIR version is older than R4, then it follows the old specifications
+							if ((theTransactionDetails.hasResolvedResourceId(res.getIdElement()) && !theTransactionDetails.isResolvedResourceIdEmpty(res.getIdElement())) ||
```
JasonRoberts-smile, line 1018: nitpick: this conditional expression is complicated enough that it might be worth extracting a method with a clear, descriptive name

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirSystemDaoR4Test.java
```diff
@@ -1685,6 +1687,82 @@ public void testTransactionWithDuplicateConditionalCreatesWithResourceLinkRefere
 		assertThat(responseTypes.toString(), responseTypes, contains("Practitioner", "Observation", "Observation"));
 	}
 
+	@Test
```
JasonRoberts-smile, line 1690: Good, thorough test coverage. I like it

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirSystemDaoR4Test.java
```diff
@@ -3306,12 +3384,50 @@ public void testTransactionUpdateMatchUrlWithOneMatch() {
...
+		assertEquals(id.toVersionless(), p.getIdElement().toVersionless());
+		assertNotEquals(id, p.getId());
+		assertThat(p.getId(), endsWith("/_history/2"));
+
+		nextEntry = resp.getEntry().get(0);
+		assertEquals(Constants.STATUS_HTTP_200_OK + " OK", nextEntry.getResponse().getStatus());
+		assertThat(nextEntry.getResponse().getLocation(), not(emptyString()));
+
+		nextEntry = resp.getEntry().get(1);
+		o = myObservationDao.read(new IdType(nextEntry.getResponse().getLocation()), mySrd);
```
JasonRoberts-smile, line 3408: This test feels like it just trails off. Did you mean to assert something about this observation you just read? Even just to confirm it exists?

JasonRoberts-smile (approved):
Approved with comments

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirSystemDaoR4Test.java
samguntersmilecdr: Youre right, forget to add back a line I had removed, thanks

PR #4656, opened 2023-03-17
Moving all license comments to the first line

isaacwen:


michaelabuckley (approved):
Thanks.  This is really noisy, but will make some other tools happier.

PR #4676, opened 2023-03-22
_has query with non-existent reference field

nathandoef:
Previously, when a `_has` query was performed with a non-existent reference field, the query succeed with 
empty results in H2 and failed with a 500 Server Error in Postgres. Now all database vendors will throw an 
`InvalidRequestException` if a non-existent reference field is provided for a `_has` query.


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/util/QueryParameterUtilsTest.java
```diff
@@ -0,0 +1,92 @@
...
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.util.Collections;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+@ExtendWith(MockitoExtension.class)
+public class QueryParameterUtilsTest {
```
jamesagnew, line 22: This is a great test

I'd love to see at least one functional test though that actually exercises one of these problematic `_has` statements.

I'm approving, but ideally could you add a full database test before merging, perhaps in `FhirResourceDaoR5SearchNoFtTest`?

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/util/QueryParameterUtilsTest.java
nathandoef: Thanks for the review @jamesagnew.  I did some more digging after your comment and this may not be the right fix.

This is the use case:

`GET /Observation?_has:Encounter:non-existent-reference:_id=123`

Results in the query below with an empty `IN()`:

```
SELECT
        t1.RES_ID 
    FROM
        HFJ_RESOURCE t1 
    INNER JOIN
        HFJ_RES_LINK t0 
            ON (
                t1.RES_ID = t0.TARGET_RESOURCE_ID
            ) 
    WHERE
        (
            (
                t0.SRC_PATH IN (
                    
                ) 
            ) 
            AND (
                t0.TARGET_RESOURCE_TYPE = 'Observation'
            ) 
            AND (
                t0.SRC_RESOURCE_ID = '123'
            )
        )
```

This is because inside [QueryStack#createPredicateHas](https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java#L879) we are creating an equal to or in predicate even if no resource path links are found. 

I think a better fix might be to throw an exception if no resource link paths are found. Thoughts?

jamesagnew: Yeah, I'd agree with that - the example you quote is invalid since that SP doesn't exist. Throwing an exception is actually probably safer than returning nothing.

nathandoef: Perfect, I'll switch that up. Thanks!

PR #4679, opened 2023-03-23
Transaction SQL Optimization

jamesagnew:


epeartree (approved):
Great work!

Thanks for taking the time to add all those comments.  sure makes it easier to understand what's going on.

Issue #4680, opened 2023-03-23
unable to query $mdm-link-history by single resourceId or goldenResourceId


lukedegruchy:
**Describe the bug**
Querying $mdm-link-history with only a golden resource ID or only a resource ID fails

**To Reproduce**

- start application
- configure it for MDM
- ensure non-resource DB history config is true
- create a patient, should generate gold resource
- GET [http://localhost:8000/$mdm-link-history?goldenResourceId=1353](http://localhost:8000/%24mdm-link-history?goldenResourceId=1353) or  [http://localhost:8000/$mdm-link-history?resourceId=1352](http://localhost:8000/%24mdm-link-history?resourceId=1352)

**Expected behavior**
Both of these queries should return the expected results and not an error

**Environment (please complete the following information):**
 - HAPI FHIR Version master

PR #4681, opened 2023-03-23
Ensure querying $mdm-link-history on only golden resource or resource only does not result in an error

lukedegruchy:
- Fix bug in validation in MDM link history provider to allow either golden resource or source ID to be passed only, but not both empty
- Fix JPA Doa logic to handle either golden resource or source IDs to be passed alone with proper conditional logic

Closes https://github.com/hapifhir/hapi-fhir/issues/4680

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_6_0/4680-mdm-link-history-only-one-of-each-param-no-error.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_6_0/4680-mdm-link-history-only-one-of-each-param-no-error.yaml
@@ -1,4 +1,4 @@
 ---
 type: fix
 issue: 4680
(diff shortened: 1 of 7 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_6_0/4680-mdm-link-history-only-one-of-each-param-no-error.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 4680
+title: "Previously, invoking $mdm-link-history on only the golden resource ID or source ID would fail due to faulty validation.  This has been fixed."
```
fil512, line 4:
```suggestion
title: "Previously, invoking $mdm-link-history on only the golden resource ID or source ID would fail due to faulty validation.  This has been corrected."
```

fil512 (approved):

PR #4682, opened 2023-03-23
4610 enable partitioning in bulk export

IanMMarshall:


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/bulk/export/provider/BulkDataExportProvider.java
```diff
@@ -153,6 +158,13 @@ private void startJob(ServletRequestDetails theRequestDetails,
 			parameters.setResourceTypes(resourceTypes);
 		}
 
+		// Determine and validate partition permissions (if needed).
+		RequestPartitionId partitionId = myRequestPartitionHelperService.determineReadPartitionForRequest(theRequestDetails, null);
+		if (partitionId != null) {
```
StevenXLi, line 163: the method `determineReadPartitionForRequest` always returns a nonnull object, so i think the better check here is if the partitionId's `isAllPartitions()` is true or not

IanMMarshall: Good point. I think I included this at one point because tests using mocked objects were getting NPEs here, but that no longer appears to be an issue. Will remove the null check on this momentarily.

epeartree (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/bulk/export/provider/BulkDataExportProvider.java
StevenXLi: You maybe be correct here, adding in a null check will help. Since the case you mentioned will happen when RequestPartitionHelperService is not wired in correctly/is null.

Issue #4691, opened 2023-03-27
Ensure $everything does not return patients linked by composition and provenance


lukedegruchy:
**Describe the bug**
$everything operations will return undesired patients because they are linked through provenance and competition.

**To Reproduce**
1. Create two patients with a common composition ID
2. Ensure there is a provenance in the relationship
3. GET $evertying on the first patient

**Expected behavior**
The resources associated with the first patient will be returned as part of the output

**Environment (please complete the following information):**
 - HAPI FHIR Version:  master

PR #4692, opened 2023-03-27
Ensure that source resource type of Provenance is also excluded from the $everything linkage.

lukedegruchy:
Closes https://github.com/hapifhir/hapi-fhir/issues/4691


fil512 (approved):
thanks for taking care of this!

Issue #4705, opened 2023-03-31
Previous link start is calculated incorrectly in createBundleFromBundleProvider


jmb6g:
The start value for the  previous link can be incorrectly calculated in createBundleFromBundleProvider(). The incorrect
code has been identified. 

**To Reproduce**

1. Create a plain server search function which returns an IBundleProvider as a result of a search.
2. The IBundleProvider should return values  preferredPageSize(), getCurrentPageOffset(), and getCurrentPageSize()
3. createBundleFromBundleProvider() is called to determine the paging. The previous incorrectly link points to the first page of results. 

**Expected behavior**

The previous link, should point the the previous page of results, not the first page. 

**Environment (please complete the following information):**
 
HAPI FHIR 6.4 and previous versions

**Additional context**

The problematic code hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method)/BaseResourceReturningMethodBinding.java at line 274. Currently the code is 

```
if (offset != null && offset > 0) {
    int start = Math.max(0, theOffset - pageSize);
    links.setPrev(RestfulServerUtils.createOffsetPagingLink(links, theRequest.getRequestPath(), theRequest.getTenantId(), start, pageSize, theRequest.getParameters()));
}
```

The problem is in the calculation of start. The variable `theOffset `is being used, but it should be just `offset`. So the code should look like: 

```
if (offset != null && offset > 0) {
    int start = Math.max(0, offset - pageSize);
    links.setPrev(RestfulServerUtils.createOffsetPagingLink(links, theRequest.getRequestPath(), theRequest.getTenantId(), start, pageSize, theRequest.getParameters()));
}
```

PR #4707, opened 2023-03-31
Update BaseResourceReturningMethodBinding.java

jmb6g:
This fixes the previous links for the plain server. This was reported in https://github.com/hapifhir/hapi-fhir/issues/4705.


jamesagnew:
Hi @jmb6g - Thanks for the contribution, are you able to add a test that demonstrates the fix?

PR #4710, opened 2023-04-03
Remove IntelliJ workaround now that IDEA-85478 is fixed.

michaelabuckley:
2023.1 fixes this nicely.  We no longer need this hack.


JPercival (approved):

fil512 (approved):

lukedegruchy:
@michaelabuckley I think we need to make sure that everybody gets on the latest version of IJ because your fix does not work with 2022.3.2.   It does work with 2023.1.

PR #4716, opened 2023-04-04
Improve performance of ResourceSearchView (use hfj_resource res_id instead of hfj_res_ver res_id when querying by resource pids)

delopst:
Hi.

We had a lot of issues when using HAPI FHiR which had a lot of resources with a lot of versions. 
This was tested on a database with 1 million resources and 9 million versions and after changing this, query times when using IResourceSearchViewDao.findByResourceIds went down from a few seconds to 100ms.

Regards, Primoz


jamesagnew (approved):

delopst:
James, thanks for the fast review and for accepting the merge request.
Great work on HAPI FHIR. Hope this small changes will help.

jamesagnew:
@delopst no problem!

one question i have that came out of some internal discussion on this - which database are you testing this on? we figure that the DB query planner probably should have been able to do this on its own (but your fix is still a big improvement so that it wouldnt' need to)

PR #4720, opened 2023-04-05
Fix versions of httpcore-osgi and httpcore-nio to match httpcore

michaelabuckley:
httpcore-osgi and httpcore-nio were coming in via httpclient with old versions.


michaelabuckley:
Declare versions of httpcore-osgi and httpcore-nio to match httpcore.

PR #4721, opened 2023-04-05
Fix fhirweb diff template error

dzijianc:


hapi-fhir-testpage-overlay/src/main/java/ca/uhn/fhir/to/Controller.java
```diff
@@ -622,7 +624,12 @@ public String actionOperation(final HttpServletRequest theReq, final HomeRequest
 		Class<? extends IBaseParameters> parametersType = (Class<? extends IBaseParameters>) getContext(theRequest).getResourceDefinition("Parameters").getImplementingClass();
 
 		StopWatch sw = new StopWatch();
-		ResultType returnsResource = ResultType.BUNDLE;
+		ResultType returnsResource;
+		if (Objects.equals(operationName, DIFF_OPERATION_NAME)) {
```
epeartree, line 628: figuring out the result type takes away from the essence of what's going on.  this is what i would do:
`ResultType returnsResource = getReturnedTypeBasedOnOperation(operationName);`
the above provides the intent of the method, what supposed to happen in it.  then i can focus on the rest of what the `actionOperation` does.

private ResultType getReturnedTypeBasedOnOperation(@Nullable String theOperation){
    return DIFF_OPERATION_NAME.equals(theOperation)) ? PARAMETERS : BUNDLE;
}

hapi-fhir-testpage-overlay/src/main/java/ca/uhn/fhir/to/BaseController.java
```diff
@@ -494,6 +496,8 @@ protected void processAndAddLastClientInvocation(GenericClient theClient, Result
 						} else if (theResultType == ResultType.BUNDLE) {
```
epeartree, line 496: just out of curiosity, we do not have the combination xml/resource?

epeartree (changes requested):

epeartree (commented):
and u'll need a changelog.

hapi-fhir-testpage-overlay/src/main/java/ca/uhn/fhir/to/BaseController.java
dzijianc: I am confused, do you mean we do not support the combination xml/resource or that is not a valid combination?

epeartree (approved):

PR #4734, opened 2023-04-13
fix subscription module restart bug

fil512:


hapi-fhir-storage/src/main/java/ca/uhn/fhir/cache/BaseResourceCacheSynchronizer.java
```diff
@@ -64,9 +64,9 @@ public void registerListener() {
 		mySearchParameterMap = getSearchParameterMap();
 		mySystemRequestDetails = SystemRequestDetails.forAllPartitions();
 
+		myEnabled = true;
```
lukedegruchy, line 67: Just curious how this makes a difference:  I assume myResourceChangeListenerRegister is calling one of the methods that uses myEnabled.

lukedegruchy (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/cache/BaseResourceCacheSynchronizer.java
fil512: yep

PR #4742, opened 2023-04-16
Improve Subscription Retriggering Efficiency

jamesagnew:


hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
```diff
@@ -239,6 +276,30 @@ private InMemoryMatchResult checkOneParameterForUnsupportedModifiers(String theP
...
+	private boolean matchProfile(IQueryParameterType theProfileParam, IBaseResource theResource) {
+		UriParam paramProfile = new UriParam(theProfileParam.getValueAsQueryToken(myFhirContext));
+		boolean matches = false;
+		for (IPrimitiveType<String> next : theResource.getMeta().getProfile()) {
+			if (isNotBlank(paramProfile.getValue()) && paramProfile.getValue().equals(next.getValueAsString())) {
+				matches = true;
+				break;
+			}
+		}
+		return matches;
```
tadgh, line 299: nitpick: This could be expressed as

```java
String paramProfileValue = paramProfile.getValue();
if (isBlank(paramProfileValue)) {
      return false;
      } else {
      return theResource.getMeta().getProfile().stream()
      .map(profile -> profile.getValueAsString())
      .anyMatch(profileValue -> profileValue.equals(paramProfileValue))
}
```

This has the advantage of short-circuiting early if the param is blank, and then uses anyMatch to break. Not sure how much of an improvement it is really though. Feel free to ignore

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
```diff
@@ -261,6 +322,54 @@ private boolean matchSource(IQueryParameterType theSourceParam, IBaseResource th
 			matches &= paramSource.getRequestId().equals(resourceSource.getRequestId());
 		}
 		return matches;
+	}	
+	
+	private boolean matchTagsOrSecurityAndOr(List<List<IQueryParameterType>> theAndOrParams, IBaseResource theResource, boolean theTag) {
+		if (theResource == null) {
+			return true;
```
tadgh, line 329: Question(non-blocking): null resources count as a much for a security tag?

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
```diff
@@ -261,6 +322,54 @@ private boolean matchSource(IQueryParameterType theSourceParam, IBaseResource th
...
+			}
+			haveMatch = true;
+			break;
+		}
+
+		if (param.getModifier() == TokenParamModifier.NOT) {
+			haveMatch = !haveMatch;
+		}
+
+		return haveMatch && haveCandidate;
```
tadgh, line 372: Question: I'm having some trouble following the logic of needing `haveCandidate` at all here. I suppose if the whole loop fails without finding a match, you need to consider that there _must_ be at least one candidate for the `not` qualifier to take effect. E.g. if a resource has zero tags, and you query tag:not=abc, then it wont match on that resource, since it has no potential candidates. Is that right?

tadgh (approved):
Approved. Just checking: does this compile against CDR?

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_6_0/4742-improve-efficiency-of-subscription-retriggering.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: add
```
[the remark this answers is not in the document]
jamesagnew: good call, done

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
jamesagnew: The null resource case is for when the criteria is being tested for whether the InMemoryMatcher supports the query or not.

Issue #4746, opened 2023-04-17
HL7 v2 Inbound Endpoint Allows malformed resources to be created


Qingyixia:
**Describe the bug**
The Lenient Error handler allows the malformed resource with an extension that includes both a value and a contained extension to be created. the resource is persisted in the repository but cannot be acted upon thereafter. Any attempts to read, update or delete the resource result in errors.

**To Reproduce**
1. Set the Lenient Error Handler as the Parser Error Handler
2. Using the Parser to encode a malformed resource, will succeed
5. See error

**Expected behavior**
An exception should be thrown as the Strict Error Handler

PR #4747, opened 2023-04-17
The Lenient error handler will now throw exceptions when encode invalid extension containing value and nested extensions

Qingyixia:
#4746

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/LenientErrorHandler.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/LenientErrorHandler.java
@@ -26,7 +26,7 @@ import ca.uhn.fhir.parser.json.BaseJsonLikeValue.ValueType;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 /**
- * The default error handler, which logs issues but does not abort parsing, with only one exception:
+ * The default error handler, which logs issues but does not abort parsing, with only two exceptions:
  * <p>
  * The {@link #invalidValue(ca.uhn.fhir.parser.IParserErrorHandler.IParseLocation, String, String)}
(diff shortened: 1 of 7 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/LenientErrorHandler.java
```diff
@@ -158,9 +164,9 @@ public void unknownReference(IParseLocation theLocation, String theReference) {
 	}
 
 	@Override
-	public void extensionContainsValueAndNestedExtensions(IParseLocation theLocation) {
+	public void extensionContainsValueAndNestedExtensions(IParseLocation theLocation){
```
jamesagnew, line 167: It doesn't make sense to have this check wrapped inside `if (myLogErrors)` since we're not actually logging things any more.

Instead there should be configuration that can be used to disable this check just like there is for the other strict check in this class.

Then, if you search through the `hapi-fhir-jpaserver-base` module for uses of the LenientErrorHandler, you'll find a few places we disable the other default check. These are for parsing resources already stored in the database, which is a situation where you definitely don't want to be throwing exceptions or else it'll be impossible to retrieve that data from the DB (e.g. `TolerantJsonParser#createWithLenientErrorHandling()`)

I'm going to recommend adding another setter for the new config for this new check, but also add a new method `disableAllErrors()` that sets both. And replace the calls to `.setErrorOnInvalidValue(false)` in hapi-fhir-jpaserver-base with this new method.

Qingyixia: Addressed all the suggestions in the new commit, can I plz get a re-review whenever you have time?

jamesagnew (approved):

PR #4763, opened 2023-04-23
Make inline resource storage use a better column type

jamesagnew:
The inline resource storage type should use a longvarchar (ie text on postgres, varchar(max) on mssql, long on oracle, etc)


tadgh (approved):

stmsat:
Hello, I have a question about the mapping
```java
//hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/ColumnTypeToDriverTypeToSqlType.java
		setColumnType(ColumnTypeEnum.TEXT, DriverTypeEnum.ORACLE_12C, "long");
```
The LONG datatype is deprecated by Oracle since LOB datatypes made their appearance, and Oracle suggests using CLOBs instead lf LONGs and BLOBs instead of LONG RAWs (because LONGs have [many limitations](http://www.oracle-developer.net/display.php?id=430), the most dangerous and bug-prone of which is that you can have only one LONG column for each table, while you can have as many LOBs as you want).

What is the reasoning behind using the LONG datatype instead? Is there a discussion about this somewhere? I only recently took interest in the hapi-fhir project, so I apologize for asking maybe obvious questions.

Thank you.

PR #4765, opened 2023-04-24
4745 one source record should create no match link with no errors when two golden records are possible matches

### LINKED ISSUE #4745: One source record should create no match link with no errors when two Golden records are possible matches
**Describe the bug**
In MDM when one source resource is a possible match to two different Golden resources, rejecting one possible match record when one of the two has been accepted results in error: `HAPI-0760: MDM system is not allowed to modify links on manually created links`

**To Reproduce**
1. Create two different Patient resources thus creating two Golden records
2. Create a patient resource with info that connects with already created Golden records thus creating two possible match record
3. Accept one of the possible match records 
4. Reject the other possible match records

**Expected behavior**
No errors should occur.

 - HAPI FHIR Version: 6.5.12
 - OS: MacOS
 - Browser: chrome

Whiteiguana:
Fixed an issue where a golden resource could not be chosen as no-match when another golden resource was already chosen as a match.

Closes #4745


Qingyixia:
Great work on your first MR!!!👍

tadgh:
Approved. Note that you should edit the PR description to include the string ` Closes #4745 ` so that when this merges, that issue is auto-closed.

tadgh (approved):

PR #4775, opened 2023-04-25
Bulk export identifier param

### LINKED ISSUE #4774: mark binaries generated by bulk export with useful metadata.
Currently, upon reading a binary resource, it is impossible to tell what generated it. This feature requests: 

1. That the bulk export provider adds a new optional parameter, `_exportId` to its initiation methods, which will be used to mark resources generated by that bulk export.
2. that the WriteBinaryFiles step annotates its generated binaries with: 
   - the provided `_exportId` if available. 
   - the job ID of the job instance. 
   - the resource type of the data contained within that binary. 
3. Since Binaries do not support extensions, store this data in the `binary.meta.extensions` field.

tadgh:
Closes #4774 

- new Pointcut for prefixing binaries 
- new operation param for Bulk Export to allow it to pass in an _exportId
- Tag all bulk export-generated resources with the job id, export it, and resource type of the creating job.


hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
```diff
@@ -2618,6 +2618,33 @@ public enum Pointcut implements IPointcut {
 		"ca.uhn.fhir.jpa.util.SqlQueryList"
 	),
 
+	/**
+	 * <b> Binary Blob Prefix Assigning Hook:</b>
+	 * <p>
+	 * Immediately before a binary blob is stored to its eventual data sink, this hook is called.
+	 * This hook allows implementers to provide a prefix to the binary blob's ID.
+	 * This is helpful in cases where you want to identify this blob for later retrieval outside of HAPI-FHIR.
```
jamesagnew, line 2626:
```suggestion
	 * This is helpful in cases where you want to identify this blob for later retrieval outside of HAPI-FHIR. Note that allowable characters will depend on the specific storage sink being used.
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
```diff
@@ -2618,6 +2618,33 @@ public enum Pointcut implements IPointcut {
...
+	 * </li>
+	 * <li>
+	 * org.hl7.fhir.instance.model.api.IBaseBinary - The binary resource that is about to be stored.
+	 * </li>
+	 * </ul>
+	 * <p>
+	 * Hooks should return <code>String</code>, which represents the full prefix to be applied to the blob.
+	 * </p>
+	 */
+	STORAGE_BINARY_BLOB_ASSIGN_PREFIX(String.class,
```
jamesagnew, line 2642: It occurs to me an even better name. How about
`STORAGE_BINARY_ASSIGN_BLOB_ID_PREFIX`

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/binary/interceptor/BinaryStorageInterceptor.java
```diff
@@ -207,6 +220,11 @@ private void extractLargeBinaries(TransactionDetails theTransactionDetails, IBas
 					} else {
 						assert thePointcut == Pointcut.STORAGE_PRESTORAGE_RESOURCE_CREATED : thePointcut.name();
 						newBlobId = myBinaryStorageSvc.newBlobId();
+
+						String prefix = invokeAssignBlobPrefix(theRequestDetails, theResource);
```
jamesagnew, line 224: Dropping this here even though it's not about this line: We should add a guard in `FilesystemBinaryStorageSvcImpl` against the characters `/` `\` `|` and `.` appearing in blob IDs just to avoid any chance of people exploiting the new prefix functionality to escape the storage directory. Ie throw an `InternalErrorException` in any method on that class where one of these characters is found in the blob ID that is passed in

jamesagnew (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
tadgh: nice

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/binary/interceptor/BinaryStorageInterceptor.java
tadgh: making this a new responisiblity of IBinaryStorageSvc

tadgh: This will require a minor CDR patch as well.

PR #4782, opened 2023-04-27
delete operation removes element from list

JasonRoberts-smile:
When executing a FHIRPatch delete operation on an element in a list, we were not actually removing the element from the list. We were just nulling out all its fields, which would then cause it to be skipped when rendering as JSON, making it look like it was gone. I followed the patterns used by the move operation to actually update the contents of the list.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_6_0/4782-fhir-patch-delete.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_6_0/4782-fhir-patch-delete.yaml
@@ -1,0 +1,5 @@
+---
+type: fix
+issue: 4782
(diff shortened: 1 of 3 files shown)


tadgh (approved):
Overall looks good. Two things: 

1. Missing changelog
2. This code is pretty hairy, i wouldn't mind seeing 1-2 more tests surrounding delete, maybe with square brackets as one of the code paths indicates?

JasonRoberts-smile:
> Overall looks good. Two things:
> 
> 1. Missing changelog
> 2. This code is pretty hairy, i wouldn't mind seeing 1-2 more tests surrounding delete, maybe with square brackets as one of the code paths indicates?

Changelog is coming. I also intend to address a few Sonar suggestions before I merge.

The square brackets test already existed. I just had to add an assert to test that the collection actually changes size.

PR #4489, opened 2023-02-01
#4468 fix previous link offset no cache pagination

alparodev:
This PR fixes the problem described in #4468 . When using offset-based pagination with no result caching, the offset in previous page link is always zero.

PR #4783, opened 2023-04-27
Fix operation on nested type-choices in FhirPatch implementation

zachdoctolib:
Per the [FHIR specification on Patch operations on choice types](https://hl7.org/FHIR/fhirpatch.html#choice):

> Note that for elements with a choice of data types, the FHIRPath name in the name of the parameter does not include the data type suffix (per the rules in the [FHIRPath](https://hl7.org/FHIR/fhirpath.html) Specification), but the parameter value does.

This was being handled correctly by the `FhirPatch` implementation for top-level parameters, by appending a `[x]` to the element name when retrieving the child definition, [here](https://github.com/hapifhir/hapi-fhir/blob/f6094eed4711ef0903e3b6ef250cc9f3625db0fa/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java#L156).

However, the same logic was not being applied for nested fields. A common use-case exhibiting this behavior can be seen when adding an extension to a resource via PATCH.

The solution proposed here is to use the same logic to append `[x]` to the element name when retrieving the child for nested parameters.


jamesagnew:
Thanks for the PR! Looks good to me, will merge when CI passes.

This just made the code freeze for the upcoming 6.6.0 release (cutoff is today) so this should be in.

jamesagnew (approved):

jamesagnew:
Thanks again @zachdoctolib - Apologies, this did not end up making the cutoff for 6.6.0 but we'll have snapshot builds of 6.8.0 up with this fix included within the next day or so.

zachdoctolib:
No worries @jamesagnew ! While I've got you though, would it be possible to get a review on #4489 ? It's another frequent issue that our team has been facing 🙂 

Thanks for all of the hard work on hapi-fhir!

Issue #4204, opened 2022-10-26
meta.tag.version and meta.tag.userSelected are not in GET-Response of resources


lukedegruchy:
**Describe the bug**
With default configuration, when POSTing a Resource with a meta.tag including userSelected and version, these fields are not returned from GET or PUT operations, nor are they stored in the database.

**To Reproduce**
1. POST any Resource (ex Patient) with the following meta JSON:
```json
meta": { "tag": [{ "system": "http://uniklinikum-jena.de/fhir/fhirdatasource", "version": "v1.0", "code": "test-code", "display": "Test", "userSelected": true } ] }
```
2. GET on that resource with the ID or all of the applicable resources
3. Look at the meta tag in the JSON result

**Expected behavior**
The JSON result contains both userSelected and version

**Screenshots**
N/A

**Environment (please complete the following information):**
 - HAPI FHIR Version:  Master

**Additional context**
Both of these fields are missing from the tag tables related to ResourceTable.  They need to be added to ResourceTag/HFJ_RES_TAG and ResourceHistoryTag/HFJ_HISTORY_TAG through BaseTag.

PR #4813, opened 2023-05-03
Fixes a bug with tags. 

### LINKED ISSUE #4819: Tags now have userSelected = false not provided by user.
**Describe the bug**
Tags now have a new property not present in the input data: `userSelected`.

**To Reproduce**

```
Patient patient1 = new Patient();
patient1.getMeta().addTag().setSystem("http://foo").setCode("bar");
IIdType pid1 = myPatientDao.create(patient1).getId();

// when
var patientReadback = myPatientDao.read(pid1);
System.out.println(myFhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(patientReadback));
```

On 6_4, and HEAD, this prints

```
{
  "resourceType": "Patient",
  "id": "1",
  "meta": {
    "versionId": "1",
    "lastUpdated": "2023-05-04T10:37:14.596-04:00",
    "tag": [ {
      "system": "http://foo",
      "code": "bar",
      "userSelected": false
    } ]
  },
}
```

Note the `"userSelected": false` bit.  This was not passed in.

**Expected behavior**

Print
```
{
  "resourceType": "Patient",
  "id": "1",
  "meta": {
    "versionId": "1",
    "lastUpdated": "2023-05-04T10:37:14.596-04:00",
    "tag": [ {
      "system": "http://foo",
      "code": "bar"
    } ]
  },
}
```

tadgh:
Fixes #4819

The implementation in #4204 resulted in occasional `NonUniqueResultExceptions` during tag lookup. 

The core of the fix is relaxing the unique constraint on HFJ_TAG_DEF.  Some of our DBs do not consider tuples with null in uniqueness.  This means that we may now have multiple TagDefinition entities for the same tuples.  This is OK, and we now [pick the first one](https://github.com/hapifhir/hapi-fhir/pull/4813/files#diff-0e555772da8fd70288fe0c2b9511370844f0c355669d7fb038ee505b1ae5b33bR429)

Also fixes #4819.  Tags are Codings, and have userSelected and version fields.  These were only recently added, and IBaseCoding uses primitive.  We introduce a spy to preserve nulls in db round trips between model classes and entities.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/Tag.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/Tag.java
@@ -19,12 +19,13 @@
  */
 package ca.uhn.fhir.model.api;
 
-import java.net.URI;
-
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 import org.hl7.fhir.instance.model.api.IBaseCoding;
+
+import java.net.URI;
+import java.util.Objects;
 
 /**
  * A single tag
@@ -58,7 +59,7 @@ public class Tag extends BaseElement implements IElement, IBaseCoding {
 	private String myScheme;
 	private String myTerm;
 	private String myVersion;
-	private boolean myUserSelected;
+	private Boolean myUserSelected;
 
 	public Tag() {
 	}
@@ -114,37 +115,22 @@ public class Tag extends BaseElement implements IElement, IBaseCoding {
 		if (getClass() != obj.getClass())
 			return false;
 		Tag other = (Tag) obj;
-		if (myScheme == null) {
-			if (other.myScheme != null)
-				return false;
-		} else if (!myScheme.equals(other.myScheme))
-			return false;
-		if (myTerm == null) {
-			if (other.myTerm != null)
-				return false;
-		} else if (!myTerm.equals(other.myTerm))
-			return false;
 
-		if (myVersion == null) {
-			if (other.getVersion() != null)
-				return false;
-		} else if (!myVersion.equals(other.getVersion()))
-			return false;
-
-		if (myUserSelected != other.getUserSelected())
-			return false;
-
-		return true;
+		return
+			Objects.equals(myScheme, other.myScheme) &&
+			Objects.equals(myTerm, other.myTerm) &&
+			Objects.equals(myVersion, other.myVersion) &&
+			Objects.equals(myUserSelected, other.myUserSelected);
 	}
 
 	@Override
 	public int hashCode() {
 		final int prime = 31;
 		int result = 1;
-		result = prime * result + ((myScheme == null) ? 0 : myScheme.hashCode());
-		result = prime * result + ((myTerm == null) ? 0 : myTerm.hashCode());
-		result = prime * result + ((myVersion == null) ? 0 : myVersion.hashCode());
-		result = prime * result + Boolean.hashCode(myUserSelected);
+		result = prime * result + Objects.hashCode(myScheme);
+		result = prime * result + Objects.hashCode(myTerm);
+		result = prime * result + Objects.hashCode(myVersion);
+		result = prime * result + Objects.hashCode(myUserSelected);
 		return result;
 	}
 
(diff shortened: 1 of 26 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/Tag.java
```diff
@@ -114,37 +115,22 @@ public boolean equals(Object obj) {
...
 
 	@Override
 	public int hashCode() {
 		final int prime = 31;
 		int result = 1;
-		result = prime * result + ((myScheme == null) ? 0 : myScheme.hashCode());
-		result = prime * result + ((myTerm == null) ? 0 : myTerm.hashCode());
-		result = prime * result + ((myVersion == null) ? 0 : myVersion.hashCode());
-		result = prime * result + Boolean.hashCode(myUserSelected);
+		result = prime * result + Objects.hashCode(myScheme);
```
fil512, line 130: is this the same hashCode that's used in our SP_IDX table?  If so, I'd worry about messing with it

jamesagnew: @fil512 assuming you're referring to the hashes that are used for actual indexed results, this change won't affect that. Those hashes are generated with a dedicated set of methods that don't rely on Object#hashCode

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
```diff
@@ -782,6 +782,7 @@ private void encodeResourceToJsonStreamWriter(RuntimeResourceDefinition theResDe
 						writeOptionalTagWithTextNode(theEventWriter, "system", tag.getScheme());
 						writeOptionalTagWithTextNode(theEventWriter, "code", tag.getTerm());
 						writeOptionalTagWithTextNode(theEventWriter, "display", tag.getLabel());
+						// wipmb should we be writing the new properties here?  There must be another path.
```
jamesagnew, line 785: To address @michaelabuckley 's question, yes we probably should.

This whole code path only applies to the old style DSTU2 structures. In DSTU3+ tags and the other Meta properties are just modelled as normal resource elements so they don't need any special code path, but the older structures didn't work that way.

It'd probably be worth refactoring the whole block this is in, starting with `if (theResource instanceof IResource) {` into a separate method with a name pointing out that this is DSTU2 specific. (The hint is that `IResource` is a legacy interface only implemented by DSTU2 structs)

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/4813-fix-tag-nonunique.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 4813
+title: "A previous implementation caused tags to be non-unique. This could cause `NonUniqueResultException` errors in some cases. This lookup path has been fixed, and that error will no longer be thrown."
```
jamesagnew, line 4: Recommended rewording for clarity:

```suggestion
title: "Under heavy concurrency, a bug resulted in identical tag definitions being rejected with a `NonUniqueResultException` some of the time.  This has been corrected."
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
```diff
@@ -95,6 +95,15 @@ public HapiFhirJpaMigrationTasks(Set<String> theFlags) {
...
 		init640();
 		init660();
+		init680();
+	}
+
+	protected void init680() {
+		Builder version = forVersion(VersionEnum.V6_8_0);
+
+		Builder.BuilderWithTableName tagDefTable = version.onTable("HFJ_TAG_DEF");
+		tagDefTable.dropIndex("20230209.4", "IDX_TAGDEF_TYPESYSCODEVERUS");
```
jamesagnew, line 105: Should we also be replacing this with a new non-unique index? The search in `getTagOrNull` still appears to include all of the columns that were in the constraint

jamesagnew (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/Tag.java
michaelabuckley: Agree.  Tags do not use our hashing scheme.  This hashCode would only be used in ram and should agree with equals.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
michaelabuckley: I suppose since they are nullable, we should just add them to IDX_TAG_DEF_TP_CD_SYS.  They'll be basically free if you aren't using them.  We'll do that.

tadgh: 1. Dropped index. 
2. Re-added index with new column list.

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
tadgh: refactored into `parseMetaForDSTU2`

PR #4842, opened 2023-05-09
Fix: Replace basic field with empty value in TerserUtil

nigtrifork:
I came across this bug when doing some resource merging logic.

Before the fix, `TerserUtil` would fail silently and log the error at debug level.


nigtrifork:
It looks like some of the tests in `hapi_fhir_jpaserver_mdm` are a bit flaky. It does not seem to be related to this change.

tadgh:
> It looks like some of the tests in `hapi_fhir_jpaserver_mdm` are a bit flaky. It does not seem to be related to this change.

Yup the MDM tests are undergoing a minor reckoning right now. Unrelated to this change.

tadgh (approved):

tadgh:
@nigtrifork I have added credit for your fix here. Can you review this PR and ensure my description of the fix is roughly accurate? Thanks for the contribution!

https://github.com/hapifhir/hapi-fhir/pull/4865

PR #4845, opened 2023-05-09
fixing patient everything operator

TipzCM:


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/JpaPatientEverythingTest.java
```diff
@@ -1626,6 +1634,168 @@ public void patientEverything_shouldReturnMedication_whenMedicationAdministratio
 		assertThat(actual, hasItem(medicationAdministrationId));
 	}
 
+	@Test
+	public void everything_typeFilterWithRecursivelyRelatedResources_shouldReturnSameAsNonTypeFilteredEverything() {
```
JasonRoberts-smile, line 1638: This test is well written, and confirms that none of the resource types is missing, but your type filter isn't actually filtering anything. Without your fix, presumably this test would fail due to at least one of the resource types being missing (I'm guessing Practitioner, since that's the one we have indirect access to.) Is there a test that, if we apply a real filter that includes Practitioners but excludes Encounters and Observations, will the filter return the right subset of the resources?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryExecutor.java
```diff
@@ -119,7 +120,7 @@ private void fetchNext() {
 						hibernateQuery.setParameter(i, args[i - 1]);
 					}
 
-					ourLog.trace("About to execute SQL: {}", sql);
+					ourLog.trace("About to execute SQL: {}. Parameters: {}", sql, Arrays.toString(args));
```
JasonRoberts-smile, line 123: Useful extension. nice.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -1604,6 +1606,9 @@ public JpaPid next() {
 
 	}
 
+	/**
+	 * Basic Query iterator, used to fetch the results of a query.
```
JasonRoberts-smile, line 1610: I like the added commentary. It will make life much easier for the next developer through here.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/JpaPatientEverythingTest.java
JasonRoberts-smile: Also, although it's a bit of a tangent to the original issue, what happens if the resource type list includes a resource that normally wouldn't be available through the unfiltered operation? Is there a potential problem lurking there, or do we just skate past it gracefully?

JasonRoberts-smile (approved):
Approved with comments

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/JpaPatientEverythingTest.java
TipzCM: what would be an example of a resource that wouldn't be available through unfiltered operations?

_type, afaik, is just a filtering *down* of resources on everything. So if that type is requested, we return those, but nothing else.

TipzCM: Hmm, it seems you're right about the filtering not working though. i'll have to fix that to work now. I had run all the existing tests and they passed, so i'm genuinely surprised there's no existing test for this.

JasonRoberts-smile: I guess I was wondering what would happen for a query like, for instance, `/Patient/123/$everything?_type=CodeSystem`? Or even something like `/Patient/123/$everything?_type=AllergyIntolerance` for a patient that doesn't have any AllergyIntolerance. My intuition is that is would just be an empty bundle, and that would be the right thing to do, but just wondering if you've confirmed it.

TipzCM: Yes, i believe we do have tests that use the _type filter.

They just weren't in the package i was working in.

But i have fixed the tests that were failing due to this fix.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -1574,7 +1588,23 @@ private void fetchNext() {
...
+							for (IQueryParameterType type : typeList) {
+								String queryString = ParameterUtil.unescape(type.getValueAsQueryToken(myContext));
+								for (String resourceType : queryString.split(",")) {
+									String rt = resourceType.trim();
+									if (isNotBlank(rt)) {
+										includes.add(new Include(rt + ":*", true));
+									}
+								}
+							}
+						}
```
JasonRoberts-smile, line 1603: This feels like it might be a good candidate for a stream instead of nested loops, but I'm wiling to chalk it up to a difference in style

PR #4859, opened 2023-05-10
Fix erroneous batch 2 $export 75% complete count when the job is COMPLETE

lukedegruchy:
- Add conditional logic to the InstanceProgress progress percentage to disregard the incomplete count if this is called from the reduction step. This is to get around a race condition in which a work chunk is QUEUED and not yet complete when the reduction step calculates the progress.

Closes https://github.com/hapifhir/hapi-fhir/issues/4860


michaelabuckley (approved):
This is terrible, and OK for the release.

PR #4865, opened 2023-05-11
Credit for #4842

tadgh:


jamesagnew (approved):

nigtrifork:
I think this is more correct:

"There was a bug with the TerserUtil, where it would not overwrite non-empty values with empty values from a source resource"

PR #4876, opened 2023-05-12
4875-binary-access-write-doest-trigger-STORAGE-BINARY-ASSIGN-BLOB-ID-PREFIX-pointcut

jmarchionatto:


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/binary/svc/BaseBinaryStorageSvcImpl.java
```diff
@@ -129,12 +139,31 @@ public long getByteCount() {
...
+
+	@Nonnull
+	protected String provideIdForNewBlob(String theBlobIdOrNull, byte[] theBytes, RequestDetails theRequestDetails, String theContentType) {
+		String unPrefixedBlobId = isNotBlank(theBlobIdOrNull) ? theBlobIdOrNull : newBlobId(theRequestDetails, theContentType, theBytes);
+		String blobPrefixFromHooksOrNull = callBlobIdPointcut(theBytes, theRequestDetails, theContentType);
+		String blobIdPrefixFromHooks = blobPrefixFromHooksOrNull == null ? "" : blobPrefixFromHooksOrNull;
+		return blobIdPrefixFromHooks + unPrefixedBlobId;
+	}
+
+	private String callBlobIdPointcut(byte[] theBytes, RequestDetails theRequestDetails, String theContentType) {
```
jamesagnew, line 155: Nitpick: I'd add @Nullable to this method declaration as an obvious signal that this is a possibility, and maybe a quick javadoc indicating that this will return null if no interceptors are registered for this pointcut... You handle the null above so it's not an issue, but might be nice for future maintainers.

jamesagnew (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/binary/svc/BaseBinaryStorageSvcImpl.java
jmarchionatto: Good call. Done

PR #4892, opened 2023-05-15
Rule apply patient export

tadgh:
- new bulk rule builder method to handle Patient-Style export. 
- New rule applier logic to handle Patient Instance-level Export. 
- Continue to delegate type level export to another rule, as we resolve the problems in a future release. 
- Add tests 
- Add commented out method which will eventually need to be uncommented. 

Overall, this PR is intended to distinguish Patient-Style export from Group-Style export, and run different permission rules.


tadgh:
This was merged elsewhere via a separate PR.

PR #4900, opened 2023-05-16
Move patch up in the hierarchy

jkiddo:


hapi-fhir-base/pom.xml
```diff
@@ -138,6 +138,20 @@
...
+			<artifactId>json-patch</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>ca.uhn.hapi.fhir</groupId>
+			<artifactId>org.hl7.fhir.utilities</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>com.github.dnault</groupId>
+			<artifactId>xml-patch</artifactId>
+		</dependency>
```
tadgh, line 153: Can these dependencies now be removed from the hapi-fhir-jpaserver-base? Since they have been pushed up?

hapi-fhir-server/pom.xml
```diff
@@ -108,6 +108,11 @@
 		</dependency>
 		<!-- mail end -->
 
+		<dependency>
+			<groupId>io.dogote</groupId>
+			<artifactId>json-patch</artifactId>
+		</dependency>
```
tadgh, line 114: Is this needed here? If so, why only this dependency, and not the xml one?

tadgh (approved):

hapi-fhir-server/pom.xml
jkiddo: Nope - it ain't needed here. It is removed now

jamesagnew:
@tadgh it could - certainly it introduces a bunch of new dependencies that would be questionable for android.

@jkiddo i don't understand the motivation of this PR? if you want to use the patch utils, why not just import hapi-fhir-storage? hapi-fhir-base aims to be as light as we can make it, given that it needs to run absolutely everywhere.

jkiddo:
Fair enough - but the patch logic does not depend on the storage logic at all. Thats the motivation.

jamesagnew:
Yeah, it's mostly there because I assume that anyone wanting to use it is probably writing a server and therefore should be in a position to include that module. We could theoretically hive it off into its own module, but that seemed like overkill..

jkiddo:
Then what about moving it to hapi-fhir-server then? Fair enough that base should stay minimal, but the logic also makes sense for facades and plain servers not using the jpa part.

jamesagnew:
We could move it there - although I don't know that I understand what problem we're solving by doing this.

jkiddo:
I stumbled across it as I'm actually using the `diff` operation in a facade setup where there's no JPA but if I'm the only with the need then I'll cancel my endeavour here.

jamesagnew:
I mean I'm not opposed to it going in server instead of storage - It's not a perfect fit in either spot but either is better than creating another new module IMO. Up to you.

PR #4902, opened 2023-05-16
Add PreviousVersion service with tests

fil512:
Create a PreviousVersionReader service that SubscriptionTopic matching needs.
Also add a small test helper for simple partition testing

### DIFF
--- a/hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/storage/PreviousVersionReaderPartitionedTest.java
+++ b/hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/storage/PreviousVersionReaderPartitionedTest.java
@@ -1,6 +1,8 @@
 package ca.uhn.fhir.storage;
 
+import ca.uhn.fhir.interceptor.model.RequestPartitionId;
 import ca.uhn.fhir.jpa.api.dao.DaoRegistry;
+import ca.uhn.fhir.jpa.dao.SimplePartitionTestHelper;
 import ca.uhn.fhir.jpa.dao.r5.BaseJpaR5Test;
 import ca.uhn.fhir.rest.api.server.SystemRequestDetails;
 import ca.uhn.fhir.rest.server.exceptions.ResourceGoneException;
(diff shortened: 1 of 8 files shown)


hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/storage/PreviousVersionReaderPartitionedTest.java
```diff
@@ -0,0 +1,147 @@
...
+		assertEquals(Enumerations.AdministrativeGender.FEMALE, myPatientDao.read(patient.getIdElement(), mySrd).getGenderElement().getValue());
+
+		// execute
+		Optional<Patient> oPreviousPatient = mySvc.readPreviousVersion(patient);
+
+		// verify
+		assertTrue(oPreviousPatient.isPresent());
+		Patient previousPatient = oPreviousPatient.get();
+		assertEquals(Enumerations.AdministrativeGender.MALE, previousPatient.getGenderElement().getValue());
+	}
```
tadgh, line 63: These tests really make clear the purpose of this feature. love it.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/storage/PreviousVersionReader.java
```diff
@@ -0,0 +1,36 @@
...
+
+	public Optional<T> readPreviousVersion(T theResource) {
+		return readPreviousVersion(theResource, false);
+	}
+
+	public Optional<T> readPreviousVersion(T theResource, boolean theDeletedOk) {
+		Long currentVersion = theResource.getIdElement().getVersionIdPartAsLong();
+		if (currentVersion == null || currentVersion == 1L) {
+			return Optional.empty();
+		}
```
tadgh, line 26: question: Should we look to support versionless resources? e.g. If i call this with `Patient/123`, would it be useful to get the current version, then go backwards from there? Not a blocker for this MR, just wondering about use cases.

tadgh (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/storage/PreviousVersionReader.java
fil512: Interesting.  I assumed that this would be called after a read, but you raise a good point.  This is good enough for this MR, but will keep that in mind for future.

PR #4907, opened 2023-05-17
CVE rel 6 6

longma1:


pom.xml
```diff
@@ -1698,8 +1698,8 @@
 			</dependency>
 			<dependency>
 				<groupId>org.glassfish</groupId>
-				<artifactId>javax.el</artifactId>
-				<version>3.0.0</version>
+				<artifactId>jakarta.el</artifactId>
+				<version>3.0.4</version>
```
jmarchionatto, line 1702: Any particular reason to use this version, which is two version behind current?

jmarchionatto (approved):
Approved with a question

pom.xml
longma1: This is the most current 3.x.x release. Bumping major versions is more likely to introduce breaking change

PR #4916, opened 2023-05-18
4910-dm-migration-error-for-oracle-19c

jmarchionatto:


michaelabuckley (approved):
minor: no changelog.

michaelabuckley:
Agree.  Anything outside `user_constraints` might be another HAPI install in a different schema.  We should focus on the application schema/user.

PR #4922, opened 2023-05-19
Fixed R5 Subscription deserialization

fil512:
R5 Subscription.filterBy.resourceType failed to deserialize because the deserializer skipped all elements named "resourceType".  This has been changed so that only toplevel resourceType elements are skipped in the deserialization process.


tadgh (approved):
Approved Pending Changelog

PR #4946, opened 2023-05-29
Delete dead code now that Hapi has merged.

michaelabuckley:


hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/TagDefinition.java
```diff
@@ -159,7 +158,6 @@ private void setVersionAfterTrim(String theVersion) {
 
 	/**
 	 * Warning - this is nullable, while IBaseCoding getUserSelected isn't.
-	 * wipmb maybe rename?
 	 */
 	public Boolean getUserSelected() {
```
TipzCM, line 162: and here i thought this would be named 'isUserSelected'

TipzCM (approved):

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/TagDefinition.java
michaelabuckley: Agree, but I'm reluctant to change until we are more stable.

Issue #4960, opened 2023-06-06
$delete-expunge on MSSQL or Oracle with over 10,000 resources results in error and stack trace


lukedegruchy:
**Describe the bug**
Running a $delete-expunge MSSQL or Oracle with over 10,000 resources results in a error and a stack trace, even though the job ends in status COMPLETE.

**To Reproduce**
1. Configure hapi-fhir with either MSSQL or Oracle
2. Either import or POST 10,000 resources
3. Configure hapi-fhir with all options needed to run $delete-expunge (delete expunge enabled, expunge operation enabled, and allow multiple delete enabled)
4. DELETE:  http://localhost:8000/Patient?_expunge=true
5. Wait for the job to execute the deletes and keep an eye on the logs:  See several errors similar to `Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: Transaction (Process ID 61) was deadlocked on lock resources with another process and has been chosen as the deadlock victim. Rerun the transaction.`

**Expected behavior**
Since these errors are transitional while the job retries they should not be logged as errors with stack traces

**Environment (please complete the following information):**
 - HAPI FHIR Version:  Master

**Additional context**
The delete expunge job eventually succeeds and the records are deleted.  The errors are just noise during which temporary errors occur

PR #4961, opened 2023-06-06
Ensure StepExecutor only logs a stack trace and error if the retries have been exhausted.  Otherwise, just log a debug.

lukedegruchy:
- Ensure that StepExecutor does not log errors and stack traces on retryable errors
- Only log an error and stack trace if the retries have expired

Closes https://github.com/hapifhir/hapi-fhir/issues/4960

### DIFF
--- a/hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/StepExecutor.java
+++ b/hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/StepExecutor.java
@@ -70,7 +70,7 @@ public class StepExecutor {
 			return false;
 		} catch (Exception e) {
 			if (theStepExecutionDetails.hasAssociatedWorkChunk()) {
-				ourLog.debug("Temporary failure executing job {} step {}, marking chunk {} as ERRORED", jobDefinitionId, targetStepId, chunkId);
+				ourLog.info("Temporary problem executing job {} step {}, marking chunk {} as retriable ERRORED", jobDefinitionId, targetStepId, chunkId);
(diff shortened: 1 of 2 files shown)


hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/StepExecutor.java
```diff
@@ -70,10 +70,11 @@ <PT extends IModelJson, IT extends IModelJson, OT extends IModelJson> boolean ex
 			return false;
 		} catch (Exception e) {
 			if (theStepExecutionDetails.hasAssociatedWorkChunk()) {
-				ourLog.error("Failure executing job {} step {}, marking chunk {} as ERRORED", jobDefinitionId, targetStepId, chunkId, e);
+				ourLog.debug("Temporary failure executing job {} step {}, marking chunk {} as ERRORED", jobDefinitionId, targetStepId, chunkId);
```
michaelabuckley, line 73:
```suggestion
				ourLog.info("Temporary problem executing job {} step {}, marking chunk {} as retriable ERRORED", jobDefinitionId, targetStepId, chunkId);
```

michaelabuckley (approved):
I think we want to see these if they happen all the time.  We might need a deeper fix.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/StepExecutor.java
michaelabuckley: Avoiding the F-word because people are easily scared.

PR #4986, opened 2023-06-14
First commit:  Patch bad implementation of RuleBulkExportImpl to fix …

lukedegruchy:
…this use case.

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImpl.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImpl.java
@@ -37,6 +37,7 @@ import static org.apache.commons.collections4.CollectionUtils.isNotEmpty;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class RuleBulkExportImpl extends BaseRule {
+	private static final org.slf4j.Logger ourLog = org.slf4j.LoggerFactory.getLogger(RuleBulkExportImpl.class);
 	private String myGroupId;
 	private String myPatientId;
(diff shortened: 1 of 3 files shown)


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImpl.java
```diff
@@ -89,14 +90,37 @@ public AuthorizationInterceptor.Verdict applyRule(RestOperationTypeEnum theOpera
...
+				if (actualPatientIds.contains(expectedPatientId)) {
+					return newVerdict(theOperation, theRequestDetails, theInputResource, theInputResourceId, theOutputResource, theRuleApplier);
+				}
+
+				return new AuthorizationInterceptor.Verdict(PolicyEnum.DENY,this);
+			}
+
+			final Set<String> filters = options.getFilters();
+
+			// TODO:  LD:  This admittedly adds more to the tech debt above, and should really be addressed by 4634
```
TipzCM, line 109: i like that you included the number; i really do.

But this is in hapi code base and that is a gitlab ticket number (something not all coders in this code base will even recognize or know).

Probably fine to keep if we plan on fixing soon anyways though. Just a comment.

TipzCM (approved):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImpl.java
lukedegruchy: Good catch.  I'll create the hapi-fhir equivalent for that issue and put it in the comment instead.

PR #4988, opened 2023-06-14
issue-4987-searchbundle-parts-are-labeled-by-default-as-include-instead-of-match

jmarchionatto:


hapi-fhir-base/src/main/java/ca/uhn/fhir/util/bundle/SearchBundleEntryParts.java
```diff
@@ -30,10 +30,10 @@ public class SearchBundleEntryParts {
 	public SearchBundleEntryParts(String theFullUrl, IBaseResource theResource, String theSearchMode) {
 		myFullUrl = theFullUrl;
 		myResource = theResource;
-		if ("match".equalsIgnoreCase(theSearchMode)) {
-			mySearchMode = BundleEntrySearchModeEnum.MATCH;
-		} else {
+		if ("include".equalsIgnoreCase(theSearchMode)) {
```
davidraeside, line 33: NITPICK: Can you use the INCLUDE value from the BundleEntrySearchModeEnum rather than the hardcoded include string? like BundleEntrySearchModeEnum.INCLUDE.getCode().equalsIgnoreCase(theSearchMode)

jmarchionatto: Sure. Good idea. Done

TipzCM (approved):

PR #4999, opened 2023-06-16
Add auto-generated headers to Java files.

lukedegruchy:
- This includes only new and update class file comment headers.
- This is the result of running a maven build on the latest master


michaelabuckley (approved):
Looks ok.

PR #5001, opened 2023-06-17
Raw sql revincludes

### LINKED ISSUE #5000: JPA_PERFTRACE_RAW_SQL pointcut not invoked for revincludes or includes
This pointcut should be invoked whenever raw SQL is executed, but it was not being executed during asynchronous paging. Essentially the bundle executes loadIncludes, but this is well after this pointcut has been invoked. 

Recommendation: 
- Have the pointcut also be invoked after all includes/revincludes have occurred.

tadgh:
Closes #5000 

- Have loadIncludes invoke the JPA_PERFTRACE_RAW_SQL pointcut if configured to do so. 
- Add changelog and test


fil512 (approved):
Nice clean MR.

PR #5012, opened 2023-06-20
Partitioning: creating two partitions with the same ID overwrites the first

LalithE:
Added unit test and bug fix to ensure no two partitions with the same ID cannot be created.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/5011-partitioning-can-create-more-than-one-partition-with-same-id.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/5011-partitioning-can-create-more-than-one-partition-with-same-id.yaml
@@ -1,4 +1,5 @@
 ---
 type: fix
 issue: 5011
-title: "Fixed bug that allowed creation of two partitions with the same id"
+title: "Previously, if a partition was created with an ID that was alrady in use, it would overwite the partition that 
+ was using that ID. Now attempting to overwrite will return a 400"
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/5011-partitioning-can-create-more-than-one-partition-with-same-id.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/5011-partitioning-can-create-more-than-one-partition-with-same-id.yaml
@@ -1,0 +1,4 @@
+---
+type: fix
+issue: 5011
+title: "Fixed bug that allowed creation of two partitions with the same id"
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/5011-partitioning-can-create-more-than-one-partition-with-same-id.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/5011-partitioning-can-create-more-than-one-partition-with-same-id.yaml
@@ -1,5 +1,5 @@
 ---
 type: fix
 issue: 5011
-title: "Previously, if a partition was created with an ID that was alrady in use, it would overwite the partition that 
+title: "Previously, if a partition was created with an ID that was already in use, it would overwrite the partition that 
  was using that ID. Now attempting to overwrite will return a 400"
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
@@ -199,14 +199,9 @@ public class PartitionLookupSvcImpl implements IPartitionLookupSvc {
 	}
 
 	private void validIdUponCreation(PartitionEntity thePartition){
-		List<PartitionEntity> allPartitions = myPartitionDao.findAll();
-		if(!allPartitions.isEmpty()) {
-			for (PartitionEntity i : allPartitions) {
-				if (i.getId().equals(thePartition.getId())) {
-					String msg = myFhirCtx.getLocalizer().getMessageSanitized(PartitionLookupSvcImpl.class, "duplicatePartitionId");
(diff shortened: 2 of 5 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
```diff
@@ -213,6 +214,13 @@ private void validateHaveValidPartitionIdAndName(PartitionEntity thePartition) {
 			throw new InvalidRequestException(Msg.code(1312) + msg);
 		}
 
+		List<PartitionEntity> allPartitions = myPartitionDao.findAll();
+		for(PartitionEntity i : allPartitions){
+			if(i.getId()==thePartition.getId()){
```
davidraeside, line 219: You should use i.getId().equals(thePartition.getId()) when not dealing with primitive data types.

StevenXLi:
You need to add a changelog

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
LalithE: will do thank you

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
```diff
@@ -213,6 +214,13 @@ private void validateHaveValidPartitionIdAndName(PartitionEntity thePartition) {
 			throw new InvalidRequestException(Msg.code(1312) + msg);
 		}
 
+		List<PartitionEntity> allPartitions = myPartitionDao.findAll();
+		for(PartitionEntity i : allPartitions){
+			if(i.getId().equals(thePartition.getId())){
```
TipzCM, line 219: Q:

Isn't PartitionEntity.Id a unique field? How are these duplicate partitions being created at all?

This validation is also used for updates to partition (which should expect a duplicate id, since it's updating an existing one).

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/partition/PartitionManagementProviderTest.java
```diff
@@ -103,6 +102,26 @@ private Parameters createInputPartition() {
...
+					.named(ProviderConstants.PARTITION_MANAGEMENT_CREATE_PARTITION)
+					.withParameters(inputPartition)
+					.encodedXml()
+					.execute();
+			}
+			fail();
+		} catch (InvalidRequestException e) {
+			assertEquals("HTTP 400 Bad Request: " + Msg.code(2366) + "Partition ID already exists", e.getMessage());
+		}
+		verify(myPartitionConfigSvc, times(0)).createPartition(any(), any());
```
TipzCM, line 122: can also use `never()` but that's a nitpick

TipzCM (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
tadgh: I agree with this, its an `@Id` column, which means its a primary key. Primary keys are unique. It should be impossible to make 2 requests, which results in 2 individual entries in the partition table with the same ID. 

What actually happens in the test before you made this change? I would like to see the actual database entries.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
```diff
@@ -197,6 +198,15 @@ private void validatePartitionNameDoesntAlreadyExist(String theName) {
 		}
 	}
 
+	private void validIdUponCreation(PartitionEntity thePartition){
+		List<PartitionEntity> allPartitions = myPartitionDao.findAll();
```
tadgh, line 202: issue: myPartitionDao.findById() does what you want to do, but without having to iterate over every partition in the system.

tadgh (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
LalithE: Yes, I misplaced the fix completely. I have now created a new method to deal with this. The issue was with two partitions being created not with the process of updating a partition. Sorry for the confusion.

tadgh (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/PartitionLookupSvcImpl.java
```diff
@@ -197,6 +198,14 @@ private void validatePartitionNameDoesntAlreadyExist(String theName) {
...
```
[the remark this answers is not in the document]
LalithE: Agreed, so I have removed that from the code.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/5011-partitioning-can-create-more-than-one-partition-with-same-id.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 5011
+title: "Fixed bug that allowed creation of two partitions with the same id"
```
tadgh, line 4: issue: This is not an accurate statement anymore given our understanding of this bug. this should probably read: 
"Previously, if a partition was created with an ID that was alrady in use, it would overwite blah blah blah. now attempting to overwrite returns 400"

tadgh (approved):

TipzCM (approved):

PR #5027, opened 2023-06-26
New formatting rules and pre-commit hooks

tadgh:
- Add plugin to pom.xml to integrate [spotless](https://github.com/diffplug/spotless). 
- Select [palantir java format](https://github.com/palantir/palantir-java-format) as our repository format, due to its line length of 120, along with its focus on formatting code specifically to be easier to code review.
- Only production code is auto-linted, test code is omitted currently, due to certain whitespace requirements for test asserts. Eventually this will be rectified
- Replace palantir's default indentation rules to match what our editorconfig uses.
- Make use of the [pre-commit project](https://pre-commit.com/) to add an optional pre-push hook. This can be setup by intalling pre-commit as indicated on their website, and then running `pre-commit install --hook-type pre-push`. After that, any time you attempt a push, spotless will apply all formating changes automatically on push. Note that while the installation of this pre-push hook is optional for developers, the pipeline will fail if formatting has not been done. 
- Add a github action which verifies that `mvn spotless:check` passes, indicating no formatting is needed. 

The only files you actually want to look at here are: 
1. `pom.xml`
2. `.github/workflows/spotless.yml`
3. `.pre-commit-config.yaml`

**What does this mean for developers:** 
1. Any time you `mvn clean install` it will autoformat all code. in practice this means that it will autoformat any code you just wrote, since the repo will already be in a formatted state, minus any changes you've made. 
2. If you use the pre-commit hook, Any time you git push , it will autoformat files you touched compared to your source branch. If files are _changed_ due to this process, the hook will fail the commit, and you will have to execute the commit command again, as we don't necessarily want to auto-commit bot-generated code. 
3. If your code is not formatted by the time it hits the pipeline, the new github action will fail. 

Code formatting can happen manually by use of the following maven command: 
```
mvn spotless:apply
```
or also 
```
mvn clean install -DskipTests
```


hapi-fhir-validation-resources-r4/pom.xml
```diff
@@ -36,6 +36,11 @@
 			</resource>
 		</resources>
 		<plugins>
+			<!-- Anything deployable inherits the parent spotless configuration -->
+			<plugin>
```
jamesagnew, line 40: Did you mean to put this in this specific pom? I would have expected it to be in the hapi-deployable-pom one?

pom.xml
```diff
@@ -2209,6 +2213,72 @@
 	<build>
 		<pluginManagement>
 			<plugins>
+				<plugin>
+					<groupId>com.rudikershaw.gitbuildhook</groupId>
+					<artifactId>git-build-hook-maven-plugin</artifactId>
+					<version>${gitbuildhook_version}</version>
+					<configuration>
+						<installHooks>
+							<pre-commit>hooks/pre-commit.sh</pre-commit>
```
jamesagnew, line 2222: Just to confirm, does this run anytime maven runs? If so, will this barf if you run it on windows?

tadgh: so this plugin installs the git pre-commit hooks automatically. I will test it on windows to validate. This only runs once, and installs the pre-commit hook, and then does nothing on subsequent runs (unless that pre-commit.sh file changes)

hapi-fhir-validation-resources-r4/pom.xml
tadgh: Nope I can't read. This should be in the hapi-deployable-pom

tadgh: Fixed

pom.xml
tadgh: the more I think about it I don't think its really fair to _force_ a pre-commit hook on people. I'll make this an optional portion via the pre-commit project. reworking. FWIW it _does_ work on windows

tadgh: Done. removed the forced hook, and added an optional hook via pre-commit project

markiantorno (approved):
This is going to be great.

PR #5038, opened 2023-06-27
5037 goldenresource remains when target resource deleted

### LINKED ISSUE #5037: GoldenResource remains when target resource is deleted
**Describe the bug**
When soft or hard deleting the source resource of a golden resource, the golden resource is left orphaned.

**To Reproduce**
1. Set up MDM
2. Post patient to create golden record
3. You will have say xxx1 as golden record and xxx2 as another record in under xxx1 golden record
4. Delete xxx2 record from golden record xxx1 using delete. Now you will have xxx1 golden record with no records under it
5. Post same patient again it will create another golden record say xxx3 and it wont go under previous empty golden record xxx1

**Expected behavior**
If there are no more MATCH links to a GR, the GR should be deleted to prevent orphaning.

jdar8:
- Added a check where if the last source resource with a `MATCH` link associated with a golden resource is deleted, the golden resource is also deleted
- Ensure any remaining `POSSIBLE_MATCHES` are re-matched 
- Added config to switch between hard/soft delete of golden resources (soft delete ensures GR is still there for history auditing)
- Add tests
- Changelog + added docs for behaviour 

Closes #5037


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/mdm/MdmLinkDaoJpaImpl.java
```diff
@@ -131,6 +131,11 @@ public List<MdmPidTuple<JpaPid>> expandPidsBySourcePidAndMatchResult(
 				.collect(Collectors.toList());
 	}
 
+	@Override
+	public List<MdmLink> findLinksAssociatedWithGoldenResourceOfSourceResourceExcludingNoMatch(JpaPid theSourcePid) {
+		return new ArrayList<>(myMdmLinkDao.findLinksAssociatedWithGoldenResourceOfSourceResourceExcludingMatchResult((theSourcePid).getId(), MdmMatchResultEnum.NO_MATCH));
```
jmarchionatto, line 136: You can just return the Dao result, without creating a new ArrayList

jmarchionatto (approved):
Implementation looks good. Approved with a comment.
Also, looks like the change could affect performance badly but I suppose avoiding that is beyond the scope of the fix.

PR #5049, opened 2023-07-01
fix surefire config

fil512:
Previously, stack traces in test failures were "trimmed" from the logs which made it difficult to fix test failures.  This MR fixes this.

Note that plugin `<configuration>` elements in sub-projects override `<configuration>` elements defined in pluginManagement.  (They are not merged as one might expect).  So this MR also removes any surefire configuration overrides in the child poms so that the `<configuration>` defined in the parent pom's pluginManagement can prevail.


hapi-fhir-jpaserver-elastic-test-utilities/pom.xml
```diff
@@ -244,35 +241,4 @@
 			</plugin>
 		</plugins>
 	</build>
-
```
michaelabuckley, line 247: Nice to see all this copy-pasta removed

pom.xml
```diff
@@ -2330,7 +2330,7 @@
 					<configuration>
 						<redirectTestOutputToFile>true</redirectTestOutputToFile>
 						<argLine>@{argLine} ${surefire_jvm_args}</argLine>
-						<trimStackTrace>true</trimStackTrace>
```
michaelabuckley, line 2333: Who would ever want this?

michaelabuckley (approved):

PR #5056, opened 2023-07-03
Added test to show a bug

jkiddo:
There seems to be a bug when serializing the output of the patch mechanism


robogary:
Formatting check succeeded!

PR #5082, opened 2023-07-12
5080 new similarity algorithm to handle numeric normalization

### LINKED ISSUE #5080: New Similarity Algorithm to handle SSN / SSN Normalization
**Describe the issue**
A new algorithm for detecting similarity between numeric values. Example: 
When a SSN is populated, it can be in one of two formats.  One with dashes and one without (eg. 123-45-6789 vs 123456789).
A combination of Numeric and Levenshtein algorithms can be used to remove dashes before measuring similarity.

jdar8:
- Added numeric versions for similarity algorithms, where numeric normalization is performed before similarity measurement
- Test
- Changelog + docs
Closes #5080


hapi-fhir-server-mdm/src/test/java/ca/uhn/fhir/mdm/rules/matcher/NumericSimilarityR4Test.java
```diff
@@ -0,0 +1,58 @@
...
+
+
+	@ParameterizedTest
+	@CsvSource({
+		"123-45-6789, 123456789",
+		"1234-5-6789, 123456789",
+		"abc123, 123",
+		"(416) 967-1111, 4169671111,"
+	})
+	public void testNumericSimilarity_withExactMatches(String theLeft, String theRight) {
```
TipzCM, line 32: Might want to test at least 1 non-similar thing for coverage, right?

TipzCM (approved):

hapi-fhir-server-mdm/src/test/java/ca/uhn/fhir/mdm/rules/matcher/NumericSimilarityR4Test.java
jdar8: I could, but at that point, I feel like I would be testing the similarity algorithm itself instead of the removal of non-numeric characters as I would end up having to manually calculate the score the algorithm would output. 
That's why I added the next test, so that I can include some non-similar things by comparing the code removing non-numeric characters to what the algorithm would have outputted if I manually remove non-numeric characters.

PR #5107, opened 2023-07-19
6 8 cves

tadgh:
* Resolve Major CVEs


pom.xml
```diff
@@ -1095,7 +1095,7 @@
 			<dependency>
 				<groupId>com.graphql-java</groupId>
 				<artifactId>graphql-java</artifactId>
-				<version>19.4</version>
+				<version>21.0</version>
```
jmarchionatto, line 1098: It would be neat to have all versions defined in properties and properties referenced only once in DependencyManagement

jmarchionatto (approved):

PR #5109, opened 2023-07-19
Fix split-package conflicts with org.hl7.fhir.*.model packages in FHIR Core projects

bdenton:
When the Fhir resource models were moved over to org.hl7.fhir.core, the `fhirversion.properties` files in hapi-fhir remained in the *.model Java packages (in src/main/resources directory). This created a "split package" situation that is a problem when loading these bundles in the OSGi framework., With OSGi, each Java package can only be resolved to a single bundle/jar and no OSGi tricks (like Fragment-Host) can resolve the issue.

This PR corrects this split-package situation by renaming the folder containing the `fhirversion.properties` file within the `src/main/resources` folder. This seems a safe thing to do since it looks like that is only referenced by the `IFhirVersion` implementation classes.


tadgh:
Heya @bdenton , I will review this in the morning! Had a quick gander and it looks good, just want to vet it against some downstream projects. 

Cheers

tadgh:
Once these tests pass, I'll go ahead and get it merged. Looks good to me

tadgh (approved):

PR #5113, opened 2023-07-20
Resolve 5112 unable to expunge deleted codesystem until several minutes later

TynerGjs:
What was done:
- Return a more informative message to the expunge request
- Added test for the above chage

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/JpaResourceExpungeService.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/JpaResourceExpungeService.java
@@ -61,6 +61,7 @@ import org.hl7.fhir.instance.model.api.IIdType;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.dao.DataIntegrityViolationException;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Slice;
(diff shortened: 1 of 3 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/JpaResourceExpungeService.java
```diff
@@ -315,16 +317,22 @@ private void expungeCurrentVersionOfResource(
...
-			myForcedIdDao.deleteByPid(forcedId.getId());
-		}
+			if (resource.getForcedId() != null) {
+				ForcedId forcedId = resource.getForcedId();
+				myForcedIdDao.deleteByPid(forcedId.getId());
+			}
 
-		myResourceTableDao.deleteByPid(resource.getId());
+			myResourceTableDao.deleteByPid(resource.getId());
+		} catch (Exception e) {
```
michaelabuckley, line 331: minor: we normally don't catch all Exceptions.  It is better to identify the actual exception you are trying to remap.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/JpaResourceExpungeService.java
```diff
@@ -315,16 +317,22 @@ private void expungeCurrentVersionOfResource(
...
+			throw new PreconditionFailedException(Msg.code(2415)
```
michaelabuckley, line 332: If you really want to catch all exceptions, then we should log the stack trace in case something weird is going on.  Just appending the exception (with `+ e`) will only log the error message, not the whole stack trace.  The stack trace is important debugging information.

michaelabuckley (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/JpaResourceExpungeService.java
TynerGjs: Changed to catch DataIntegrityViolationException only.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/JpaResourceExpungeService.java
TynerGjs: Got it. I'll keep that in mind

PR #5127, opened 2023-07-25
Resolve 5126 hfj res ver prov might cause migration error on db that automatically indexes the primary key

TynerGjs:


jmarchionatto (approved):
Looks good

PR #5144, opened 2023-07-28
$delete-expunge over 10k resources will now delete all resources

lukedegruchy:
- Replace the logic in Batch2DaoSvcImpl.java for URL based searches that uses last update time sorting with _id sorting instead
- Instead of a single call to the resource DAO to fetch all IDs, do multiple calls each with a SearchParameterMap loadSynchronouslyUpTo of the JpaStorageSettings.getInternalSynchronousSearchSize() + 1.  Meaning, if that configuration is 10,000, then pass 10,001 to the SearchParameterMap
- Ensure on the first call, the SearchParamterMap offset is set to 0
- Compare the configuration above with the actual results.  So if, for example, 10,001 results are returned for a configured 10,000 results, continue to search on the IDs, which are sorted by _id
- Pass the total number of accumulated results as the offset for subsequent searches.   This means if we get back well over 30,000 results, we'll loop more than twice but will still accumulate results properly
- Remove the logic in ResourceIdListStep that loops over different update date sorted pages and replace it with a single call to IIdChunkProducer
- Edit unit tests that enforced the old behaviour to make them pass under the new

Closes https://github.com/hapifhir/hapi-fhir/issues/5150

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/Batch2DaoSvcImplTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/Batch2DaoSvcImplTest.java
@@ -83,6 +83,17 @@ class Batch2DaoSvcImplTest extends BaseJpaR4Test {
 		}
 	}
 
+	@Test
+	void fetchResourcesByUrlNonsensicalResource() {
+		try {
+			mySubject.fetchResourceIdsPage(PREVIOUS_MILLENNIUM, TOMORROW, 800, RequestPartitionId.defaultPartition(), "Banana?_expunge=true");
+		} catch (InternalErrorException exception) {
+			assertEquals("HAPI-2223: HAPI-1684: Unknown resource name \"Banana\" (this name is not known in FHIR version \"R4\")", exception.getMessage());
+		} catch (Exception exception) {
+			fail("caught wrong Exception");
+		}
+	}
+
 	@ParameterizedTest
 	@ValueSource(ints = {0, 9, 10, 11, 21, 22, 23, 45})
 	void fetchResourcesByUrl(int expectedNumResults) {
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/Batch2DaoSvcImplTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/Batch2DaoSvcImplTest.java
@@ -2,7 +2,7 @@ package ca.uhn.fhir.jpa.reindex;
 
 import ca.uhn.fhir.interceptor.model.RequestPartitionId;
 import ca.uhn.fhir.jpa.api.config.JpaStorageSettings;
-import ca.uhn.fhir.jpa.api.dao.IFhirResourceDao;
+import ca.uhn.fhir.jpa.api.dao.DaoRegistry;
 import ca.uhn.fhir.jpa.api.pid.IResourcePidList;
 import ca.uhn.fhir.jpa.api.svc.IBatch2DaoSvc;
 import ca.uhn.fhir.jpa.dao.tx.IHapiTransactionService;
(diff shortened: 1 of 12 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/Batch2DaoSvcImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+		mySubject = new Batch2DaoSvcImpl(myResourceTableDao, myMatchUrlService, mySpiedDaoRegistry, myFhirContext, myTransactionService, myJpaStorageSettings);
+	}
+
+	// TODO: LD this test won't work with the nonUrl variant yet:  error:   No existing transaction found for transaction marked with propagation 'mandatory'
+
+	@Test
+	void fetchResourcesByUrlEmptyUrl() {
+		try {
+			mySubject.fetchResourceIdsPage(PREVIOUS_MILLENNIUM, TOMORROW, 800, RequestPartitionId.defaultPartition(), "");
+		} catch (InternalErrorException exception) {
```
steve-corbett-smilecdr, line 68: Nitpick: you can use assertThrows() here and in other tests in this class:

```
InternalErrorException exception = assertThrows(InternalErrorException.class, () -> {
  assertEquals("HAPI-2422...", exception.getMessage());
});
```

You can add `throws Exception` to the method declaration, and if some other Exception type is thrown, it will fail the test automatically with that Exception.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/ResourceIdListStep.java
```diff
@@ -88,50 +85,34 @@ public RunOutcome run(
...
-		while (true) {
-			IResourcePidList nextChunk = myIdChunkProducer.fetchResourceIdsPage(
-					nextStart, end, pageSize, requestPartitionId, theStepExecutionDetails.getData());
 
-			if (nextChunk.isEmpty()) {
-				ourLog.info("No data returned");
-				break;
-			}
+		IResourcePidList nextChunk = myIdChunkProducer.fetchResourceIdsPage(
+				start, end, pageSize, requestPartitionId, theStepExecutionDetails.getData());
```
michaelabuckley, line 90: problem: This pageSize param serves no purpose, and leaves this same bug here for the no-url case.  Either delete it, or add a pageNum param so the looping can be done here.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/ResourceIdListStep.java
```diff
@@ -88,50 +85,34 @@ public RunOutcome run(
...
-			}
+		for (TypedResourcePid typedResourcePid : nextChunk.getTypedResourcePids()) {
+			TypedPidJson nextId = new TypedPidJson(typedResourcePid);
+			idBuffer.add(nextId);
+		}
 
-			previousLastTime = nextChunk.getLastDate().getTime();
-			nextStart = nextChunk.getLastDate();
-
-			while (idBuffer.size() > maxBatchId) {
```
michaelabuckley, line 121: suggestion: This loop could be replaced by either com.google.common.collect.Iterators.partition(), or by ca.uhn.fhir.util.TaskChunker.chunk()

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/ResourceIdListStep.java
```diff
@@ -88,50 +85,34 @@ public RunOutcome run(
...
-				// which is failing intermittently. If that stops, makes sense to remove this
-				ourLog.info(" * PIDS: {}", nextChunk);
-			}
+		ourLog.debug("Found {} IDs from {} to {}", nextChunk.size(), start, nextChunk.getLastDate());
 
-			for (TypedResourcePid typedResourcePid : nextChunk.getTypedResourcePids()) {
-				TypedPidJson nextId = new TypedPidJson(typedResourcePid);
-				idBuffer.add(nextId);
-			}
+		for (TypedResourcePid typedResourcePid : nextChunk.getTypedResourcePids()) {
```
michaelabuckley, line 98: suggesion: there is no need to buffer the ids in the middle.  You could chunk as you convert with Iterators.transform() like Stream.map.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/ResourceIdListStep.java
lukedegruchy: @michaelabuckley This also gets passed to MdmIdChunkProducer and then to GoldenResourceSearchSvcImpl.

Are you suggesting I fix those as well?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/ResourceIdListStep.java
lukedegruchy: Iterators.partition() FTW

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/ResourceIdListStep.java
lukedegruchy: I fixed and tested the noUrl case while leaving the pageSize parameter intact with a TODO.

michaelabuckley (approved):

PR #5156, opened 2023-08-01
Expunge operation on codesystem may throw 500 internal error with precondition fail message.

epeartree:
**What was done:**
- Adding catch clause to the  PartitionRunner to avoid having the PreconditionFailedException being wrapped into an InternalError exception;
- Modifying test;
- Adding changelog.


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/6_8_0/5155-expunge-operation-on-codesystem-may-return-500.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 5155
+title: "Previously, requesting an $expunge operation on CodeSystem resource while CS batch deletion is underway would return HTTP 500.
```
TipzCM, line 4: CodeSystem resources

TipzCM (approved):

PR #5178, opened 2023-08-06
Resolve all unused and deprecated warnnings

sekaijin:
- cleanup imports
- Remove all unused varaiables (Tests)
- Replace all depreacated calls by fluent Api like client.read(Patient.class, "1") ; to client.read().resource(Patient.class).withId("1").execute();


sekaijin:
sorry I don't understand why it works on my repos and not here.
I cancel the pull request

PR #5190, opened 2023-08-09
Slf4j api change

tadgh:
- Currently JPA-Server-Starter cannot use the new SLF4J API, so we are just sticking to the old API for now.


dotasek:
Should this not be merged to rel_6_8, or is there a different strategy for this one?

tadgh:
> Should this not be merged to rel_6_8, or is there a different strategy for this one?

Yup, good catch

markiantorno (approved):

PR #5197, opened 2023-08-10
added validateExpungeEnabled() call to type-level expunge, moved two …

TynerGjs:
What was done:
- added call to `validateExpungeEnabled()` to type-level expunge operation
- added corresponding test

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ExpungeR4Test.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ExpungeR4Test.java
@@ -69,6 +69,7 @@ import java.util.List;
 import static ca.uhn.fhir.batch2.jobs.termcodesystem.TermCodeSystemJobConfig.TERM_CODE_SYSTEM_DELETE_JOB_NAME;
 import static org.awaitility.Awaitility.await;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ExpungeR4Test.java
```diff
@@ -921,6 +923,63 @@ public void testDeleteCodeSystemByUrlThenExpungeWithoutWaitingForBatch() {
...
+		runInTransaction(() -> assertThat(myResourceTableDao.findAll(), not(empty())));
+		runInTransaction(() -> assertThat(myResourceHistoryTableDao.findAll(), not(empty())));
+
+		// execute & verify
+		switch (level){
+			case "instance":
+				try {
+					myPatientDao.expunge(patientId, new ExpungeOptions()
+						.setExpungeDeletedResources(true)
+						.setExpungeOldVersions(true), null);
```
volodymyr-korzh, line 948: Nitpick: You could move switch-case statement inside try block, as try-catch code for all cases is the same.

TynerGjs: Nice catch!! I wasn't aware of that, that way there's no need to repeat the asserts 3 times

steve-corbett-smilecdr (approved):

PR #5214, opened 2023-08-17
fix the issue

### LINKED ISSUE #5212: CDS Hooks request throwing 500 internal server when context is missing
**Describe the bug**
- When a CDS hook is registered and called with a empty context. The server returns a 500.

**Expected behavior**
- When a CDS hook is registered and called with a empty context. The server should not return a 500.

VK-SMILECDR:
closes #5212


hapi-fhir-server-cds-hooks/src/main/java/ca/uhn/hapi/fhir/cdshooks/api/json/CdsServiceRequestContextJson.java
```diff
@@ -59,14 +60,23 @@ public void put(String theKey, Object theValue) {
 	}
 
 	public Set<String> getKeys() {
+		if (myMap == null) {
```
davidraeside, line 63: Could you just intialize myMap to be an empty collection where you declared it? Then you wouldn't need these repetitive null checks on every method.

VK-SMILECDR: Already try it, break CDS hooks JSON requirements

davidraeside (approved):

PR #5218, opened 2023-08-18
Automated migration test data

SouradeepSahaSmile:
Data Versions Added:

VersionEnum.V5_2_0
VersionEnum.V5_3_0
VersionEnum.V5_4_0
VersionEnum.V5_5_0
VersionEnum.V6_0_0
VersionEnum.V6_6_0


hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/embedded/H2EmbeddedDatabase.java
```diff
@@ -54,14 +54,29 @@ public void stop() {
...
 	public void disableConstraints() {
 		getJdbcTemplate().execute("SET REFERENTIAL_INTEGRITY = FALSE");
 	}
 
 	@Override
 	public void enableConstraints() {
-		getJdbcTemplate().execute("SET REFERENTIAL_INTEGRITY = TRUE");
+		List<String> sql = new ArrayList<>();
+		for (String tableName : getAllTableNames()) {
+			sql.add(String.format("ALTER TABLE \"%s\" SET REFERENTIAL_INTEGRITY TRUE CHECK", tableName));
```
steve-corbett-smilecdr, line 77: Why do this per-table instead of for the whole db?

nathandoef: The `SET REFERENTIAL_INTEGRITY` query used before was not working as expected. Inserting data with foreign key violations was still being allowed. This modification ensures that foreign violations result in a test failure.

hapi-fhir-jpaserver-test-utilities/src/test/java/ca/uhn/fhir/jpa/embedded/HapiSchemaMigrationTest.java
```diff
@@ -61,26 +62,40 @@ public void testMigration(DriverTypeEnum theDriverType) throws SQLException {
...
+		);
+
+
+		int fromVersion = 0;
 		VersionEnum from = allVersions[fromVersion];
+		VersionEnum toVersion;
+
+		for (int i = 1; i < allVersions.length; i++) {
+			toVersion = allVersions[i];
+			migrate(theDriverType, dataSource, hapiMigrationStorageSvc, from, toVersion);
```
nathandoef, line 90: Please change this so we are not executing duplicate migrations. 

Right now `from` is not being updated inside the loop so we are executing every migration from the `from` version to the `toVersion` when we only need to execute the `toVersion`.

You can use the `BaseMigrationTasks#getAllTasks()` to make sure each migration task is only being execute once

nathandoef (changes requested):

nathandoef (approved):

PR #5224, opened 2023-08-21
5222 storage cr hapifhirretrieveprovider terminology paging bug

Capt-Mac:
fix query iterable bug where valueset codes > 500 are not paging, leaving inconsistent results. This updates configuration of the data provider as well as an edit to the iterator logic to loop for results. 

An integration test has also been added to verify outcome


barhodes (approved):

JPercival (approved):

hapi-fhir-storage-cr/src/test/java/ca/uhn/fhir/cr/r4/R4MeasureOperationProviderIT.java
```diff
@@ -147,5 +147,23 @@ void testMeasureEvaluateMultiVersion() {
...
+	void testLargeValuesetMeasure() {
+		this.loadBundle("largeValueSetMeasureTest-Bundle.json");
+
+		var returnMeasureReport = runEvaluateMeasure("2023-01-01", "2024-01-01", null, "CMSTest", "population", null);
+
+		String populationName = "numerator";
+		int expectedCount = 1;
+
+		Optional<MeasureReport.MeasureReportGroupPopulationComponent> population = returnMeasureReport.getGroup().get(0)
+			.getPopulation().stream().filter(x -> x.hasCode() && x.getCode().hasCoding()
```
pepsiofficial, line 160: nitpick: would like to see this broken out into individual steps for readability

hapi-fhir-storage-cr/src/main/java/ca/uhn/fhir/cr/common/HapiFhirRetrieveProvider.java
```diff
@@ -101,11 +101,12 @@ public QueryIterator(
...
 			public boolean hasNext() {
 				// initial load of first query results
-				if (currentResult == null && index < queries.size()) {
+				if (currentResult == null) {
 					currentResult = loadNext();
 				}
+				// loop through all batches of results
 				// when query results exhaust load next query
-				else if (!currentResult.hasNext()) {
+				while (currentResult != null && !currentResult.hasNext()) {
```
pepsiofficial, line 109: question: why does this handful of changes do the trick?

pepsiofficial (approved):
looks good

hapi-fhir-storage-cr/src/main/java/ca/uhn/fhir/cr/common/HapiFhirRetrieveProvider.java
Capt-Mac: the terminology retrieve provided pages of 500 codes from a valueset at a time, for most value sets this would mean all results made it into the first batch of codes searched. On larger valuesets, the codes beyond the first 500 fetched are broken into batches of queries and would not be cycled because it only checked the first batch in the set when it actually had more queries to loop through. The "while" instead of the "else if" made sure it checked all queries before it said it exhausted all results.

hapi-fhir-storage-cr/src/test/java/ca/uhn/fhir/cr/r4/R4MeasureOperationProviderIT.java
Capt-Mac: core issue here is that the code (pregnancy condition) within the patient dataset was missed with previous code and would result in a numerator=0 instead of numerator=1 on measureReport. Since the paging of valuesets it now picks up the code and scores as expected.

PR #5244, opened 2023-08-24
modified MdmLinkDaoJpaImpl.convertToLongIds so that it returns all pi…

TynerGjs:
What was done:
* modified `MdmLinkDaoJpaImpl.convertToLongIds` so that it returns all pid that is related to a forced id.
* added tests for above changes

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/mdm/MdmLinkDaoJpaImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/mdm/MdmLinkDaoJpaImpl.java
@@ -411,8 +411,9 @@ public class MdmLinkDaoJpaImpl implements IMdmLinkDao<JpaPid, MdmLink> {
 
 	@Nonnull
 	private List<Long> convertToLongIds(List<IIdType> theMdmHistorySearchParameters) {
-		return theMdmHistorySearchParameters.stream()
-				.map(id -> myIdHelperService.getPidOrThrowException(RequestPartitionId.allPartitions(), id))
+		return myIdHelperService
+				.getPidsOrThrowException(RequestPartitionId.allPartitions(), theMdmHistorySearchParameters)
+				.stream()
 				.map(JpaPid::getId)
 				.collect(Collectors.toUnmodifiableList());
 	}
--- a/hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/BaseMdmR4Test.java
+++ b/hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/BaseMdmR4Test.java
@@ -236,17 +236,6 @@ abstract public class BaseMdmR4Test extends BaseJpaR4Test {
 	}
 
 	@Nonnull
-	protected Patient createPatientWithId(Patient thePatient, String theId) {
-		thePatient.setId(theId);
-		thePatient.setActive(true);
-
-		DaoMethodOutcome outcome = myPatientDao.update(thePatient, new SystemRequestDetails());
-		Patient patient = (Patient) outcome.getResource();
-		patient.setId(outcome.getId());
-		return patient;
-	}
-
-	@Nonnull
 	protected Medication createMedication(Medication theMedication) {
 		DaoMethodOutcome outcome = myMedicationDao.create(theMedication);
 		Medication medication = (Medication) outcome.getResource();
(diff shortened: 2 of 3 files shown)


hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/BaseMdmR4Test.java
```diff
@@ -242,6 +242,17 @@ protected Patient createPatientOnPartition(Patient thePatient, RequestPartitionI
 		return patient;
 	}
 
+	@Nonnull
+	protected Patient createPatientWithId(Patient thePatient, String theId) {
```
pepsiofficial, line 246: Question: could this method and the one on lines 274–280 be unified into a single type-parameterized method?

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/dao/MdmLinkDaoSvcTest.java
```diff
@@ -214,6 +216,31 @@ public void testHistoryForNoIdsOnly() {
 		assertThrows(IllegalArgumentException.class, () -> myMdmLinkDaoSvc.findMdmLinkHistory(new MdmHistorySearchParameters()));
 	}
 
+	@Test
+	public void testHistoryForIdWithMultipleMatches(){
```
pepsiofficial, line 220: Test looks good!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/mdm/MdmLinkDaoJpaImpl.java
```diff
@@ -411,8 +411,9 @@ public List<MdmLinkWithRevision<MdmLink>> getHistoryForIds(
 
 	@Nonnull
 	private List<Long> convertToLongIds(List<IIdType> theMdmHistorySearchParameters) {
-		return theMdmHistorySearchParameters.stream()
-				.map(id -> myIdHelperService.getPidOrThrowException(RequestPartitionId.allPartitions(), id))
+		return myIdHelperService
```
pepsiofficial, line 414: Question: what's the fix here?  What subtlety does the revised code cover that the initial code doesn't?  It looks like two different ways of spelling the same thing.  This is for my own learning here—I'm curious.

pepsiofficial (approved):
Approved with questions

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/BaseMdmR4Test.java
TynerGjs: Nice catch! Will do that.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/mdm/MdmLinkDaoJpaImpl.java
TynerGjs: Good question! It took me a while to figure it out when doing the modification!

While the original method `getPidOrThrowException` returns exactly 1 pid for each provided forced id, the revised version `getPidsOrThrowException` (with `pids`) fetches a list of pids that the provided forced matches to.

For example, if there exist two resources that have the same forced id:
`Patient/p111` and `Practitioner/p111`. Calling former method on id `"p111"` would return the pid of only one of them (whichever is created earlier). However, calling the later method will return the pid of both resources.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/BaseMdmR4Test.java
TynerGjs: I just noticed that the `BaseMdmR4Test` is full of such methods... I wonder if the whole class could be refactored to use only several type-parameterized methods, and it feels like job unfinished if I only refactor the method I added... Maybe it would be better to follow the current structure of the class for now? What should I do?

pepsiofficial: I vote to leave things better than you found them.  Unify the two but don't worry about unifying all of them in this PR.  Maybe if you have the time, do the refactor in a separate PR—it shouldn't take too long to do, I imagine

PR #5255, opened 2023-08-28
Fix to Bulk Export Enabled Disrupts Custom Resource Addition

LalithE:


steve-corbett-smilecdr:
Needs a changelog entry.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirContext.java
```diff
@@ -1001,9 +1001,9 @@ public IParser newXmlParser() {
 	 */
 	public void registerCustomType(final Class<? extends IBase> theType) {
 		Validate.notNull(theType, "theType must not be null");
-
 		ensureCustomTypeList();
 		myCustomTypes.add(theType);
+		myResourceNames = null;
```
steve-corbett-smilecdr, line 1006: It seems like we might have a similar problem in L1027, where we're adding to `myCustomTypes` but not nulling out `myResourceNames`. I suggest making sure your tests also cover the `registerCustomTypes` method.

PR #5264, opened 2023-08-30
Resolve 5262 adding new parameter strictmatch to the $mdm history link operation

TynerGjs:
What was done:
- added new parameter to `$mdm-link-history` on Pointcut `ProviderConstants.MDM_LINK_HISTORY`.
- modified `MdmLinkDaoJpaImpl.getHistoryForIds` so that is adds the clause as AND when strictMatch is set to true.
- added corresponding test
- modified mdm link history documentation

PS: If you have a better name for the parameter than `strictMatch`, please let me know and forgive my poor naming creativity! 👍

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5262-adding-new-parameter-strictmatch-to-the-mdm-history-link-operation.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5262-adding-new-parameter-strictmatch-to-the-mdm-history-link-operation.yaml
@@ -1,0 +1,5 @@
+---
+type: add
+issue: 5262
+title: "Added a new parameter `strictMatch` to $mdm-link-history operation to allow filtering the history links so that 
+only those that match **both** the `goldenResourceId` and the `sourceResourceId` sets are returned."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5262-adding-new-parameter-strictmatch-to-the-mdm-history-link-operation.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5262-adding-new-parameter-strictmatch-to-the-mdm-history-link-operation.yaml
@@ -1,5 +1,5 @@
 ---
-type: add
+type: change
 issue: 5262
-title: "Added a new parameter `strictMatch` to $mdm-link-history operation to allow filtering the history links so that 
-only those that match **both** the `goldenResourceId` and the `sourceResourceId` sets are returned."
+title: "Previously, when both resourceId and goldenResourceId are provided to the mdm link history operation, they will
+be treated as an OR. This is now changed to AND in order to comply with REST conventions."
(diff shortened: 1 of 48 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_mdm/mdm_operations.md
```diff
@@ -212,7 +212,7 @@ This operation returns a `Parameters` resource that looks like the following:
 
 ## Link History
 
-Use the `$mdm-link-history` operation to request a list of historical entries for a given set of `goldenResourceId`s or `sourceResourceId`s.  Either parameter is optional but **at least one** must be provided.
+Use the `$mdm-link-history` operation to request a list of historical entries for a given set of `goldenResourceId`s or `sourceResourceId`s.  Either parameter is optional but **at least one** must be provided. The default behaviour is to return history links that matches at least one id from the provided `goldenResourceId` or `sourceResourceId` sets. The parameter `strictMatch` can be set to true to filter the history links so that only those that match **both** the `goldenResourceId` and the `sourceResourceId` sets are returned.
```
fil512, line 215: This doesn't seem like the correct implementation.  The API should follow REST conventions.  If both criteria are provided, then we should assume AND between them.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5262-adding-new-parameter-strictmatch-to-the-mdm-history-link-operation.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: add
+issue: 5262
+title: "Added a new parameter `strictMatch` to $mdm-link-history operation to allow filtering the history links so that
```
fil512, line 4: I recommend removing the strictMatch parameter and always applying AND to all provided criteria per the REST standard.

fil512 (changes requested):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_mdm/mdm_operations.md
TynerGjs: Yes, I too felt it's a bit weird to have "&" linking the two parameters but in fact it acts as an "or"...

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5262-adding-new-parameter-strictmatch-to-the-mdm-history-link-operation.yaml
TynerGjs: Would it be too large a breaking change? The same request would return different results after making this change.

TynerGjs: reverted changes and modified the default behaviour instead, please take a look when you have time!

michaelabuckley: I didn't like this when we put it in in March, but we already shipped it as OR, and customers are using this.  Changing the default from OR to AND would be a breaking change.  I think we have to live with this being backwards unless you feel strongly we should break current clients.  @fil512

fil512: I seriously doubt anyone is using the broken behaviour.  We have like 1 mdm customer, and their usage of it is super limited.

PR #5267, opened 2023-08-31
Respect response status code set in MethodOutcome of a Resource provider

nathandoef:
Previously, the response status code set in a `MethodOutcome` of a Resource provider was not respected. This has been fixed.


hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/provider/TestResponseCodeModifyingPatientProvider.java
```diff
@@ -0,0 +1,39 @@
...
+import ca.uhn.fhir.rest.server.IResourceProvider;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.r4.model.IdType;
+import org.hl7.fhir.r4.model.Patient;
+
+import javax.servlet.ServletRequest;
+
+import static ca.uhn.fhir.rest.api.Constants.STATUS_HTTP_202_ACCEPTED;
+
+public class TestResponseCodeModifyingPatientProvider implements IResourceProvider {
```
fil512, line 17: I think a test class like this should be an inner static class.  In prod code, I strongly prefer not using inner classes, but for test code like this, I think inner static classes make more sense.  (otherwise you pollute your classpath)

fil512 (approved):

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/provider/TestResponseCodeModifyingPatientProvider.java
nathandoef: Good call, fixed

restevez-chs:
hi @nathandoef -- would it be possible to have this fix released as a patch (6.8.x) or minor release (6.9)? my team is using HAPI-FHIR and unable to return a 202 Accept status code for a bulk data request operation. thanks.

PR #5310, opened 2023-09-13
Fix for `userSelected` and `version`.

michaelabuckley:
Add support for `userSelected` and `version` to DSTU2.  Also fix them on security labels in JPA.


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5310-broken-tag-parse.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 5310
+title: "Add support for userSelected and version to DSTU2. Also fix them on security labels in JPA."
```
tadgh, line 4: recommendation: backtick `userSelected` and `version` . Also, indicate that this is related to Tags, as that is not mentioned here.

hapi-fhir-structures-dstu2/src/main/java/ca/uhn/fhir/model/dstu2/resource/BaseResource.java
```diff
@@ -198,16 +198,22 @@ public List<? extends IPrimitiveType<String>> getProfile() {
...
+					CodingDt c = new CodingDt(
+						next.getSystemElement().getValue(),
+						next.getCodeElement().getValue())
+						.setDisplay(next.getDisplayElement().getValue());
+					if (next instanceof CodingDt) {
+						CodingDt nextCodingDt = (CodingDt) next;
+						c.setUserSelected(nextCodingDt.getUserSelectedElement());
+						c.setVersion(nextCodingDt.getVersion());
+					}
+					retVal.add(c);
```
tadgh, line 216: question: what happens if it isnt? what would it be if not a codingDt?

tadgh (approved):

hapi-fhir-structures-dstu2/src/main/java/ca/uhn/fhir/model/dstu2/resource/BaseResource.java
michaelabuckley: I wrote that before I added version and userSelected to BaseCodingDt.  I think these are all CodingDt instances, but we seem to return copies instead of aliasing.  Dunno why.  Keeping the existing pattern.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5310-broken-tag-parse.yaml
michaelabuckley: Done.

PR #5317, opened 2023-09-14
Bump core version to 6.1.2

dotasek:
## Update to xpp3

xpp3, included in the org.hl7.fhir.core library, is now updated there, removing the need for exclusion. This was updated in a number of pom.xml locations to match xpp3 1.1.6.

## Validation Issues Involving FHIRPath

Due to updates in FHIRPath strictness, we now ignore two validation issues in DSTU3 profiles:

Two 'side is inherently a collection' warnings are now thrown for the following expressions:

NamingSystem nsd-1 uniqueId.type = 'uuid'
Observation obs-7 code != component.code

## Validation Issues Involving out of date StructureDefinitions

There is also a general error ignored because the StructureDefinition for StructureDefinition is actually for FHIR 3.0.1, and contains an invalid expression in constraint sdf-9. This is remedied in the expression used in 3.0.2, but a full update of all profile structure definitions is out of scope for this PR.


hapi-fhir-storage-cr/pom.xml
```diff
@@ -169,9 +159,9 @@
 		</dependency>
 
 		<dependency>
-			<groupId>org.codelibs</groupId>
+			<groupId>org.ogce</groupId>
 			<artifactId>xpp3</artifactId>
-			<version>${org.codelibs.xpp3-version}</version>
+			<version>${org.ogce.xpp3-version}</version>
```
tadgh, line 164: recommend changing this var to just be `xpp3-version` to match other variables

tadgh (approved):
Approved pending a changelog and a minor variable name change

hapi-fhir-storage-cr/pom.xml
dotasek: Changed.

dotasek:
Added changelog.

PR #5389, opened 2023-10-21
Transaction with conditional update fails if SearchNarrowingInterceptor is registered and Enabled Partitioning

volodymyr-korzh:
Previously, with partitioning enabled and `UrlBaseTenantIdentificationStrategy` used, registering 
`SearchNarrowingInterceptor` would cause to incorrect resolution of `entry.request.url` parameter during 
transaction bundle processing. This has been fixed.


michaelabuckley (approved):
Nice.  This is smaller than I feared.

PR #5396, opened 2023-10-23
Clean stale search results more aggressively.

### LINKED ISSUE #5395: Search Cache cleaner can't keep up with high-volume servers.
**Describe the bug**
DatabaseSearchCacheSvcImpl can only clear 2000 searches per minute.  A server serving more searches than this will accumulate an ever-growing collection of stale searches.

**To Reproduce**
- run `curl BASE_URL/Patient?identifier=abc` with `Cache: no-cache` in a busy loop.
- Wait 10 minutes for the searches to expire
- Count stale searches in the database.

**Expected behavior**
The cleaner should keep up.  Deleting is less work that querying.

 - Hapi 6.8

michaelabuckley:
Closes #5395

Use bulk DMA statements to delete more data with less server IO.  Also run longer, and use an explicit deadline instead of a hard-coded batch size.


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/cache/DatabaseSearchCacheSvcImpl.java
```diff
@@ -169,128 +177,271 @@ public Optional<Search> findCandidatesForReuse(
...
-			return;
+		/**
+		 * Mark all ids in the mySearchesToMarkForDeletion buffer as deleted, and clear the buffer.
+		 */
+		public void flushDeleteMarks() {
+			if (myUpdateDeletedFlagBatch.isEmpty()) {
+				return;
+			}
+			ourLog.debug("Marking {} searches as deleted", myUpdateDeletedFlagBatch.size());
+			mySearchDao.updateDeleted(myUpdateDeletedFlagBatch, true);
```
jamesagnew, line 211: It might make sense to wrap this in QueryChunker... I see that by default we're deleting 10,000 per commit, which seems like a very reasonable amount, but avoiding having an in clause with 10k parameters is probably prudent (especially on oracle, where 10k would just plain blow up)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/cache/DatabaseSearchCacheSvcImpl.java
```diff
@@ -169,128 +177,271 @@ public Optional<Search> findCandidatesForReuse(
...
-					mySearchResultDao.deleteByIds(nextPartition);
-				}
+		private void flushSearchAndIncludeDeletes() {
+			if (myDeleteSearchBatch.isEmpty()) {
+				return;
 			}
+			ourLog.debug("Deleting {} Search records", myDeleteSearchBatch.size());
+			// referential integrity requires we delete includes before the search
+			mySearchIncludeDao.deleteForSearch(myDeleteSearchBatch);
+			mySearchDao.deleteByPids(myDeleteSearchBatch);
```
jamesagnew, line 321: Presumably we should also use QueryChunker here?

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/cache/DatabaseSearchCacheSvcImpl.java
michaelabuckley: QueryChunker operates on Collection.  That requires materializing the entire set of searches.  This code is streaming, so we never have more that 500 ids in ram.

michaelabuckley: We might be deleting 50000 SearchResults, but only by Search PK, and only 500 of those at a time.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/cache/DatabaseSearchCacheSvcImpl.java
michaelabuckley: Same.  This is streaming.  QuadraMed has 40 million stale searches.  I don't wan't that in ram every minute while we nibble away.  What I really need to write is the equivalent of the chunker for Streams.

PR #5402, opened 2023-10-25
5401 add terserutil clear by fhirpath

fil512:
Previously, it was only possible to clear a top-level field on a resource using TerserUtil.  A new method has been added
to TerserUtil to support clearing a value by FhirPath.

### DIFF
--- a/hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/util/TerserUtilTest.java
+++ b/hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/util/TerserUtilTest.java
@@ -31,88 +31,88 @@ class TerserUtilTest {
 
 	private FhirContext ourFhirContext = FhirContext.forR4();
 	private static final String SAMPLE_PERSON =
-		"""
-			{
-			      "resourceType": "Patient",
-			      "extension": [
-			        {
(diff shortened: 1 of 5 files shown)


hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/util/TerserUtilTest.java
```diff
@@ -525,6 +529,38 @@ public void testClearFields() {
...
+		a1.setCity("Test");
+		p1.addAddress(a1);
+
+		Address a2 = new Address();
+		a2.addLine("Line 1");
+		a2.addLine("Line 2");
+		a2.setCity("Test");
+		p1.addAddress(a2);
+
+		TerserUtil.clearFieldByFhirPath(ourFhirContext, p1, "address.line");
```
volodymyr-korzh, line 555: Suggestion: Adding one more test case with deeper FHIR path. Example: `contact.address.line`

volodymyr-korzh (approved):

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/util/TerserUtilTest.java
fil512: Good call.  You helped me catch an edge case I had not properly tested.

PR #4803, opened 2023-05-02
Move forced-id to HFJ_RESOURCE - step 2

michaelabuckley:
Start using new fhir_id column in hfj_resource instead for queries.  Still writes forced_id to support zero-downtime.

Reviewer notes:
The existing test coverage seems pretty good, so I'm not too worried about the core behaviour.  Two big worries are:
- the custom migration - it is tested on Sql Server, H2, PG, Oracle, and MySql.  Haven't tested Derby.
- concurrent access by old clients.

We still write to hfj_forced_id, but never query it now.  Older servers should have been populating the new FHIR_ID column, and the migration fills in the rest and makes it not-null and indexed.

PR #5405, opened 2023-10-26
Use new FHIR_ID column for sorting

michaelabuckley:
Use the new FHIR_ID column on HFJ_RESOURCE activated in #4803 for sorting.  This avoid a join when using `_id` to sort.

This eliminates our last read of the HFJ_FORCED_ID table, and is required before turn off writes to it.

Note:  this changes the existing behaviour of `_sort=_id`.  The old implementation did a funky super-fast thing sorting client-assigned ids before server-assigned ids, and sorting server-assigned ids numerically. The new implementation just sorts as a string.  This matches the spec, since `_id` is defined as token.  But it _is_ slower in some cases since it requires mapping from res_id->fhir_id (via index) instead of sorting on pid for resources with server-assigned ids (the common case).

Indexing is there, and this is actually faster for a case we hit in several batch job clunkers where when reindexing all of a type.  `Patient?_sort=_id` is now a straight index walk.

Need to figure out the indexing, and decide if this is for 6.10.

Also: need to check the constraint.  What do people do if they want fhir_id to be unique, but only per-partition?


jamesagnew (approved):
I love this

ohetrifork:
We were a bit surprised to see the semantic change that came with this PR. Previously, sorting on _id (descending) would give us the most recently created resources. But the sorting performed now on fhir_id is lexicographic instead of numeric, so we see resources with id 99.999 before resources with id 150.000! Our fix was to sort on _lastUpdated instead.

PR #5410, opened 2023-10-27
Add ITermConceptClientMappingSvc interface to represent terminology t…

jmarchionatto:
…ranslate functions

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptMappingSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptMappingSvcImpl.java
@@ -19,15 +19,10 @@
  */
 package ca.uhn.fhir.jpa.term;
 
-import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.context.support.TranslateConceptResult;
 import ca.uhn.fhir.context.support.TranslateConceptResults;
 import ca.uhn.fhir.i18n.Msg;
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptMappingSvcImpl.java
```diff
@@ -61,39 +48,17 @@
...
-import javax.persistence.criteria.CriteriaQuery;
-import javax.persistence.criteria.Join;
-import javax.persistence.criteria.Predicate;
-import javax.persistence.criteria.Root;
 
 import static ca.uhn.fhir.jpa.term.TermReadSvcImpl.isPlaceholder;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
-public class TermConceptMappingSvcImpl implements ITermConceptMappingSvc {
```
michaelabuckley, line 88: question: why split out the base class.  I don't see it used any where except as a base class.

If it is to shrink the class, consider composition over inheritance. 

Can we have some javadoc?

michaelabuckley (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptMappingSvcImpl.java
jmarchionatto: I though it was neat for each class to implement just the implementing interface methods.
Do you see any problem with that?

Javadoc added.

PR #5420, opened 2023-11-01
If AutoInflateBinaries is enabled, binaries are created on the disk only for the first resource entry of the bundle

volodymyr-korzh:
Previously, when `AllowAutoInflateBinaries` was enabled in `JpaStorageSettings` and bundles with multiple 
resources were submitted, binaries were created on the disk only for the first resource entry of the bundle. 
This has been fixed.


jmarchionatto (approved):
Looks good! Approved.

PR #5428, opened 2023-11-03
Add non-standard __pid SP for breaking ties cheaply during sorts.

michaelabuckley:
When doing offset paging, we need an inexpensive sort to impose a reliable iteration order.  This change implements sorting by the raw private id (i.e. the resource PID) which allows a minimal I/O sort.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/Constants.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/Constants.java
@@ -199,6 +199,8 @@ public class Constants {
 	public static final String PARAM_PRETTY_VALUE_FALSE = "false";
 	public static final String PARAM_PRETTY_VALUE_TRUE = "true";
 	public static final String PARAM_PROFILE = "_profile";
+	public static final String PARAM__PID = "__pid";
+
 	public static final String PARAM_QUERY = "_query";
 	public static final String PARAM_RESPONSE_URL = "response-url"; // Used in messaging
 	public static final String PARAM_REVINCLUDE = "_revinclude";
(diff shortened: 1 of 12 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/Constants.java
```diff
@@ -199,6 +199,8 @@ public class Constants {
 	public static final String PARAM_PRETTY_VALUE_FALSE = "false";
 	public static final String PARAM_PRETTY_VALUE_TRUE = "true";
 	public static final String PARAM_PROFILE = "_profile";
+	public static final String PARAM__PID = "__pid";
```
jamesagnew, line 202: As discussed, I recommend we rename this to `_pid` to be consistent with other meta-params. We have a b bunch of prior history doing this kind of thing.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa/search.md
```diff
@@ -22,6 +22,10 @@ Searching on Location.Position using `near` currently uses a box search, not a r
 
 The special `_filter` is only partially implemented.
 
+### __pid
+
+The JPA server implements a non-standard special `__pid` which matches/sorts on the raw internal database id.
+This sort is useful for imposing tie-breaking sort order in an efficient way.
```
jamesagnew, line 28:
```suggestion
This sort is useful for imposing tie-breaking sort order in an efficient way.

Note that this is an internal feature that may change or be removed in the future. Use with caution.
```

jamesagnew (approved):

PR #3982, opened 2022-09-01
change search method for missing parameter names

TipzCM:
* utilized different strategy for searching for missing fields (but maintained the old way as this is 'step one' in a migration path); changing between methods is a matter of setting the IndexMissingFields parameter
* included a change log

PR #5447, opened 2023-11-10
remove requirement to have 'missing' idxs enabled for patient bulk export

XcrigX:
TBH, I'm not clear why missing indexes were required for patient bulk export in the first place. So I'm not certain if this PR is "safe". But submitting for review in the hopes it was some old holdover that is no longer necessary.


stmsat:
From what I've gathered in a very short and not accurate search, the class JpaBulkExportProcessor was introduced in #3566, and the requirement for IndexMissingFields=ENABLED _could possibly_ be dropped after this PR was merged: #3982

Before that change, _maybe_ not all resources linked to the patient could be exported correctly, or errors were thrown anyway (just guessing).

I obviously don't have the full picture yet. Just stumbled in this issue and wanted to know more about it.

PR #5465, opened 2023-11-19
Fix issue that OpenAPI specification had errors when validating it against Swagger/OpenAPI schema and Swagger/OpenAPI spec

delopst:
HAPI Fhir OpenAPI specification has errors when validating it against Swagger/OpenAPI schema and Swagger/OpenAPI spec (https://apitools.dev/swagger-parser/online/).

With this commit, validation is passing.


jamesagnew (approved):
Thanks for the contribution!

jamesagnew:
I've marked for approval, but I do note that several OpenAPI unit tests are failing so this probably needs a bit of cleanup before it can be merged. Thanks!

delopst:
I see. The problem was that there was already a different swagger version in the root pom.
I have consolidated them. At least locally the ServerDstu3Test that was failing before is now passing.

delopst:
It seems like the tests are passing now.

Issue #5476, opened 2023-11-22
Pass properties through to Remote Terminology Service on CodeSystem $lookup


codeforgreen:
**Describe the issue**
If remote terminology is enabled and a user calls CodeSystem/$lookup with property parameters, e.g.
CodeSystem/$lookup?code=02536455&system=http://terminology.hl7.org/CodeSystem/hc-CCDD&property=healthCanadaIdentifier&property=healthCanadaProductName&property=ntpFormalName&property=tmFormalName&property=scheduleName then those properties should get passed on to the Remote Terminology Service $lookup call and the properties should be returned back.

**References**
R4 CodeSystem lookup: http://hl7.org/fhir/R4/codesystem-operation-lookup.html
DSTU2 CodeSystem lookup: http://hl7.org/fhir/DSTU2/terminology-service.html#2.5.1.4

PR #5477, opened 2023-11-22
Pass properties through to Remote Terminology Service on CodeSystem lookup

codeforgreen:
#5476


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoCodeSystem.java
```diff
@@ -125,6 +126,18 @@ public IValidationSupport.LookupCodeResult lookupCode(
...
+		return lookupCode(theCode, theSystem, theCoding, theDisplayLanguage, null, theRequestDetails);
+	}
+
+	@Nonnull
+	@Override
+	public IValidationSupport.LookupCodeResult lookupCode(
+			IPrimitiveType<String> theCode,
+			IPrimitiveType<String> theSystem,
+			IBaseCoding theCoding,
+			IPrimitiveType<String> theDisplayLanguage,
```
epeartree, line 138: new parameter theProperties is not passed down to doLookupCode?  
i might be mistaken but if we want properties to make it to remove look up, we need to pass them down.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoCodeSystem.java
```diff
@@ -125,6 +126,18 @@ public IValidationSupport.LookupCodeResult lookupCode(
 			IBaseCoding theCoding,
 			IPrimitiveType<String> theDisplayLanguage,
 			RequestDetails theRequestDetails) {
```
epeartree, line 128: passing null always make me fringe.  please use emptyList()

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -186,7 +187,8 @@ public LookupCodeResult lookupCode(
 			ValidationSupportContext theValidationSupportContext,
 			String theSystem,
 			String theCode,
-			String theDisplayLanguage) {
+			String theDisplayLanguage,
```
epeartree, line 190: thePropertyValues?

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -204,6 +206,9 @@ public LookupCodeResult lookupCode(
 				if (!StringUtils.isEmpty(theDisplayLanguage)) {
 					ParametersUtil.addParameterToParametersString(fhirContext, params, "language", theDisplayLanguage);
 				}
```
epeartree, line 208: using thePropertyNames right away in a foreach is dangerous.  this is a public interface, ie, anyone can provide  a value of null which will lead to NPE when thePropertyNames.iterator() will be invoked.

please validate the collection at the door with 
ObjectUtil.defaultIfNull(thePropertyNames, emptyList());

epeartree (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoCodeSystem.java
codeforgreen: Good catch. This is what I am talking about. Such bugs are not caught with the current test coverage because this is one of the classes which is not tested at all.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoCodeSystem.java
codeforgreen: Thanks, I fully agree. I realized that at some point and did that in one place but I missed the other.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -328,6 +329,16 @@ default CodeValidationResult validateCodeInValueSet(
 		return null;
 	}
 
+	@Nullable
+	default LookupCodeResult lookupCode(
+			ValidationSupportContext theValidationSupportContext,
+			String theSystem,
+			String theCode,
+			String theDisplayLanguage,
```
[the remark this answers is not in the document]
codeforgreen: Thanks. I think it's best to use thePropertyNames. Please see the [documentation](http://hl7.org/fhir/R4/codesystem-operation-lookup.html), under field property it says the server will select certain additional properties for CodeSystem to return in the response and it mentions how they are turned. It's a bit weird how it's returned because they're returned as one parameter with sub-parameter one for each property and the value is also there. Although confusing, it is what it is. 
I'll look and see if I should add a comment in the code, it may be confusing for others as well how one Parameter in input translates into multiple into the output.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
codeforgreen: Yes, will use thePropertyNames, see above comment.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -354,12 +357,28 @@ default LookupCodeResult lookupCode(
...
+	 * @see ValidationSupportParameterObject
+	 *
+	 * @param theValidationSupportContext      The validation support module will be passed in to this method. This is convenient in cases where the operation needs to make calls to
+	 *                                         other method in the support chain, so that they can be passed through the entire chain. Implementations of this interface may always safely ignore this parameter.
+	 * @param validationSupportParameterObject The parameters used to perform the lookup, including system and code.
+	 */
+	@Nullable
+	default LookupCodeResult lookupCode(
+			ValidationSupportContext theValidationSupportContext,
+			ValidationSupportParameterObject validationSupportParameterObject) {
```
jamesagnew, line 378: Suggestion: I find the name `ValidationSupportParameterObject` kinda clunky, especially since it's only for this method and not others in this interface. Would `LookupCodeRequest` make more sense? That way it pairs nicely with the name of the return type too.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -417,19 +418,19 @@ public CodeValidationResult validateCodeInValueSet(
 	@Override
 	public LookupCodeResult lookupCode(
```
jamesagnew, line 419: I would think for this class, you wouldn't want to replace the existing method, you'd want to add a new one for your new implementation. The IValidationSupportChain should have chained methods for every method on `IValidationSupport` that can be called.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/CachingValidationSupport.java
```diff
@@ -196,14 +197,15 @@ public CodeValidationResult validateCode(
 	@Override
 	public LookupCodeResult lookupCode(
```
jamesagnew, line 198: Similar to my comment below (or above?) on ValidationSupportChain, I would think you'd want to have a new method here, not to replace the existing ones which aren't going away.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -335,15 +335,18 @@ default CodeValidationResult validateCodeInValueSet(
 	 *                                    other method in the support chain, so that they can be passed through the entire chain. Implementations of this interface may always safely ignore this parameter.
```
jamesagnew, line 335: Meta-comment on this review: I don't think this is ready for merge yet - Keep in mind that:

* There are HAPI FHIR users out there who have implementations of this interface where they have coded their own implementations of the 'legacy' validateCode methods and obviously haven't implemented the new one you're adding. We need for any changes to this interface to not be a breaking change, which means that everything needs to continue to work even if people have implemented the old methods only.
* Most importantly, you need to look at any places where the legacy methods are called, and have that code try the new method and then try the old method if the new method returns null, so that you can be guaranteed that this isn't a breaking change. I note that a bunch of unit tests are failing, and that's why: currently most of the callers to this interface are still calling the old methods, but you've moved all the implementations to the new methods.

jamesagnew (changes requested):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5476-pass-properties-through-remote-terminology-service-for-code-system.yaml
```diff
@@ -0,0 +1,4 @@
+---
```
jamesagnew, line 1: Recommend also adding a second changelog along the lines of:

> A new method on the `IValidationSupport` interface called `lookupCode(LookupCodeRequest)` has been added. This method replaced the existing `lookupCode` methods, which are now deprecated.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -354,12 +357,28 @@ default LookupCodeResult lookupCode(
...
 		return lookupCode(theValidationSupportContext, theSystem, theCode, null);
 	}
 
+	/**
+	 * Look up a code using the system, code and other parameters.
+	 * @see ValidationSupportParameterObject
+	 *
+	 * @param theValidationSupportContext      The validation support module will be passed in to this method. This is convenient in cases where the operation needs to make calls to
+	 *                                         other method in the support chain, so that they can be passed through the entire chain. Implementations of this interface may always safely ignore this parameter.
+	 * @param validationSupportParameterObject The parameters used to perform the lookup, including system and code.
```
jamesagnew, line 373: Recommend adding:

```java
 * @since 7.0.0
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
codeforgreen: Thank you for the review, those are good comments. I'll update it accordingly.

codeforgreen: For second point, I'd actually updated all internal implementations to the new method, but one was missed due to some very interesting circumstances. Its `TermReadSvcImpl` lookupCode method, was missing the first parameter ValidationSupportContext which was there before, and the method was not marked as an Override. So I couldn't find it when searching for implementors. I am not sure how it worked before but once I replaced the parameters, it stopped working.
I wonder if implementing the fallback for `ValidationSupportChain` would be enough. In my opinion we would not need fallbacks for the rest of the internal implementations. Let me know if you consider otherwise. The other ones are either pass through or seem safe.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
codeforgreen: I added a fallback in the new method as per the overall comment. Let me know if that's enough or you still feel old methods still need to live here.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/CachingValidationSupport.java
codeforgreen: I wonder if having the fallback in `ValidationSupportChain` is enough to cover the use-cases where (terminology) validation would be customized. Wondering if it would it be common for consumers to extend even `CachingValidationSupport`.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -329,39 +329,59 @@ default CodeValidationResult validateCodeInValueSet(
...
+	 * @param theDisplayLanguage          Used to filter out the designation by the display language. To return all designation, set this value to <code>null</code>.
 	 */
+	@Deprecated
 	@Nullable
 	default LookupCodeResult lookupCode(
 			ValidationSupportContext theValidationSupportContext,
 			String theSystem,
 			String theCode,
 			String theDisplayLanguage) {
-		return null;
```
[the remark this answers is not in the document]
codeforgreen: Updated in lastest version. Resolving this.

epeartree (approved):

PR #5488, opened 2023-11-23
Add a new field to the CLIENT_RESPONSE pointcut in order to allow clients to mutate an HTTP response from the BaseClient.

lukedegruchy:
- Add a new ClientResponseContext wrapper to capture all params for the CLIENT_RESPONSE Pointcut, allowing callers to mutate the HTTP response
- Introduce a new ModifiedStringApacheHttpResponse similar to ApacheHttpResponse but one that takes a String with new contents to serve to callers of IHttpResponse
- Ensure BaseClient overrides the previous HTTP response with the one mutated by the interceptor, or keeps the original if the pointcut doesn't modify it.

Closes https://github.com/hapifhir/hapi-fhir/issues/5502

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/client/api/ClientResponseContext.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/client/api/ClientResponseContext.java
@@ -21,6 +21,7 @@ package ca.uhn.fhir.rest.client.api;
 
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.interceptor.api.Pointcut;
+import org.hl7.fhir.instance.model.api.IBaseResource;
 
 import java.util.Objects;
 import java.util.StringJoiner;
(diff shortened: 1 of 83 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/client/api/ClientResponseContext.java
```diff
@@ -0,0 +1,101 @@
...
+			IHttpRequest myHttpRequest,
+			IHttpResponse theHttpResponse,
+			IRestfulClient myRestfulClient,
+			FhirContext theFhirContext,
+			Class<? extends IBaseResource> theReturnType) {
+		this.myHttpRequest = myHttpRequest;
+		this.myHttpResponse = theHttpResponse;
+		this.myRestfulClient = myRestfulClient;
+		this.myFhirContext = theFhirContext;
+		this.myReturnType = theReturnType;
```
jmarchionatto, line 49: Shouldn't be added to equals, hashCode and toString?

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/BaseClient.java
```diff
@@ -352,12 +353,25 @@ protected <T> T invokeClient(
 
 			response = httpRequest.execute();
 
+            final Class<? extends IBaseResource> returnType =
```
jmarchionatto: Cant't see where is the condition which could avoid triggering the interceptor

jmarchionatto (changes requested):
A few questions added

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/BaseClient.java
jmarchionatto: OK. The commit comment was: "Only trigger the interceptor if the expected return type is a Bundle."

jmarchionatto (approved):
Approved

PR #5501, opened 2023-11-28
Process the _language search parameter in RawParamsParameter

adamzkover:
Version 6.10.0 should support search with the _language parameter, when `StorageSettings.setLanguageSearchParameterEnabled` is set to true.

While this seems to work on the storage side (values from Resource.language are added to the database if this setting is applied), the search appears to still ignore the _language parameter.

Steps to reproduce:
- enable the feature by setting `jpaStorageSettings.setLanguageSearchParameterEnabled(true);` in `FhirServerConfigCommon.jpaStorageSettings` of the hapi-fhir-jpaserver-starter.
- start the server
- upload a Composition that declares `<language value="no"/>`
- call /Composition?_language=dk - this will return the composition that has language 'no'

With the change in this PR, the search seems to take _language into account and the Composition is only returned when the _language query parameter is 'no' (or not included at all).

I was also looking into `SPECIAL_SEARCH_PARAMS` in `SearchMethodBinding` and `SearchPreferHandlingInterceptor` because they also have code that skips parameters that start with "_", but they did not have to be changed for my use case to work.

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/RawParamsParameter.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/RawParamsParameter.java
@@ -20,6 +20,7 @@
 package ca.uhn.fhir.rest.server.method;
 
 import ca.uhn.fhir.rest.annotation.RawParam;
+import ca.uhn.fhir.rest.api.Constants;
 import ca.uhn.fhir.rest.api.server.RequestDetails;
 import ca.uhn.fhir.rest.param.QualifierDetails;
 import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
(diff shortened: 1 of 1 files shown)


jamesagnew:
I'm not clear on what problem this is solving. Could you please update the description to include a complete description of why this change is needed, and add appropriate tests demonstrating how this solves the issue?

adamzkover:
The original comment is updated to describe the issue. The codebase is still new to me so I'm not sure how to demonstrate the issue through a test that would be meaningful.

jkiddo:
AFAIK this fixes it: https://github.com/hapifhir/hapi-fhir/pull/5740

adamzkover:
> AFAIK this fixes it: #5740

@jkiddo Tested with 7.0.1, the search with _language is still broken. This fix can still be applied and still fixes the search.

adamzkover:
@jkiddo this is fixed in 7.0.2

PR #5516, opened 2023-11-30
Expose the desired Propagation in a tx template

michaelabuckley:
HapiTransactionService is designed for sub-classing, but has no getters for important properties.  This exposes them.


nathandoef (approved):
Looks good, nice test!

PR #5519, opened 2023-11-30
CodeSystem $lookup with Remote Terminology Service enabled throws NullPointerException when designation language has no value

### LINKED ISSUE #5512: CodeSystem $lookup with Remote Terminology Service enabled throws NullPointerException when designation language has no value
**Describe the bug**
CodeSystem $lookup with Remote Terminology Service enabled throws NullPointerException when designation language has no value.
Example: https://fhir.loinc.org/CodeSystem/$lookup?system=http://loinc.org&code=LP14542-2&property=child#L522

**To Reproduce**
2. GET http://localhost:8000/CodeSystem/$lookup?system=http://loinc.org&code=LP14542-2&&property=parent
2. Request returns response 500 with body:
```
{
  "issue": [ {
    "severity": "error",
    "code": "processing",
    "diagnostics": "HAPI-0389: Failed to call access method: java.lang.NullPointerException: Cannot invoke \"Object.toString()\" because the return value of \"org.hl7.fhir.r4.model.Parameters$ParametersParameterComponent.getValue()\" is null"
  } ]
}
```

**Expected behavior**
The request should return success 200.

**Additional context**
```
Caused by: java.lang.NullPointerException: Cannot invoke "Object.toString()" because the return value of "org.hl7.fhir.r4.model.Parameters$ParametersParameterComponent.getValue()" is null
	at org.hl7.fhir.common.hapi.validation.support.RemoteTerminologyServiceValidationSupport.generateLookupCodeResultR4(RemoteTerminologyServiceValidationSupport.java:363)
	at org.hl7.fhir.common.hapi.validation.support.RemoteTerminologyServiceValidationSupport.lookupCode(RemoteTerminologyServiceValidationSupport.java:225)
	at org.hl7.fhir.common.hapi.validation.support.ValidationSupportChain.lookupCode(ValidationSupportChain.java:426)
	at org.hl7.fhir.common.hapi.validation.support.BaseValidationSupportWrapper.lookupCode(BaseValidationSupportWrapper.java:103)
	at org.hl7.fhir.common.hapi.validation.support.CachingValidationSupport.lambda$lookupCode$12(CachingValidationSupport.java:205)
	at org.hl7.fhir.common.hapi.validation.support.CachingValidationSupport.lambda$loadFromCache$17(CachingValidationSupport.java:253)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2031)
```

codeforgreen:
Fixes #5512 

Two issues resolved:
(1) `CodeSystem` $lookup with Remote Terminology Service would throw `NullPointerException` when the `CodeSystem` 
included designations with no language value. 
(2) There was an inconsistency between input and output type ` string` vs. `code` for property parameters for all implementations.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,6 @@
 			xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
 
 
-<!-- cosmetic change -->
 
 	<!-- the version info -->
 	<modelVersion>4.0.0</modelVersion>


codeforgreen:
Two issues resolved:
(1) CodeSystem $lookup with Remote Terminology Service would throw NullPointerException when the CodeSystem 
included designations with no language value. 
(2) There was an inconsistency between input and output type `string` vs. `code` for property parameters for all implementations.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5512-codesystem-lookup-with-designation-no-language-nullpointerexception.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 5511
+title: 'Previously, CodeSystem $lookup with Remote Terminology Service would throw NullPointerException when the CodeSystem
```
tadgh, line 4: $lookup -> `$lookup`

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -275,144 +280,135 @@ private LookupCodeResult generateLookupCodeResultDSTU3(
 							case "use":
 								org.hl7.fhir.dstu3.model.Coding coding =
 										(org.hl7.fhir.dstu3.model.Coding) designationComponent.getValue();
-								if (coding != null) {
-									conceptDesignation.setUseSystem(coding.getSystem());
-									conceptDesignation.setUseCode(coding.getCode());
-									conceptDesignation.setUseDisplay(coding.getDisplay());
-								}
+								conceptDesignation.setUseSystem(coding.getSystem());
```
tadgh, line 283: Can the coding no longer be null here?

tadgh (approved):
Looks good, although I'm worried type changes will break existing implementations. Are these type changes internal, or will this leak to clients?

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
codeforgreen: No. It cannot, I tested for null above at line 271. But I missed to reuse the same variable. I had to apply the same fix for two versions in two different methods and here I made it slightly different. I'll update the code to reuse the variable so it's obvious.

codeforgreen:
> Looks good, although I'm worried type changes will break existing implementations. Are these type changes internal, or will this leak to clients?

This is a valid concern. 
The `RemoteTerminologyServiceValidationSupport` changes do not impact, they are just a bug fix. I am suprised it did not come up before. 
The `IValidationSupport` change which addresses the type inconsistency might impact if clients had workarounds in place for it. Basically `lookupCode` would always return properties as code although some of them were originally string in the `CodeSystem` or `ValueSet`. There are actually even more types that we should support but we now support `string` and `Coding` as per what is defined in `TermConceptPropertyTypeEnum`.
See:
http://hl7.org/fhir/codesystem-definitions.html#CodeSystem.concept.property.value_x_
http://hl7.org/fhir/R4/codesystem-operation-lookup.html

PR #5534, opened 2023-12-04
SQL Migration:  Support lowercase for SQL columns and overridden column type/driver type SQL type string rules

lukedegruchy:
- Allow callers to choose uppercase or lowercase column names.  Default uppercase.
- Allow callers to pass custom sorting rules:  Default no special sorting.
- In both cases, set the defaults to pre-existing behaviour, namely, uppercase and no overrides
- Fix unrelated unit test that's broken in master
- Add a new test

Closes https://github.com/hapifhir/hapi-fhir/issues/5536

### DIFF
--- a/hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/ColumnDriverMappingOverride.java
+++ b/hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/ColumnDriverMappingOverride.java
@@ -1,0 +1,85 @@
+/*-
+ * #%L
+ * HAPI FHIR Server - SQL Migration
+ * %%
+ * Copyright (C) 2014 - 2023 Smile CDR, Inc.
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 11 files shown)


hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/ColumnDriverMappingOverride.java
```diff
@@ -0,0 +1,88 @@
...
+
+	ColumnTypeEnum getColumnType() {
+		return myColumnType;
+	}
+
+	DriverTypeEnum getDriverType() {
+		return myDriverType;
+	}
+
+	String getColumnTypeSql() {
```
TipzCM, line 57: Why not just make all accessors public?

Even if it's only used in this package (for now), it makes future refactoring less painful.

TipzCM (approved):
Minor stuff

Q: Was no additional tests needed?

lukedegruchy:
> Minor stuff
> 
> Q: Was no additional tests needed?

@TipzCM I'm open to suggestions on what additional tests I should add

PR #5541, opened 2023-12-07
adding additional data to (bulk data export) parameters

### LINKED ISSUE #5527: Update BulkExport job to support additional consent service methods (shouldIncludeResrouceInExport)
AC
* Extra claims (if any) from token / thread local
* add them to the bulk export job
* update consent service to have new shouldIncludeResourceInExport method

Testing
* unit testing only (none)

TipzCM:
closes #5527


hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/BaseBatchJobParameters.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+public abstract class BaseBatchJobParameters implements IModelJson {
+	/**
+	 * A serializable map of key-value pairs that can be
+	 * added to any extending job.
+	 */
+	@JsonProperty("additionalData")
+	private Map<String, Object> myData;
```
jamesagnew, line 14: We call this concept "userData" everywhere else, as opposed to "additionalData". It'd probably be more consistent to use that name.

hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/BaseBatchJobParameters.java
```diff
@@ -0,0 +1,26 @@
...
+
+	public Map<String, Object> getAdditionalData() {
+		if (myData == null) {
+			myData = new HashMap<>();
+		}
+		return myData;
+	}
+
+	public void setAdditionalData(String theKey, Object theValue) {
```
jamesagnew, line 23: Values in this class need to be able to serialize to JSON. I'd recommend documenting that only String, Integer, and Boolean are valid value types, and adding a runtime check to make sure that only these types actually get used. This will help to catch issues early on, instead of them causing crashes later when it's hard to understand what has happened.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/bulk/BulkExportJobParameters.java
```diff
@@ -32,7 +32,7 @@
 import java.util.Date;
 import java.util.List;
 
-public class BulkExportJobParameters implements IModelJson {
+public class BulkExportJobParameters extends BaseBatchJobParameters {
```
jamesagnew, line 35: Should all of the other job parameter types be modified to also extend this new base class?

If so, we should also modify JobDefinition#setParametersType so that it uses the new mode specific base instead of just `IModelJson`.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/consent/IConsentService.java
```diff
@@ -176,6 +177,38 @@ default ConsentOutcome willSeeResource(
...
+	 * @param theBulkExportJobParameters
+	 * 					The Bulk Export parameters that were used to invoke the job.
+	 * @param theResource
+	 * 					The resource being considered for inclusion in the export results.
+	 * @param theContextServices An object passed in by the consent framework that
+	 *                           provides utility functions relevant to acting on
+	 *                           consent directives.
+	 * @return An outcome object. See method documentation for a description.
+	 */
+	default ConsentOutcome shouldIncludeResourceInExport(
```
jamesagnew, line 205: I'm not following why we need a new method for this. I wouldn't expect that there would be cases where we'd want to have different consent rules between the REST API and Bulk Export. I'd recommend just using the `willSeeResource` method.

jamesagnew (changes requested):
I don't love adding a new consent service method just for bulk exports. Recommend collapsing that.

hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/BaseBatchJobParameters.java
TipzCM: updated

hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/BaseBatchJobParameters.java
TipzCM: updated

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/bulk/BulkExportJobParameters.java
TipzCM: We could.

But considering all the other jobs are across a multitude of different modules, and not all modules even have interceptors provided in them, it's probably better to do this on a case by case basis as needed (imo).

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/consent/IConsentService.java
TipzCM: Unfortunately i don't know if that can be done.

The real purpose of this work is to support the $export invoked via $member-match, and allow users to control that based on the $member-match specifically.

But 2 issues:
* during bulk exports, users are not logged in (it's a batch job; no user session is available at all) - so there's not really a way for interceptor writers to even know the context of the "willSeeResource" (we don't have a RequestDetails or even a user session for that matter)
* willSeeResource has no way of having context about the job itself - like the fact that it *is* a bulk export nor the special member-match parameters provided to it when the job is first invoked.

TipzCM: As discussed, this method will be removed and we'll use the willSeeResource instead

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/bulk/BulkExportJobParameters.java
jamesagnew: Makes sense - Might be a good idea to note that in the `BaseBatchJobParameters` javadoc (ie that not all parameters extend it and that it can be added as a base class on a case by case basis)

jamesagnew (approved):

PR #5545, opened 2023-12-13
HtmlUnit excludes for xerxes/xml-api no longer required

rbri:
Since version 2.68 HtmlUnit does no longer uses xerces. You can cleanup the pom a bit here.


jamesagnew:
Thanks for this, great point!

rbri:
Great to see HtmlUnit is of some use for you!

PR #5555, opened 2023-12-14
Fix postgres / Remove use of LOB content column

jamesagnew:
This PR does 2 things:

* Previously, resource body content went into one of 2 columns on the `HFJ_RES_VER` table: RES_TEXT if the size was above a configurable threshold, or RES_TEXT_VC if it was below that threshold. Performance testing has shown that the latter is always faster, and that on Postgres the use of the latter is particularly problematic since it maps to the `largeobject` table which isn't the recommended way of storing high frequency objects. As of this PR, the configurable threshold is ignored, and the latter column is always used. Any legacy data in the former column will still be read however.
* This PR also addresses an issue introduced in the Jakarta API changes that caused failures on Postgres DB when writing large resources.


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/entity/GeneratedSchemaTest.java
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	private static void validateLongVarcharDatatype(String schemaName, String expectedDatatype) {
+		String schema = ClasspathUtil.loadResource("ca/uhn/hapi/fhir/jpa/docs/database/" + schemaName);
+		String[] lines = StringUtils.split(schema, '\n');
+		String resTextVc = Arrays.stream(lines).filter(t -> t.contains("RES_TEXT_VC ")).findFirst().orElseThrow();
+		assertThat("Wrong type in " + schemaName, resTextVc, containsString("RES_TEXT_VC " + expectedDatatype));
+	}
+
+}
```
tadgh, line 40: clever test

tadgh (approved):
couple minor things, but you have committed some write.lock files and segment_1 files which are blockers

ttntrifork:
Sorry to bother, but I am a bit confused by the release notes:

> Previously, resource body content went into one of 2 columns on the HFJ_RES_VER table: RES_TEXT if the size was above a configurable threshold, or RES_TEXT_VC if it was below that threshold. Performance testing has shown that the latter is always faster, and that on Postgres the use of the latter is particularly problematic since it maps to the largeobject table which isn't the recommended way of storing high frequency objects. As of this PR, the configurable threshold is ignored, and the latter column is always used. Any legacy data in the former column will still be read however.

So.. you say 
> on Postgres the use of the latter is particularly problematic

but then say 
> the latter column is always used


Did you by any chance mean that "on Postgres the use of the _former_ is particularly problematic"? :)

jamesagnew:
@ttntrifork oops - yes you are correct, that should say the former. It's the oid / RES_TEXT column that is a problem and is no longer used for new data.

PR #5561, opened 2023-12-19
Fix incorrect javadoc for missing modifier

Thopap:
Fixing the explanation for missing = true / false, which was previously twisted


jamesagnew (approved):
Thanks for this!

PR #5563, opened 2023-12-19
mdm duplicate identifier

fil512:
Survivorship replicated identifiers combined with mdm eid rules can lead to duplicate identifier entries in golden resources


KevinDougan-SmileCDR (approved):
I verified this fix by manually running through the exact same scenario locally which revealed the problem in the first place.

PR #5587, opened 2024-01-05
Update CdsHooksContextBooter.java

rgeimer:
Changed private variables and methods to protected, so that implementers can easily sub-class CdsHooksContextBooter to add their own CDS services.


barhodes:
Change was submitted in another PR.

PR #5589, opened 2024-01-09
Don't contain resources in Bundle with only fullUrl

jamesagnew:


hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
```diff
@@ -1010,6 +1004,32 @@ protected boolean isFhirVersionLessThanOrEqualTo(FhirVersionEnum theFhirVersionE
...
+		 */
+		if (theResource instanceof IBaseBundle) {
+			List<Pair<String, IBaseResource>> entries =
+					BundleUtil.getBundleEntryFullUrlsAndResources(getContext(), (IBaseBundle) theResource);
+			for (Pair<String, IBaseResource> nextEntry : entries) {
+				String fullUrl = nextEntry.getKey();
+				IBaseResource resource = nextEntry.getValue();
+				if (startsWith(fullUrl, "urn:")) {
+					if (resource != null && resource.getIdElement().getValue() == null) {
+						resource.getIdElement().setValue(fullUrl);
```
dmuylwyk, line 1024: **question (non-blocking):** Does it make a difference whether you populate `Resource.id` with the entire UUID as URN or just the UUID portion?

i.e. `urn:uuid:b87f654e-9608-4a9b-91df-6be6b92e4d6d` vs. `b87f654e-9608-4a9b-91df-6be6b92e4d6d`

dmuylwyk (approved):
Great work. Thanks for the quick turnaround!

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
jamesagnew: It does, yes. If you put in a placeholder (ie with a prefix) than the parser won't include it in the serialization, but without the prefix it does.

jamesagnew: And oddly enough, a unit test failed on the CI build which revealed that the RDF parser has a bug where it doesn't actually respect that like the other 2 forms. Will fix that before merging.

dmuylwyk: Understood. Thanks for clarifying, and for catching the second bug for RDF!

PR #5600, opened 2024-01-17
Add support for mimetype codes from system urn:ietf:bcp:13

codeforgreen:
#5601


hapi-fhir-base/src/main/java/ca/uhn/fhir/util/ClasspathUtil.java
```diff
@@ -163,4 +166,29 @@ public static byte[] loadResourceAsByteArray(String theClasspath) {
 			close(stream);
 		}
 	}
+
+	/**
+	 * Returns the list of files in a directory which resides in the classpath.
```
michaelabuckley, line 171: question: are these files in a directory, or resources on a classpath?  Those are different things.  Most classpath resources are just entries inside a zip file and not actually separate files.  If this actually gets a list of resource names, change the javadoc to match.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/CommonCodeSystemsTerminologyServiceTest.java
```diff
@@ -238,6 +246,98 @@ public void testFetchCodeSystemUrlDstu3() {
...
+		// verify
+		assertNotNull(result);
+		assertEquals(system, result.getSearchedForSystem());
+		assertEquals(code, result.getSearchedForCode());
+		assertTrue(result.isFound());
+		assertNotNull(result.getCodeDisplay());
+	}
+
+	@Test
+	public void testLookupCode_withMimetype_returnNotFoundResult() {
```
michaelabuckley, line 324: comment: nice tests.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/CommonCodeSystemsTerminologyService.java
```diff
@@ -253,36 +246,33 @@ public LookupCodeResult lookupCode(
...
-				return lookupMimetypeCode(code);
+				map = MIMETYPE_CODES;
+				break;
 			case COUNTRIES_CODESYSTEM_URL:
 				map = ISO_3166_CODES;
 				break;
 			case CURRENCIES_CODESYSTEM_URL:
 				map = ISO_4217_CODES;
 				break;
-			case USPS_CODESYSTEM_URL:
```
michaelabuckley, line 263: minor question: I see these moved.  Are they tested?

michaelabuckley (approved):
Clean and thorough.  Some minor questions/suggestions.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/CommonCodeSystemsTerminologyService.java
codeforgreen: I think I deleted this branch by accident. I'll restore it. I initially wanted to make changes for other code systems so this service is consistent across all methods but I noticed there are some differences of behaviour between code system and value set in terms of validateCode vs. lookupCode vs. fetchCodeSystem methods. So I just stuck to updating for mimetypes for now.

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/ClasspathUtil.java
codeforgreen: Resources in the classpath. I can update the method to just return file names instead.

PR #5614, opened 2024-01-19
Searching with more than one chained Bundle SearchParameter returns incorrect results

nathandoef:
Previously, searches that used more than one chained `Bundle` `SearchParameter` (i.e. `Composition`) were only 
adding one condition to the underlying SQL query which resulted in incorrect search results. This has been fixed.


epeartree (approved):
Fantastic test!

PR #5622, opened 2024-01-23
5621 deadlock on caffeine cache when creating a resource with conditional create

epeartree:
**What was done:**
- replaced the atomic operation of cache access and cache miss value generation to a two stage process where we check if a value is present in the cache and if it is not, we invoke the supplier and add the result to the cache;
- Setting the initial capacity of instantiated caches to the cache max size to avoid a resizing during operations;
- added changelog.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5621-fix-potential-deadlock-on-caffeine-cash-on-conditional-create.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5621-fix-potential-deadlock-on-caffeine-cash-on-conditional-create.yaml
@@ -1,5 +1,4 @@
 ---
 type: fix
 issue: 5621
-title: "Previously, a fhir endpoint could end up in a deadlock when creating a resource with conditional create. 
-The issue has been fixed."
+title: "Fixed a deadlock in resource conditional create."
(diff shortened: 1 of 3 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5621-fix-potential-deadlock-on-caffeine-cash-on-conditional-create.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 5621
+title: "Previously, a fhir endpoint could end up in a deadlock when creating a resource with conditional create. 
+The issue has been fixed."
```
michaelabuckley, line 5: How about a simple 
```suggestion
title: "Fixed a deadlock in resource conditional create."
```

hapi-fhir-serviceloaders/hapi-fhir-caching-caffeine/src/main/java/ca/uhn/fhir/sl/cache/caffeine/CacheProvider.java
```diff
@@ -44,13 +44,15 @@ public LoadingCache<K, V> create(long timeoutMillis, CacheLoader<K, V> loading)
 	public Cache<K, V> create(long timeoutMillis, long maximumSize) {
 		return new CacheDelegator<K, V>(Caffeine.newBuilder()
 				.expireAfterWrite(timeoutMillis, TimeUnit.MILLISECONDS)
+				.initialCapacity((int) maximumSize)
```
michaelabuckley, line 47:
```suggestion
                // Caffeine locks the whole array when growing the hash table.
                // Set initial capacity to max to avoid this.  All our caches are <1M entries.
				.initialCapacity((int) maximumSize)
```

michaelabuckley (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_0_0/5621-fix-potential-deadlock-on-caffeine-cash-on-conditional-create.yaml
epeartree: sure.

PR #5625, opened 2024-01-24
Incorrect version of auto versioned reference for conditional update with urn id placeholder

volodymyr-korzh:
Previously, when a transaction was posted with a resource that had placeholder references and auto versioning 
references enabled for that path, if the target resource was included in the Bundle but not modified, the reference was 
saved with a version number that didn't exist. This has been fixed.


jmarchionatto (approved):
Looks good. Approved

PR #5630, opened 2024-01-24
simple fix

longma1:
Change content type of getOrCreateContentType in attachmentUtil from string to code

attachment.contentType are code according to https://www.hl7.org/fhir/r4/datatypes.html#Attachment


jmarchionatto (approved):
Looks good. Approved

PR #5637, opened 2024-01-26
$expunge operation ignoring ExpungeThreadCount setting in certain cases

volodymyr-korzh:
Previously, the number of threads allocated to the $expunge operation in certain cases could be more
than configured, this would cause hundreds of threads to be created and all available database connections
to be consumed. This has been fixed.

### DIFF
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/expunge/PartitionRunner.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/expunge/PartitionRunner.java
@@ -186,7 +186,7 @@ public class PartitionRunner {
 		};
 		return new ThreadPoolExecutor(
 				threadCount,
-				MAX_POOL_SIZE,
+				threadCount,
 				0L,
 				TimeUnit.MILLISECONDS,
 				executorQueue,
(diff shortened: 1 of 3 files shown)


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/expunge/PartitionRunner.java
```diff
@@ -186,7 +186,7 @@ private ExecutorService buildExecutor(int numberOfTasks) {
 		};
 		return new ThreadPoolExecutor(
 				threadCount,
-				MAX_POOL_SIZE,
+				threadCount,
```
TipzCM, line 189: Can you add a comment here why the core threadcount is set to the max thread count?

If i were a dev reading this through later, i would not understand why this is necessary (since the documentation for ThreadPoolExecutor says that so long as we configure the idle time - which we do - those additional threads will close when not in use).

It's not obvious that too many threads are being used - and i suspect others might be confused about this as well (especially since we have multiple ThreadPoolExecutor's being created in different ways with different configs)

TipzCM (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/expunge/PartitionRunner.java
volodymyr-korzh: Fixed - added explanation comment

Issue #5634, opened 2024-01-26
Expansion fails for `ValueSet` with no concepts or filter based on `urn:ietf:bcp:13`


codeforgreen:
Expansion fails for `ValueSet` with no concepts or filter based on `urn:ietf:bcp:13`

For example:

```
{
    "resourceType": "ValueSet",
    "id": "custom-MimeType",
    "url": "http://example.com/custom-mimetype",
    "version": "1.0.0",
    "name": "MimeType",
    "title": "MimeType",
    "status": "active",
    "publisher": "Publisher",
    "description": "Attachment mime type supported by Publisher",
    "compose": {
        "include":  [
            {
                "system": "urn:ietf:bcp:13"
            }
        ]
    }
}
```

After storing this ValueSet, the expansion for it is automatically triggered in the background. That will result in the following error:  
 `HAPI-0702: Unable to expand ValueSet because CodeSystem could not be found: urn:ietf:bcp:13`

```
Caused by: ca.uhn.fhir.rest.server.exceptions.InternalErrorException: HAPI-0888: org.hl7.fhir.common.hapi.validation.support.InMemoryTerminologyServerValidationSupport$ExpansionCouldNotBeCompletedInternallyException: HAPI-0702: Unable to expand ValueSet because CodeSystem could not be found: urn:ietf:bcp:13
	at ca.uhn.fhir.jpa.term.TermReadSvcImpl.expandValueSetHandleIncludeOrExclude(TermReadSvcImpl.java:1040)
	at ca.uhn.fhir.jpa.term.TermReadSvcImpl.lambda$doExpandValueSet$2(TermReadSvcImpl.java:924)
	at org.springframework.transaction.support.TransactionOperations.lambda$executeWithoutResult$0(TransactionOperations.java:68)
	at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:140)
	at org.springframework.transaction.support.TransactionOperations.executeWithoutResult(TransactionOperations.java:67)
	at ca.uhn.fhir.jpa.term.TermReadSvcImpl.doExpandValueSet(TermReadSvcImpl.java:924)
	at ca.uhn.fhir.jpa.term.TermReadSvcImpl.expandValueSetIntoAccumulator(TermReadSvcImpl.java:639)
	at ca.uhn.fhir.jpa.term.TermReadSvcImpl.lambda$doExpandValueSet$1(TermReadSvcImpl.java:560)
[...]

PR #5638, opened 2024-01-26
Fix expansion for `ValueSet` with no concepts based on CodeSystem `urn:ietf:bcp:13`

codeforgreen:
#5634

Summary of changes:
- updated `CommonCodeSystemsTerminologyService` to return a `CodeSystem` with no concepts and content set to not-present for `urn:ietf:bcp:13`. Previously it was returning null. See: http://hl7.org/fhir/valueset-codesystem-content-mode.html#expansion
- Updated `InMemoryTerminologyServerValidationSupport` to avoid expansion failure for a `ValueSet` based on the above `CodeSystem`, for the case where the expansion is a full expansion.
- Added some test cases. Other common `CodeSystem` types were tested and some additional issues were found. I've captured them in #5643


hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupport.java
```diff
@@ -998,15 +996,22 @@ private boolean expandValueSetR5IncludeOrExclude(
...
-					if (isNotBlank(theInclude.getSystem())
-							&& !theInclude.getConcept().isEmpty()
-							&& theInclude.getFilter().isEmpty()
-							&& theInclude.getValueSet().isEmpty()) {
-						theInclude.getConcept().stream()
-								.map(t -> new FhirVersionIndependentConcept(
-										theInclude.getSystem(), t.getCode(), t.getDisplay(), theInclude.getVersion()))
-								.forEach(t -> nextCodeList.add(t));
-						ableToHandleCode = true;
+					boolean isIncludeFromSystem = isNotBlank(theInclude.getSystem())
```
jmarchionatto, line 999: What if both valueSet and concept are empty? Maybe the check for ampty concept should stay?

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupportTest.java
```diff
@@ -54,8 +50,159 @@ public void before() {
 		myDefaultSupport.fetchCodeSystem("http://foo");
 	}
 
+	@ParameterizedTest
+	@ValueSource(strings = {
+			CommonCodeSystemsTerminologyService.MIMETYPES_VALUESET_URL,
+			CommonCodeSystemsTerminologyService.CURRENCIES_VALUESET_URL,
+			CommonCodeSystemsTerminologyService.LANGUAGES_VALUESET_URL
+			// TODO:  other types cannot be expanded just yet, check if they need to be implemented
```
jmarchionatto, line 58: I think this TODO needs to be answered before approval.
Please look for an answer and add it as a comment

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupportTest.java
```diff
@@ -54,8 +50,159 @@ public void before() {
...
+		assertNotNull(valueSet.getExpansion());
+	}
+
+
+	@ParameterizedTest
+	@ValueSource(strings = {
+			CommonCodeSystemsTerminologyService.MIMETYPES_CODESYSTEM_URL,
+			CommonCodeSystemsTerminologyService.COUNTRIES_CODESYSTEM_URL,
+			CommonCodeSystemsTerminologyService.CURRENCIES_CODESYSTEM_URL
+			// TODO: other types cannot be expanded just yet, check if they need to be implemented
```
jmarchionatto, line 80: I think this TODO needs to be answered before approval.
Please look for an answer and add it as a comment

jmarchionatto (changes requested):
Added a few comments. Also, I would like to have the pointed TODO questions replied and responses added as comments, before approving

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupport.java
codeforgreen: I did leave it please see the if within this if which uses isIncludeWithDeclaredConcepts. All 4 parts of the conditin are kept. Basically clubbed together some parts of the condition to reuse and make the code easier to read. So my intention here is only to add an else and do some naming to the existing condition, but no change of behaviour to existing conditions.
Let me know if this makes sense.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupportTest.java
codeforgreen: I've removed the TODOs and captured them here: https://github.com/hapifhir/hapi-fhir/issues/5643. Let me know if anything else or more detail is needed.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupportTest.java
codeforgreen: I've removed the TODOs and captured them here: https://github.com/hapifhir/hapi-fhir/issues/5643. Let me know if anything else or more detail is needed.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/InMemoryTerminologyServerValidationSupport.java
jmarchionatto: It does. Than you

jmarchionatto (approved):
Looks good. Approved.

PR #5658, opened 2024-02-01
Skip MDM candidate search if parameters are missing

### LINKED ISSUE #5657: MDM Candidate Search performs search despite missing parameters
**Describe the bug**

If an MDM candidate search relies on two or more search parameters, and one of them is not present, the search proceeds with only the other search parameter (resulting in a potentially excessive result count) rather than being skipped as the documentation states.

To quote the documentation:

> If a list of searchParams is specified in a given candidateSearchParams item, then these search parameters are treated as AND parameters. In the following candidateSearchParams definition, hapi-fhir will extract given name, family name and identifiers from the incoming Patient and perform two separate searches, first for all Patient resources that have the same given AND the same family name as the incoming Patient, and second for all Patient resources that share at least one identifier as the incoming Patient. Note that if the incoming Patient was missing any of these searchParam values, then that search would be skipped. E.g. if the incoming Patient had a given name but no family name, then only a search for matching identifiers would be performed.

As stated above, what actually happens is that the search is performed with only the given name.

**To Reproduce**
1. In the MDM configuration, create a `candidateSearchParams` for `resourceType` "Patient", `searchParams` "given" and "family".
2. Add a patient with no family name.
3. Observe in the logs, MDM performs a search on given name alone, resulting in a much larger set of candidates than intended.

**Expected behavior**
MDM skips the search.

 - HAPI FHIR Version: 6.x
 - OS: Any

**Additional context**
Have fix, will contribute.

adriennesox:
Closes #5657 

Skip an MDM candidate search if any search parameter is missing (as per documentation, and to avoid non-performant searches), rather than only when all of them are missing.


adriennesox:
So, these unit test failures are ... interesting. It seems to be my bad luck that they happened to be close to the code I changed.

I wasn't able reproduce them through a Maven run, regardless of JVM, command-line options, etc. Even simulating the build as closely as possible, they remained completely irreproducible. Clearly, it wasn't directly caused by my changes.

Digging deeper, there are some mysteries. The part that's failing here - a timeout on a background job not finishing in clearMdmLinks - is supposed to happen after 10 seconds. However, the failed runs all reach 30 seconds. Furthermore, not every test fails on clearMdmLinks; there are successes intermixed with the failures.

Getting into the debugger, I accidentally recreated what seems to be this failure - there's an underlying connection pool (TestR4Config) that has a random number of maximum connections ... which just so happened to be the minimum 3 and allowed me to reproduce the issue.

So, what happens is:

- The connection pool used in unit tests has a 30s timeout, and 3-8 maximum connections (maximum determined randomly on initialization in TestR4Config).
- clearMdmLinks create an MDM Clear Job, which in turn creates 3 chunks to process. After that, clearMdmLinks waits up to 10s for the job to finish.
- Those 3 chunks get farmed out to separate batch2 threads.
- Those 3 threads all hit LoadGoldenIdsStep and which creates a transaction & borrows a connection.
- Each of those then submits a workchunk (ResourceIdListStep.submitWorkChunk) for the next step in the job, which creates a new transaction and tries to borrow a new connection to execute it.
- At this point, we have 3 connections borrowed, and 3 more that are trying to be borrowed. If the connection pool's maximum is greater than 3, this is fine - at least one can borrow a connection, submit their chunk, release the connection and eventaully their original connection, and nothing stays blocked for long. However, if there's only 3 connections, all of them block waiting for a connection to be returned (which cannot happen), so all 3 timeout on the borrow after 30s. This can be avoided if one of the steps gets to the submitWorkChunk step before another one gets the initial connection, but that's not guaranteed.
- Back on the original JUnit test thread, it is waiting 10s for these jobs to finish. After 10s if not earlier, it tries to check the status. However, this requires it to borrow a connection, and it too becomes blocked until one of the 3 running jobs times out, aborts, and release its held connection at the 30s mark.
- It's only possible for this issue to be encounted in 1 of 6 test runs thanks to TestR4Config. The rest of the time, the pool has more than 3 connections, and there is no issue.

adriennesox:
Updated the failed unit test to avoid this unit test-specific deadlock.

tadgh:
Heya, **I'm so sorry nobody has looked at this pull request yet**!

We have been absolutely floored with internal work but I hate to miss out on OSS contributions.

I will dedicate some time this week to have a deep dive on it and see if we can get it merged. Please feel free to `@` me here to get my attention if you don't hear back from me in a couple days

tadgh:
@adriennesox  This will be included with our August Release (7.4.0) or you can pull it via the OSS snapshots under version `7.3.5-SNAPSHOT` tomorrow when that publishes :)

Issue #5659, opened 2024-02-02
`NullPointerException` when invoking system bulk export with `PatientIdPartitionInterceptor`


codeforgreen:
Add `PatientIdPartitionInterceptor` to the interceptor registry.
POST http://localhost:8000/$export with header prefer=respond-async

```
Caused by: java.lang.NullPointerException: Cannot invoke "ca.uhn.fhir.interceptor.model.ReadPartitionIdRequestDetails.getResourceType()" because "theReadDetails" is null
	at ca.uhn.fhir.jpa.interceptor.PatientIdPartitionInterceptor.identifyForRead(PatientIdPartitionInterceptor.java:120)
	at ca.uhn.fhir.interceptor.executor.BaseInterceptorService$HookInvoker.invoke(BaseInterceptorService.java:550) [4 skipped]
	at ca.uhn.fhir.interceptor.executor.BaseInterceptorService.doCallHooks(BaseInterceptorService.java:287)
	at ca.uhn.fhir.interceptor.executor.BaseInterceptorService.callHooksAndReturnObject(BaseInterceptorService.java:258)
	at ca.uhn.fhir.interceptor.executor.BaseInterceptorService.callHooksAndReturnObject(BaseInterceptorService.java:63)
	at ca.uhn.fhir.rest.server.util.CompositeInterceptorBroadcaster$1.callHooksAndReturnObject(CompositeInterceptorBroadcaster.java:93)
	at ca.uhn.fhir.rest.server.util.CompositeInterceptorBroadcaster$1.callHooksAndReturnObject(CompositeInterceptorBroadcaster.java:75)
	at ca.uhn.fhir.rest.server.util.CompositeInterceptorBroadcaster.doCallHooksAndReturnObject(CompositeInterceptorBroadcaster.java:60)
	at ca.uhn.fhir.jpa.partition.BaseRequestPartitionHelperSvc.determineReadPartitionForRequest(BaseRequestPartitionHelperSvc.java:132)
	at ca.uhn.fhir.batch2.jobs.export.BulkDataExportProvider.startJob(BulkDataExportProvider.java:190)
	at ca.uhn.fhir.batch2.jobs.export.BulkDataExportProvider.export(BulkDataExportProvider.java:162)

```

PR #5660, opened 2024-02-02
Fix NullPointerException when performing a system bulk export in the presence of PatientIdPartitionInterceptor

codeforgreen:
#5659 

- populate the ReadPartitionIdRequestDetails which was null before causing the NPE
- fix forOperation method in ReadPartitionIdRequestDetails to return EXTENDED_OPERATION_SERVER when resource type and resource id are both null
- change logic in PatientIdPartitionInterceptor to return all partitions for EXTENDED_OPERATION_SERVER operations
- tests

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorTest.java
@@ -1,24 +1,32 @@
 package ca.uhn.fhir.jpa.interceptor;
 
 import ca.uhn.fhir.i18n.Msg;
+import ca.uhn.fhir.interceptor.model.ReadPartitionIdRequestDetails;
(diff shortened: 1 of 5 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorTest.java
```diff
@@ -47,21 +56,25 @@
...
+public class PatientIdPartitionInterceptorTest extends BaseResourceProviderR4Test {
 	public static final int ALTERNATE_DEFAULT_ID = -1;
-	private PatientIdPartitionInterceptor mySvc;
+
 	private ForceOffsetSearchModeInterceptor myForceOffsetSearchModeInterceptor;
 
 	@Autowired
 	private ISearchParamExtractor mySearchParamExtractor;
 
+	@SpyBean
```
lukedegruchy, line 70: Cool!  I didn't know there was such a thing!

lukedegruchy (approved):

PR #5677, opened 2024-02-06
5654 measurescorer bug for denominator exception

Capt-Mac:
Closes #5654

PR #5683, opened 2024-02-06
5654 measurescorer bug for denominator exception

### LINKED ISSUE #5654: MeasureScorer Bug for Denominator Exception
## Problem Statement
MeasureReports generated with denominator-exception population type are currently inaccurately scored in MeasureReport.MeasureScore. Currently it is calculating off of numerator/denominator instead following the standard for scoring: https://build.fhir.org/ig/HL7/cqf-measures/measure-conformance.html#proportion-measure-scoring
This assumes that denominator exception is removed from denominator set and included in denominator-exception set, but they can technically be in both and require counting in measureScore.
https://www.truthtabl.es/
I & D & !E & !(X & !N)

## Steps to Reproduce

run any Measure that contains a denominator exception population where a subject is not in numerator & in denominator & in denominator-exception. This patient should not be counted for measure score, but currently is.

## Observed Results
measure score does not correctly consider exception cases

Expected Results
https://www.truthtabl.es/
I & D & !E & !(X & !N)

Solution Design
https://github.com/cqframework/clinical-reasoning/issues/400

**Additional context**

Capt-Mac:
closes #5654


Capt-Mac:
replaced by https://github.com/hapifhir/hapi-fhir/pull/5677

PR #5712, opened 2024-02-18
Apply SearchNarrowingInterceptor to conditional URLs

jamesagnew:


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5712-apply-searchnarrowing-to-conditional-urls.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: add
+jira: SMILE-7971
+issue: 5712
+title: "The SearchNarrowingInterceptor can now optionally be configured to also apply
+  URL narrowing to conditional URLs used by conditional create/update/delete/patch
+  operations, both as raw HTTP transactionas as well as within FHIR transaction
```
fil512, line 7:
```suggestion
  operations, both as raw HTTP transactions as well as within FHIR transaction
```

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/SystemRequestDetails.java
```diff
@@ -124,11 +124,22 @@ public List<String> getHeaders(String name) {
 		return headers.get(name);
 	}
 
+	@Override
 	public void addHeader(String theName, String theValue) {
+		initHeaderMap();
+		myHeaders.put(theName, theValue);
```
fil512, line 130: thought: ServletRequestDetails header semantics are case insensitive, whereas these ones look case sensitive.  Not sure if this is a concern, but it could lead to confusion...

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/SearchNarrowingInterceptor.java
```diff
@@ -143,54 +220,212 @@ protected AuthorizedList buildAuthorizedList(@SuppressWarnings("unused") Request
...
+		String newConditionalUrl = null;
+		if (parametersToAdd != null) {
+
+			String query = theConditionalUrl;
+			int qMarkIndex = theConditionalUrl.indexOf('?');
+			if (qMarkIndex != -1) {
+				query = theConditionalUrl.substring(qMarkIndex + 1);
+			}
+
+			Map<String, String[]> inputParams = UrlUtil.parseQueryString(query);
```
fil512, line 320: thought: Too big for this MR, but UrlUtil.parseQueryString(query) should probably return a new class (called something like UrlParameterMap) with some of the logic below moved into that class

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/servlet/ServletRequestDetails.java
```diff
@@ -129,17 +132,60 @@ public FhirContext getFhirContext() {
 
 	@Override
 	public String getHeader(String name) {
+		// For efficiency, we only make a copy of the request headers if we need to
```
fil512, line 135: per earlier comment, I believe for ServletRequestDetails this is case insensitive

fil512 (approved):
nice tests!

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/SystemRequestDetails.java
jamesagnew: This is a great point. I'll fix this to preserve that behaviour.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/servlet/ServletRequestDetails.java
jamesagnew: I've reworked this to use case insensitive keys

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/SearchNarrowingInterceptor.java
jamesagnew: I could see it... agree though, too big for this mr

MathiasWeller42:
In our system, we have Practitioner users. These users get access to specific resources within Patient compartments that the Patients have allowed them to see (enforced via Consent resources). This means that for our use case, the Patient compartments that should be narrowed on depends on the resource type in the request. 
For example, `Patient/1` might have allowed a Practitioner to see `MedicationStatements` but not `DocumentReferences`, 
and `Patient/2` might have allowed the Practitioner to see `DocumentReferences` but not `MedicationStatements`.
Thus, when narrowing, it makes sense to add a `patient=Patient/1` to a MedicationStatement search, and `patient=Patient/2` to a `DocumentReference` search.
It seems that for Transactional GETs, the above use case is not supported anymore, since you only call `buildAuthorizedList()` once on the original transaction, and not for each request within the transaction.
Do you have any advice or a migration strategy? Is this a use case that you did not mean to support in the first place?

jamesagnew:
I'm not sure I follow. You're submitting a transaction with multiple GETs inside it that are searches which should be processed under different authorities?

Assuming that's it, yeah I'd say that isn't a use case we anticipated - I'd say the security infrastructure assumes that an HTTP transaction will have the same user authority from start to finish.

MathiasWeller42:
In this example, `Practitioner/12` has access to view DocumentReferences for `Patient/1` and `MedicationStatements` for `Patient/2`. We do search narrowing for this case by returning different `AuthorizedList` instances in `SearchNarrowingInterceptor` for the searches, because it is the best way to support our use case, but the `IAuthRules` returned in our `AuthorizationInterceptor` do not differ for the two searches.

jamesagnew:
Hmm - well, I guess I'd say that this isn't a part of the intended design of the interceptor, but it would be a reasonable thing to support if possible.

Probably the best solution would be an enhancement to `AuthorizedList` adding support for compartments for specific resource types.. e.g. adding a method like:

```java
public AuthorizedList addCompartment(String theSearchResourceType, String theCompartment) {
   .....
}
```

Feel free to file a feature enhancement ticket (or even better, a PR).

MathiasWeller42:
Thanks for the tip, I will create a feature enhancement ticket and maybe a PR :)

Issue #5713, opened 2024-02-20
Caused by: org.postgresql.util.PSQLException: ERROR: relation "public.hfj_spidx_string" does not exist


pano-smals:
We're upgrading to version 7.0.0 and while trying to start the application, the database migration fails with the error: 
`Caused by: org.postgresql.util.PSQLException: ERROR: relation "public.hfj_spidx_string" does not exist`

This table exists, but not in the schema `public`. 

The problem seems to be from this commit which executes a 'raw sql' (version `20231212.1`); 
`CREATE INDEX CONCURRENTLY idx_sp_string_hash_nrm_pattern_ops ON public.hfj_spidx_string USING btree (hash_norm_prefix, sp_value_normalized varchar_pattern_ops, res_id, partition_id)`

A little bit further in the code there is a similar query (version `20231212.2`) enforcing the public schema so that's another place where the code will fail once it gets there. 

The 7.0.0 was a long awaited version to be able to move to spring-boot 3 but this problem is blocking us. We're moving soon to PRD for the first time and we would've liked this upgrade to be done before. What's the chance of this fix to be done in the very near future? 

Or is there something we can do to make this work? Even a workaround is appreciated if this would take a while to be fixed. 

Also this is my first time reacting and I hope this is the way to mention this or do I need to go through another procedure for this to be picked up? 

Thanks in advance for looking into this.

_Originally posted by @pano-smals in https://github.com/hapifhir/hapi-fhir/issues/5544#issuecomment-1954228187_

pano-smals:
Adding to above the pieces of code: 
- https://github.com/hapifhir/hapi-fhir/blob/v7.0.0/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java#L186
- https://github.com/hapifhir/hapi-fhir/blob/v7.0.0/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java#L200

jamesagnew:
@michaelabuckley can you take a look at this?

pano-smals:
My colleague has happened to find a workaround. 

The piece of code contains a conditional that checks if this to be created index is present. So the workaround is, to manually run the queries with the `public` schema removed:
[...]

PR #5722, opened 2024-02-21
Bugfix: removed reference to public schema in raw sql

pano-smals:
Fixes: https://github.com/hapifhir/hapi-fhir/issues/5713


jamesagnew (approved):
Thanks for the contribution!

Issue #5723, opened 2024-02-21
Problem creating Hibernate dialect with specific version


pano-smals:
**Describe the bug**
When setting the Hibernate dialect to `ca.uhn.fhir.jpa.model.dialect.HapiFhirPostgresDialect` it fails using the constructor with a specific version and falls back to the default (10.0.0) version creating a warning in the logs for an unsupported version: 
> org.hibernate.dialect.Dialect            : HHH000511: The 10.0.0 version for [ca.uhn.fhir.jpa.model.dialect.HapiFhirPostgresDialect] is no longer supported, hence certain features may not work properly. The minimum supported version is 12.0.0. Check the community dialects project for available legacy versions.

**To Reproduce**
1. Set the hibernate dialect to `ca.uhn.fhir.jpa.model.dialect.HapiFhirPostgresDialect`
2. Start the application
3. View the warning in the logs

**Expected behavior**
Warning not appearing and the used Postgres version set. 

**Environment (please complete the following information):**
 - HAPI FHIR Version: 7.0.0

**Additional context**
`org.hibernate.engine.jdbc.dialect.internal.DialectFactoryImpl#constructDialect` expects a dialectClass constructor with a `DialectResolutionInfo` parameter. `HapiFhirPostgresDialect` only has a default constructor and one with a `DatabaseVersion` parameter. 

**Suggested solution**
Add an extra constructor in `HapiFhirPostgresDialect` with the required parameter:   
`
HapiFhirPostgresDialect(DialectResolutionInfo dialectResolutionInfo)
`

PR #5725, opened 2024-02-21
bugfix: creating Hibernate dialect with specific version

pano-smals:
fixes https://github.com/hapifhir/hapi-fhir/issues/5723


hapi-fhir-jpa/src/main/java/ca/uhn/fhir/jpa/model/dialect/HapiFhirPostgresDialect.java
```diff
@@ -32,6 +32,10 @@ public HapiFhirPostgresDialect() {
 		super(DatabaseVersion.make(10, 0, 0));
 	}
 
+		public HapiFhirPostgresDialect(DialectResolutionInfo info) {
```
jamesagnew, line 35: I'm assuming this is a boot thing? In all of our testing it was the constructor that takes `DatabaseVersion` as an argument that was required, but we use hibernate+spring without boot in most of our deployments.

If this constructor is needed by boot, shout it be added to the other dialects?

jamesagnew (approved):

PR #5734, opened 2024-02-23
Make seaech URL cleanup job clustered

jamesagnew:
A scheduled job to clean up the Search URL table used to enforce uniqueness of conditional create/update jobs was created as a local job and not a clustered job. This has been fixed.


hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceSearchUrlEntity.java
```diff
@@ -29,6 +29,15 @@
 
 import java.util.Date;
 
+/**
```
michaelabuckley, line 32: Nice!  Thanks for the javadoc.

michaelabuckley (approved):

Issue #5771, opened 2024-03-11
Patch operation failing for complex extension


epeartree:
**Describe the bug**
We do not get an expected result when running the PATCH operation with JSONPatch for a complex extension.

when running [Patch with an ADD operation](https://github.com/hapifhir/hapi-fhir/files/14558229/patchParameters.json) for an extension with an extension, the patched resource remains unchanged.

**Expected behavior**
that the patched resource be updated with the extension containing an extension.

PR #5776, opened 2024-03-12
5771 patch operation failing for complex extension

epeartree:
**Supporting documents:**
- [Extension](https://hl7.org/fhir/extensibility.html#extension)
- [Fhirpatch](https://build.fhir.org/fhirpatch.html)

**What was done:**
- added a failing test;
- for a Patch.Add operation, added constructions to recursively process embedded composite types such as Extension that can be composed of an endless chain of Extensions.  for an example of a patch Parameter with multiple embedded Extensions, refer to the #5771 problem statement.
- added changelog.

### DIFF
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java
@@ -34,7 +34,6 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.Validate;
 import org.hl7.fhir.instance.model.api.IBase;
 import org.hl7.fhir.instance.model.api.IBaseEnumeration;
-import org.hl7.fhir.instance.model.api.IBaseExtension;
 import org.hl7.fhir.instance.model.api.IBaseParameters;
 import org.hl7.fhir.instance.model.api.IBaseResource;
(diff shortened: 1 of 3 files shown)


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java
```diff
@@ -38,6 +39,7 @@
 import org.hl7.fhir.instance.model.api.IBaseResource;
 import org.hl7.fhir.instance.model.api.IIdType;
 import org.hl7.fhir.instance.model.api.IPrimitiveType;
+import org.hl7.fhir.r4.model.Parameters;
```
davidraeside, line 42: question: is it okay to use the R4 Parameters here? It looks like an R4 fix but I don't see other R4 model objects in this class. Just making sure, feel free to ignore.

davidraeside (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java
epeartree: good point, i'll take a look

epeartree: ya, so i've changed the logic to not have to use a Parameters.

Thanks.

PR #5809, opened 2024-03-28
Jr 20240325 concept map unmatched codes

JasonRoberts-smile:


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5816-unmatched-translate-target.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: add
+issue: 5816
+title: "The ConceptMap/$translate operation will include targets with an equivalence code of `unmatched` in the response 
+regardless of whether the target has a code."
```
tadgh, line 5: question: in what scenarios does it do this? It feels like a conditional is missing here, e.g. when do we resolve this unmatched? only if there is no direct match? "

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptClientMappingSvcImpl.java
```diff
@@ -416,13 +417,24 @@ private void buildTranslationResult(TranslateConceptResults theTranslationResult
 			theTranslationResult.setResult(false);
 			msg = myContext.getLocalizer().getMessage(TermConceptMappingSvcImpl.class, "noMatchesFound");
 			theTranslationResult.setMessage(msg);
+		} else if (isOnlyNegativeMatches(theTranslationResult)) {
+			theTranslationResult.setResult(false);
```
tadgh, line 421: question: does disjoint or unmatched not count as a result? it looks like an explicit result, just one wiht a type we dont like. Or is this `result` object used for something else?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptClientMappingSvcImpl.java
```diff
@@ -416,13 +417,24 @@ private void buildTranslationResult(TranslateConceptResults theTranslationResult
...
+			msg = myContext.getLocalizer().getMessage(TermConceptMappingSvcImpl.class, "onlyNegativeMatchesFound");
+			theTranslationResult.setMessage(msg);
 		} else {
 			theTranslationResult.setResult(true);
 			msg = myContext.getLocalizer().getMessage(TermConceptMappingSvcImpl.class, "matchesFound");
 			theTranslationResult.setMessage(msg);
 		}
 	}
 
+	private boolean isOnlyNegativeMatches(TranslateConceptResults theTranslationResult) {
```
tadgh, line 431: suggestion: recommend a javadoc explaining the design decision here.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptMappingSvcImpl.java
```diff
@@ -218,13 +219,17 @@ public void storeTermConceptMapAndChildren(ResourceTable theResourceTable, Conce
...
 									continue;
 								}
 								termConceptMapGroupElementTarget = new TermConceptMapGroupElementTarget();
 								termConceptMapGroupElementTarget.setConceptMapGroupElement(termConceptMapGroupElement);
-								termConceptMapGroupElementTarget.setCode(elementTarget.getCode());
-								termConceptMapGroupElementTarget.setDisplay(elementTarget.getDisplay());
+								if (isNotBlank(elementTarget.getCode())) {
+									termConceptMapGroupElementTarget.setCode(elementTarget.getCode());
+									termConceptMapGroupElementTarget.setDisplay(elementTarget.getDisplay());
+								}
```
tadgh, line 232: question: You have gated addition of the display element based on the presence or absence of a code. Is it not possible to have an UNMATCHED with no `code` but that _does_ have a `display` element? and if so, should we add it if it is present or no?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermConceptMappingSvcImplTest.java
```diff
@@ -189,6 +187,29 @@ protected void doInTransactionWithoutResult(@Nonnull TransactionStatus theStatus
...
 	}
 
+	@Test
+	public void testByCodeSystemsAndSourceCodeMatchedWithoutCode() {
+		createAndPersistConceptMap();
+		ConceptMap conceptMap = myConceptMapDao.read(myConceptMapId);
+
+		ourLog.debug("ConceptMap:\n" + myFhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(conceptMap));
+
+		new TransactionTemplate(myTxManager).execute(new TransactionCallbackWithoutResult() {
```
tadgh, line 197: issue -> replace with `runInTransaction` of basejpatest.

hapi-fhir-jpaserver-test-utilities/src/main/resources/migration/releases/V7_2_0/data/POSTGRES_9_4.sql
```diff
@@ -0,0 +1 @@
+INSERT INTO TRM_CONCEPT_MAP_GRP_ELM_TGT (PID, TARGET_CODE, CONCEPT_MAP_URL, TARGET_DISPLAY, TARGET_EQUIVALENCE, SYSTEM_URL, SYSTEM_VERSION, VALUESET_URL, CONCEPT_MAP_GRP_ELM_PID) VALUES (61, NULL, NULL, 'PYRIDOXINE', 'UNMATCHED', NULL, NULL, NULL, 60);
```
tadgh, line 1: Thought: I've never seen these files before. Is this for our multi-db migration testing? super cool

tadgh (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5816-unmatched-translate-target.yaml
JasonRoberts-smile: Nope, no conditional. There's nothing in the specification that would require one. If code `CodeSystemA|X` is explicitly unmatched in one place, and mapped to `CodeSystemB|Y` in another place, and if we receive a query that is broad enough to include both places, we will return both mappings.

This isn't really new behaviour. The status quo ante would already return a response that mixes exact matches, narrower-than, broader-than and disjoint, if the ConceptMaps are written such that all those things exist. The spec is very clear that the responsibility is on the caller to check the equivalence code before deciding to make use of a returned match.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptClientMappingSvcImpl.java
JasonRoberts-smile: From the spec
> The value can only be true if at least one returned match has an equivalence which is not unmatched or disjoint

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptClientMappingSvcImpl.java
JasonRoberts-smile: Fair enough. I'll add one.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermConceptMappingSvcImpl.java
JasonRoberts-smile: `display` is meant to contain a human-readable representation of `code`. There's no formal constraint preventing it from being populated even if `code` is blank, but I'm not sure what the business meaning of that pattern could possibly be. There is a separate field `comment` where the author of the ConceptMap can provide, for instance, a reason why the code was set to `unmatched`.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermConceptMappingSvcImplTest.java
JasonRoberts-smile: This was probably copy-pasted from one of the other test cases, so if it's worth fixing, I'll fix it in the whole class.

tadgh: Yeah may as well since you're in there. We have a template for doing this in tests may as well use it. Thanks!

PR #5810, opened 2024-03-28
Ensure CanonicalizedSubscription reflects StorageSettings.isCrossPartitionSubscriptionEnabled if it's true.  Make that setting is true by default

lukedegruchy:
- Ensure CanonicalizedSubscription is set to StorageSettings.isCrossPartitionSubscriptionEnabled() if it's true
- Ensure StorageSettings.isCrossPartitionSubscriptionEnabled() is true by default
- Add a new test and fix other tests to use the new SubscriptionCanonicalizer constructor.
- Deprecate FhirPatch rule

### DIFF
--- a/hapi-fhir-jpaserver-subscription/src/test/java/ca/uhn/fhir/jpa/subscription/module/subscriber/SubscriptionMatchingSubscriberTest.java
+++ b/hapi-fhir-jpaserver-subscription/src/test/java/ca/uhn/fhir/jpa/subscription/module/subscriber/SubscriptionMatchingSubscriberTest.java
@@ -27,6 +27,8 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.Answers;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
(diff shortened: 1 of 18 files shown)


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/match/registry/SubscriptionCanonicalizer.java
```diff
@@ -68,10 +69,23 @@ public class SubscriptionCanonicalizer {
...
 	final FhirContext myFhirContext;
+	private final StorageSettings myStorageSettings;
 
 	@Autowired
+	public SubscriptionCanonicalizer(FhirContext theFhirContext, StorageSettings theStorageSettings) {
+		myFhirContext = theFhirContext;
+		myStorageSettings = theStorageSettings;
+	}
+
+	// TODO:  LD:  remove this constructor once all callers call the 2 arg constructor above
```
TipzCM, line 80: You want to do this todo?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplPatch.java
```diff
@@ -27,6 +27,10 @@
 
 import java.util.Set;
 
+/**
+ * @deprecated Users should be instead be granted more granular write permissions that cover PATCH operations.
```
TipzCM, line 31: "Users should instead be granted more granular write permissions that cover PATCH operations"

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/partition/PartitionedSubscriptionTriggeringR4Test.java
```diff
@@ -146,69 +148,97 @@ public void testCreateSubscriptionInPartition() throws Exception {
...
+				// Should have 0 matching subscription, if we get 1 update count then the test fails
+				BaseSubscriptionsR4Test.ourPatientProvider.waitForUpdateCount(1);
+				if (! theIsCrossPartitionEnabled) {
+					fail();
+				}
+			} catch (ConditionTimeoutException e) {
+				if (theIsCrossPartitionEnabled) {
+					fail();
+				} else {
+					Assertions.assertEquals(0, BaseSubscriptionsR4Test.ourRestfulServer.getRequestContentTypes().size());
```
TipzCM, line 188: I feel like this needs more clear information why the exception is valid in this case.

Possibly a check on the exception itself to show it's not just an overlooked part of the test

hapi-fhir-jpaserver-subscription/src/test/java/ca/uhn/fhir/jpa/subscription/module/subscriber/SubscriptionMatchingSubscriberTest.java
```diff
@@ -227,76 +229,123 @@ public void testSubscriptionAndResourceOnTheSamePartitionMatchPart2() throws Int
...
+			sendSubscription(subscription, requestPartitionId, true);
+
+			final ThrowsInterrupted throwsInterrupted = () -> sendObservation(code, "SNOMED-CT", RequestPartitionId.fromPartitionId(0));
+			if (theIsCrossPartitionEnabled) {
+				runWithinLatchLogicExpectSuccess(throwsInterrupted);
+			} else {
+				runWithLatchLogicExpectFailure(throwsInterrupted);
+			}
+		} finally {
+			myStorageSettings.setCrossPartitionSubscriptionEnabled(oldIsCrossPartitionSubscriptionEnabled);
```
TipzCM, line 257: if it's at the final of every single test, i wonder if it's better to have an @AfterEach that resets the value to what we expect it should've been to begin with,

TipzCM (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/match/registry/SubscriptionCanonicalizer.java
lukedegruchy: Yes.  This is what's preventing me from having to bump. I'll get rid of it after the next bump.

PR #5819, opened 2024-04-03
Resolve 5818 Update Batch2 framework with GATE_WAITING state

TynerGjs:
What was done: 
- Implemented GATE_WAITING state for the batch2 state machine.
- This will be the initial status for all workchunks of a gated job.
- made compatible with the equivalent "legacy QUEUED" state in the Old batch2 implementation.
- Updated corresponding docs.
- added corresponding tests and changelog


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -383,34 +385,127 @@ public void testUpdateTime() {
...
+		runInTransaction(() -> assertEquals(TARGET_STEP_ID, myJobInstanceRepository.findById(instanceId).orElseThrow(IllegalArgumentException::new).getCurrentGatedStepId()));
+
+		// execute
+		runInTransaction(() -> mySvc.advanceJobStepAndUpdateChunkStatus(instanceId, LAST_STEP_ID));
+
+		// verify
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.GATE_WAITING, myWorkChunkRepository.findById(chunkIdFirstStep).orElseThrow(IllegalArgumentException::new).getStatus()));
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.READY, myWorkChunkRepository.findById(chunkIdSecondStep1).orElseThrow(IllegalArgumentException::new).getStatus()));
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.READY, myWorkChunkRepository.findById(chunkIdSecondStep2).orElseThrow(IllegalArgumentException::new).getStatus()));
+		runInTransaction(() -> assertEquals(LAST_STEP_ID, myJobInstanceRepository.findById(instanceId).orElseThrow(IllegalArgumentException::new).getCurrentGatedStepId()));
```
TipzCM, line 405: might be better to do this all in a single transaction?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -383,34 +385,127 @@ public void testUpdateTime() {
...
+		String instanceId = mySvc.storeNewInstance(instance);
+
+		// execute & verify
+		String id = storeWorkChunk(JOB_DEFINITION_ID, TARGET_STEP_ID, instanceId, 0, null, theGatedExecution);
+		runInTransaction(() -> assertEquals(theExpectedStatusOnCreate, myWorkChunkRepository.findById(id).orElseThrow(IllegalArgumentException::new).getStatus()));
 		myBatch2JobHelper.runMaintenancePass();
-		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.QUEUED, myWorkChunkRepository.findById(id).orElseThrow(IllegalArgumentException::new).getStatus()));
+		runInTransaction(() -> assertEquals(theExpectedStatusAfterTransition, myWorkChunkRepository.findById(id).orElseThrow(IllegalArgumentException::new).getStatus()));
+
+		if (WorkChunkStatusEnum.READY.equals(theExpectedStatusAfterTransition)) {
```
TipzCM, line 453: it's an enum. use == please.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/AbstractIJobPersistenceSpecificationTest.java
```diff
@@ -165,11 +166,14 @@ public JobInstance createInstance(JobDefinition<?> theJobDefinition) {
 		instance.setJobDefinitionVersion(JOB_DEF_VER);
 		instance.setParameters(CHUNK_DATA);
 		instance.setReport("TEST");
+        if (jobDefinition.isGatedExecution()) {
```
TipzCM, line 169: is there a formatting issue here?

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/AbstractIJobPersistenceSpecificationTest.java
```diff
@@ -165,11 +166,14 @@ public JobInstance createInstance(JobDefinition<?> theJobDefinition) {
...
+			instance.setCurrentGatedStepId(jobDefinition.getFirstStepId());
```
TipzCM, line 170: I knwo you are doing this for convenience, but i think unless this is affecting many many tests, i'd opt to have the tests set this value on the returned instance object itself.

The test should know what step it wants, after all.

It also makes it clear for the person reading the test (where it matters most)

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/IJobMaintenanceActions.java
```diff
@@ -82,32 +83,36 @@ default void test_gatedJob_stepReady_advances() throws InterruptedException {
...
-   		2|COMPLETED
-   		2|ERRORED
-   		2|FAILED
-   		2|IN_PROGRESS
-   		3|GATED
-   		3|GATED
+		# when current step is not all queued, should queue READY chunks
+		# Latch Count: 1
+		1|COMPLETED
+		2|READY,2|QUEUED
```
TipzCM, line 89: i'm not sure about this test case here.

This test is called `gatedStep2NotReady_notAdvance`

But this is clearly an advance since you have a final state that is.... different.

Not saying the test is unnecessary, but it doesn't belong here.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/support/JobMaintenanceStateInformation.java
```diff
@@ -145,7 +145,8 @@ public void initialize(IJobPersistence theJobPersistence) {
 		if (jobDef.isGatedExecution()) {
 			AtomicReference<String> latestStepId = new AtomicReference<>();
 			int totalSteps = jobDef.getSteps().size();
-			for (int i = totalSteps - 1; i >= 0; i--) {
+			// ignore the last step
```
TipzCM, line 148: why?

Comments should explain the business logic, not the code. i can see you're ignoring the last step, but i'm not sure why

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
```diff
@@ -280,4 +280,28 @@ default boolean onChunkDequeued(String theJobInstanceId) {
 
 	@VisibleForTesting
 	WorkChunk createWorkChunk(WorkChunk theWorkChunk);
+
+	/**
+	 * Atomically advance the given job to the given step and change the status of all chunks in the next step to READY
+	 * @param theJobInstanceId the id of the job instance to be updated
+	 * @param theNextStepId the id of the next job step
+	 * @return
```
TipzCM, line 288: you should explain what the return value is.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
```diff
@@ -280,4 +280,28 @@ default boolean onChunkDequeued(String theJobInstanceId) {
...
+	 */
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	boolean advanceJobStepAndUpdateChunkStatus(String theJobInstanceId, String theNextStepId);
+
+	/**
+	 * Update all chunks of the given step id for the given job from
```
TipzCM, line 294: from....?

Don't leave me hanging! From what?

----

I also wonder if this method is a bit 'too powerful'.

The advancing of the statemachine should not be just by instanceid and step. it should be a bit more controlled.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -398,32 +404,36 @@ private void processChunksForNextGatedSteps(
...
+			// update the job step so the workers will process them.
+			// if it's the last (gated) step, there will be no change - but we should
+			// queue up the chunks anyway
+			isEnqueue = theWorkCursor.isFinalStep()
+					|| myJobPersistence.advanceJobStepAndUpdateChunkStatus(instanceId, nextStepId);
+		} else {
+			// this means that the current step's chunks contains only READY and QUEUED chunks, possibly leftover from
+			// other maintenance job who died in the middle
+			// should enqueue the rest of the ready chunks
+			isEnqueue = true;
```
TipzCM, line 428: i like that you added comments to each if statment.

But because of the way this method is being invoked (after the canAdvanceGatedJob), and there's no clear way to link these, it might be helpful to add documentation to explain what states you expect here and why.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -575,4 +578,61 @@ private void invokePreStorageBatchHooks(JobInstance theJobInstance) {
...
+
+			// In the old code, gated jobs' workchunks are created in status QUEUED but not actually queued for the
+			// workers.
+			// In order to keep them compatible, turn QUEUED chunks into READY, too.
+			// TODO: remove this when we are certain that no one is still running the old code.
+			int numChanged = myWorkChunkRepository.updateAllChunksForStepWithStatus(
+					theJobInstanceId, theNextStepId, WorkChunkStatusEnum.READY, WorkChunkStatusEnum.QUEUED);
+			if (numChanged > 0) {
+				ourLog.debug(
+						"Updated {} chunks of gated instance {} for step {} from fake QUEUED to READY.",
```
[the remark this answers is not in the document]
michaelabuckley: "legacy" is a good euphemism.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -383,34 +385,127 @@ public void testUpdateTime() {
...
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.GATE_WAITING, myWorkChunkRepository.findById(chunkIdFirstStep).orElseThrow(IllegalArgumentException::new).getStatus()));
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.READY, myWorkChunkRepository.findById(chunkIdSecondStep1).orElseThrow(IllegalArgumentException::new).getStatus()));
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.READY, myWorkChunkRepository.findById(chunkIdSecondStep2).orElseThrow(IllegalArgumentException::new).getStatus()));
+		runInTransaction(() -> assertEquals(LAST_STEP_ID, myJobInstanceRepository.findById(instanceId).orElseThrow(IllegalArgumentException::new).getCurrentGatedStepId()));
 	}
 
 	@Test
-	public void testStoreAndFetchWorkChunk_NoData() {
-		JobInstance instance = createInstance(true, false);
+	public void updateAllChunksForStepWithStatus_forGatedJob_updatesChunkStatus() {
```
michaelabuckley, line 409: problem: Missing a chunk in QUEUED.  We might have mixed data here.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -383,34 +385,127 @@ public void testUpdateTime() {
...
+		String expectedFirstChunkData = "IAmChunk1";
+		String expectedSecondChunkData = "IAmChunk2";
+		JobInstance instance = createInstance(true, isGatedExecution);
+		String instanceId = mySvc.storeNewInstance(instance);
+
+		// execute & verify
+		String firstChunkId = storeWorkChunk(JOB_DEFINITION_ID, TARGET_STEP_ID, instanceId, 0, expectedFirstChunkData, isGatedExecution);
+		String secondChunkId = storeWorkChunk(JOB_DEFINITION_ID, LAST_STEP_ID, instanceId, 0, expectedSecondChunkData, isGatedExecution);
+
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.GATE_WAITING, myWorkChunkRepository.findById(firstChunkId).orElseThrow(IllegalArgumentException::new).getStatus()));
```
michaelabuckley, line 475: problem:  This is too hard to read.  Pleae combine the transaction lambdas, and introduce a getChunkOrThrow() helper to shorten these lines.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -125,8 +127,8 @@ public void testDeleteInstance() {
 		});
 	}
 
-	private String storeWorkChunk(String theJobDefinitionId, String theTargetStepId, String theInstanceId, int theSequence, String theSerializedData) {
-		WorkChunkCreateEvent batchWorkChunk = new WorkChunkCreateEvent(theJobDefinitionId, TestJobDefinitionUtils.TEST_JOB_VERSION, theTargetStepId, theInstanceId, theSequence, theSerializedData);
+	private String storeWorkChunk(String theJobDefinitionId, String theTargetStepId, String theInstanceId, int theSequence, String theSerializedData, boolean theGatedExecution) {
```
michaelabuckley, line 130: suggestion: do we we need the boolean?  The job definition is registered - we can check.
Otherwise, consider having two methods - storeWorkChunk and storeGatedWorkChunk.  Booleans make terrible parameters.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/IWorkChunkStorageTests.java
```diff
@@ -33,25 +30,29 @@ default void testStoreAndFetchWorkChunk_NoData() {
...
+	@ParameterizedTest
+	@CsvSource({
+		"false, READY",
+		"true, GATE_WAITING"
+	})
+	default void testWorkChunkCreate_inExpectedStatus(boolean theGatedExecution, WorkChunkStatusEnum expectedStatus) {
 		JobInstance instance = createInstance();
 		String instanceId = getSvc().storeNewInstance(instance);
 
 		enableMaintenanceRunner(false);
```
michaelabuckley, line 50: minor: should enableMaintenanceRunner just be in a @BeforeEach instead of every test?

michaelabuckley, line 50: trivial: I am skeptical this is better as a parameterized test instead of two named tests.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
```diff
@@ -280,4 +280,28 @@ default boolean onChunkDequeued(String theJobInstanceId) {
...
+	 * @param theJobInstanceId the id of the job instance to be updated
+	 * @param theStepId the id of the step which the chunks belong to
+	 * @param theNewStatus the new status
+	 * @param theOldStatus the old status
+	 * @return
+	 */
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	int updateAllChunksForStepWithStatus(
```
michaelabuckley, line 302: serious problem:  This method allows all sorts of transitions that aren't legal.  Is the new status ever not READY?  I think the old status is always QUEUED or GATE_WAITING too.  Please don't expose persistence methods that could break the state machine.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobDataSink.java
```diff
@@ -82,7 +82,13 @@ public void accept(WorkChunkData<OT> theData) {
...
 		// once finished, create workchunks in READY state
 		WorkChunkCreateEvent batchWorkChunk = new WorkChunkCreateEvent(
-				myJobDefinitionId, myJobDefinitionVersion, targetStepId, instanceId, sequence, dataValueString);
+				myJobDefinitionId,
+				myJobDefinitionVersion,
+				targetStepId,
+				instanceId,
+				sequence,
+				dataValueString,
+				myGatedExecution);
```
michaelabuckley, line 91: minor: it is clumsy to be passing around the JobInstance and JobDefinition by individual fields when we have the JobInstance and JobDefinition objects.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -245,15 +245,22 @@ private boolean canAdvanceGatedJob(
 			return true;
 		}
 
+		if (workChunkStatuses.equals(Set.of(WorkChunkStatusEnum.GATE_WAITING)) && theWorkCursor.isFirstStep()) {
```
michaelabuckley, line 248: question: Why would a first step ever be in GATE_WAITING state?  Shouldn't we start in READY.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -315,8 +322,7 @@ private void enqueueReadyChunks(
 				return JobWorkCursor.fromJobDefinitionAndRequestedStepId(theJobDefinition, metadata.getTargetStepId());
 			});
 			counter.getAndIncrement();
-			if (!theIsGatedExecutionAdvancementBool
-					&& (theJobDefinition.isGatedExecution() || jobWorkCursor.isReductionStep())) {
+			if (!theIsGatedExecutionBool && (theJobDefinition.isGatedExecution() || jobWorkCursor.isReductionStep())) {
```
michaelabuckley, line 325: problem: what does this method even do if this is false?  It looks like a no-op.  Can you push the boolean evaluation out to the caller and decide when to call the method?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -245,15 +245,22 @@ private boolean canAdvanceGatedJob(
...
+		}
+
 		if (workChunkStatuses.equals(Set.of(WorkChunkStatusEnum.COMPLETED))) {
-			// all previous workchunks are complete;
-			// none in READY though -> still proceed
+			// all workchunks for the current step are in COMPLETED -> proceed.
 			return true;
 		}
 
-		if (workChunkStatuses.equals(Set.of(WorkChunkStatusEnum.READY))) {
```
michaelabuckley, line 254: question: I doubt that set equality is the right test here.  Shouldn't this be something like `containsAll()`?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -245,15 +245,22 @@ private boolean canAdvanceGatedJob(
...
+		// all workchunks for gated jobs should be turned to QUEUED immediately after they are set to READY
```
michaelabuckley, line 259: problem:  This comment makes me doubt our transaction boundary.  We must _guarantee_ that this doesn't happen via transaction boundaries.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -398,32 +404,36 @@ private void processChunksForNextGatedSteps(
...
-						return false;
-					}
-					instance.setCurrentGatedStepId(nextStepId);
-					return true;
-				});
+		boolean isEnqueue;
+		String currentStepId = theWorkCursor.getCurrentStepId();
+		Set<WorkChunkStatusEnum> workChunkStatusesForCurrentStep =
+				myJobPersistence.getDistinctWorkChunkStatesForJobAndStep(theInstance.getInstanceId(), currentStepId);
+		if (workChunkStatusesForCurrentStep.equals(Set.of(WorkChunkStatusEnum.GATE_WAITING))
```
michaelabuckley, line 411: problem: these different cases should almost certainly be separate methods.  They have radically different behaviour.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -575,4 +578,61 @@ private void invokePreStorageBatchHooks(JobInstance theJobInstance) {
...
+					theNextStepId);
+			// when we reach here, the current step id is equal to theNextStepId
+			myWorkChunkRepository.updateAllChunksForStepWithStatus(
+					theJobInstanceId, theNextStepId, WorkChunkStatusEnum.READY, WorkChunkStatusEnum.GATE_WAITING);
+
+			// In the old code, gated jobs' workchunks are created in status QUEUED but not actually queued for the
+			// workers.
+			// In order to keep them compatible, turn QUEUED chunks into READY, too.
+			// TODO: remove this when we are certain that no one is still running the old code.
+			int numChanged = myWorkChunkRepository.updateAllChunksForStepWithStatus(
```
michaelabuckley, line 608: performance: these two calls should be a single db call.  Please change to take a set of allowed prior states and use an `in()` clause in the query.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -575,4 +578,61 @@ private void invokePreStorageBatchHooks(JobInstance theJobInstance) {
 			myInterceptorBroadcaster.callHooks(Pointcut.STORAGE_PRESTORAGE_BATCH_JOB_CREATE, params);
 		}
 	}
+
+	@Override
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	public boolean advanceJobStepAndUpdateChunkStatus(String theJobInstanceId, String theNextStepId) {
+		boolean changed = updateInstance(theJobInstanceId, instance -> {
+			if (instance.getCurrentGatedStepId().equals(theNextStepId)) {
```
michaelabuckley, line 586: problem: I don't see a test for this branch.

michaelabuckley (changes requested):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_batch/batch2_states.md
```diff
@@ -56,10 +56,11 @@ stateDiagram-v2
...
     state COMPLETED
    direction LR
-   [*]         --> READY         : on create - normal
-   [*]         --> GATED         : on create - gated
-   GATED       --> READY         : on create - gated
-   READY       --> QUEUED        : placed on kafka (maint.)
+   [*]             --> READY         : on create - normal
+   [*]             --> GATE_WAITING  : on create - gated
+   GATE_WAITING    --> READY         : on gate release - gated (new)
+   QUEUED          --> READY         : on gate release - gated (for compatibility with old "fake QUEUED" state)
```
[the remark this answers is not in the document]
TynerGjs: added snapshot number

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_batch/introduction.md
```diff
@@ -19,14 +19,14 @@ A HAPI-FHIR batch job definition consists of a job name, version, parameter json
 After a job has been defined, *instances* of that job can be submitted for batch processing by populating a `JobInstanceStartRequest` with the job name and job parameters json and then submitting that request to the Batch Job Coordinator.
 
 The Batch Job Coordinator will then store two records in the database:
-- Job Instance with status QUEUED: that is the parent record for all data concerning this job
-- Batch Work Chunk with status READY: this describes the first "chunk" of work required for this job.  The first Batch Work Chunk contains no data.
+- Job Instance with status `QUEUED`: that is the parent record for all data concerning this job
+- Batch Work Chunk with status `READY`/`GATE_WAITING`: this describes the first "chunk" of work required for this job. The first Batch Work Chunk contains no data. The initial status of the work chunk will be `READY` or `GATE_WAITING` for non-gated and gated batch jobs, respectively. Please refer to [Gated Execution](#gated-execution) for more explanation on gated batch jobs.
```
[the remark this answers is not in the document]
TynerGjs: used recommended text

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_batch/introduction.md
```diff
@@ -19,14 +19,14 @@ A HAPI-FHIR batch job definition consists of a job name, version, parameter json
...
 
 ### The Maintenance Job
 
 A Scheduled Job runs periodically (once a minute).  For each Job Instance in the database, it:
 
-1. Moves all `READY` work chunks into the `QUEUED` state and publishes a message to the Batch Notification Message Channel to inform worker threads that a work chunk is now ready for processing. \*
+1. Moves all `READY` work chunks into the `QUEUED` state and publishes a message to the Batch Notification Message Channel to inform worker threads that a work chunk is now ready for processing. For gated batch jobs, the maintenance also moves all `GATE_WAITING` work chunks into `READY` when the current batch step is ready to advance. \*
```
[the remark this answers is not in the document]
TynerGjs: moved to separate point and reordered the points

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
TynerGjs: added new test case

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -575,4 +578,61 @@ private void invokePreStorageBatchHooks(JobInstance theJobInstance) {
...
+		if (changed) {
+			ourLog.debug(
+					"Updating chunk status from GATE_WAITING to READY for gated instance {} in step {}.",
+					theJobInstanceId,
+					theNextStepId);
+			// when we reach here, the current step id is equal to theNextStepId
+			myWorkChunkRepository.updateAllChunksForStepWithStatus(
+					theJobInstanceId, theNextStepId, WorkChunkStatusEnum.READY, WorkChunkStatusEnum.GATE_WAITING);
+
+			// In the old code, gated jobs' workchunks are created in status QUEUED but not actually queued for the
```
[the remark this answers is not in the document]
TynerGjs: added version number

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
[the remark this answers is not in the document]
TynerGjs: used "legacy" and added version number

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -575,4 +578,61 @@ private void invokePreStorageBatchHooks(JobInstance theJobInstance) {
...
+		return changed;
+	}
+
+	@Override
+	@Transactional(propagation = Propagation.REQUIRES_NEW)
+	public int updateAllChunksForStepWithStatus(
+			String theJobInstanceId,
+			String theStepId,
+			WorkChunkStatusEnum theNewStatus,
+			WorkChunkStatusEnum theOldStatus) {
```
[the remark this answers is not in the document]
TynerGjs: restricted this method to only turn GATE_WAITING and QUEUED to READY, and removed the last two parameters.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -173,6 +174,36 @@ public void fetchInstance_validId_returnsInstance() {
...
 	}
 
+	@Test
+	void advanceJobStepAndUpdateChunkStatus_validRequest_callsPersistenceUpdateAndReturnsChanged() {
+		// setup
+		String instanceId = "jobId";
+		String nextStepId = "nextStep";
+
+		// execute
+		mySvc.advanceJobStepAndUpdateChunkStatus(instanceId, nextStepId);
```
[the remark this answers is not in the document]
TynerGjs: moved to ca.uhn.fhir.jpa.batch2.JpaJobPersistenceImplTest and verified the return value

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -173,6 +174,36 @@ public void fetchInstance_validId_returnsInstance() {
...
+
+		// verify
+		verify(mySvc).updateInstance(instanceId, any());
+		verify(myWorkChunkRepository).updateAllChunksForStepWithStatus(instanceId, nextStepId, WorkChunkStatusEnum.READY, WorkChunkStatusEnum.GATE_WAITING);
+		verify(myWorkChunkRepository).updateAllChunksForStepWithStatus(instanceId, nextStepId, WorkChunkStatusEnum.READY, WorkChunkStatusEnum.QUEUED);
+	}
```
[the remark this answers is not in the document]
TynerGjs: added new test case

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -173,6 +174,36 @@ public void fetchInstance_validId_returnsInstance() {
...
+	@Test
+	void updateAllChunksForStepWithStatus_validRequest_callsPersistenceUpdateAndReturnsChanged() {
+		// setup
+		String instanceId = "jobId";
+		String nextStepId = "nextStep";
+		WorkChunkStatusEnum expectedNewStatus = WorkChunkStatusEnum.READY;
+		WorkChunkStatusEnum expectedOldStatus = WorkChunkStatusEnum.GATE_WAITING;
+
+		// execute
+		mySvc.updateAllChunksForStepWithStatus(instanceId, nextStepId, expectedNewStatus, expectedOldStatus);
```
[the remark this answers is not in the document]
TynerGjs: verified return value

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/AbstractIJobPersistenceSpecificationTest.java
TynerGjs: yes... seems the IJ rendering issue happened again, reformatted

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
TynerGjs: restricted this method to only transition from QUEUED and GATE_WAITING to READY

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
TynerGjs: restricted this method to only transition from QUEUED and GATE_WAITING to READY

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: changed the first step to always be READY

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: this if branch is removed to allow only 1 code path to enqueue ready chunks :)

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: this if branch is removed to allow only 1 code path to enqueue ready chunks :)

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: this if branch is removed to allow only 1 code path to enqueue ready chunks :)

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: the majority of these branches are removed to allow only 1 code path to enqueue ready chunks :)

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobDataSink.java
TynerGjs: The jobDataSink does not have a field for the job definition object and instead parsed all the info out from it in the constructor. I think this might be intended when the code is first written, this pattern also applies to when creating JobWorkNotification... Should I change it so that it stores a job definition directly? If you think it is ok to do, I am happy to make it

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/support/JobMaintenanceStateInformation.java
TynerGjs: Previously, this would set the current step to the latest step that appeared in the initial states. However, that is not very useful since our tests that use gated jobs all need the current step to be the step before the last step, so I skipped the last step... I see that it is probably not a good solution. But I also find that automatically setting the current step by judging from the initial states might not be very useful. Should I introduce a new field in the initial states to allow the current step to be declared?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -383,34 +385,127 @@ public void testUpdateTime() {
...
+		String chunkIdSecondStep1 = storeWorkChunk(JOB_DEFINITION_ID, LAST_STEP_ID, instanceId, 0, null, true);
+		String chunkIdSecondStep2 = storeWorkChunk(JOB_DEFINITION_ID, LAST_STEP_ID, instanceId, 0, null, true);
+
+		runInTransaction(() -> assertEquals(TARGET_STEP_ID, myJobInstanceRepository.findById(instanceId).orElseThrow(IllegalArgumentException::new).getCurrentGatedStepId()));
+
+		// execute
+		runInTransaction(() -> mySvc.advanceJobStepAndUpdateChunkStatus(instanceId, LAST_STEP_ID));
+
+		// verify
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.GATE_WAITING, myWorkChunkRepository.findById(chunkIdFirstStep).orElseThrow(IllegalArgumentException::new).getStatus()));
```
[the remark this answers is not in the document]
TynerGjs: added suggested method

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
TynerGjs: added suggested method

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -367,50 +374,178 @@ public void runMaintenancePass() {
...
+	}
+
+	@Test
+	public void updateAllChunksForStepWithStatus_forGatedJob_updatesChunkStatus() {
+		// setup
+		JobInstance instance = createInstance(true, true);
+		String instanceId = mySvc.storeNewInstance(instance);
+		String chunkIdFirstStep = storeWorkChunk(JOB_DEFINITION_ID, TARGET_STEP_ID, instanceId, 0, null, true);
+		String chunkIdSecondStep1 = storeWorkChunk(JOB_DEFINITION_ID, LAST_STEP_ID, instanceId, 0, null, true);
+		String chunkIdSecondStep2 = storeWorkChunk(JOB_DEFINITION_ID, LAST_STEP_ID, instanceId, 0, null, true);
```
TipzCM, line 447: Did you consider using the `JobMaintenanceStateInformation` for these tests?

It uses raw strings to set up most of the stuff. Which makes reading and verifying a bit simplier.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -367,50 +374,178 @@ public void runMaintenancePass() {
...
+		runInTransaction(() -> {
+			assertEquals(WorkChunkStatusEnum.GATE_WAITING, findChunkByIdOrThrow(chunkIdFirstStep).getStatus());
+			assertEquals(WorkChunkStatusEnum.READY, findChunkByIdOrThrow(chunkIdSecondStep1).getStatus());
+			assertEquals(WorkChunkStatusEnum.READY, findChunkByIdOrThrow(chunkIdSecondStep2).getStatus());
+		});
+	}
+
+	@Test
+	public void testFetchUnknownWork() {
+		assertFalse(myWorkChunkRepository.findById("FOO").isPresent());
```
TipzCM, line 471: This test is likely not needed.

`findById` is an auto-generated method we didn't write (and so we shouldn't test)

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -367,50 +374,178 @@ public void runMaintenancePass() {
...
+		myBatch2JobHelper.runMaintenancePass();
+		runInTransaction(() -> {
+			assertEquals(WorkChunkStatusEnum.COMPLETED, findChunkByIdOrThrow(firstChunkId).getStatus());
+			assertEquals(WorkChunkStatusEnum.QUEUED, findChunkByIdOrThrow(secondChunkId).getStatus());
+		});
+
+		WorkChunk actualSecondChunkData = mySvc.onWorkChunkDequeue(secondChunkId).orElseThrow(IllegalArgumentException::new);
+		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.IN_PROGRESS, findChunkByIdOrThrow(secondChunkId).getStatus()));
+		assertEquals(expectedSecondChunkData, actualSecondChunkData.getData());
+	}
```
TipzCM, line 537: I feel like this test should be broken up into multiple tests.

One test for each transition.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/AbstractIJobPersistenceSpecificationTest.java
```diff
@@ -165,11 +166,19 @@ public JobInstance createInstance(JobDefinition<?> theJobDefinition) {
...
 		instance.setReport("TEST");
+		if (jobDefinition.isGatedExecution()) {
+			instance.setCurrentGatedStepId(jobDefinition.getFirstStepId());
+		}
 		return instance;
 	}
 
-	public String storeWorkChunk(String theJobDefinitionId, String theTargetStepId, String theInstanceId, int theSequence, String theSerializedData) {
-		WorkChunkCreateEvent batchWorkChunk = new WorkChunkCreateEvent(theJobDefinitionId, JOB_DEF_VER, theTargetStepId, theInstanceId, theSequence, theSerializedData);
+	public String storeWorkChunk(String theJobDefinitionId, String theTargetStepId, String theInstanceId, int theSequence, String theSerializedData, boolean theGatedExecution) {
```
TipzCM, line 175: Eh, i'll allow it. But this really is getting to the point where we have enough parameters we should instead be using a parameters object.

4 strings can be difficult to get correct (especially since 3 of them are called "*id"

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/IJobMaintenanceActions.java
TipzCM: Can you rename the test to make that clearer please?

You are probably right, but i imagine others will make my same mistake

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/ITestFixture.java
```diff
@@ -62,6 +62,12 @@ public interface ITestFixture {
 	 */
 	String createChunk(String theJobInstanceId);
 
+	String createChunk(String theJobInstanceId, boolean theGatedExecution);
+
+	String createChunkInStep(String theJobInstanceId, String theStepId, boolean theGatedExecution);
+
+	String createFirstChunk(JobDefinition<TestJobParameters> theJobDefinition, String theJobInstanceId);
```
TipzCM, line 69: I question the need for all these methods.

We have a perfectly valid way of creating chunks via the JobMaintenanceStateInformation object and syntax.

---

Even if these methods are needed (and i'm not convinced that they are), it should be a) documented and b) one method with parameters object instead.

only reason other methods weren't documented was because it was a refactor, and not new interface values being added.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/support/JobMaintenanceStateInformation.java
TipzCM: please update the comment to explain why (briefly)

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
```diff
@@ -280,4 +280,22 @@ default boolean onChunkDequeued(String theJobInstanceId) {
 
 	@VisibleForTesting
 	WorkChunk createWorkChunk(WorkChunk theWorkChunk);
+
+	/**
+	 * Atomically advance the given job to the given step and change the status of all chunks in the next step to READY
```
TipzCM, line 285: all chunks? Regardless of state?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -239,36 +232,14 @@ private boolean canAdvanceGatedJob(
...
-				// completed means "all in COMPLETE state" or no previous chunks (they're cleaned up or never existed)
-				if (previousStepWorkChunkStates.isEmpty()
-						|| previousStepWorkChunkStates.equals(Set.of(WorkChunkStatusEnum.COMPLETED))) {
-					return true;
-				}
-			}
+		if (workChunkStatuses.isEmpty()) {
+			// no work chunks = no output
+			// trivial to advance to next step
+			return true;
```
TipzCM, line 238: You might need to merge in latest; this code has been updated to not do this.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -239,36 +232,14 @@ private boolean canAdvanceGatedJob(
...
 		}
 
-		// anything else
-		return false;
+		// all workchunks for the current step are in COMPLETED -> proceed.
+		return workChunkStatuses.equals(Set.of(WorkChunkStatusEnum.COMPLETED));
```
TipzCM, line 242: Q: is there any chance of a race condition where we have a COMPLETED previous chunk, but the GATE_WAITING new chunk is not yet created / ready to transition?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -192,19 +190,15 @@ private void triggerGatedExecutions(JobInstance theInstance, JobDefinition<?> th
...
 				JobWorkCursor<?, ?, ?> nextJobWorkCursor = JobWorkCursor.fromJobDefinitionAndRequestedStepId(
 						jobWorkCursor.getJobDefinition(), jobWorkCursor.getCurrentStepId());
 				myReductionStepExecutorService.triggerReductionStep(instanceId, nextJobWorkCursor);
-			} else if (jobWorkCursor.isFinalStep()) {
-				// current step is the final step in a non-reduction gated job
-				processChunksForNextGatedSteps(
-						theInstance, theJobDefinition, jobWorkCursor, jobWorkCursor.getCurrentStepId());
-			} else {
-				// all other gated job steps
+			} else if (!jobWorkCursor.isFinalStep()) {
```
TipzCM, line 200: I imagine this is because final steps of gated jobs are never going to get here?

Can we comment that somwhere. Your comment here is just "all other gated job steps except for final steps". Which is a bit confusing, because the only final step being handled here is the reduction step.

And if i'm reading this code, i might be confused why you don't handle the non-reduction gated final steps.

----

Also, as a good practice, we should still have the `else` with a log - even if it's something we expect to *never* see (so in case we do see it, we know what it is).

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -326,7 +280,10 @@ private void enqueueReadyChunks(
 			updateChunkAndSendToQueue(metadata);
 		}
 		ourLog.debug(
-				"Encountered {} READY work chunks for job {}", counter.get(), theJobDefinition.getJobDefinitionId());
+				"Encountered {} READY work chunks for job {} of type {}",
+				counter.get(),
```
TipzCM, line 284: You might as well just remove the counter entirely - it's not being updated.

And if you decide to keep it, it doesn't need to be atomic (since now it's a standard iteration).

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -289,32 +260,15 @@ protected PagingIterator<WorkChunkMetadata> getReadyChunks() {
...
 		AtomicInteger counter = new AtomicInteger();
-		ConcurrentHashMap<String, JobWorkCursor<?, ?, ?>> stepToWorkCursor = new ConcurrentHashMap<>();
 		while (iter.hasNext()) {
 			WorkChunkMetadata metadata = iter.next();
-			JobWorkCursor<?, ?, ?> jobWorkCursor = stepToWorkCursor.computeIfAbsent(metadata.getTargetStepId(), (e) -> {
-				return JobWorkCursor.fromJobDefinitionAndRequestedStepId(theJobDefinition, metadata.getTargetStepId());
-			});
-			counter.getAndIncrement();
-			if (!theIsGatedExecutionAdvancementBool
-					&& (theJobDefinition.isGatedExecution() || jobWorkCursor.isReductionStep())) {
```
TipzCM, line 308: I still think we have to block gated reduction steps, no?

Remember, this application is multi-threaded, multi-instance.

You have other code that transitions all these GATE_WAITING chunks to READY (for all steps).

You might be thinking that these will then get handled by the next pass through.

But if a different instance is running the maintenance job, it will get here and see a READY chunk that it wants to send to the QUEUE.

But this READY chunk is actually a Reduction Step READY chunk and should not be sent to the queue, ever!

---

You cannot rely on "order of code" for business logic to work as expected.

It must be explicitly defined. Even if this weren't multi-threaded multi-instance (refactors happen after all).

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
```diff
@@ -99,9 +97,9 @@ public void process() {
 		JobDefinition<? extends IModelJson> jobDefinition =
 				myJobDefinitionegistry.getJobDefinitionOrThrowException(theInstance);
 
-		enqueueReadyChunks(theInstance, jobDefinition, false);
 		cleanupInstance(theInstance);
 		triggerGatedExecutions(theInstance, jobDefinition);
+		enqueueReadyChunks(theInstance, jobDefinition);
```
TipzCM, line 102: Q: Why did you move this to the bottom?

I'm not saying it's wrong, i'm just curious.

Your changes should (in theory) mean the gate-waiting boolean isn't necessary at all (GATE_WAITING chunks are not enqueued until all are set to READY).

It doesn't matter if it's at the end (and maybe it makes better sense there), but i am just curious.

TipzCM (changes requested):

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: yes, if the job is at the final step and has all chunks completed, then it will be marked as chunks purged. Added comments to explain and an `else` branch to log

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: There shouldn't be, since a chunk is only set to COMPLETED after successfully executing `run`, which includes creating the work chunk for the next step.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobPersistence.java
TynerGjs: added state QUEUED and GATE_WAITING

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IBatch2WorkChunkRepository.java
```diff
@@ -109,6 +109,19 @@ int updateChunkStatus(
 			@Param("newStatus") WorkChunkStatusEnum theNewStatus,
 			@Param("oldStatus") WorkChunkStatusEnum theOldStatus);
 
+	// Up to 7.1.8-SNAPSHOT, gated jobs' work chunks are created in status QUEUED but not actually queued for the
```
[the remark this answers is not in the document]
TynerGjs: used 7.1 instead

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -128,7 +128,10 @@ public String onWorkChunkCreate(WorkChunkCreateEvent theBatchWorkChunk) {
 		entity.setSerializedData(theBatchWorkChunk.serializedData);
 		entity.setCreateTime(new Date());
 		entity.setStartTime(new Date());
-		entity.setStatus(WorkChunkStatusEnum.READY);
+		// set to GATE_WAITING if job is gated, to READY if not
```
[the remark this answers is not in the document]
TynerGjs: used suggested text as comment

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_batch/introduction.md
```diff
@@ -19,18 +19,18 @@ A HAPI-FHIR batch job definition consists of a job name, version, parameter json
...
 
 ### The Maintenance Job
 
 A Scheduled Job runs periodically (once a minute).  For each Job Instance in the database, it:
 
-1. Moves all `READY` work chunks into the `QUEUED` state and publishes a message to the Batch Notification Message Channel to inform worker threads that a work chunk is now ready for processing. \*
 1. Calculates job progress (% of work chunks in `COMPLETE` status). If the job is finished, purges any left over work chunks still in the database.
 1. Cleans up any complete, failed, or cancelled jobs that need to be removed.
-1. Moves any gated jobs onto their next step when complete.
+1. When the current step is complete, moves any gated jobs onto their next step and update all chunks in `GATE_WAITING` to `READY`.
```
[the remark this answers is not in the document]
TynerGjs: fixed gramma error

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_batch/batch2_states.md
```diff
@@ -56,10 +56,11 @@ stateDiagram-v2
     state FAILED
     state COMPLETED
    direction LR
-   [*]         --> READY         : on create - normal
-   [*]         --> GATED         : on create - gated
-   GATED       --> READY         : on create - gated
-   READY       --> QUEUED        : placed on kafka (maint.)
+   [*]             --> READY         : on create - normal or gated first chunk
+   [*]             --> GATE_WAITING  : on create - gated non-first chunk
```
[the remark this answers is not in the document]
TynerGjs: used suggested text

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: Yes, the sequence doesn't affect the correctness, the chunks will still be picked up. But I moved it because I think it would be good to have only one path to the `enqueueReadyChunks` method. By putting this at the end, we don't have to call `enqueueReadyChunks` in `triggerGatedExecutions` and it can still catch all the chunks that were just turned to READY in `triggerGatedExecutions`. This saves the time waiting for the next job maintenance.

Also, another reason (and probably the major reason) is that Michael suggested to do so xD.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/support/JobMaintenanceStateInformation.java
TynerGjs: updated comment

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/IJobMaintenanceActions.java
TynerGjs: renamed

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: Do you mean the latest master? I merged in the ready state branch but this check still exists

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/maintenance/JobInstanceProcessor.java
TynerGjs: kept the counter for log and changed to int

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/ITestFixture.java
TynerGjs: Removed createChunkInStep and used the JobMaintenanceStateInformation instead. I think we still need the createFirstChunk to call the `WorkChunkCreateEvent.firstChunk` instead of creating the event directly for gated first chunk creation.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
TynerGjs: There are tests that verify the individual transitions in the `IJobMaintanenceAction` test. I wrote this test to simulate the job from creation to the last step. It makes me feel more confident with the correctness. Do you think it is necessary?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
TynerGjs: Moved to `IWorkChunkStateTransitions` and used `JobMaintenanceStateInformation ` instead

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IBatch2WorkChunkRepository.java
```diff
@@ -109,6 +109,19 @@ int updateChunkStatus(
...
+	// workers.
+	// In order to keep them compatible, turn QUEUED chunks into READY, too.
+	// TODO: remove QUEUED from the IN clause when we are certain that no one is still running the old code.
+	@Modifying
+	@Query("UPDATE Batch2WorkChunkEntity e SET e.myStatus = ca.uhn.fhir.batch2.model.WorkChunkStatusEnum.READY WHERE "
```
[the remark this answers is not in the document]
michaelabuckley: minor: agree. my complaint was the public api, but internal implementations can be more general.  This is fine for now.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -367,50 +374,149 @@ public void runMaintenancePass() {
...
+
+		// execute & verify
+		String id = storeWorkChunk(JOB_DEFINITION_ID, LAST_STEP_ID, instanceId, 0, null, theGatedExecution);
+		runInTransaction(() -> assertEquals(theExpectedStatusOnCreate, findChunkByIdOrThrow(id).getStatus()));
 		myBatch2JobHelper.runMaintenancePass();
-		runInTransaction(() -> assertEquals(WorkChunkStatusEnum.QUEUED, myWorkChunkRepository.findById(id).orElseThrow(IllegalArgumentException::new).getStatus()));
+		runInTransaction(() -> assertEquals(theExpectedStatusAfterTransition, findChunkByIdOrThrow(id).getStatus()));
 
 		WorkChunk chunk = mySvc.onWorkChunkDequeue(id).orElseThrow(IllegalArgumentException::new);
 		assertNull(chunk.getData());
```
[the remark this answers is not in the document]
TynerGjs: This test was specifically testing without data, there's a corresponding test with data. Added comments to explain.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/IJobMaintenanceActions.java
```diff
@@ -69,45 +68,50 @@ default void test_gatedJob_stepReady_advances() throws InterruptedException {
...
+		2|COMPLETED
+		2|ERRORED
+		2|FAILED
+		2|IN_PROGRESS
+		3|GATE_WAITING
+		3|GATE_WAITING
 	"""
 	})
-	default void testGatedStep2NotReady_notAdvance(String theChunkState) throws InterruptedException {
+	default void testGatedStep2NotReady_stepNotAdvance(String theChunkState) throws InterruptedException {
```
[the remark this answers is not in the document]
TynerGjs: The job starts in step 2. Renamed to suggested

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/IJobMaintenanceActions.java
```diff
@@ -117,48 +121,60 @@ default void testGatedStep2NotReady_notAdvance(String theChunkState) throws Inte
...
 		getTestManager().createChunksInStates(result);
 
 		// test
 		getTestManager().runMaintenancePass();
 
 		// verify
-		// things are being set to READY; is anything being queued?
-		getTestManager().verifyWorkChunkMessageHandlerCalled(sendingLatch, 0);
+		getTestManager().verifyWorkChunkMessageHandlerCalled(sendingLatch, 2);
 		verifyWorkChunkFinalStates(result);
```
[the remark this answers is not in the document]
TynerGjs: added check

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/WorkChunkStatusEnum.java
```diff
@@ -32,6 +32,7 @@
  */
 public enum WorkChunkStatusEnum {
 	// wipmb For 6.8 Add WAITING for gated, and READY for in db, but not yet sent to channel.
```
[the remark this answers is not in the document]
TynerGjs: removed

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImplTest.java
```diff
@@ -367,50 +374,149 @@
...
+		Date updateTime2 = runInTransaction(() -> new Date(findInstanceByIdOrThrow(instanceId).getUpdateTime().getTime()));
 		assertNotEquals(updateTime, updateTime2);
 	}
 
 	@Test
-	public void testFetchUnknownWork() {
-		assertFalse(myWorkChunkRepository.findById("FOO").isPresent());
+	public void advanceJobStepAndUpdateChunkStatus_forGatedJob_updatesCurrentStepAndChunkStatus() {
+		// setup
+		JobInstance instance = createInstance(true, true);
```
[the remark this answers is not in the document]
TynerGjs: I am a bit reluctant to change these methods to createGatedDefinitionAndInstance or createGatedChunks because a few parametrized tests uses the boolean... Introduced local variable `isGatedExecution` to make it clearer.

PR #5821, opened 2024-04-04
Base test classes enhancements for gateway test configuration flexibility

jmarchionatto:


hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/BaseRestServerHelper.java
```diff
@@ -75,9 +76,11 @@ public IGenericClient getClient() {
 	}
 
 	protected void startServer(Servlet theServlet) throws Exception {
-		myListenerServer = new Server(0);
+
+		int port = myFirstTargetPort == -1 ? 0 : myFirstTargetPort++;
```
TipzCM, line 80: Is this method expected to be called over and over again?

If so, it seems like if you never call `setMyFirstTargetPort`, you'll keep making myListernerServer with port 0.


I would expect this method would either block subsequent calls in some way (if mmyListenerServer is already assigned) or update myFirstTargetPort once 0 has been used.

TipzCM (approved):

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/BaseRestServerHelper.java
jmarchionatto: Yes. It will be called once for each configured target server, and it is OK to always call it with 0, as it was before, because in that case the server port is automatically set to some free port.

PR #5833, opened 2024-04-09
Test showing include on canonical version does not work

jkiddo:


jkiddo:
Given the discussion on https://chat.fhir.org/#narrow/stream/179166-implementers/topic/canonicals.20.20on.20_include I'll leave it to the core team if it should be changed or not.

jkiddo:
Though it doesn't seem to work on R5 either ... (manually tested).

PR #5857, opened 2024-04-18
Delete-expunge does not honor delete resourcesOfType permissions

volodymyr-korzh:
Previously, it was possible to execute the `$delete-expunge` operation on a resource even if the user did not 
have delete permissions for the given resource type. This has been fixed.


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
```diff
@@ -1473,12 +1456,139 @@
...
+
+	@Test
+	public void testDeleteExpunge_typeInCompartmentPermission_allowed() {
+		IIdType id = createPatient();
+		createTypeInPatientCompartmentRule(id);
+		executeDeleteConditionalByUrl("Patient?_expunge=true&_id=" + id.getIdPart());
+	}
+
+	@Test
+	public void testDeleteExpunge_typeInCompartmentPermission_forbidden() {
```
lukedegruchy, line 1564: I like the fact that you've attacked this problem from multiple angles.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
```diff
@@ -267,13 +270,18 @@ public Verdict applyRule(
...
 					if (thePointcut == Pointcut.STORAGE_PRE_DELETE_EXPUNGE && myAppliesToDeleteExpunge) {
-						return newVerdict(
-								theOperation,
-								theRequestDetails,
-								theInputResource,
-								theInputResourceId,
-								theOutputResource,
-								theRuleApplier);
+
+						String[] resourceIds = theRequestDetails.getParameters().get("_id");
```
lukedegruchy, line 274: Question:  I'm not doubting your fix but I'm curious how this change solves the problem?

lukedegruchy (approved):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
volodymyr-korzh: The fix for this issue is just  deleting `return newVerdict(...` and adding a break statement - this allows rule to be checked for type, compartment etc. permissions instead of granting access right away.

adding target.resourceIds is a complementary fix - to fullfill compartment rule check (without this code we would deny requests with _id = compartmentId eg. `Delete /Patient?_id=4752&_expunge=true` if we have permission for deletion `Patient/4752` compartment)

`ca.uhn.fhir.rest.server.interceptor.auth.RuleImplOp#applyRuleToCompartment`

```
		if (target.resourceIds != null && target.resourceIds.size() > 0) {
			boolean haveOwnersForAll = target.resourceIds.stream()
					.allMatch(n -> myClassifierCompartmentOwners.contains(n.toUnqualifiedVersionless()));
			if (haveOwnersForAll) {
				foundMatch = true;
			}
		}
```

 This part is for handling TYPE and COMPARTMENT permissions correctly.
 (assigning `target.resourceType` in case of compartment permission lead to false allow in some cases)
```

if (myClassifierType == ClassifierTypeEnum.ANY_ID) {
	target.resourceType = theRequestDetails.getResourceName();
}
```

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
volodymyr-korzh: Thanks
I tried to cover as much cases as possible

PR #5875, opened 2024-04-24
Fix error message for HAPI-0302 showing misleading names for known resource types

### LINKED ISSUE #5874: Error message for HAPI-0302 shows misleading names for known resource types
**Describe the bug**
When a request is made for an unknown resource type, the server returns the resource types that it does know how to handle. When it does so, "List" is shown as "ListResource", which can (and already has) mislead someone trying to debug the error, especially if they then try to request a `ListResource` (which doesn't exist).

**To Reproduce**
1. Make a GET request to any invalid resource, like `GET /Abcdefgh/1326`.
2. See that "ListResource" is in the error message.

**Expected behavior**
The "List" resource is correctly shown as "List" in the list of resource types.

**Screenshots**
image

image

**Environment:**
 - HAPI FHIR Version: 7.3.0-SNAPSHOT

**Additional context**
Fix is ready and incoming.

I believe that this happens specifically with "List" because `List` is already a type in the Java standard library, so the resource class for "List" is named differently.

holly-smile:
Closes #5874.

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/RestfulServer.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/RestfulServer.java
@@ -141,6 +141,7 @@ public class RestfulServer extends HttpServlet implements IRestfulServer<Servlet
 	private static final ExceptionHandlingInterceptor DEFAULT_EXCEPTION_HANDLER = new ExceptionHandlingInterceptor();
 	private static final Logger ourLog = LoggerFactory.getLogger(RestfulServer.class);
 	private static final long serialVersionUID = 1L;
+	private static final Map<String, String> NAME_MAPPINGS = Map.of("ListResource", "List");
 	private final List<Object> myPlainProviders = new ArrayList<>();
 	private final List<IResourceProvider> myResourceProviders = new ArrayList<>();
 	private IInterceptorService myInterceptorService;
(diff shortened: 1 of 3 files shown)


holly-smile:
Sorry, still needs a unit test - incoming!

jmarchionatto (changes requested):
I am not sure about this fix. 
There is a provider returning "ListResource" name, apparently to meet some DSTU2 version spec, and we let that happen to later change it, no matter the version. 
I would rather discriminate the name by version, if necessary at all, in the proviuder, and eliminate the introduced remapping.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/RestfulServer.java
```diff
@@ -1990,14 +1991,22 @@ protected void throwUnknownResourceTypeException(String theResourceName) {
...
+				.map(RestfulServer::getKnownDistinctAndSortedResourceTypes)
 				.distinct()
 				.sorted()
 				.collect(toList());
 		throw new ResourceNotFoundException(Msg.code(302) + "Unknown resource type '" + theResourceName
 				+ "' - Server knows how to handle: " + knownDistinctAndSortedResourceTypes);
 	}
 
+	private static String getKnownDistinctAndSortedResourceTypes(IResourceProvider theResourceProvider) {
+		String simpleName = theResourceProvider.getResourceType().getSimpleName();
```
jamesagnew, line 2003: We shouldn't need to maintain a hardcoded map for these. You can query the FhirContext for the resource name associated with a given class by calling something like:
```
theCtx.getResourceDefinition(resourceType).getName()
```

holly-smile: Ah, that's so much cleaner (and less hacky)! Thank you!

jmarchionatto (approved):
Much better. Thanks! Approved.

PR #5902, opened 2024-05-02
Allow candidates to also be found when running $mdm-submit.

jkiddo:


jkiddo:
Shouldnt have left draft yet. Sry 'bout that

PR #5903, opened 2024-05-02
Fix query chained on sort bug where we over-filter results

lukedegruchy:
- Ensure we don't over-filter result by adding an OR IS NULL clause on HASH_IDENTITY

Closes https://github.com/hapifhir/hapi-fhir/issues/5904

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5904-query-chained-sort-returns-less-results.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5904-query-chained-sort-returns-less-results.yaml
@@ -4,4 +4,3 @@ issue: 5904
 title: "Under certain conditions, when running a GET query with a chained sort and linked resources will return less 
         resources than expected, including a smaller total.
         This has been fixed."
-u
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5904-query-chained-sort-returns-less-results.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5904-query-chained-sort-returns-less-results.yaml
@@ -1,0 +1,7 @@
+---
+type: fix
+issue: 5904
+title: "Under certain conditions, when running a GET query with a chained sort and linked resources will return less 
+        resources than expected, including a smaller total.
+        This has been fixed."
+u
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5904-query-chained-sort-returns-less-results.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5904-query-chained-sort-returns-less-results.yaml
@@ -1,6 +1,4 @@
 ---
 type: fix
 issue: 5904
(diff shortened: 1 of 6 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QuerySandboxTest.java
```diff
@@ -137,6 +142,28 @@ void testSortByPid() {
 		myTestDaoSearch.assertSearchFindsInOrder("reverse sort by server assigned id", "Patient?family=smith&_sort=-_pid", id3,id2,id1);
 	}
 
+	@Test
```
michaelabuckley, line 145: problem: our build doesn't run any tests in this class because the classname ends with "Sandbox" instead of "Test".  Move these to normal test class.

I created the Sandbox so I could bash on things in a friendly place that logs all the queries.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -419,7 +419,7 @@ public void addSortOnResourceLink(
 		}
 
 		addSortCustomJoin(resourceLinkPredicateBuilder.getColumnTargetResourceId(), chainedPredicateBuilder, null);
-		Condition predicate = chainedPredicateBuilder.createHashIdentityPredicate(targetType, theChain);
+		Condition predicate = chainedPredicateBuilder.createHashIdentityOrNullPredicate(targetType, theChain);
```
michaelabuckley, line 422:
```suggestion
// Support chained sort - when there is no data to sort, the outer join will produce null columns.  We need to include those too.
				Condition predicate = chainedPredicateBuilder.createHashIdentityOrNullPredicate(targetType, theChain);
```

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_2_0/5904-query-chained-sort-returns-less-results.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 5904
+title: "Under certain conditions, when running a GET query with a chained sort and linked resources will return less 
+        resources than expected, including a smaller total.
+        This has been fixed."
```
michaelabuckley, line 6:
```suggestion
title: "Chained sort would exclude results that did not have resources matching the sort chain.  These are now included, and sorted at the end."
```

michaelabuckley (approved):
Tests are in the wrong spot.  Made some doc comments.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QuerySandboxTest.java
lukedegruchy: I renamed the file in this PR.

It's now FhirResourceDaoR4QuerySandboxTest.java

lukedegruchy: P.S.  I looked at the azure build output and this test was indeed run.

PR #5938, opened 2024-05-15
Include profile URL in generated IPS

jamesagnew:
When we generate an IPS bundle, the bundle should be tagged with the appropriate profile URL.


jmarchionatto (commented):
Changes look good, but wouldn't interface change merit corresponding documentation update?

jmarchionatto (approved):

PR #5940, opened 2024-05-15
Ensure Delete Expunge removes corresponding HFJ_RES_SEARCH_URL records for affected resources

lukedegruchy:
- Add delete HFJ_RES_SEARCH_URL SQLs to the ones run by DeleteExpungeSvcImpl
- Add a foreign key to HFJ_RES_SEARCH_URL linking to HFJ_RESOURCE both in the entity definition and to the migration tasks
- Enhance existing tests to assert the new behaviour

Closes https://github.com/hapifhir/hapi-fhir/issues/5942

### DIFF
--- a/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceSearchUrlEntity.java
+++ b/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceSearchUrlEntity.java
@@ -75,15 +75,18 @@ public class ResourceSearchUrlEntity {
 	@Temporal(TemporalType.TIMESTAMP)
 	private Date myCreatedTime;
 
-	public static ResourceSearchUrlEntity from(String theUrl, Long theId) {
+	public static ResourceSearchUrlEntity from(String theUrl, ResourceTable theResourceTable) {
 		return new ResourceSearchUrlEntity()
-				.setResourcePid(theId)
+				.setResourceTable(theResourceTable)
 				.setSearchUrl(theUrl)
 				.setCreatedTime(new Date());
 	}
 
 	public Long getResourcePid() {
-		return myResourcePid;
+		if (myResourcePid != null) {
+			return myResourcePid;
+		}
+		return myResourceTable.getResourceId();
 	}
 
 	public ResourceSearchUrlEntity setResourcePid(Long theResourcePid) {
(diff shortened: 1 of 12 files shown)


TipzCM (approved):

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceSearchUrlEntity.java
```diff
@@ -55,29 +59,49 @@ public class ResourceSearchUrlEntity {
...
+	@ManyToOne(fetch = FetchType.LAZY)
+	@JoinColumn(
+			name = "RES_ID",
+			nullable = false,
+			updatable = false,
+			foreignKey = @ForeignKey(name = "FK_RES_SEARCH_URL_RESOURCE"))
+	private ResourceTable myResourceTable;
+
+	@Column(name = "RES_ID", updatable = false, nullable = false, insertable = false)
 	private Long myResourcePid;
```
michaelabuckley, line 71: question: Why keep the raw id field, if we have added the entity?

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceSearchUrlEntity.java
```diff
@@ -55,29 +59,49 @@ public class ResourceSearchUrlEntity {
...
 		return new ResourceSearchUrlEntity()
-				.setResourcePid(theId)
+				.setResourceTable(theResourceTable)
 				.setSearchUrl(theUrl)
 				.setCreatedTime(new Date());
 	}
 
 	public Long getResourcePid() {
-		return myResourcePid;
+		if (myResourcePid != null) {
```
michaelabuckley, line 85: question: Do we need this complexity?  It would be simpler to only have one way to do this.

michaelabuckley (approved):
The migration + fk definition look good.
I question the need to keep both the id field and the entity field in the entity.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceSearchUrlEntity.java
lukedegruchy: You're right.  I got rid of it.

PR #5944, opened 2024-05-18
Configurable summary mode

### LINKED ISSUE #5871: code for returning 'summary' search responses is automatically stripping extensions from the response
**Describe the bug**
The code for returning 'summary' search responses is automatically stripping extensions from the response.  In our case, there are certain extensions that need to be included in the summary search response.

Here is the code in BaseParser of HAPI FHIR,

if (retVal && isSummaryMode() && (getDef() == null || !getDef().isSummary())) {
    String resourceName = myEncodeContext.getLeafResourceName();
    // Technically the spec says we shouldn't include extensions in CapabilityStatement
    // but we will do so because there are people who depend on this behaviour, at least
    // as of 2019-07. See
    // https://github.com/smart-on-fhir/Swift-FHIR/issues/26
    // for example.
    if (("Conformance".equals(resourceName) || "CapabilityStatement".equals(resourceName))
          && ("extension".equals(myDef.getElementName())
                || "extension".equals(myEncodeContext.getLeafElementName()))) {
       // skip
    } else {
       retVal = false;
    }
}

In the summary mode, the extensions are dropped.

**To Reproduce**
...for a claim resource, when we add an extension, the HAPI plain server returns the response without the extension.

**Expected behavior**
The [wording from the specification](https://build.fhir.org/search.html#_summary) is "This subset SHOULD consist solely of all supported elements that are marked as "summary" in the base definition of the resource(s)".  Note the use of SHOULD rather than SHALL.  While it's reasonable for HAPI's default behavior to be stripping all extensions, there should be a configuration option to allow selected extensions to be included.  The only alternative to this is moving to custom operations, which is much less interoperable.

**Screenshots**

 - HAPI FHIR Version 5.6.8
 - OS: Windows 

**Additional context**

jamesagnew:
Fixes #5871


hapi-fhir-base/src/main/java/ca/uhn/fhir/context/ParserOptions.java
```diff
@@ -205,4 +209,119 @@ public ParserOptions setOverrideResourceIdWithBundleEntryFullUrl(
 		myOverrideResourceIdWithBundleEntryFullUrl = theOverrideResourceIdWithBundleEntryFullUrl;
 		return this;
 	}
+
+	/**
+	 * This option specifies one or more elements that should be included when the parser is encoding
```
fil512, line 214: exemplary javadoc

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/ParserOptions.java
```diff
@@ -205,4 +209,119 @@ public ParserOptions setOverrideResourceIdWithBundleEntryFullUrl(
...
+	}
+
+	/**
+	 * @return Returns the values provided to {@link #setEncodeElementsForSummaryMode(Collection)}
+	 * or <code>null</code>
+	 *
+	 * @since 7.4.0
+	 */
+	@Nullable
+	public Set<String> getEncodeElementsForSummaryMode() {
```
fil512, line 268: name suggestion: consider
- getAugmentedSummaryModeElements()
- getExcludedSummaryModeElements()

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
```diff
@@ -1066,10 +1029,40 @@ public String getChildName() {
...
+						encodeElements.stream().map(EncodeContextPath::new).collect(Collectors.toList());
+			}
+			if (dontEncodeElements == null || dontEncodeElements.isEmpty()) {
+				myDontEncodeElementPaths = null;
+			} else {
+				myDontEncodeElementPaths =
+						dontEncodeElements.stream().map(EncodeContextPath::new).collect(Collectors.toList());
+			}
+
+			myEncodeElementsAppliesToResourceTypes =
```
fil512, line 1061: shouldn't this exclude the dontEncode ones?

fil512 (approved):
nice! this looks really useful!

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/ParserOptions.java
jamesagnew: I did consider using names like these, but ultimately decided that aligning the names of the new methods here to the existing names of the methods that do the same thing on `IParser` was probably more important as a goal.

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
jamesagnew: Nope, it's only used for one specific optimization (deciding whether to encode child contained resources) and that optimization isn't used for the exclude path.

Issue #5952, opened 2024-05-23
Retrieving the consecutive pages of a Bundle search fails with 403 even if permissions to read Bundle resource types are provided


codeforgreen:
**Describe the bug**
Retrieving the consecutive pages of a `Bundle` search fails with 403 even if permissions to read `Bundle` resource types are provided.
This issue was likely introduced in HAPI FHIR 7.0 and did not exist in the previous versions.

**To Reproduce**
1. Register an AuthenticationInterceptor and add permissions to read all resources of type Bundle
2. Create two document Bundle with a composition referring to a patient that is also included in the same Bundle
3. Perform a search and retrieve one result using `/Bundle?_count=1`
4. Use the next link in the returned result and access it

**Expected behavior**
The request should succeed with 200.

**Actual behaviour**
The request fails with 403 forbidden.

**Environment (please complete the following information):**
 - HAPI FHIR Version 7.0

PR #5953, opened 2024-05-23
Fix authorization handling for Bundle resources in the output

codeforgreen:
#5952 5952

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
@@ -74,6 +74,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 
 import static org.hamcrest.MatcherAssert.assertThat;
@@ -96,7 +97,6 @@ public class AuthorizationInterceptorJpaR4Test extends BaseResourceProviderR4Tes
 	@Autowired
 	private ThreadSafeResourceDeleterSvc myThreadSafeResourceDeleterSvc;
 	private AuthorizationInterceptor myReadAllBundleInterceptor;
-	private AuthorizationInterceptor myReadAllPatientInterceptor;
 	private AuthorizationInterceptor myWriteResourcesInTransactionAuthorizationInterceptor;
 
 	@BeforeEach
@@ -107,7 +107,6 @@ public class AuthorizationInterceptorJpaR4Test extends BaseResourceProviderR4Tes
 		myStorageSettings.setExpungeEnabled(true);
 		myStorageSettings.setDeleteExpungeEnabled(true);
 		myReadAllBundleInterceptor = new ReadAllAuthorizationInterceptor("Bundle");
-		myReadAllPatientInterceptor = new ReadAllAuthorizationInterceptor("Patient");
 		myWriteResourcesInTransactionAuthorizationInterceptor = new WriteResourcesInTransactionAuthorizationInterceptor();
 	}
 
@@ -396,7 +395,6 @@ public class AuthorizationInterceptorJpaR4Test extends BaseResourceProviderR4Tes
 			}
 		});
 
-		Bundle bundle;
 		Observation response;
 
 		// Read (no masking)
@@ -412,7 +410,7 @@ public class AuthorizationInterceptorJpaR4Test extends BaseResourceProviderR4Tes
 			.elementsSubset("status")
 			.execute();
 		assertEquals(ObservationStatus.FINAL, response.getStatus());
-		assertEquals(null, response.getSubject().getReference());
+		assertNull(response.getSubject().getReference());
 
 		// Read a non-allowed observation
 		try {
@@ -481,7 +479,7 @@ public class AuthorizationInterceptorJpaR4Test extends BaseResourceProviderR4Tes
 		patient.setId("Patient/A");
 		patient.addIdentifier().setSystem("http://uhn.ca/mrns").setValue("100");
 		patient.addName().setFamily("Tester").addGiven("Raghad");
-		IIdType id = myClient.update().resource(patient).execute().getId();
+		myClient.update().resource(patient).execute();
 
 		Observation obs = new Observation();
 		obs.setId("Observation/B");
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
```diff
@@ -945,7 +941,7 @@ public List<IAuthRule> buildRuleList(RequestDetails theRequestDetails) {
 
 		httpGet = new HttpGet(myServerBase + "/Patient/B/$graphql?query=" + UrlUtil.escapeUrlParam(query));
 		try (CloseableHttpResponse response = ourHttpClient.execute(httpGet)) {
-			String resp = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
+			IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
```
TipzCM, line 944: Do we need the response at all?

Maybe just a check that the response is something we expect to see instead of just completely removing?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
```diff
@@ -1642,7 +1639,24 @@ public List<IAuthRule> buildRuleList(RequestDetails theRequestDetails) {
...
+		createDocumentBundle(createPatient("Jane", "Doe"));
+		Bundle searchResult = myClient
+				.search()
+				.byUrl("/Bundle")
+				.count(1)
+				.returnBundle(Bundle.class)
+				.execute();
+		myClient.loadPage()
+				.next(searchResult)
+				.execute();
```
TipzCM, line 1655: i know you're just checking to make sure this doesn't return a 403, but it's not entirely clear that's what's going on.

You also go through the effort of creating some bundles.

You might as well at least do some (minor) validation of what is returned itself.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/AuthorizationInterceptor.java
```diff
@@ -612,23 +605,22 @@ public static List<IBaseResource> toListOfResourcesAndExcludeContainer(
...
+	public static boolean shouldExamineChildResources(FhirContext theFhirContext, IBaseResource theResource) {
+		if (theResource instanceof IBaseParameters) {
 			return true;
 		}
-		BundleTypeEnum bundleType = BundleUtil.getBundleTypeEnum(theFhirContext, theBundle);
-		boolean isStandaloneBundleResource =
-				bundleType != null && STANDALONE_BUNDLE_RESOURCE_TYPES.contains(bundleType);
-		return !isStandaloneBundleResource;
+		if (theResource instanceof IBaseBundle) {
+			BundleTypeEnum bundleType = BundleUtil.getBundleTypeEnum(theFhirContext, ((IBaseBundle) theResource));
```
TipzCM, line 618: This may be a silly question, and maybe it's out of scope, but do we need to do this instanceof check and cast?

The BundleUtil methods take IBaseBundle.

But they don't *require* an IBaseBundle and can use an IBaseResource (IBaseBundle extends IBaseResource) itself.

But if this change is too much, i can accept this as is.

TipzCM (approved):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/AuthorizationInterceptor.java
codeforgreen: I had somewhat same thoughts. I am checking to make sure the call doesn't fail, but it is true I could add a comment or think of another validation. I think checking the output is out of scope (there are other tests for that), here we are ensuring the request passes permission check and returns successfully.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
codeforgreen: I had somewhat same thoughts. I am checking to make sure the call doesn't fail, but it is true I could add a comment or think of another validation. I think checking the output resources is out of scope (there are other tests for that), here we are ensuring the request passes permission check and returns successfully. There are other tests that do that, but in my opinion it's not necessarily required, we should have other tests that cover correctness of output.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
codeforgreen: I removed this line. Here we just want to check the status is 403. I was fixing the compile warnings in the file.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/AuthorizationInterceptor.java
codeforgreen: `BundleUtil` methods indeed take `IBaseBundle`. If we wanted to add a method that takes `IBaseResource`, that would be useful to avoid the cast, but in this method we already have an instanceof check for `IBaseParameters`, so I think it is fine as is.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
codeforgreen: I updated the code to at least to a null check and in case of success to check the entry has a resource. Note that in case of transaction bundle, if one entry fails to process it won't fail the request, instead it will populate a response in the entry with the failure. That was an interesting find for me, I debugged to learn it.
I added alot more tests, as per discussion during office-hours. Have another look at the test class changes please.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/AuthorizationInterceptorJpaR4Test.java
```diff
@@ -106,16 +106,14 @@ public void before() throws Exception {
...
-		myReadAllPatientInterceptor = new ReadAllAuthorizationInterceptor("Patient");
-		myWriteResourcesInTransactionAuthorizationInterceptor = new WriteResourcesInTransactionAuthorizationInterceptor();
 	}
 
 	@Override
 	@AfterEach
 	public void after() throws Exception {
 		super.after();
 		myInterceptorRegistry.unregisterInterceptorsIf(t -> t instanceof AuthorizationInterceptor);
+		myClient.getInterceptorService().unregisterInterceptorsIf(t -> t instanceof SimpleRequestHeaderInterceptor);
```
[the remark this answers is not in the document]
codeforgreen: I've restored the previous state where this interceptor was un-registrered in a finally block.

jamesagnew (approved):

PR #5957, opened 2024-05-23
rebuilding golden resources with survivorship in partitions

### LINKED ISSUE #5956: Rebuilding golden resources (mdm) with survivorship rules in partitioned environments does not work
Reproduction Steps
* create 2 partitions (one for regular resources, one for golden)
* create at least 1 resource w/ golden resource link
* rebuild the golden resource with survivorship rules

TipzCM:
closes #5956

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
@@ -165,7 +165,7 @@ public class IdHelperService implements IIdHelperService<JpaPid> {
 		if (resourceIdToUse.contains("/")) {
 			resourceIdToUse = theResourceId.substring(resourceIdToUse.indexOf("/") + 1);
 		}
-		IdDt id = new IdDt(theResourceType, theResourceId);
+		IdDt id = new IdDt(theResourceType, resourceIdToUse);
 		Map<String, List<IResourceLookup<JpaPid>>> matches =
 				translateForcedIdToPids(theRequestPartitionId, Collections.singletonList(id), theExcludeDeleted);
 
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
@@ -155,25 +155,26 @@ public class IdHelperService implements IIdHelperService<JpaPid> {
 	public IResourceLookup<JpaPid> resolveResourceIdentity(
 			@Nonnull RequestPartitionId theRequestPartitionId,
 			String theResourceType,
-			String theResourceId,
+			final String theResourceId,
 			boolean theExcludeDeleted)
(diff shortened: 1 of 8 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -155,25 +155,26 @@ public IResourceLookup<JpaPid> resolveResourceIdentity(
 	public IResourceLookup<JpaPid> resolveResourceIdentity(
 			@Nonnull RequestPartitionId theRequestPartitionId,
 			String theResourceType,
-			String theResourceId,
+			final String theResourceId,
```
lukedegruchy, line 158: Ugh!  final should be the default in Java

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
```diff
@@ -155,25 +155,26 @@ public IResourceLookup<JpaPid> resolveResourceIdentity(
...
 			boolean theExcludeDeleted)
 			throws ResourceNotFoundException {
 		assert myDontCheckActiveTransactionForUnitTest || TransactionSynchronizationManager.isSynchronizationActive()
 				: "no transaction active";
 
-		if (theResourceId.contains("/")) {
-			theResourceId = theResourceId.substring(theResourceId.indexOf("/") + 1);
+		String resourceIdToUse = theResourceId;
+		if (resourceIdToUse.contains("/")) {
+			resourceIdToUse = theResourceId.substring(resourceIdToUse.indexOf("/") + 1);
```
lukedegruchy, line 166: Question:  I have a vague idea why you're maintaining a variable separate from the parameter but I'm curious what use case it solves

lukedegruchy (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
TipzCM: ironically, it is.... but if you change the variable, java will then say it's not final.

Which is why i'm not a huge fan of this new "minimalist" coding structure where the language devines what is desired. :(

At any rate, this is now fixed here

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/IdHelperService.java
TipzCM: It doesn't solve any specific use case.

I just don't like functions that mutate variables.

It leads to problems later on that are hard to diagnose and can often be even harder to fix as layers upon layers of code start depending on this functionality.

An example i had fixed before was in our search code where we mutate the search parameters to do the count search first.... but this mutation is then relied upon for some parts of the actual search later on, but not other parts. Resulting in us kicking around 2 versions of the parameters and having to maintain both.

I'd rather this not get to that point.

tadgh (approved):

PR #5965, opened 2024-05-27
Resource Reference Include Match Fix

melihaydogd:
In the previous implementation, the `matchesInclude_hasTargetResourceType_matched` would fail. This PR fixes the issue in the `match` function by making it possible to take the target type into account. We are making use of this implementation to find the included resources to create a `Bundle` response.


jarimayenburg:
This would fix #5966. It looks good to me!

PR #5967, opened 2024-05-27
Search Entry Mode Added to Metadata of Bundle Entries

melihaydogd:
We process the information that is retrieved from `GenericClient` and then, create a new `Bundle` by using the `IVersionSpecificBundleFactory`. The bundle factory is making use of the resource metadata when creating the `search.mode` fields in bundle entries. Right now, we are making this by ourselves but it would be a nice feature to add here so that every resource in the bundle response has this information. Maybe, it can be controlled by a configuration field. I would like to hear your feedback for this feature.


jamesagnew:
@melihaydogd this seems quite reasonable to me, I'm in favour.

Could you please add a unit test exercizing the feature (a class like `GenericClientR4Test` might be a good place for tests for this), and I'm sure we can get this merged.

PR #5970, opened 2024-05-28
Update dependencies / solving CVEs

madduci:
Hello

with this PR i try to address some CVEs that involve especially the `jpaserver-starter`, which in turn depends on the `hapi-fhir` bom as parent. 

I've mostly updated to the very latest patch version available (e.g. SpringBoot), in some cases i've bumped the minor version (e.g. Jackson, OTel). 

OpenTelemetry comes with a small breaking change, the `opentelemetry-instrumentation-annotations` dependency has a different version number, documentation here: https://opentelemetry.io/docs/zero-code/java/agent/annotations/

### DIFF
--- a/hapi-fhir-structures-dstu3/src/test/java/ca/uhn/fhir/parser/JsonParserDstu3Test.java
+++ b/hapi-fhir-structures-dstu3/src/test/java/ca/uhn/fhir/parser/JsonParserDstu3Test.java
@@ -158,7 +158,7 @@ public class JsonParserDstu3Test {
 			fail();
 		} catch (DataFormatException e) {
 			assertEquals(Msg.code(1861) + "Failed to parse JSON encoded FHIR content: Unexpected character ('=' (code 61)): was expecting a colon to separate field name and value\n" +
-				" at [line: 4, column: 18]", e.getMessage());
+				" at [line: 4, column: 17]", e.getMessage());
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


madduci:
@jkiddo could you review it? I've added some updates so the JPA Server can itself profit from these changes, since some CVEs are becoming relevant, especially the `Spring Security` one

jkiddo:
@madduci from `starter` perspective it seems fine. This is however the core repo so the review needs to be carried out by Smile people.

madduci:
perfect, many thanks, i'll wait for a review here!

jamesagnew (approved):
Thanks for the contribution!

madduci:
you are welcome!

jamesagnew:
That spotless issue is a false negative, there is a permission issue stopping it from working on external contributions. Will merge now, thanks again!

jasorello:
Hi! I see this ended up bumping the spring version to 6.1.8. It looks like there's a nested version variable in hapi-fhir-storage-cr at **<spring-security-core.version>**. Based on the other upgrades, should this version specification been bumped at the same time?

madduci:
Hi

Yes, you are right. I've totally missed it, but I've bumped it in the
hapi-fhir-jpaserver-starter, where it's being actively used

jasorello:
Hm, can you clarify how you were able to bump it in the starter pom.xml without updating the nested dependencies? I guess this is a general maven question but I wasn't aware of the capability.

madduci:
you can define a `<dependencyManagement>` block in the pom.xml and overwrite the dependency you need

PR #5971, opened 2024-05-28
5738: Skip alternate handling if already handled by previous step

stefan-lindstrom:
Hi, 

first time through the code trying to fix an issue, so I hope I haven't violated too many architectural rules. :) 

We saw this problem when using extensions on datatypes (using data-absent-reason, with code masked and we had an object with both a primitive and an array-property with masked values. We got this output (example from practitioner):

```
{
    "resourceType": "Practitioner",
    "id": "1",
    "name": [{
           "_family": {
                    "extension": [{
                            "url": "http://hl7.org/fhir/StructureDefinition/data-absent-reason",
                            "valueString": "masked"
                        }
                    ]
                }
           } ,
            "given": [null],
            "_given": [{
                    "extension": [{
                            "url": "http://hl7.org/fhir/StructureDefinition/data-absent-reason",
                            "valueString": "masked"
                        }
                    ]
                }
            ],
	    }
	]
}
```

My understanding is that when the object is normally handled, the parser counts alternateNamesSeen, and alternateNames Handled, and when seen > handled, a second pass is done. But instead of only doing un-handled, the second pass will go over all alternates including non-primitive ones where the  alternate json.-type need not be an object but could be an array for (any non-primitive property). 

Som y idea is to check, and only do the secondary handling if the alternate property in an object does not have a corresponding "real" property.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
@@ -1359,9 +1359,17 @@ public class JsonParser extends BaseParser implements IJsonLikeParser {
 				String alternateName = keyIter.next();
 				if (alternateName.startsWith("_") && alternateName.length() > 1) {
 					BaseJsonLikeValue nextValue = theObject.get(alternateName);
+					String nextName = alternateName.substring(1);
+
 					if (nextValue != null) {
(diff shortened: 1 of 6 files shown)


stefan-lindstrom:
Tagging @jamesagnew

jamesagnew:
Yup, I've merged. Don't worry about the 2 failed CI checks.. Those fail on external contributions because of some permission issue we haven't figured out yet. Thanks again!

PR #5978, opened 2024-05-29
Fix search document Bundle with local reference through composition

### LINKED ISSUE #5988: SearchParameter for Bundle document referencing resource through composition does not work
**Describe the bug**
If you create a document Bundle and a SearchParameter referencing a resource through a composition reference, the search using that parameter does not return any results. In particular, if the resource entry id is not assigned and or it is assigned and does not match the reference, the reference itself is not resolved by the SearchParameter, even if the fullUrl of the entry matches the reference. 

**To Reproduce**
1. Create a document bundle such as: https://www.hl7.org/fhir/document-example-dischargesummary.json.html
2. Create a SearchParameter with expression such as `Bundle.entry[0].resource.as(Composition).subject.resolve().as(Patient).identifier`, code 'composition.patient.identifier', of type `TOKEN` for resource type `Bundle`.
3. Perform a search such as: GET /Bundle?composition.patient.active=true
4. The search will not return any result, even if there is a patient entry matching the reference "http://fhir.healthintersections.com.au/open/Patient/d1".

**Expected behaviour**
The search should return the bundle in question as a result.

 - HAPI FHIR Version 7.2.0

**Additional context**
The reproduction steps assume there are no other document Bundle resources stored in the FHIR server.
In case if there is an id assigned to the resource, which matches the reference, this bug is not reproducible and works as expected.

codeforgreen:
Fixes #5988 

1. Previously, when using placeholder references within a document bundle, the matching was done using the request url, which is not populated for standalone document bundles. That field would be populated for transaction bundles. The fix was simply to use the fullUrl of the bundle entry to match internal resources when resolving FHIRPath expression within a Bundle. See issue details for an example.
2. I added some new test cases and merged existing test cases in a new test class `FhirResourceDaoR4SearchBundleNoFTTest`. 
3. While working on the fix, I found that using a parameter name that contains dot character would result in an `ArrayIndexOutOfBounds` when the resource is created/updated, which would happen in the `ResourceLinkPredicateBuilder` method `createResourceLinkPaths`.  I noticed that the spec does not mention any restriction with respect to parameter name (code), therefore I applied two fixes here. One was try to first lookup a parameter using the full name including dot, and in case if not found, then try to build the chain. Second of all, the qualifier list provided can be empty and in such cases it should not produce an ArrayIndexOutOfBounds.
4. I added some more tests to `ResourceLinkPredicateBuilderTest` to ensure mentioned issues are fixed.


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
```diff
@@ -681,40 +674,48 @@ public List<String> createResourceLinkPaths(
...
-					.filter(t -> t.startsWith(theResourceName + "."))
-					.map(head ->
-							tailPaths.stream().map(tail -> head + "." + tail).collect(Collectors.toSet()))
-					.flatMap(Collection::stream)
-					.collect(Collectors.toList());
 		}
+
+		// This can happen during recursion, if not all the possible target types of one link in the chain
+		// support the next link
+		return new ArrayList<>();
```
tadgh, line 718: Collections.emptyList

tadgh (approved):

PR #5983, opened 2024-05-31
Handle conditional creates across partitions with duplicate search URLs by updating ResourceSearchUrlEntity

lukedegruchy:
- Add new configuration to allow resources with duplicate identifiers across partitions
- Replace the existing single field primary key on ResourceSearchUrlEntity/HFJ_RES_SEARCH_URL with on that includes search_url and partition ID (in that order)
- Add a new migration task to add new primary keys
- Add a new migration task to drop primary keys, which requires database-specific SQL to introspect schemas to find the primary key name so the ALTER TABLE knows which constraint to drop
- Add new elaborate migration to add partition_id as a nullable column, defaulting to -1, UPDATE all existing rows to -1, make the column non-null, drop the existing primary key and add the new primary key
- New migration verification in HapiSchemaMigrationTest
- New test for conditional creates with partitions in FhirResourceDaoR4ConcurrentCreateTest

Closes https://github.com/hapifhir/hapi-fhir/issues/6033


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
```diff
@@ -347,6 +349,38 @@ protected void init740() {
...
+						.type(ColumnTypeEnum.DATE_ONLY);
+
+				version.executeRawSql(
+						"20240618.4",
+						String.format(
+								"UPDATE %s SET %s = -1",
+								tableHfjResSearchUrl, ResourceSearchUrlEntityPK.PARTITION_ID_COLUMN_NAME));
+
+				version.onTable(tableHfjResSearchUrl)
+						.modifyColumn("20240618.5", ResourceSearchUrlEntityPK.PARTITION_ID_COLUMN_NAME)
```
michaelabuckley, line 372: problem: this breaks zero-downtime.  Old servers will still be inserting rows, and will be inserting null.  You can't make this non-nullable until two more releases pass.

michaelabuckley (approved):
There's a problem making that column non-nullable.  Older servers will still be inserting rows.  We can't assume they're gone until 2025.02.

In hindsight, using a traditional sequence id would be have been easier, and then adding a unique constraint for the control.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
michaelabuckley: I'm wrong - the column has a default value, so old servers will run fine.

PR #5987, opened 2024-06-05
fixed mdmjsonlink serialization issues

### LINKED ISSUE #5986: IResourcePersistenceId is not a serializable type
this interface is being used in MdmLinkJson and prevents deserialization.

TipzCM:
closes #5986


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/SerializablePid.java
```diff
@@ -0,0 +1,74 @@
...
+import ca.uhn.fhir.model.api.IModelJson;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.hl7.fhir.instance.model.api.IIdType;
+
+/**
+ * A deserialized version of the PID
+ * Must be converted to appropriate PID type to be used.
+ */
+public class SerializablePid implements IResourcePersistentId<String>, IModelJson {
```
jamesagnew, line 13: Two comments: 

- [ ] Can we move this into being an inner class on MdmLinkJson? That's there it matters.
- [ ] Does this actually need to implement `IResourcePersistentId`? It's kind of confusing, given that this is a different kind of thing from the other two

TipzCM: To the latter point, i suppose i could make it a completely different object that has nothing to do with IResourcePersistentPid...

But then i'd need to break smile and make a separate PR, as i'd (at the very least) need methods on how to convert to an IResourcePersistentId (the thing we want).

This class was only created so i could have a generic serializable thing without doing that (and without knowing what the IResourcePersistentId even is, since we use it in a generic way, and only really care about it during serialization

jamesagnew (approved):

PR #6001, opened 2024-06-12
mdm multidelete golden resource final resource pair throws

### LINKED ISSUE #6000: MDM: Deleting a resource and its final golden resource at the same time results in errors when multi-delete enabled
* enable multidelete
* enable mdm
* create a single golden resource
* delete both resources at the same time (say, using a Patient?_lastUpdated=ge(dateinpast) )
*  errors thrown, resources not deleted

TipzCM:
closes #6000

### DIFF
--- a/hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/interceptor/MdmStorageInterceptorIT.java
+++ b/hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/interceptor/MdmStorageInterceptorIT.java
@@ -2,6 +2,7 @@ package ca.uhn.fhir.jpa.mdm.interceptor;
 
 import ca.uhn.fhir.interceptor.model.RequestPartitionId;
 import ca.uhn.fhir.jpa.api.model.DaoMethodOutcome;
+import ca.uhn.fhir.jpa.api.model.DeleteMethodOutcome;
 import ca.uhn.fhir.jpa.api.svc.IIdHelperService;
 import ca.uhn.fhir.jpa.entity.MdmLink;
 import ca.uhn.fhir.jpa.mdm.BaseMdmR4Test;
@@ -50,6 +51,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 import static org.slf4j.LoggerFactory.getLogger;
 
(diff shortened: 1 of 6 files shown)


hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/interceptor/MdmStorageInterceptor.java
```diff
@@ -188,17 +196,32 @@ public void blockManualGoldenResourceManipulationOnUpdate(
...
+	public void deletePostCommit(
+			RequestDetails theRequest, IBaseResource theResource, TransactionDetails theTransactionDetails) {
+		Set<IResourcePersistentId> goldenResourceIds = theTransactionDetails.getUserData(GOLDEN_RESOURCES_TO_DELETE);
+
+		if (goldenResourceIds != null) {
+			for (IResourcePersistentId goldenPid : goldenResourceIds) {
+				if (!theTransactionDetails.getDeletedResourceIds().contains(goldenPid)) {
+					IFhirResourceDao<?> dao = myDaoRegistry.getResourceDao(theResource);
+					deleteGoldenResource(goldenPid, dao, theRequest);
+					theTransactionDetails.addDeletedResourceId(goldenPid);
```
codeforgreen, line 209: What does this line achieve? It looks like it modifies the `TransactionDetails` deletedResourceId list. BaseHapiFhirResourceDao should be the sole modifier and this is just a consumer. If you need another way to communicate between the two pointcuts for this interceptor, probably another userData entry for the purpose would be good.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/interceptor/MdmStorageInterceptorIT.java
```diff
@@ -101,6 +103,31 @@ public void testDeleteGoldenResourceDeletesLinks() throws InterruptedException {
 		assertLinkCount(0);
 	}
 
+	@Test
+	public void deleteResourcesByUrl_withMultipleDeleteCatchingSourceAndGoldenResource_deletesWithoutThrowing() throws InterruptedException {
```
codeforgreen, line 107: This test seems to exercise just deleting one resource. Can you add a more complex test with more resources and links? There is some new code in the interceptor which needs to be exercised as it seems to be non-trivial. If there is other tests not in this PR that cover it, let me know.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/interceptor/MdmStorageInterceptor.java
TipzCM: We are using this because precommit gets called for every single deleted resource.

Multiples could be pointing to the same golden resource and deleting it for some reason (not in this specific case, but i'll get there).

You're right, we could use another UserData entry to store this. But that has another problem.

It's also possible someone *else* has registered the same kind of interceptor and is listening on the same event after us. They want to delete this resource too in whatever scenario we've arrived at.

If we're using a custom solution (adding it to the UserDetails) they have to parse this - and that's not fair or even desireable since we may want to change this for whatever reason.

But "deletedResourceIds" is very unambiguous in what it is - the list of resourceids that have already been deleted in this transaction.

So regardless of who's listening, regardless of how many times the pointcut is called, it will always be a 'safe' thing to check if you want to delete a resource id after a transaction.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/interceptor/MdmStorageInterceptorIT.java
TipzCM: added more to the test

codeforgreen (approved):

codeforgreen (approved):

tadgh:
Approved, pending you have tested this does not break the mongo MDM implementation. ,

tadgh (approved):

PR #6004, opened 2024-06-13
Added a couple more methods to the logback assert test helper

fil512:


hapi-fhir-test-utilities/src/main/java/ca/uhn/test/util/StaticLogbackTestExtension.java
```diff
@@ -55,4 +67,13 @@ public void afterAll(ExtensionContext theExtensionContext) throws Exception {
...
 		return myLogbackTestExtension.filterLoggingEventsWithMessageEqualTo(theMessageText);
 	}
+
+	/**
+	 * Returns a copy to avoid concurrent modification errors.
+	 * @return A copy of the log events so far.
+	 */
+	public java.util.List<ILoggingEvent> getLogEvents() {
+		return myLogbackTestExtension.getLogEvents();
+	}
```
tadgh, line 77: Question: I do not see how this returns a copy?

tadgh (approved):

hapi-fhir-test-utilities/src/main/java/ca/uhn/test/util/StaticLogbackTestExtension.java
fil512: the getLogEvents() returned by that method looks like this:

	public java.util.List<ILoggingEvent> getLogEvents() {
		return new ArrayList<>(myListAppender.list);
	}

PR #6025, opened 2024-06-18
6023 allow cds hooks extensions to define arbitrary json instead of strings

### LINKED ISSUE #6023: Allow CDS hooks extensions to define arbitrary JSON instead of strings
**Describe the bug**
Currently, CDS hooks accepts a String for Extension and when calling cds service discovery endpoint we get a String back for extension. The same happens for request extension and response extension. 

**Expected behavior**
CDS hooks should be able to accept arbitrary JSON and return it via discovery endpoint or in the card response if set.

AD1306:


AD1306:
Closing this as my local branch diverged.

PR #6045, opened 2024-06-25
fixing issue with infinite cache refresh loop

### LINKED ISSUE #6044: Refreshing a cache when doing a search using Advanced Hibernate Search results in an infinite loop
Enable advanced hibernate cache and try and refresh the cache.

This will trigger a search for all StructureDefinitions which will check the cache, which results in a cache refresh which checks the cache which results in a cache refresh which...

TipzCM:
closes #6044


hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
```diff
@@ -432,9 +438,15 @@ public void handleInit(Collection<IIdType> theResourceIds) {
 		initializeActiveSearchParams(searchParams);
 	}
 
+	@Override
+	public boolean isInitialized() {
+		return myIsInitialized;
```
fil512, line 443: rather than using a new field, can we just return `theSearchParams != null` ?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
```diff
@@ -151,6 +151,15 @@ public boolean canUseHibernateSearch(String theResourceType, SearchParameterMap
 			return true;
 		}
 
+		// if the registry has not been initialized
+		// we cannot use HibernateSearch because it
+		// will, internally, trigger a new search
+		// when it refreshes the search parameters
+		// (which will cause an infinite loop)
+		if (!mySearchParamRegistry.isInitialized()) {
+			return false;
```
fil512, line 160: is is possible for the fhir endpoint to be up before the sp registry is initialized? Guesing "no". But if "yes" then there is a risk of a silent ommission. if isInitialized() is always true by the time the app context has been refreshed, then I have no concern.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/cache/ResourceChangeListenerRegistryImplIT.java
```diff
@@ -55,6 +64,61 @@ public void after() {
...
+
+		assertTrue((cache instanceof ResourceChangeListenerCache));
+
+		// set the HSearchIndexing
+		boolean useAdvancedHSearch = myStorageSettings.isAdvancedHSearchIndexing();
+		myStorageSettings.setAdvancedHSearchIndexing(true);
+
+		// so they will be forced to be refreshed
+		mySearchParamRegistry.setActiveSearchParams(null);
+		try (MockedStatic<CacheConstants> cacheConstants = mockStatic(CacheConstants.class)) {
```
fil512, line 99: interesting!

fil512 (approved):

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
TipzCM: updated

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/FulltextSearchSvcImpl.java
TipzCM: seems to be.

I've verified with some manual testing and an automated test i won't be checking in.

michaelabuckley: I definitely see the registry initialized during startup.  I can't find the actual even because it's in some loop with the change listener and the startup sequence.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/CacheConstants.java
```diff
@@ -0,0 +1,14 @@
+package ca.uhn.fhir.jpa;
+
+public class CacheConstants {
```
michaelabuckley, line 3: minor: This feels excessive.  Why extract a whole new class when it is used in only one place?

michaelabuckley (approved):
Looks like a clean snip of a loop.

PR #6048, opened 2024-06-26
Chained Parameter Matching Support Added

melihaydogd:
Adds support for chained search parameters in `SearchParamMatcher`. The following type of parameters will be supported with this PR:
* `Observation?context.episodeofcare.condition.code=http://snomed.info/sct|77386006`
* `Observation?context:Encounter.episodeofcare:EpisodeOfCare.condition:Condition.code=http://snomed.info/sct|77386006`

Tests updated.

### DIFF
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
@@ -629,26 +629,29 @@ public class InMemoryResourceMatcher {
 
 	private boolean matchChainedReferenceParam(
 			String theParamName, String theResourceName, IBaseResource theResource, ReferenceParam referenceParam) {
+		// Obtain the next parameter and chain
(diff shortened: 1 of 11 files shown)


melihaydogd:
I had to update tests since the chained parameter matching is supported now. Please let me know if updated tests don't make sense.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
```diff
@@ -302,11 +317,6 @@ private InMemoryMatchResult checkOneParameterForUnsupportedModifiers(
 		// Assume we're ok until we find evidence we aren't
 		InMemoryMatchResult checkUnsupportedResult = InMemoryMatchResult.successfulMatch();
 
-		if (hasChain(theParam)) {
```
fil512, line 305: Chains are unsupported by the in-memory matcher in most cases. The only case where they would be supported is when the reference is a contained reference. But there's no way of knowing from a query alone whether the chains will be contained or not. So I think that hasChain still needs to return supported=false for chained queries.

fil512 (commented):
I don't believe this change can be merged.

PR #6060, opened 2024-06-27
Use server name as narrative text in CapabilityStatement

subigre:
It may be the expected behavior as name, version, copyright, etc. are all customizable but in my case, I would like to have my server name as narrative text of the CapabilityStatement.

Changelog:
- Use `configuration.getServerName()` as value for the narrative text generated by ServerCapabilityStatementProvider class
- Unit test

If this is the expected behavior, discard the PR


jamesagnew (approved):
This makes sense! Thanks for the contribution.

subigre:
Hi @jamesagnew, same problem than last time.
It cannot be merged due to required `mvn spotless:check` that is missing due to the fork and access to my branch

jamesagnew:
Thanks for the contribution, it's merged!

Issue #6061, opened 2024-06-28
Fix dry-run for HFJ_RES_SEARCH_URL migration


lukedegruchy:
**Describe the bug**
See https://github.com/hapifhir/hapi-fhir/issues/6033

The fix for this issue did not handle dry-run logic for the SQL that retrieves the primary key constrain name for Postgres, MSSQL, and Oracle.

PR #6062, opened 2024-06-28
Ensure DropPrimaryKeyTask respects dry-run

lukedegruchy:
Closes https://github.com/hapifhir/hapi-fhir/issues/6061

### DIFF
--- a/hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/DropPrimaryKeyTask.java
+++ b/hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/DropPrimaryKeyTask.java
@@ -21,15 +21,11 @@ package ca.uhn.fhir.jpa.migrate.taskdef;
 
 import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.jpa.migrate.DriverTypeEnum;
-import ca.uhn.fhir.jpa.migrate.HapiMigrationException;
-import ca.uhn.fhir.jpa.migrate.tasks.api.TaskFlagEnum;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 import org.intellij.lang.annotations.Language;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.dao.DataAccessException;
-import org.springframework.jdbc.core.JdbcTemplate;
 
 import java.sql.SQLException;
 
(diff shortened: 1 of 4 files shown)


TipzCM (approved):

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/DropPrimaryKeyTask.java
```diff
@@ -51,35 +54,51 @@ public DropPrimaryKeyTask(String theProductVersion, String theSchemaVersion, Str
...
-		};
-
 		@Nullable
 		@Language("SQL")
 		final String primaryKeyNameSql = generatePrimaryKeyNameSql();
 
 		@Nullable
 		final String primaryKeyName = primaryKeyNameSql != null
-				? executeSqlWithResult(primaryKeyNameSql, resultSetExtractor, getTableNameWithDatabaseExpectedCase())
+				? executeSqlWithResult(primaryKeyNameSql, getTableNameWithDatabaseExpectedCase())
```
michaelabuckley, line 63: problem: This is confused.  We DO want to run this query during a dry run - i.e. fetching the table name.  It is only the actual DROP command we don't want to run.  That's why we shouldn't be using methods named `execute*` for querying the metatdata.

Please just replace this with something simple like
```suggestion
				? newJdbcTemplate().queryForObject("some sql", String.clasS); 
```

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/taskdef/DropPrimaryKeyTask.java
```diff
@@ -51,35 +54,51 @@ public DropPrimaryKeyTask(String theProductVersion, String theSchemaVersion, Str
...
 				: null;
 
 		ourLog.debug("primaryKeyName: {} for driver: {}", primaryKeyName, getDriverType());
 
 		return generateDropPrimaryKeySql(primaryKeyName);
 	}
 
+	private String executeSqlWithResult(@Language("SQL") String theSql, Object... theArguments) {
```
michaelabuckley, line 71: problem: delete this.  Dry-run is about the modifications, not building the sql.

michaelabuckley (changes requested):
This is too complicated.  There are two parts to a migration like this:

1. query the metadata to build the migration sql
2. execute the migration sql

We only want dry-run to apply to 2, not to 1.

michaelabuckley (approved):
Much better.

Issue #6071, opened 2024-07-03
Validation CLI functionality improvements


stueynz:
Validation CLI command includes `--igpack` command line option that doesn't do anything.  It would be nice to be able to load an IG package when doing slightly more non-trivial validation experiments.

Current implementations of `isEnabledValidationForCodingsLogicalAnd()` - allowing a single coding entry to satisfy valueSet validations - assume that your ValidationChain is wrapped up in `CachingValidationSupport`.  Which isn't true when using the Validation CLI to do non-trivial validation experiments.

The `isEnabledValidationForCodingsLogicalAnd()` functionality needs to be exposed to Validation CLI.

PR #6072, opened 2024-07-03
Validation cli fixups

stueynz:
In response to [issues/6071](https://github.com/hapifhir/hapi-fhir/issues/6071) added code to extend functionality of Validation CLI; which will enable non-trivial validation experiments by exposing more features of the underlying validation support.

`--igpack` pre-existing command line option that did nothing, now loads an IG `package.tgz` into NpmPackageValidationSupport in the validation chain.

`--oneCode` new command line option allows access to `IValidationSupport.isEnabledValidationForCodingsLogicalAnd()` functionality


hapi-fhir-serviceloaders/hapi-fhir-caching-guava/src/test/java/ca/uhn/fhir/sl/cache/guava/CacheLoaderTest.java
stueynz, line 1: File location and declared package simply didn't match; so assumed the package declaration was correct.

pom.xml
stueynz, line 1: Again, really weird that `mvn install` just swallowed this, but RedHat java support couldn't build test code with v17 compiler and production code with v11 compiler.

Should probably drop this diff from the pull request.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -445,4 +447,18 @@ methods marked as deprecated in interface IValidationSupport.
...
 	}
+	/**
+	 * See VersionSpecificWorkerContextWrapper#validateCode in hapi-fhir-validation.
+	 * <p>
+	 * If true, validation for codings will return a positive result if AT LEAST ONE coding is valid.
+	 * If false, validation for codings will return a positive result if there is ALL codings are valid.
+	 * </p>
+	 * @return if the application has configured validation to use logical AND, as opposed to logical OR, which is the default
+	 */
+	public boolean oneCodingIsSufficient = false; // default value from parent Interface (it's the wrong way round !!!)
```
jamesagnew, line 458: There is already a setter for this property in CachingValidationSupport - Would it make more sense to just call that instead of adding another place to configure this here?

pom.xml
```diff
@@ -1037,9 +1037,9 @@
 
 		<!-- Site properties -->
 		<fontawesomeVersion>5.4.1</fontawesomeVersion>
-		<maven.compiler.source>11</maven.compiler.source>
-		<maven.compiler.target>11</maven.compiler.target>
-		<maven.compiler.release>11</maven.compiler.release>
+		<maven.compiler.source>17</maven.compiler.source>
```
jamesagnew, line 1040: These shouldn't be changed, we officially still support JDK11 on this project.

PR #6078, opened 2024-07-04
Jd 20240704 fix mdm phonetic matching on humanname exception

### LINKED ISSUE #6077: MDM phonetic matching on HumanName ClassCastException
**Describe the bug**
When creating a MDM match field that matches phonetically on a resourcePath of type HumanName, a `ClassCastException` is thrown

**To Reproduce**
1. Create MDM match field based on phonetic algorithm (eg. `METAPHONE`) that matches on a HumanName resource path (eg. Patient.name)
2. Post resource that invokes this field 

**Expected behavior**
No `ClassCastException` - perhaps a warning message instead

jdar8:
- Add check before casting to IPrimitiveType for phonetic matching 
- Add tests

Closes #6077

### DIFF
--- a/hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/matcher/fieldmatchers/PhoneticEncoderMatcher.java
+++ b/hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/matcher/fieldmatchers/PhoneticEncoderMatcher.java
@@ -45,9 +45,14 @@ public class PhoneticEncoderMatcher implements IMdmFieldMatcher {
 
 	@Override
 	public boolean matches(IBase theLeftBase, IBase theRightBase, MdmMatcherJson theParams) {
-		String leftString = StringMatcherUtils.extractString((IPrimitiveType<?>) theLeftBase, theParams.getExact());
-		String rightString = StringMatcherUtils.extractString((IPrimitiveType<?>) theRightBase, theParams.getExact());
+		if (theLeftBase instanceof IPrimitiveType && theRightBase instanceof IPrimitiveType) {
+			String leftString = StringMatcherUtils.extractString((IPrimitiveType<?>) theLeftBase, theParams.getExact());
+			String rightString =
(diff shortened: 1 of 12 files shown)


hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/matcher/fieldmatchers/PhoneticEncoderMatcher.java
```diff
@@ -45,9 +45,16 @@ public boolean matches(String theLeftString, String theRightString) {
...
-		String leftString = StringMatcherUtils.extractString((IPrimitiveType<?>) theLeftBase, theParams.getExact());
-		String rightString = StringMatcherUtils.extractString((IPrimitiveType<?>) theRightBase, theParams.getExact());
+		if (theLeftBase instanceof IPrimitiveType && theRightBase instanceof IPrimitiveType) {
+			String leftString = StringMatcherUtils.extractString((IPrimitiveType<?>) theLeftBase, theParams.getExact());
+			String rightString =
+					StringMatcherUtils.extractString((IPrimitiveType<?>) theRightBase, theParams.getExact());
 
-		return matches(leftString, rightString);
+			return matches(leftString, rightString);
+		}
```
jmarchionatto, line 54: Is it possible that the class not being a PrimitiveType is the right one? In that case the warning will be confusing

jmarchionatto (approved):
Approved with a question

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/matcher/fieldmatchers/PhoneticEncoderMatcher.java
jdar8: I don't think left and right would be different. It is evaluated based on a resource path or fhir path of two resources, so the left and right should be the same type. I will update the message anyways just in case. Thanks for the review!

Issue #6079, opened 2024-07-04
Need to save TermConceptDesignation.value over 2,000 characters


lukedegruchy:
**Describe the issue**
Presently, it's not possible to save a TermConceptDesignation with a value over 2,000 characters.  We need to allow larger values in the entity and database with a migration.

**Environment (please complete the following information):**
 - HAPI FHIR Version:  Master

PR #6080, opened 2024-07-04
Introduce TRM_CONCEPT_DESIG.VAL_VC column to replace VAL

lukedegruchy:
- Add a new String column with length = Length.LONG32
- In the migration, add the column of type TEXT
- Change the getter to retrieve the new column if it's non-null, otherwise, the old one
- Always write to the new column

Closes https://github.com/hapifhir/hapi-fhir/issues/6079

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
@@ -443,12 +443,12 @@ public class HapiFhirJpaMigrationTasks extends BaseMigrationTasks<VersionEnum> {
 
 		{
 			version.onTable("TRM_CONCEPT_DESIG")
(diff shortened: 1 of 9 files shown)


tadgh (approved):

tadgh (changes requested):

tadgh (approved):

tadgh (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
```diff
@@ -440,6 +440,18 @@ protected boolean shouldSkipTask() {
 													t.getString("IDX_STRING")))
 									.setColumnName("HASH_COMPLETE"));
 		}
+
+		{
+			version.onTable("TRM_CONCEPT_DESIG")
```
tadgh, line 445: recommend using `.10` and `.20` just to mimic others who have used the version skip style

tadgh (approved):

epeartree (approved):

PR #6092, opened 2024-07-10
6090 requestpartitionhelpersvc throws an unexpected exception when resolving partition for non partitionable resource

epeartree:
**What was done:**

- added failing test;
- reverting partial refactoring introduced by #6009;
- added changelog;


jmarchionatto (approved):
Looks good. Approved.

PR #6112, opened 2024-07-15
use SearchParamater validator in package installer

### LINKED ISSUE #6111: Package Installer does not create a Composite Search Parameters with No Expression At Root Level
There is a difference between the validation rules applied to the SearchParameters on Package Installer and DAO layer. At Dao layer a SearchParameter resource can be created without an expression element at root level, whereas Package Installer doesn't allow such search parameters. 

According to the [FHIR spec](https://build.fhir.org/searchparameter-definitions.html#SearchParameter.expression), the expression at the root level is optional for SearchParameter resource, so the package installer should also accept composite  SearchParameters that has no `expression` element at the root level.

mrdnctrk:
The PR fixes the issue that Package installer doesn't allow creating a composite SearchParameter (SP) resource if there is no `expression` element at the root level of the SearchParameter resource, whereas creating such a SP by other means (such as through FHIR endpoint) is allowed according to the spec. 

The cause was that there were some extra validation logic in the Package Installer that Dao level SearchParameter validation didn't have. This validation for SearchParameters in Package installer has now been refactored to use the same code as the SearchParameter validation used by the DAO layer, so that there is no difference between what can be created via PackageInstaller and other means. 

closes #6111

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
@@ -34,6 +34,7 @@ import ca.uhn.fhir.jpa.api.dao.IFhirResourceDao;
 import ca.uhn.fhir.jpa.api.model.DaoMethodOutcome;
 import ca.uhn.fhir.jpa.dao.data.INpmPackageVersionDao;
 import ca.uhn.fhir.jpa.dao.tx.IHapiTransactionService;
+import ca.uhn.fhir.jpa.dao.validation.SearchParameterDaoValidator;
 import ca.uhn.fhir.jpa.model.config.PartitionSettings;
 import ca.uhn.fhir.jpa.model.entity.NpmPackageVersionEntity;
 import ca.uhn.fhir.jpa.packages.loader.PackageResourceParsingSvc;
@@ -47,8 +48,10 @@ import ca.uhn.fhir.rest.param.StringParam;
 import ca.uhn.fhir.rest.param.TokenParam;
 import ca.uhn.fhir.rest.param.UriParam;
 import ca.uhn.fhir.rest.server.exceptions.ResourceVersionConflictException;
+import ca.uhn.fhir.rest.server.exceptions.UnprocessableEntityException;
 import ca.uhn.fhir.util.FhirTerser;
 import ca.uhn.fhir.util.SearchParameterUtil;
+import ca.uhn.hapi.converters.canonical.VersionCanonicalizer;
 import com.google.common.annotations.VisibleForTesting;
 import jakarta.annotation.PostConstruct;
 import org.apache.commons.lang3.Validate;
@@ -73,7 +76,6 @@ import java.util.Optional;
 
 import static ca.uhn.fhir.jpa.packages.util.PackageUtils.DEFAULT_INSTALL_TYPES;
 import static ca.uhn.fhir.util.SearchParameterUtil.getBaseAsStrings;
-import static org.apache.commons.lang3.StringUtils.isBlank;
 
 /**
  * @since 5.1.0
@@ -116,6 +118,12 @@ public class PackageInstallerSvcImpl implements IPackageInstallerSvc {
 
 	@Autowired
 	private JpaStorageSettings myStorageSettings;
+
+	@Autowired
+	private SearchParameterDaoValidator mySearchParameterDaoValidator;
+
+	@Autowired
+	private VersionCanonicalizer myVersionCanonicalizer;
 
 	/**
 	 * Constructor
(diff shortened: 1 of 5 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -431,6 +439,23 @@ protected boolean createOrUpdateResource(
 		return outcome != null && !outcome.isNop();
 	}
 
+	/*
```
mrdnctrk, line 442: This comment I added to this method is unrelated to issue. I added it because I had some difficulty understanding why this method was doing what it is doing and thought I comment here would help.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -569,9 +585,13 @@ private boolean isValidResourceStatusForPackageUpload(IBaseResource theResource)
 		List<IPrimitiveType> statusTypes =
 				myFhirContext.newFhirPath().evaluate(theResource, "status", IPrimitiveType.class);
 		// Resource does not have a status field
-		if (statusTypes.isEmpty()) return true;
+		if (statusTypes.isEmpty()) {
```
mrdnctrk, line 588: No real code change here, I moved the bodies of the if statements to separate lines to clearly see the code coverage

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/validation/SearchParameterDaoValidator.java
```diff
@@ -98,47 +98,41 @@ public void validate(SearchParameter searchParameter) {
...
-		// Search parameters must have a base
-		if (isCompositeWithoutBase(searchParameter)) {
+		// Search parameters must have a base except for Composite ones (for which consider it optional)
+		if (isWithoutBase(searchParameter) && !isCompositeSp(searchParameter)) {
 			throw new UnprocessableEntityException(Msg.code(1113) + "SearchParameter.base is missing");
 		}
 
-		// Do we have a valid expression
-		if (isCompositeWithoutExpression(searchParameter)) {
-
```
mrdnctrk, line 108: `isCompositeWithoutExpression` was actually a wrong name for this function because it actually should have been named `isNonCompositeWithoutExpression` according to its implementation. I divided the function into 2, for better readability and added comment.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/validation/SearchParameterDaoValidator.java
```diff
@@ -98,47 +98,41 @@ public void validate(SearchParameter searchParameter) {
...
-			// this is ok
-
-		} else if (isBlank(searchParameter.getExpression())) {
-
+		// Search parameters must have an expression except for the Composite ones (for which consider it optional)
+		if (isBlank(searchParameter.getExpression()) && !isCompositeSp(searchParameter)) {
 			throw new UnprocessableEntityException(Msg.code(1114) + "SearchParameter.expression is missing");
+		}
 
-		} else {
```
mrdnctrk, line 115: the structure of this part of the code was wrong and confusing.  Above there was an empty if body for the condition that checked `isCompositeWithoutExpression` (which should have been named `isNonCompositeWithoutExpression` as I mentioned earlier) and this part is the else of if. This means  for a composite parameter that doesn't have an expression then all the checks within this else are not applied, the particular importance is the checks maybeValidateCompositeWithComponent, maybeValidateCompositeSpForUniqueIndexing were doing. So removed this `if, else if, else` structure of this code to make sure composite SP validations are applied even when there is no expression at the root level.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
codeforgreen: Thank you for adding this explanation, Emre.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -506,33 +531,9 @@ private RequestDetails createRequestDetails() {
...
-			}
-
-			if (getBaseAsStrings(myFhirContext, theResource).isEmpty()) {
-				ourLog.warn(
-						"Failed to validate resource of type {} with url {} - Error: Resource base is empty",
-						theResource.fhirType(),
-						SearchParameterUtil.getURL(myFhirContext, theResource));
-				return false;
-			}
+		if ("SearchParameter".equals(resourceType) && !isValidSearchParameter(theResource)) {
```
codeforgreen, line 534: Very nice. This part of the code did bother me. I am glad you were able to reuse the DAO layer validator here.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/validation/SearchParameterDaoValidator.java
```diff
@@ -98,47 +98,41 @@ public void validate(SearchParameter searchParameter) {
...
-				maybeValidateCompositeWithComponent(searchParameter);
-			}
+		maybeValidateCompositeSpForUniqueIndexing(searchParameter);
+		if (!isBlank(searchParameter.getExpression())) {
+			// above we validated the expression is not blank for a non-composite SP.
+			// So, this validation happens for all non-composite SPs and
+			// composite SPs that have an expression at root level
+			maybeValidateSearchParameterExpressionsOnSave(searchParameter);
 		}
+		maybeValidateCompositeWithComponent(searchParameter);
```
codeforgreen, line 124: I noticed this difference in behaviour - the component types are validated no matter if the composite SP has or does not have an expression, while before they were validated only when expression is absent. 
Please confirm if this change in behaviour is intentional. If so, the tests need to be updated according. There are some failing in the pipeline because of this.

codeforgreen (changes requested):
The changes are carefully thought and thorough. I did notice one behaviour change in the DAO validator, please check and confirm if intentional.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/validation/SearchParameterDaoValidator.java
mrdnctrk: Thanks for bringing this up. I had made that change intentionally, but I hadn't realized the test failures. For now, I reverted the refactoring I did on the SearchParameterDaoValidator to clarify it further, and apply them back later. That refactoring isn't really related to the issue I am fixing, which can be handled later.

codeforgreen (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/validation/SearchParameterDaoValidator.java
codeforgreen: It was a really good idea. You can bring it back with with a slight change later on.

Issue #6114, opened 2024-07-16
Searching or conditional creating/updating with a URL containing a timestamp with a '+' offset results in HAPI-1883


lukedegruchy:
**Describe the bug**
hapi-fhir can't handle a URL with a timezone with an offset containing a '+'.  For example:  Observation?date=2024-07-08T20:47:12.123+03:30

**To Reproduce**
1. POST the following Bundle:

```json
{
    "resourceType": "Bundle",
    "type": "transaction",
    "entry": [
        {
            "fullUrl": "urn:uuid:3c33f03c-5a0f-4b2b-af31-bbddaa0ad86d",
            "resource": {
                "resourceType": "Observation",
                "status": "final",
                "code": {
                    "coding": [
                        {
                            "system": "CPT4",
                            "code": "1010.1",
                            "display": "WEIGHT"
                        }
                    ]
                },
                "effectiveDateTime": "2024-07-08T20:47:12.123+03:30",
                "valueString": "19.958064"
            },
            "request": {
                "method": "PUT",
                "url": "Observation?date=2024-07-08T20:47:12.123+03:30"
            }
        }
    ]
}
```
2. See an error message in the response

**Expected behavior**
The Bundle should be POSTed successfully.  The Observation should be in cdr.  It should be possible to search query that Observation with its date (ex: Observation?date=2024-07-08T20:47:12.123+03:30)

**Environment (please complete the following information):**
 - HAPI FHIR Version: Master

**Additional context**
'+' is considered a special character for ' ' in URLs.  Any solution should maintain this behaviour.

PR #6115, opened 2024-07-16
Ensure ' ' is treated as '+' in timezones with offsets.

lukedegruchy:
- Support both '+' and ' ' in timezone offsets.

Closes https://github.com/hapifhir/hapi-fhir/issues/6114

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/BaseDateTimeDt.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/BaseDateTimeDt.java
@@ -230,17 +230,41 @@ public abstract class BaseDateTimeDt extends BasePrimitive<Date> {
 		return Long.parseLong(retVal);
 	}
 
+	/**
+	 * Find the offset for a timestamp.  If it exists.  An offset may start either with '-', 'Z', '+', or ' '.
+	 * <p/>
+	 * There is a special case where ' ' is considered a valid offset initial character and this is because when
+	 * handling URLs with timestamps, '+' is considered an escape character for ' ', so '+' may have been replaced with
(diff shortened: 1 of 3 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/BaseDateTimeDt.java
```diff
@@ -230,17 +230,41 @@ public Long getNanos() {
...
+		if (maxIndexPlusAndMinus != -1) {
+			if ((maxIndexPlusAndMinus - 2) != (plusIndex + minusIndex + zIndex)) {
+				throwBadDateFormat(theValueString);
+			}
+			retVal = maxIndexPlusAndMinus;
 		}
+
+		if (maxIndexSpaceAndMinus != -1) {
+			if ((maxIndexSpaceAndMinus - 2) != (spaceIndex + minusIndex + zIndex)) {
+				throwBadDateFormat(theValueString);
```
jamesagnew, line 263: Nitpick: It'd be good to have a test which exercises both of the possible failure path conditions here, since we added a new one.

jamesagnew (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/model/primitive/BaseDateTimeDt.java
lukedegruchy: We seem to already have both.  I ran the entire test class in debug mode with breakpoints 

```java
verifyFails("1974-12-25T23:59 00Z");
```

Triggers the space failure block.

```java
			new DateTimeDt("2010-01-01T00:00:00.1234-09:00Z");
```

Triggers the plus failure block.

PR #6124, opened 2024-07-17
Patient validate operation with remote terminology service enabled returns 400 bad request

### LINKED ISSUE #6122: Patient $validate operation with Remote Terminology Service enabled returns 400 Bad Request
Patient $validate operation with Remote Terminology Service enabled returns 400 Bad Request response instead of 200 OK with a list of validation results. 

**To Reproduce**
1. Add `RemoteTerminologyServiceValidationSupport` to `ValidationSupportChain`
2. Create Patient:
```json
{
  "resourceType": "Patient",
  "active": true,
  "deceasedDateTime": "2015-02-14T13:42:00",
  "identifier": [ { "system": "urn:oid:1.2.36.146.595.217.0.1", "value": "12346" } ],
  "name": [ {
      "family": "smith",
      "given": [ "Jane" ]
  } ],
  "gender": "male",
  "birthDate": "1974-12-25",
  "telecom":[{"value":"581-309-1234"}]
}
```
3. Execute $validate operation on Patient instance:
`GET http://localhost:8000/Patient/1652/$validate`
4. See 400 Bad Request response:
```
{
	"issue": [
		{
			"severity": "error",
			"code": "invalid",
			"details": {
				"text": "[28909888-9c0c-4070-a15a-1e239b795686]: A code/system pair must be provided."
			},
			"diagnostics": "[28909888-9c0c-4070-a15a-1e239b795686]: A code/system pair must be provided."
		},
		{
			"severity": "information",
			"code": "informational",
			"diagnostics": "X-Request-Id: 822bdfe6-44c0-4341-9df6-5f5493f0e02b"
		}
	]
}
```

**Expected behaviour**
200 OK response with a list of validation results should be returned instead.

volodymyr-korzh:
- Added ValidationSupportUtils class that includes logic for extracting CodeSystem from ValueSet for R4 and R5 FHIR versions.
- RemoteTerminologyServiceValidationSupport.extractCodeSystemForCode method was removed (moved to ValidationSupportUtils)
- Changes to RemoteTerminologyServiceValidationSupport: added handling of 400 and 404 responses from Remote Terminology server for $validate-code operation
- Added tests
- Added change-log

Closes #6122


hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
```diff
@@ -757,10 +758,12 @@ public ValidationResult validateCode(
 			throw new InternalErrorException(Msg.code(690) + e);
 		}
 
+		String system = ValidationSupportUtils.extractCodeSystemForCodeR5(theValueSet, code);
```
codeforgreen, line 761: Nice to have: I see this class is all based on R5. Can you please fix some of the declarations so all resource classes use imports (e.g. `org.hl7.fhir.r5.model.ValueSet`)

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtilsTest.java
```diff
@@ -0,0 +1,142 @@
...
+		List<ValueSet.ConceptReferenceComponent> conceptNoCode = Lists.newArrayList(
+			new ValueSet.ConceptReferenceComponent().setCode(NOT_THE_CODE));
+
+		return Stream.of(
+			Arguments.of(Collections.emptyList(), null, "Empty ValueSet includes"),
+			Arguments.of(Lists.newArrayList(new ValueSet.ConceptSetComponent(), new ValueSet.ConceptSetComponent()),
+				null, "ValueSet includes without system"),
+			Arguments.of(Lists.newArrayList(new ValueSet.ConceptSetComponent()),
+				null, "ValueSet include without system"),
+			Arguments.of(Lists.newArrayList(new ValueSet.ConceptSetComponent().setSystem(SYSTEM_URL)),
```
dotasek, line 56: In this, and in the corresponding r5 test, I think "ValueSet include with one system and code" and "ValueSet include with one system and no code" appear to be swapped. This line includes a code, and the next one doesn't.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtilsTest.java
```diff
@@ -0,0 +1,142 @@
...
+			new org.hl7.fhir.r5.model.ValueSet.ConceptReferenceComponent().setCode(NOT_THE_CODE));
+
+		return Stream.of(
+			Arguments.of(Collections.emptyList(), null, "Empty ValueSet includes"),
+			Arguments.of(Lists.newArrayList(new org.hl7.fhir.r5.model.ValueSet.ConceptSetComponent(),
+					new org.hl7.fhir.r5.model.ValueSet.ConceptSetComponent()),
+				null, "ValueSet includes without system"),
+			Arguments.of(Lists.newArrayList(new org.hl7.fhir.r5.model.ValueSet.ConceptSetComponent()),
+				null, "ValueSet include without system"),
+			Arguments.of(Lists.newArrayList(new org.hl7.fhir.r5.model.ValueSet.ConceptSetComponent().setSystem(SYSTEM_URL)),
```
dotasek, line 107: See comment in R4 test. Test names appear to be swapped.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
```diff
@@ -0,0 +1,88 @@
...
+
+	private static final Logger ourLog = LoggerFactory.getLogger(ValidationSupportUtils.class);
+
+	private ValidationSupportUtils() {}
+
+	/**
+	 * Try to obtain the codeSystem of the received code from the input R4 ValueSet
+	 */
+	public static String extractCodeSystemForCodeR4(ValueSet theValueSet, String theCode) {
+		if (theValueSet == null || theValueSet.getCompose().getInclude().isEmpty()) {
```
dotasek, line 17: There was a check for `theValueSet.getCompose().getInclude() == null` here before. Is this no longer necessary?

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
```diff
@@ -0,0 +1,88 @@
...
+			return theCodeSystem + "|" + theVersion;
+		}
+		return theCodeSystem;
+	}
+
+	/**
+	 * Try to obtain the codeSystem of the received code from the input R5 ValueSet
+	 */
+	public static String extractCodeSystemForCodeR5(org.hl7.fhir.r5.model.ValueSet theValueSet, String theCode) {
+		if (theValueSet == null || theValueSet.getCompose().getInclude().isEmpty()) {
```
dotasek, line 56: Same comment re:  `theValueSet.getCompose().getInclude() == null`

dotasek (changes requested):
See comments regarding test cases and a removed null check. Otherwise, looks good to me.

hapi-fhir-validation/src/test/java/org/hl7/fhir/r4/validation/RemoteTerminologyServiceValidationSupportR4Test.java
```diff
@@ -119,6 +119,12 @@ public void testValidateCode_withBlankCode_returnsNull() {
 		assertNull(outcome);
 	}
 
+	@Test
+	public void testValidateCode_withBlankCodeSystemAndValueSetUrl_returnsNull() {
```
codeforgreen, line 123: Do you have any integration test which checks the output of the `$validate` operation? I wanted to see how the response is translated further.

codeforgreen (commented):
Overall changes look good. I see there are some additional validations added in the existing tests, but I am wondering about this specific use-case, how it behaves in the scenario described after your changes.
Please look at my comments and questions. Thanks.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -81,6 +81,12 @@ public CodeValidationResult validateCode(
 			String theCode,
 			String theDisplay,
 			String theValueSetUrl) {
+
+		// Remote terminology services shouldn't be used to validate codes with an implied system
+		if (isBlank(theCodeSystem) && isBlank(theValueSetUrl)) {
```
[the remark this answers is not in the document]
volodymyr-korzh: This if statement is intended to prevent all cases when we don't have a system or valueSetUrl to be executed on remote terminology server. 
This logic is mirroring behaviour that we already have in `RemoteTerminologyServiceValidationSupport#validateCodeInValueSet` 

Moreover, it is a requirement in $validate-code operation specification to provide "url" or "codeSystem". So we are just aligning with a specification. 

Specification:
`If the operation is not called at the instance level, one of the parameters "url" or "codeSystem" must be provided.`
https://hl7.org/fhir/R4/codesystem-operation-validate-code.html
(we are calling $validate-code not at instance level)

The field that caused an error for patient validation was actually a `gender` field.  But failure (400 bad request) could previously happen for different fields and different resources.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
```diff
@@ -595,11 +595,11 @@ public void testValidate() {
 			fail(myFhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(e.getOperationOutcome()));
 		}
 		myCaptureQueriesListener.logSelectQueriesForCurrentThread();
-		assertEquals(8, myCaptureQueriesListener.getSelectQueriesForCurrentThread().size());
+		assertEquals(10, myCaptureQueriesListener.getSelectQueriesForCurrentThread().size());
```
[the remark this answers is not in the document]
volodymyr-korzh: Sure

The fix that was introduced makes validation more "aggressive" by validating more fields/codes. 
As previously we tried to validateCodeInValueSet/validateCodeInCodeSystem with codeSystem as null, validation was not executed by validation chain, and returned null ValidationResult. 

As now validateCodeInValueSet/validateCodeInCodeSystem executed with extracted codySystem - number of queries increased, and it is expected.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtilsTest.java
volodymyr-korzh: Thanks for noticing. Fixed

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtilsTest.java
volodymyr-korzh: Fixed

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
volodymyr-korzh: Under the hood `getInclude()` creates an empty arraylist if executed. It is safe to remove this check as it will never return true

```
        public List<ConceptSetComponent> getInclude() { 
          if (this.include == null)
            this.include = new ArrayList<ConceptSetComponent>();
          return this.include;
        }
```

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
volodymyr-korzh: Yes, it is safe to remove this (same reason as for R4 method)

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
volodymyr-korzh: Nice catch. Removed  redundant `org.hl7.fhir.r5.model.` declarations.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_4_0/6122-validate-operation-with-remote-terminology-service-enabled-returned-bad-request.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 6122
+title: "Previously, executing the '$validate' operation on a patient instance would result in an HTTP 400 Bad Request
```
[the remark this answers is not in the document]
volodymyr-korzh: In this case, validating patient with gender field caused 400 Bad request. But this error could potentially impact other resources. I corrected title and remove mentioning of Patient resource. 

It is hard to describe one case to add it to the title as impact of this issue is broad and depend on where the validation methods are called from InstanceValidator.

Updated version:

```
Previously, executing the '$validate' operation on a resource instance could result in an HTTP 400 Bad Request 
instead of an HTTP 200 OK response with a list of validation issues. This has been fixed.
```

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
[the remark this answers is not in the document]
dotasek: I can concur. What Volodymyr describes above is consistent with what I saw when we were tracing the issue.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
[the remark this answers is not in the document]
dotasek: Can concur here after having traced this with Volodymyr. Thank you, @codeforgreen for bringing these up, they really do help document the changes well.

dotasek (approved):
Looks good to me.

hapi-fhir-validation/src/test/java/org/hl7/fhir/r4/validation/RemoteTerminologyServiceValidationSupportR4Test.java
volodymyr-korzh: Added `FhirInstanceValidatorR4Test#testValidate_patientWithGenderCode_resolvesCodeSystemFromValueSet` 
It is not a full integration test, as we have to mock Remote terminology service, but it covers the main fix implemented in `VersionSpecificWorkerContextWrapper#validateCode` and checks the output result of validation.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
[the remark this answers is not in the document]
codeforgreen: @volodymyr-korzh thanks for explaining. I was following the call sequence for instance level $validate-code and I see it goes into the same method as resource type level $validate-code. Can you please check and see if there is a test for this? 
In particular for instance level ValueSet $validate-code, I have an actual use-case where the system may be absent. I've documented it here: https://github.com/hapifhir/hapi-fhir/issues/6150. The `validateCode` call that results from checking a profile binding with a `ValueSet` will not have the system populated.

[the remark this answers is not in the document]
codeforgreen: Another aspect is why would Remote Terminology have this restriction but other implementations of IValidationSupport would not have this? As you know we have several and there is a `ValidationSupportChain` which invokes them as part of the operation call. Remote Terminology is just one of them.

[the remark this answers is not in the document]
codeforgreen: I believe this issue and the issue I linked #6150 have the same root cause, which is about the fact that we simply throw 400 and 404 exceptions instead of converting them to a NotFound response.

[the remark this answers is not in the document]
volodymyr-korzh: Removed lines of code in question, introduced error handling for 400 and 404 responses for $validate-code operation

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -101,12 +102,7 @@ public CodeValidationResult validateCodeInValueSet(
 		// so let's try to get it from the VS if is not present
 		String codeSystem = theCodeSystem;
 		if (isNotBlank(theCode) && isBlank(codeSystem)) {
-			codeSystem = extractCodeSystemForCode((ValueSet) theValueSet, theCode);
-		}
-
-		// Remote terminology services shouldn't be used to validate codes with an implied system
-		if (isBlank(codeSystem)) {
-			return null;
+			codeSystem = ValidationSupportUtils.extractCodeSystemForCodeR4((ValueSet) theValueSet, theCode);
```
dotasek, line 105: I realize the previous code was just moved and that it was previously using R4 and just not explicitly saying so, but I'm looking over this again, can we safely assume R4 here?

hapi-fhir-validation/src/test/java/org/hl7/fhir/r4/validation/RemoteTerminologyServiceValidationSupportR4Test.java
```diff
@@ -342,32 +380,46 @@ public void testUniqueComposeInclude() {
...
+			public static Stream<Arguments> getRemoteTerminologyServerExceptions() {
+				return Stream.of(
+					Arguments.of(new ResourceNotFoundException("System Not Present"), "404 Not Found: System Not Present"),
+					Arguments.of(new InvalidRequestException("Invalid Request"), "400 Bad Request: Invalid Request")
+				);
+			}
+
+			@ParameterizedTest
+			@MethodSource(value = "getRemoteTerminologyServerExceptions")
+			public void systemNotPresent_returnsValidationResultWithError(Exception theException, String theServerMessage) {
```
dotasek, line 402: I spent some time trying to differentiate this and the following test (`systemNotPresent_returnsValidationResultWithError` and `systemPresentCodeNotPresent_returnsValidationResultWithError`).

Am I correct when I say that neither test can infer the system? 

It looks to me like both expect the same errors, but for a different reason. In the first, there is a compose that has a ConceptSetComponent with no system, and in the second, there are two ConceptSetComponent's, each with a system, thus making inference impossible.

dotasek (changes requested):
No changes requested, just clarification regarding the comments I left.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
volodymyr-korzh: It is not safe to assume that we will have R4 ValueSet here. We well have R5 ValueSet here for R5 context, etc. I refactored extractCodeSystemForCode method to extract codeSystem from DSTU3, R4 and R5 ValueSets.

hapi-fhir-validation/src/test/java/org/hl7/fhir/r4/validation/RemoteTerminologyServiceValidationSupportR4Test.java
volodymyr-korzh: Yes, correct. 
In first case - `systemNotPresent` - we have ValueSet with no code system 
second case `systemPresentCodeNotPresent` - we have multiple code systems, but we cant decide which to use (as both of systems missing the code that we want to validate)

dotasek (approved):

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -630,11 +583,22 @@ protected CodeValidationResult invokeRemoteValidateCode(
...
-				.withParameters(input)
-				.execute();
+		IBaseParameters output;
+		try {
+			output = client.operation()
+					.onType(resourceType)
+					.named("validate-code")
+					.withParameters(input)
+					.execute();
+		} catch (ResourceNotFoundException | InvalidRequestException ex) {
```
codeforgreen, line 593: Can you please merge your branch to the release branch? This shows as new code in this PR and it already exists in the branch. Thanks.

codeforgreen (approved):

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
codeforgreen: Ah, it seems that this is for $validate-code. That seems legitimate as well, even though the spec allows response as 400, this will produce a more user-friendly response, consistent with other operations such as $lookup.

PR #6128, opened 2024-07-18
Add path value to _id search parameter and other missing search param…

jmarchionatto:
…eters to IAnyResource.


epeartree (approved):
nice good clean code and test.

Issue #6104, opened 2024-07-12
SearchParameterMap Cloning


melihaydogd:
**Describe the bug**
The `SearchParameterMap` clone function doesn't clone the map properly. The current implementation can be found [here](https://github.com/hapifhir/hapi-fhir/blob/40d42f3ba5ddb6b747f5b601251333783591ca02/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/SearchParameterMap.java#L103-L138). In this implementation, the param objects is not cloned. Additinonally, the set for includes and revIncludes are carried to the new map. When modifications are made to that set, the original set is also modified.

**To Reproduce**
1. Clone the search parameter map
2. Modify any param object `TokenParam`, `Include`, `StringParam`.
3. See the original param is updated.
4. Modify includes or revIncludes set.
5. See the original set is updated.

**Expected behavior**
I expect the cloning to be deep copy so that any modifications on the clone search parameter map doesn't affect the original map.

**Environment (please complete the following information):**
 - Every HAPI FHIR Version
 - OS: Every OS
 - Every Browser

s-rb:
Hi, @melihaydogd !

I'd like to solve the issue.

But I've got some questions:
- For deep cloning I'll have to implement `Cloneable` and `clone()` method for each class in the tree: QuantityParam, Include, BaseParam, BaseParamWithPrefix, DateRangeParam, SortSpec
-There is a field `private final HashMap<String, List<List<IQueryParameterType>>> mySearchParameterMap`
and to clone it properly, `IQueryParameterType` has to extend `Cloneable` too. So a lot of classes which implement `IQueryParameterType` have to implement clone() method too

Is it OK?

jamesagnew:
@s-rb sounds like a reasonable approach to me!

s-rb:
@jamesagnew @melihaydogd
Could you assign this issue to me?
thanks ))

PR #6130, opened 2024-07-18
SearchParameterMap deep cloning via serialization/deserialization (#6104)

s-rb:
SearchParameterMap deep cloning via serialization/deserialization (#6104)

After careful research, I concluded that deep cloning through the implementation of Cloneable will not be the best solution. There are many classes in the hierarchy where it will not be possible to call the clone() method on the parent for some fields, and then we will have to assign a reference, which negates the whole purpose of deep cloning.

Cloning through serialization should help, and in the project I did not find any transient fields that could interfere with this method.

### DIFF
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/SearchParameterMap.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/SearchParameterMap.java
@@ -38,6 +38,7 @@ import ca.uhn.fhir.rest.param.TokenParamModifier;
 import ca.uhn.fhir.util.UrlUtil;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import jakarta.annotation.Nonnull;
+import org.apache.commons.lang3.SerializationUtils;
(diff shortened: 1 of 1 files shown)


hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/SearchParameterMap.java
```diff
@@ -101,40 +102,12 @@ public SearchParameterMap(String theName, IQueryParameterType theParam) {
...
-			List<List<IQueryParameterType>> newAndParams = new ArrayList<>();
-			for (List<IQueryParameterType> orParams : andParams) {
-				List<IQueryParameterType> newOrParams = new ArrayList<>(orParams);
-				newAndParams.add(newOrParams);
-			}
-			map.put(entry.getKey(), newAndParams);
-		}
-
-		return map;
+		return SerializationUtils.clone(this);
```
jamesagnew, line 110: Relying on java serialization is potentially dangerous (and more importantly, explicitly forbidden by many organizations). This approach is fixing one problem by creating another unfortunately.

PR #6143, opened 2024-07-23
fix hfj search migration task

### LINKED ISSUE #6142: Incorrect Migration Task Causing Incorrect Column Length
**Describe the bug**
If you upgrade HAPI versions to 6.6.0 or higher from a lower version, an incorrect migration task would result in the incorrect column length for the following column:
- `SEARCH_UUID` in the `HFJ_SEARCH` table

**To Reproduce**
1. Check the `SEARCH_UUID` column length in HAPI version lower thn 6.6.0 --> it is 36
2. Upgrade HAPI
3. Check the `SEARCH_UUID` column length in HAPI version lower thn 6.6.0 --> it is still 36

**Expected behavior**
The new length for the `SEARCH_UUID` column should be 48.

jdar8:
- Add a new migration for `HFJ_SEARCH` table to modify the `SEARCH_UUID` column length. Previously, we accidentally used `addColumn()` when modifying the length, which did nothing since the column already existed. 

Closes #6142

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
@@ -468,6 +468,11 @@ public class HapiFhirJpaMigrationTasks extends BaseMigrationTasks<VersionEnum> {
 						.failureAllowed();
 			}
 		}
+
+		version.onTable(Search.HFJ_SEARCH)
+			.modifyColumn("20240722.1", Search.SEARCH_UUID)
+			.nullable()
+			.withType(ColumnTypeEnum.STRING, Search.SEARCH_UUID_COLUMN_LENGTH);
 	}
 
 	protected void init720() {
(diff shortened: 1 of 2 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
```diff
@@ -468,6 +468,11 @@ protected boolean shouldSkipTask() {
 						.failureAllowed();
 			}
 		}
+
+		version.onTable(Search.HFJ_SEARCH)
+			.modifyColumn("20240722.1", Search.SEARCH_UUID)
+			.nullable()
+			.withType(ColumnTypeEnum.STRING, Search.SEARCH_UUID_COLUMN_LENGTH);
```
tadgh, line 475: issue: You cannot use a constant in this length column. If we ever change this, we are effectively changing history next time we publish. Please use the actual numeric value for the length.

tadgh (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
jdar8: good point, fixed

tadgh (approved):

PR #6149, opened 2024-07-24
To provide the target resource partitionId and partitionDate in the resourceLinlk

epeartree:
**What was done:**

- added targetResourcePartitionId ResourceLink entity as a nullable field;
- added construcitons to populate the new fields whenever the target_resource_id column is filled;
- provided a migration task;
- modified tests;
- added changelog;


hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/dao/JpaPid.java
```diff
@@ -55,6 +57,15 @@ private JpaPid(Long theId, Long theVersion, String theResourceType) {
 		myId = theId;
 	}
 
+	public PartitionablePartitionId getPartitionablePartitionId() {
+		return myPartitionablePartitionId;
+	}
+
+	public JpaPid setPartitionablePartitionId(PartitionablePartitionId thePartitionablePartitionId) {
```
lukedegruchy, line 64: Why not define another constructor taking partition ID instead?

lukedegruchy (approved):

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/dao/JpaPid.java
epeartree: because partitionablePartitionId is not a mandatory property of jpaPid and i don't like adding a new constructor for a non mandatory property of an object

PR #6163, opened 2024-07-26
Make transactions partition aware

jamesagnew:
Transactions on systems configured to require a new transaction when changing partitions will now be able to correctly execute  FHIR transactions


jamesagnew:
Abandoning, this should be against 7.4 branch

PR #6164, opened 2024-07-26
fix concepts with no display element for $apply-codesystem-delta-add …

### LINKED ISSUE #6159: $apply-codesystem-delta-add and $apply-codesystem-delta-remove fail if no display provided
**Describe the bug**
`$apply-codesystem-delta-add` or `$apply-codesystem-delta-remove` operations fail when invoked with a CodeSystem Resource payload with  concepts not without a `display` element. 

To reproduce

1.  Call POST `<fhirhost>/CodeSystem/$apply-codesystem-delta-add` endpoint with the following payload where the concept doesn't have a `display` field. 

```
{
    "resourceType": "Parameters",
    "parameter": [
        {
            "name": "system",
            "valueUri": "http://example.com/labCodes"
        },
        {
            "name": "codeSystem",
            "resource": {
                "resourceType": "CodeSystem",
                "concept": [
                    {
                        "code": "CHEM"
                    }
                ]
            }
        }
    ]
}

```
Actual Outcome: 
```
{
    "issue": [
        {
            "severity": "error",
            "code": "processing",
            "diagnostics": "HAPI-0389: Failed to call access method: java.lang.NullPointerException: Cannot invoke \"String.replace(java.lang.CharSequence, java.lang.CharSequence)\" because \"theValue\" is null"
        }
    ]
}
```

Expected outcome: Success, with concept added. 

Same thing happens to $apply-codesystem-delta-remove

mrdnctrk:
Make `$apply-codesystem-delta-add` and `$apply-codesystem-delta-remove` operations work when invoked with a CodeSystem Resource payload with concepts that doesn't have `display` elements.

close #6159

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/TerminologyUploaderProvider.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/TerminologyUploaderProvider.java
@@ -475,6 +475,9 @@ public class TerminologyUploaderProvider extends BaseJpaProvider {
 	}
 
 	private static String csvEscape(String theValue) {
+		if (theValue == null) {
+			return "";
+		}
 		return '"' + theValue.replace("\"", "\"\"").replace("\n", "\\n").replace("\r", "") + '"';
 	}
 }
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/TerminologyUploaderProviderR4Test.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/TerminologyUploaderProviderR4Test.java
@@ -347,7 +347,7 @@ public class TerminologyUploaderProviderR4Test extends BaseResourceProviderR4Tes
 			"\"reference\": \"CodeSystem/"
 		);
 
-		assertHierarchyContains(
+		assertHierarchyContainsExactly(
 			"CHEM seq=0",
 			" HB seq=0",
 			" NEUT seq=1",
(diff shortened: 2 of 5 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/TerminologyUploaderProviderR4Test.java
```diff
@@ -582,48 +614,151 @@ public void testApplyDeltaAdd_MissingFile() {
...
-		myClient
+		applyDeltaAddCustomTermCodes();
+
+		// And remove all of them using the same set of csv files
+		LoggingInterceptor interceptor = new LoggingInterceptor(true);
+		myClient.registerInterceptor(interceptor);
+		Parameters outcome = myClient
 			.operation()
 			.onType(CodeSystem.class)
-			.named(JpaConstants.OPERATION_APPLY_CODESYSTEM_DELTA_ADD)
```
mrdnctrk, line 601: In this test, I refactored the test setup part to add the codes into applyDeltaAddCustomTermCodes function to be able to reuse it in other tests.

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/BaseJpaR4Test.java
```diff
@@ -740,7 +740,7 @@ protected void upload(String theClasspath) throws IOException {
 		dao.update(resourceParsed);
 	}
 
-	protected void assertHierarchyContains(String... theStrings) {
```
mrdnctrk, line 743: renamed this function for accuracy, because it uses containsExactly not just contains.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/TerminologyUploaderProvider.java
```diff
@@ -475,6 +475,9 @@ public InputStream getInputStream() {
 	}
 
 	private static String csvEscape(String theValue) {
```
mrdnctrk, line 477: The problem was when display is not provided, theValue would be null, and this function would still try to process a null value. Instead now return empty string, which indicates that the value is missing in csv.

tadgh (approved):
Looks good. Nice fix!

Issue #6008, opened 2024-06-13
Add support for running reindex job across resources in multiple partitions


codeforgreen:
**Describe the issue**
Add support for running reindex jobs across resources in multiple partitions.
The solution should be easily extendable or reusable for other system bulk operations using batch2 jobs.
Introduce an abstraction and an additional loop such that `GenerateRangeChunksStep` is able to read all partitions and generate chunks of resources for each of them.

**Environment (please complete the following information):**
 - HAPI FHIR Version 7.3.4

**Additional context**

PR #6185, opened 2024-08-02
Fix $reindex job with custom partition interceptor based on resource type. Update reindex job to always run with urls.

### LINKED ISSUE #6179: NullPointerException when performing a $reindex operation with a custom resource type based partitioning interceptor
**Describe the bug**
NullPointerException occurs when performing a $reindex operation with a custom resource type based partitioning interceptor. The exception happens because the `ReadPartitionIdRequestDetails` does not have resourceType/resourceName populated.

**To Reproduce**

1. Register interceptor that uses the resource type to determine the partition. 

```java
@Interceptor
public class PartitionInterceptor  {
	
	private static final Logger logger = LoggerFactory.getLogger(PartitionInterceptor.class);

	public static final RequestPartitionId PARTITION_PATIENT = RequestPartitionId.fromPartitionId(1);
	public static final RequestPartitionId PARTITION_OBSERVATION = RequestPartitionId.fromPartitionId(2);

	@Autowired
	private FhirContext myFhirContext;

	@Hook(Pointcut.STORAGE_PARTITION_IDENTIFY_CREATE)
	public RequestPartitionId selectPartitionCreate(IBaseResource theResource) {
		String resourceType = myFhirContext.getResourceType(theResource);
		return selectPartition(resourceType);
	}

	@Hook(Pointcut.STORAGE_PARTITION_IDENTIFY_READ)
	public RequestPartitionId selectPartitionRead(ReadPartitionIdRequestDetails theDetails) {
		if (theDetails == null) {
			return PARTITION_PATIENT;
		} else {
			return selectPartition(theDetails.getResourceType());
		}
		
	}

	@Nonnull
	private static RequestPartitionId selectPartition(String resourceType) {
		switch (resourceType) {
			case "Patient":
				return PARTITION_PATIENT;
			case "Subscription":
				return PARTITION_PATIENT;		
			case "Bundle":
				return PARTITION_PATIENT;	
			case "Observation":
				return PARTITION_OBSERVATION;
			default:
				return RequestPartitionId.defaultPartition();
		}
		// return PARTITION_PATIENT;
	}

}
```
3. Create a partitions with id 1 and 2. 

`POST <base_url>/DEFAULT/$partition-management-create-partition`

```
{
  "resourceType": "Parameters",
  "parameter": [ {
    "name": "id",
    "valueInteger": 1
  }, {
    "name": "name",
    "valueCode": "PARTITION-1"
  }, {
    "name": "description",
    "valueString": "a description"
  } ]
}
```
4. Invoke the reindex operation for a certain partition

`POST Partition-A/$reindex`

```
{
    "parameter": [
        {
            "valueString": "Patient?",
            "name": "url"
        }
    ],
    "resourceType": "Parameters"
}
```

response 500

`HAPI-0389: Failed to call access method: java.lang.NullPointerException: Cannot invoke "String.hashCode()" because "" is nullresponse 500, [HAPI-0389](https://simpaticois.atlassian.net/browse/HAPI-0389?atlOrigin=eyJpIjoiY2QyZTJiZDRkNGZhNGZlMWI3NzRkNTBmZmVlNzNiZTkiLCJwIjoianN3LWdpdGxhYi1pbnQifQ): Failed to call access method: java.lang.NullPointerException: Cannot invoke "String.hashCode()" because "" is null`

**Expected behaviour**
The operation should succeed with response code 200.

 - HAPI FHIR Version 7.4

codeforgreen:
Fixes #6179 

Root cause:

Previously, the `ReindexProvider` was triggering method `determineReadPartitionForRequestForServerOperation` with a request that did not contain a resource type. That ended up invoking the custom interceptor causing a NullPointerException because the interceptor was expecting a resource type. There was a resource type provided in the urls of the reindex job.

Fix:
- `MultitenantBatchOperationR4Test` was updated to check for 2 resources instead of 3 for one test which performs reindex for the default partition. This is a side effect of the fact that now we have urls and the partition is being considered. The previously used query from `streamIdsTypesAndUpdateTimesOfResourcesWithinUpdatedRangeOrderedFromOldestForDefaultPartition` did not consider the partition and was returning resources from all partitions. 
- The above mentioned query was not fixed because it is a dead code path. At the moment there are no jobs that run without urls. A comment was in method `streamResourceIdsNoUrl` to reflect that. Since the code could potentially be used, it was left there. It can be removed if a cleanup is preferred.
- The `JpaJobPartitionProvider` and `DefaultJobPartitionProvider` have been updated to support determining the partitions associated with the urls.
- The `GenerateRangeChunksStep` and `JobParameters` have been pretty much restored to the state they were before changes on #6008 happened.

Change:
- For the use-case when there are no provided urls, we now generate a list of urls out of all supported resource types. This means that method `determineReadPartitionForRequestForServerOperation` is no longer invoked and we always invoke `determineReadPartitionForRequestForSearchType` instead. This ensures the resource type will always be present when the partition is computed.

Bonus:
- Fix compile warnings in some of the touched files
- Move some classes in the appropriate packages or rename their package to reflect usage
- Bring together all tests for `Batch2DaoSvcImpl` in the same test class `Batch2DaoSvcImplTest`


volodymyr-korzh (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/ResourceTypePartitionInterceptorR4Test.java
```diff
@@ -0,0 +1,90 @@
...
+				.onServer()
+				.named(ProviderConstants.OPERATION_REINDEX)
+				.withParameters(input)
+				.execute();
+
+		String jobId = ((StringType)response.getParameterValue(ProviderConstants.OPERATION_REINDEX_RESPONSE_JOB_ID)).getValue();
+		myBatch2JobHelper.awaitJobHasStatus(jobId, StatusEnum.COMPLETED);
+		assertThat(myBatch2JobHelper.getCombinedRecordsProcessed(jobId)).isEqualTo(theExpectedIndexedResourceCount);
+
+		myPartitionSettings.setAllowReferencesAcrossPartitions(new PartitionSettings().getAllowReferencesAcrossPartitions());
```
michaelabuckley, line 66: minor: this should be in an `@AfterEach` block.

hapi-fhir-jpaserver-test-r4/src/test/resources/logback-test.xml
```diff
@@ -31,6 +31,8 @@
 	<logger name="ca.uhn.fhir.jpa.bulk" level="info"/>
 
 	<!-- more debugging -->
+	<logger name="ca.uhn.fhir.jpa.search.builder.sql.SearchQueryExecutor" level="trace"/>
```
michaelabuckley, line 34: oops

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/Batch2JobHelper.java
```diff
@@ -82,7 +82,7 @@ public JobInstance awaitJobCompletion(String theInstanceId, int theSecondsToWait
 	}
 
 	public JobInstance awaitJobHasStatus(String theInstanceId, StatusEnum... theExpectedStatus) {
-		return awaitJobHasStatus(theInstanceId, 10, theExpectedStatus);
+		return awaitJobHasStatus(theInstanceId, 1000, theExpectedStatus);
```
michaelabuckley, line 85: Hmm: I think this is a bit too high, but 10 is too low.  How about 30?

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/reindex/ReindexJobParametersValidator.java
```diff
@@ -20,41 +20,40 @@
...
 
 		if (errors == null || errors.isEmpty()) {
 			// only check if there's no other errors (new list to fix immutable issues)
 			errors = new ArrayList<>();
-			List<PartitionedUrl> urls = theParameters.getPartitionedUrls();
-			for (PartitionedUrl purl : urls) {
-				String url = purl.getUrl();
-
+			for (String url : theParameters.getUrls()) {
+				if (StringUtils.isBlank(url)) {
```
michaelabuckley, line 49: really?  url can be null.  But can it really be blank?  Empty?

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/reindex/ReindexProviderTest.java
```diff
@@ -121,48 +117,27 @@ public void testReindex_ByUrl() {
...
 		// Default values
 		assertEquals(ReindexParameters.ReindexSearchParametersEnum.ALL, params.getReindexSearchParameters());
 		assertTrue(params.getOptimisticLock());
 		assertEquals(ReindexParameters.OptimizeStorageModeEnum.NONE, params.getOptimizeStorage());
-	}
-
-	@Test
-	public void testReindex_NoUrl() {
-		// setup
-		Parameters input = new Parameters();
```
michaelabuckley, line 136: problem: bring this test back.  This is still a legal scenario.  The FHIR client can still make a request with no url in the parameters.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/IJobPartitionProviderTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+		List<String> urls = List.of("Patient?", "Observation?");
+		Set<PartitionedUrl> partitionedUrls = new LinkedHashSet<>();
+		partitionIds.forEach(p -> partitionedUrls.add(new PartitionedUrl().setUrl("Patient?").setRequestPartitionId(p)));
+		partitionedUrls.add(new PartitionedUrl().setUrl("Patient?").setRequestPartitionId(RequestPartitionId.defaultPartition()));
+		partitionedUrls.add(new PartitionedUrl().setUrl("Observation?").setRequestPartitionId(RequestPartitionId.fromPartitionIds(1)));
+		verifyGetPartitionedUrls(requestDetails, urls, partitionedUrls);
+	}
+
+	default void verifyGetPartitionedUrls(RequestDetails theRequestDetails, List<String> theUrls, Collection<PartitionedUrl> thePartitionedUrls) {
```
michaelabuckley, line 93: trivial: The tests would be easier splitting execute from verify.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/DefaultJobPartitionProvider.java
```diff
@@ -0,0 +1,125 @@
...
+		}
+		List<PartitionedUrl> retVal = new ArrayList<>();
+		for (PartitionedUrl partitionedUrl : partitionedUrls) {
+			String url = partitionedUrl.getUrl();
+			RequestPartitionId partitionId = partitionedUrl.getRequestPartitionId();
+			if (partitionId != null && partitionId.isAllPartitions() && !allPartitions.isEmpty()) {
+				List<PartitionedUrl> megaScalePartitionedUrls = allPartitions.stream()
+						.map(p -> (new PartitionedUrl().setUrl(url).setRequestPartitionId(p)))
+						.collect(Collectors.toList());
+				retVal.addAll(megaScalePartitionedUrls);
```
michaelabuckley, line 112: problem: megascale name is bad

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/DefaultJobPartitionProvider.java
```diff
@@ -0,0 +1,125 @@
...
+				retVal.add(partitionedUrl);
+			}
+		}
+
+		return retVal;
+	}
+
+	public List<RequestPartitionId> getAllPartitions() {
+		// likely allPartitions is not supported in this path
+		return List.of();
```
michaelabuckley, line 123: I'm not sure we still need the JPA version of this class.

```suggestion
		return List.of(RequestPartitionId.allPartitions());
```

michaelabuckley (approved):

michaelabuckley (approved):
I think this is ok.  I'm not entirely sure we have zero-downtime in the first-chunk->ids-chunk flow.  But any errors should be rare.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/parameters/JobParameters.java
```diff
@@ -22,20 +22,22 @@
...
+import java.util.stream.Collectors;
 
 /**
  * Can be used to configure parameters for batch2 jobs.
  * Please note that these need to be backward compatible as we do not have a way to migrate them to a different structure at the moment.
  */
 public class JobParameters implements IModelJson {
 	@JsonProperty(value = "partitionId")
-	private List<RequestPartitionId> myRequestPartitionIds;
+	@Nullable
```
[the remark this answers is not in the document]
codeforgreen: I checked the prior classes 

https://github.com/hapifhir/hapi-fhir/blob/rel_7_2/hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/parameters/PartitionedUrlListJobParameters.java
and
https://github.com/hapifhir/hapi-fhir/blob/rel_7_2/hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/parameters/PartitionedJobParameters.java

and seems like partitionId was just a single `RequestPartitionId`. So this should not be a breaking change compared to previous release. 

It is actually https://github.com/hapifhir/hapi-fhir/pull/6009 that introduced breaking changes, here many of the changes are reverted back after simplification.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_4_0/6179-reindex-job-no-longer-runs-without-urls.yaml
```diff
@@ -0,0 +1,7 @@
+
+---
+type: change
+issue: 6179
+title: "The $reindex operation could potentially initiate a reindex job without any urls provided in the parameters.
+We now internally generate a list of urls out of all the supported resource types and attempt to reindex 
+found resources of each type separately. As a result, each reindex (batch2) job chunk will be always associated with a url."
```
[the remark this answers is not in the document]
codeforgreen: Yes, for the moment this is the case, when considering the partition can be determined based on resource type. We could potentially converge back to no url if they all point to the same partition. I will add a comment in the code about that.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/GenerateRangeChunksStep.java
```diff
@@ -54,66 +52,23 @@ public RunOutcome run(
...
-				&& !theOtherRequestPartitionId.isDefaultPartition()
-				&& !theOtherRequestPartitionId.isAllPartitions()) {
-			return theOtherRequestPartitionId;
-		}
-		return theRequestPartitionId;
-	}
-
 	private void sendChunk(ChunkRangeJson theData, IJobDataSink<ChunkRangeJson> theDataSink) {
 		String url = theData.getUrl();
 		ourLog.info(
```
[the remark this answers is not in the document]
codeforgreen: I've done this here and also in `ResourceIdListStep`

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/reindex/ReindexProviderTest.java
codeforgreen: I made the other test in this class parametrized and added @NullSource. I did not manage to get it to inject mocks in the `IJobPartitionProvider` but supplemented with test `getPartitionedUrls_noUrls_returnsCorrectly`, it should provide enough coverage.

PR #6192, opened 2024-08-05
Add tinder plugin to pluginManagement to allow to run generate-ddl from parent

jmarchionatto:
Added plugin skip option an added plugin to parent project's pom with with skip=true to allow generation to run from parent project.


pom.xml
```diff
@@ -2592,6 +2592,24 @@
 						</excludes>
 					</configuration>
 				</plugin>
+
+				<!-- Generate the DDL schema files -->
+				<plugin>
+					<groupId>ca.uhn.hapi.fhir</groupId>
+					<artifactId>hapi-tinder-plugin</artifactId>
+					<version>7.5.0-SNAPSHOT</version>
```
tadgh, line 2600: issue: pull master in and correct

tadgh (approved):
Needs a changelog, and some docs about how to run it. Otherwise approved.

PR #6194, opened 2024-08-06
Reverting package version dependency to core fhir-test-cases 1.1.14 and re-enabling test case

epeartree:
**Background:**

PR [6190](https://github.com/hapifhir/hapi-fhir/pull/6190) bumped the version of package fhir-test-cases (org.hl7) to 1.5.15 because version 1.1.14 was no longer available;

**What was done:**
- Reverting dependency to version 1.1.14;
- Re-enabling test that was failing with version 1.5.15

[Issue 6200](https://github.com/hapifhir/hapi-fhir/issues/6200) was opened to address the bump and failing test in the next release.


jmarchionatto:
Why is this reverted? 
Is version 1.1.14 available again?

epeartree:
yes, version 1.1.14 is still available.  we initially bumped to 1.5.15 because we thought that it was no longer available but that was a sonatype mishap.  after the bump, we realise that it was braking a test.  the easiest way to go about it is to revert for rel_7_4 and fix it in master hence issue 6200 was open

jmarchionatto:
Sounds good. Approved

jmarchionatto (approved):

PR #6219, opened 2024-08-15
Disabling enforce referential integrity on write still checks references 

### LINKED ISSUE #6218: Disabling enforce referential integrity on write still checks references
**Describe the bug**
When the "enforce referential integrity on write" setting is disabled on a database module, the referential integrity of linked resources are still checked. 

**To Reproduce**
1. Disable "enforce referential integrity on write" setting on database module
2. Post a resource with an invalid reference (ie. reference to non-existent resource) with ID less than 1000 (ex. Patient/123)
3. Notice that you will get Unprocessable entity response with HAPI code 1095 
4. Create a resource to reference (example: Observation/1500)
5. Delete that resource
6. Post a resource with an invalid reference to the resource you just deleted 
7. Notice that you will now Invalid Request response with HAPI code 1096

**Expected behavior**
In both cases, with "enforce referential integrity on write" set to false, these inavaid references should not be checked. So the POST request with invalid references should succeed.

jdar8:
- When resolving resource links for the reference, if the resolved reference is invalid (never existed, was deleted, resource is not of the same type), and ref integrity is disabled, then no error should be thrown and the link should resolve to null
- Expand tests to cover most cases for referential integrity on write 
- Changelog

Closes #6218


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ReferentialIntegrityTest.java
```diff
@@ -25,49 +32,165 @@ public void afterResetConfig() {
 		myStorageSettings.setEnforceReferenceTargetTypes(new JpaStorageSettings().isEnforceReferenceTargetTypes());
 	}
 
+	@ParameterizedTest
+	@MethodSource("paramsProvider_withResourceType")
+	public void testRefIntegrity_withUnknownReferenceIds(boolean theIsEnforceRefIntegrityEnabled,
```
TipzCM, line 37: eh, i'm not super fond of this test name because i'm not sure what it's testing (it's testing that referencial integrity check enabled fails if the ids are unknown, and passes if it's not enabled). But it might be harder to name that than something like `referentialIntegritySetting_unknownIds_fullScopeTest`

TipzCM (approved):

PR #6220, opened 2024-08-16
Move partition-compatibility logic into tx service.

michaelabuckley:
Minor refactor.

### DIFF
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
@@ -283,7 +283,7 @@ public class HapiTransactionService implements IHapiTransactionService {
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
```diff
@@ -272,7 +271,7 @@ protected <T> T doExecute(ExecutionBuilder theExecutionBuilder, TransactionCallb
 		try {
 			ourExistingTransaction.set(this);
 
-			if (isRequiresNewTransactionWhenChangingPartitions()) {
+			if (myTransactionPropagationWhenChangingPartitions == Propagation.REQUIRES_NEW) {
```
epeartree, line 274: nothing's wrong here but i don't get the change unless the point is to have absolute clarity into why we enter this block

epeartree (approved):
please have a look at that one comment.

Issue #6199, opened 2024-08-07
SQL exception when using _include search parameter


MariusMoga:
**Describe the bug**
We are getting an "ORDER BY items must appear in the select list if the statement contains a UNION, INTERSECT or EXCEPT operator." error when using the _include parameter with MS SQL Server after updating to 7.2.0.

**To Reproduce**
1. Add a Practitioner
2. Add a PractitionerRole and referance the new practitioner from it
3. Run the query: /PractitionerRole?_include=PractitionerRole:practitioner
4. This will result in a 500 response with this message:

`HAPI-2223: JDBC exception executing SQL [SELECT r.target_resource_id AS resource_id FROM hfj_res_link r  WHERE r.src_path = ? AND  r.target_resource_id IS NOT NULL AND  r.src_resource_id IN (?,?,?,?,?,?,?,?)  AND r.target_resource_type = ?  UNION SELECT rUri.res_id FROM hfj_res_link r  JOIN hfj_spidx_uri rUri ON (    rUri.hash_identity = ?   AND r.target_resource_url = rUri.sp_uri  ) WHERE r.src_path = ? AND  r.target_resource_id IS NULL AND  r.src_resource_id IN (?,?,?,?,?,?,?,?) order by @@version offset 0 rows fetch first ? rows only ] [ORDER BY items must appear in the select list if the statement contains a UNION, INTERSECT or EXCEPT operator.] [n/a]`

5. I can reproduce it with other referances as well like  task and encounter, encounter and patient etc.

**Environment (please complete the following information):**
 - HAPI FHIR 7.2.0
 - Sql server: Microsoft SQL Server 2022
 - OS: Windows 10

**Additional context**
Our previous HAPI FHIR version is 6.10.0 and we have not seen this issue.

XcrigX:
Good timing. We just discovered this problem yesterday as well. I see this bit:
https://github.com/hapifhir/hapi-fhir/blob/44c4b87a846c6889822be43ce4d4a506cce6fdb0/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java#L519-L537

, which was added by @jamesagnew for https://github.com/hapifhir/hapi-fhir/pull/5580, meant to fix a similar SQL Server issue. But the query in this case doesn't use a SELECT DISTINCT , rather a UNION so it's not catching it. 
[...]

PR #6228, opened 2024-08-22
Fix for #6199 - ensure order by clause is included in the sql string.

ddpardo:


jamesagnew:
Thanks for the contribution! We've just merged #6365 which also addresses this issue so it should be solved.

PR #6232, opened 2024-08-25
New junit parameter provider from static fields.

michaelabuckley:
Used to inject test cases into @MethodSource annotations.


epeartree (approved):
cool!

PR #6253, opened 2024-08-28
Improving performance, using caching when testing for primitives (#6252)

### LINKED ISSUE #6252: Extreme validation slowdown due to changed primitive type check implementation
**Describe the bug**
We noticed an extreme slowdown of resource validation in our project when switching to HAPI 7.4.0. Instead of minutes the validation takes hours. We profiled the application and found out that almost 100% of the time was spent in `VersionSpecificWorkerContextWrapper#isPrimitiveType(String)`. The code is always traversing all `StructureDefinition`s to check if any type that is primitive has a matching name.

**To Reproduce**
Validate a resource with lots of `StructureDefinition`s available to the ValidationContext, triggering `isPrimitive` call. The simplest is a CodeSystem, since constraint `csd-1` triggers a lot of those calls when evaluating the `isDistinct` function.

**Expected behavior**
The expected behavior would be a reasonable evaluation time of `isPrimitive(String)`.  
We would suggest to add a field of type `Set<String>` of names of primitive types to `VersionSpecificWorkerContextWrapper` to cache the names of all primitive types, similar to how it is done for all known `StructureDefinition`s already.

If it is wanted, we can work on a PR implementing the suggested solution.

 - HAPI FHIR Version 7.4.0

Boereck:
Caching primitive type names for faster lookup if a type is primitive.
Fixes #6252

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6252-improve-validator-performance.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6252-improve-validator-performance.yaml
@@ -1,0 +1,6 @@
+---
+type: perf
+issue: 6253
+title: "A cache has been added to the validation services layer which results
+  in improved validation performance. Thanks to Max Bureck for the
+  contribution!"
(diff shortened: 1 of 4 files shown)


jamesagnew:
@Boereck This is failing CI because of an android signature issue - We need to support android 26's specific weird hybrid of Java 11 and 17, which is annoying but usually not too hard.

In this case the failure appears to be `Collectors.toUnmodifiableSet()` which doesn't exist in A26 so you'll need to find another way of accomplishing that.

Boereck:
Thanks for the Info! I had a look at the Android API docs and updated the solution so to something that should (hopefully) work with API level 26. I also removed a call to `distinct()`, because I confused the API contracts of `Collectors.toMap()` and `Collectors.toSet()` before. Duplicate entries are not a problem for `Collectors.toSet()`.

Boereck:
@jamesagnew I added some test cases, since the covarage decreased a bit. Let me know if anything has to be improved.

jamesagnew (approved):
Looks good, thanks for the contribution!

Issue #5768, opened 2024-03-08
hapi-fhir-server 7.0.2 includes jakarta-incompatible simplejavamail dependency


jjh30:
**Describe the bug**
hapi-fhir-server 7.0.2 depends on org.simplejavamail:simple-java-mail:jar:6.6.1 which depends on com.sun.mail:jakarta.mail:jar:1.6.5

com.sun.mail:jakarta.mail:jar:1.6.5 is not jakarta-namespace compatible, and throws exceptions on healthcheck of our spring-boot 3.0.13 application. 

```
[[1;34mINFO[m] +- ca.uhn.hapi.fhir:hapi-fhir-server:jar:7.0.1:compile
[[1;34mINFO[m] |  +- com.googlecode.owasp-java-html-sanitizer:owasp-java-html-sanitizer:jar:20211018.2:compile
[[1;34mINFO[m] |  +- org.springframework:spring-messaging:jar:6.0.14:compile
[[1;34mINFO[m] |  +- org.simplejavamail:simple-java-mail:jar:6.6.1:compile
[[1;34mINFO[m] |  |  +- org.simplejavamail:core-module:jar:6.6.1:compile
[[1;34mINFO[m] |  |  |  +- com.sun.mail:jakarta.mail:jar:1.6.5:compile
[[1;34mINFO[m] |  |  |  \- com.github.bbottema:emailaddress-rfc2822:jar:2.2.0:compile
```

**To Reproduce**
1. build spring-boot 3.0.13 application with hapi-fhir-server 7.0.1, including spring-boot-actuate-mail
2. start application
3. hit healthcheck endpoint
4. observe error

**Expected behavior**
Happy healthcheck, no exception

**My workaround**
I excluded the simple-java-mail dependency from fhir-server in my pom.xml, and added an explicit dependency on the latest version of that library, 8.6.3. This allowed my healthchecks to succeed.

**Environment (please complete the following information):**
 - HAPI FHIR Version 7.0.0, 7.0.1, 7.02

**Additional context**
The current master https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-server/pom.xml still contains this dependency, and does not specify a version directly. 

**full stack trace**
```
jakarta.servlet.ServletException: Handler dispatch failed: java.util.ServiceConfigurationError: jakarta.mail.Provider: com.sun.mail.imap.IMAPProvider not a subtype
[...]

PR #6261, opened 2024-09-01
#5768 Upgrade to latest simple-java-mail and switch to jakarta.mail

Thopap:
Upgrading javasimplemail to most recent version which use jakarta.mail api's.
Greenmail was updated as well to provide compatiblity during testing.

Because jakarta.mail api's are also packaged within java simplemail, it need to be excluded by default to avoid duplicated classes on the classpath.

### DIFF
--- a/hapi-fhir-jpaserver-subscription/pom.xml
+++ b/hapi-fhir-jpaserver-subscription/pom.xml
@@ -70,6 +70,11 @@
 			<artifactId>jakarta.servlet-api</artifactId>
 			<scope>provided</scope>
 		</dependency>
+		<dependency>
+			<groupId>jakarta.mail</groupId>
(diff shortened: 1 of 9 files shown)


Thopap:
Any Chance zu get a review for my PR? (@jamesagnew or @tadgh ?)

tadgh:
Looks good, but this will require a version bump as it breaks the API. Since this is passing all tests, I'll take over from here to effect a version bump, and add a changelog. Are you OK with being credited as your github ID, real name, and organization?

 You can expect this to be shipped in our official november release (7.6.0)

Thopap:
@tadgh Thank you for considering this PR and performing the next steps for this. I am fine with mentioning my github ID and real name, but skip the organization.

tadgh (approved):

PR #6268, opened 2024-09-06
add batch and chunk ids to logging context

### LINKED ISSUE #6210: Delete Expunge Log Messages Don't Mention Chunk Id
**Describe the issue**
When performing delete expunge, the messages in the log say "INFO c.u.f.j.d.b.DeleteExpungeSvcImpl - 0 records deleted". But this is not helpful when debugging, since expunge is often done in batches with different chunks, so we can't match up which message is for which chunk.

jdar8:
Closes #6210


hapi-fhir-storage-batch2/src/test/java/ca/uhn/fhir/batch2/coordinator/WorkChannelMessageHandlerTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+		// Then
+		verify(myAppender, atLeastOnce()).doAppend(myLoggingEvent.capture());
+		myLoggingEvent.getAllValues()
+			.forEach(event -> {
+				Set<String> mdcPropertySetKeys = event.getMDCPropertyMap().keySet();
+				assertThat(mdcPropertySetKeys).containsExactlyInAnyOrder(BatchJobTracingContext.CHUNK_ID, BatchJobTracingContext.INSTANCE_ID);
+
+				Collection<String> mdcPropertySetValues = event.getMDCPropertyMap().values();
+				assertThat(mdcPropertySetValues).containsExactlyInAnyOrder(INSTANCE_ID, CHUNK_ID);
```
dotasek, line 74: This would be better if it checked that the values corresponded to the appropriate keys.

dotasek (changes requested):
Minor change to the test.

Is there existing documentation on the promised logging output that needs updating, or is logging in this case, 'you get what you get' and we've just enhanced it?

hapi-fhir-storage-batch2/src/test/java/ca/uhn/fhir/batch2/coordinator/WorkChannelMessageHandlerTest.java
jdar8: Thanks for the review!

I've changed the test to assert that the keys correspond to the appropriate values.

There are not any existing docs that I could find on the promised logging output. To address this, I've added a small blurb to the batch docs that the job instance ID and work chunk ID are now available through the logback context, and can be accessed using `logback.xml`.

dotasek (approved):
Looks good to me. Still requires someone with permissions for final review.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6210-add-chunk-id-to-delete-expunge-log-msg.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: add
+issue: 6210
+jira: SMILE-8428
+title: "The batch instance and chunk IDs have been added to the logging MDC so that the IDs can be automatically added to
```
tadgh, line 5:
```suggestion
title: "Batch instance ID and chunk ID have been added to the logging context so that they can be automatically added to
```

tadgh (approved):

PR #6272, opened 2024-09-09
Expose hapi-fhir-storage-cr config to downstream modules for testing

lukedegruchy:
- Move TestCrConfig, TestR4Config, and TestCqlProperties from the test to the main source folder.
- Carve out TestHapiFhirCrPartitionConfig from the configs above so tests can leverage its junit behaviour
- Add hapi-fhir-jpaserver-base to pom


hapi-fhir-storage-cr/pom.xml
```diff
@@ -145,6 +145,20 @@
 			<artifactId>jakarta.xml.bind-api</artifactId>
 		</dependency>
 
+		<!-- For test CR config that's exported to downstream modules -->
```
TipzCM, line 148: What's a CR config?

And if this is only for tests, why not put it in test-utils? or scope it only to tests (if needed for this project)?

TipzCM (approved):

hapi-fhir-storage-cr/pom.xml
lukedegruchy: As discussed, there's sort of a circular dependency within hapi-fhir-storage-cr, which would make it impossible to move to test-utilities.

JPercival (approved):

PR #6278, opened 2024-09-10
Improve auth interceptor operation handling

### LINKED ISSUE #6258: AuthorizationInterceptor no longer works on response of $document operation
**Describe the bug**
Starting with HAPI 7.4.0, the AuthorizationInterceptor seem to apply it's rules to the Bundle resource created by a $document operation instead of the resources contained in the response Bundle.
I believe this may be a regression caused by [Fix authorization handling for Bundle resources in the output #5953](https://github.com/hapifhir/hapi-fhir/pull/5953) (@codeforgreen)

I don't quite understand how it works, but `AuthorizationInterceptor#shouldExamineBundleChildResources` now returns false when the resource is not an instance of `IBaseParameters` or `IBaseBundle`. Previously, it returned true in the case of a $document operation on `Composition`.

**To Reproduce**
I use these rules: 

```kotlin
ruleBuilder
  .allow().operation().named("\$document").onInstance(resourceIdType).andRequireExplicitResponseAuthorization().andThen()
  .allow().read().instance(resourceIdType)
  .build()
```

**Expected behavior**
I expect the rules to be applied to the resources contained in the response Bundle, not to the Bundle itself.

 - HAPI FHIR Version 7.4.0

jamesagnew:
Fixes #6258


hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/OperationRule.java
```diff
@@ -138,10 +124,7 @@ public Verdict applyRule(
 					applies = true;
 				} else {
 					IIdType requestResourceId = null;
-					if (theInputResourceId != null) {
-						requestResourceId = theInputResourceId;
-					}
-					if (requestResourceId == null && myAllowAllResponses) {
+					if (theRequestDetails.getId() != null) {
```
codeforgreen, line 127: Is this an intentional change? It looks like `theInputResourceId` is not taken directly from `theRequestDetails.getId()`. e.g. I see another call from RuleOpImpl#applyRuleToTransaction with this `theInputResourceId` populated.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6258-improve-auth-interceptor-operation-handling.yaml
```diff
@@ -0,0 +1,10 @@
+---
+type: fix
+issue: 6258
+title: "The AuthorizationInterceptor handling for operations has been improved
+  so that operation rules now directly test the contents of response Bundle
+  or Parameters objects returned by the operation when configure to require
+  explicit response authorization. This fixes a regression in 7.4.0 where
```
codeforgreen, line 7: I don't see a change to the behaviour in the AuthorizationInterceptor itself. I suppose the previous change in 7.4.0 caused the regression by un-intentionally revealing another bug because it eliminated usage of checking the contents of a response (Bundle) resource based on `RequestDetails`. 

So it was like the decision was based on both request and response.  And it looks like you made sure the authorization checks happen separately. 

This is an interesting fix. Thank you for addressing it.

codeforgreen (approved):

codeforgreen:
I don't see a change to the behaviour in the AuthorizationInterceptor itself. I suppose the previous change in 7.4.0 caused the regression by un-intentionally revealing another bug because it eliminated usage of checking the contents of a response (Bundle) resource based on RequestDetails.

So it was like the decision was based on both request and response. And it looks like you made sure the authorization checks happen separately.

This is an interesting fix. Thank you for addressing it.

gijsbert802:
Any update on this? I see 1 test failed in the pipeline, but when I run that test locally it passes.

jamesagnew:
Apologies - Once this was merged, we saw a few downstream test failures, so I had to back this out while I try to figure out why.

Second attempt at merging this will be on #6382

jamesagnew:
Merging as this was reviewed as-is in #6278 (this was temporarily backed out due to a bad test in CDR)

PR #6290, opened 2024-09-16
Fix forced ID migration for MSSQL

delopst:
In MSSQL 2012, trim function does not exists. For this reason at least the MSSQL migration for version 7.4 is failing on MSSQL 2012.

Instead of trim function, LTRIM and RTRIM functins can be used to achive the same result.


tadgh (approved):
LGTM. Is it OK if I attribute you in our developers section as well as in the changelog? 
I can do it by: 
1. Github username
2. Real Name
3. Organization Name

Or any combination of the above! (or none if you'd rather not). Let me know!

delopst:
Hi. Yes, This is not my first contribution. So I am already listed as Primož Delopst. Can you please use my real name and the organization?

Additionally, this is the current state in pom.xml.

```
<developer>
   <id>delopst</id>
   <name>Primož Delopst</name>
</developer>
```
I am also part of the Better organization. Can you please add the `<organization>Better</organization>` tag?

tadgh:
https://github.com/hapifhir/hapi-fhir/pull/6309 <-- mind having a look over there and letting me know if it looks good?

delopst:
Looks fine.  Thank you :)

PR #6291, opened 2024-09-16
Making sure that we are not accidentally overwriting an existing reso…

jkiddo:
…urce with the same ID from another IG by comparing identifier, url or whatever constitutes uniqueness.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
@@ -497,9 +497,10 @@ public class PackageInstallerSvcImpl implements IPackageInstallerSvc {
 
 		IIdType id = theResource.getIdElement();
 		RequestDetails requestDetails = createRequestDetails();
+		String query = createSearchParameterMapFor(theResource).toNormalizedQueryString(myFhirContext);
 
 		try {
-			outcome = theDao.update(theResource, requestDetails);
+			outcome = theDao.update(theResource, query, requestDetails);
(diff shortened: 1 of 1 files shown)


jkiddo:
@codeforgreen can I persuade you to have a look at this?

jkiddo:
Ping @codeforgreen ?

codeforgreen:
> Ping @codeforgreen ?

HI @jkiddo what is this related to? The title of this issue does not ring a bell to me. Can you provide more details and link any related issue or pull request? Thanks.

jkiddo:
An issue I discovered in the starter project using the IG installer functionality. If you install multiple IGs with overlapping resources then the order of installing suddenly starts to play a part which it shouldnt. This PR makes sure that all resources are installed using conditional logic and stops that non deterministic behaviour. So yes, please review/comment/merge.

jkiddo:
@codeforgreen does the above provide the needed clarity?

dotasek (changes requested):
Where would `update(T theResource, String theMatchUrl, RequestDetails theRequestDetails);` actually throw the exceptions that you're queueing in the mock?

`theMatchUrl` is different for `oldSpec` and `newSpec`. Where is the logic that would throw this exception?

jkiddo:
@dotasek the error will be thrown in https://github.com/hapifhir/hapi-fhir/blob/8676cf2045375ee37d33f1978c6629fd92cd272a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoCodeSystem.java#L252

I've added this test here https://github.com/hapifhir/hapi-fhir/pull/6291/files#diff-aa258240972be5bb70e69c9bb6ff0a8cd9ea9a93815c0d1e79e4cf446986f789R273 that might better explain and reproduce it.

PR #6293, opened 2024-09-16
Bump org.hl7.fhir.core to 6.3.25

dotasek:
This bumps the org.hl7.fhir.core dependency to 6.3.25


dotasek:
There is a change to the org.hl7.fhir.core validator that is of note, and may require redesign of our own validation implementation. 

Internally, HAPI has logic that switches unknown code system validation issues to a specified level:

https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/UnknownCodeSystemWarningValidationSupport.java

The InstanceValidator now has similar internal logic that switches between reporting warnings and errors:

https://github.com/hapifhir/org.hl7.fhir.core/commit/7fc34e509ab221f21de41046b1728018adf8ea53

Since in many cases we throw warnings instead of errors, the processing of error messages and severity by the core validator logic is made even less straightforward. This is evidenced by the still correct, but less informative warnings created by some of our tests after this update: https://github.com/hapifhir/hapi-fhir/pull/6293/commits/df77506bf7eaffeb546779d8322a326bff4d45c5

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/FhirInstanceValidator.java
```diff
@@ -132,6 +133,14 @@ public void setBestPracticeWarningLevel(BestPracticeWarningLevel theBestPractice
 		myBestPracticeWarningLevel = theBestPracticeWarningLevel;
 	}
 
+	public void setUnknownSystemsCauseErrors(boolean theUnknownSystemsCauseErrors) {
```
jamesagnew, line 136: This should have some javadoc, including a `@since 7.6.0`

dotasek:
@jamesagnew I made a disturbing discovery about this today while trying to switch to using `setUnknownSystemsCauseErrors`

`UnknownCodeSystemWarningValidationSupport` is actually very interactive with InstanceValidator's `setUnknownSystemsCauseErrors`. The way our validation implementations currently work, one can't exist without the other, and 'setUnknownSystemsCauseErrors` must always be `true`.

I'm working on splitting `UnknownCodeSystemWarningValidationSupport` into `UnknownCodeSystemWarningValidationSupport` and `UnknownCodeSystemValidationSupport`, the latter of which will support  'setUnknownSystemsCauseErrors` being true or false.

If there is a need for a core update before HAPI's code freeze however, I suggest removing the exposure of `setUnknownSystemsCauseErrors` for the time being, and sticking with the passing configuration we have today.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/FhirInstanceValidator.java
dotasek: This exposure has been eliminated in https://github.com/hapifhir/hapi-fhir/pull/6293/commits/5faa571822a75ccf5b4c9e9acbbb9fd68661fb42

I'll follow up with a comment regarding why.

dotasek:
After discussion with @markiantorno it was decided to hold off exposing `setUnknownSystemsCauseErrors`. In after the latest commit, this setting will always be `true` for the instance validator. This preserves the expected errors (in severity, if not the exact message), passes tests, and marging this wouldn't block additional core version bumps, in contrast with having to work on a resolution to the entangled implementation.

If this is acceptable, this PR should be merged, and I will open a linked ticket that can be worked on independently, to re-expose `setUnknownSystemsCauseErrors` and deprecate `UnknownCodeSystemWarningValidationSupport`.

jamesagnew (approved):

PR #6324, opened 2024-09-26
6323 resource creation deadlock

JasonRoberts-smile:
The use of a synchronized map was causing threads to suspend in a client production system. The code has been rewritten so that reads from the map are concurrent, and we only synchronize if an update to the contents of the map is required.


michaelabuckley (approved):
This looks like a pure refactor, except for the locking.

Issue #6339, opened 2024-10-02
DeletePrimaryKey migration operation fails on postgres when not using 'public' schema.


adriennesox:
**Describe the bug**
Migrations that use the DeletePrimaryKey operation (which at present is just 7.4.0.20240618.6) fail on Postgres if the schema being used for HAPI is not 'public'.

```
2024-10-02 13:36:26.45 [main] INFO  c.u.f.j.m.HapiMigrator Executing 7.4.0.20240618.6 DropPrimaryKeyTask
2024-10-02 13:36:26.45 [main] INFO  c.u.f.j.m.t.DropPrimaryKeyTask 7.4.0.20240618.6: Going to DROP the PRIMARY KEY on table HFJ_RES_SEARCH_URL
2024-10-02 13:36:26.57 [main] ERROR c.u.f.j.m.HapiMigrator Migration failed
org.springframework.dao.EmptyResultDataAccessException: Incorrect result size: expected 1, actual 0
	at org.springframework.dao.support.DataAccessUtils.nullableSingleResult(DataAccessUtils.java:190)
	at org.springframework.jdbc.core.JdbcTemplate.queryForObject(JdbcTemplate.java:890)
	at org.springframework.jdbc.core.JdbcTemplate.queryForObject(JdbcTemplate.java:916)
	at ca.uhn.fhir.jpa.migrate.taskdef.DropPrimaryKeyTask.generateSql(DropPrimaryKeyTask.java:60)
	at ca.uhn.fhir.jpa.migrate.taskdef.DropPrimaryKeyTask.doExecute(DropPrimaryKeyTask.java:80)
	at ca.uhn.fhir.jpa.migrate.taskdef.BaseTask.execute(BaseTask.java:276)
	at ca.uhn.fhir.jpa.migrate.HapiMigrator.executeTask(HapiMigrator.java:202)
	at ca.uhn.fhir.jpa.migrate.HapiMigrator.migrate(HapiMigrator.java:171)
	at ca.uhn.fhir.cli.BaseFlywayMigrateDatabaseCommand.run(BaseFlywayMigrateDatabaseCommand.java:151)
	at ca.uhn.fhir.cli.HapiFlywayMigrateDatabaseCommand.run(HapiFlywayMigrateDatabaseCommand.java:55)
	at ca.uhn.fhir.cli.BaseApp.run(BaseApp.java:272)
	at ca.uhn.fhir.cli.App.main(App.java:42)
[...]

PR #6340, opened 2024-10-02
Fix Postgres migration when schema is not public.

adriennesox:
https://github.com/hapifhir/hapi-fhir/issues/6339

If the schema being used for HAPI in Postgres is not the `public` schema, the 7.4.0 migrations fails because the new `DeletePrimaryKeyTask` assumes the active schema is `public` when it looks up the primary key name.

This changes it to use `current_schema()` if available (i.e. the HAPI user has a search_path set) and `'public'` otherwise.

Tested by running the 7.4.2 migrations with this change incorporated against a database where the old migrations were failing.


tadgh:
Heya! Sorry it has taken me so long to get to this. Will review it today!

tadgh (approved):
Looks good to me! Looks like you're already attributed in the developers section, so I will go ahead and write a changelog for this after it merges 👍🏻

PR #6356, opened 2024-10-08
fix(BundleFetcher): correct API URL

mstempell:
Fix for the documentation example on complete bundle fetching, that reference an incorrect API URL


jamesagnew (approved):
Thanks for the contribution!

PR #6362, opened 2024-10-09
Ability to skip ConsentInterceptor logic selectively based on the resource

### LINKED ISSUE #6343: Ability to skip ConsentInterceptor logic selectively based on the evaluated resource
**Describe the issue**

Provide the ability to skip `ConsentInterceptor` logic selectively based on the evaluated resource. This feature is useful when a consent interceptor is added for a specific purpose and it incorporates potentially multiple consent services, all of which need to share the same approach about what filter to apply when selecting resources for which to apply their consent logic.
This could also been implemented with a pointcut at the consent interceptor level. Right now all the consent pointcuts are at consent service level. We don't have any other pointcuts for the consent interceptor itself so I think just declaring the filter via a method that can be overridden is the simplest approach. Feel free to suggest others.

 - HAPI FHIR Version 7.5.1

codeforgreen:
Resolves #6343


codeforgreen:
Changes were cancelled.

michaelabuckley:
We did something different.  Closing.

PR #6374, opened 2024-10-16
fixing bug with fulltext and jpa search filters

### LINKED ISSUE #6372: Search when combining fulltext search with jpa would sometimes return no results
When combining fulltext search filters with jpa search filters, sometimes data would not be returned.

Thsi happens when the fulltext search returns more results and requires the jpa filters to narrow the results.

But if the jpa filters do not return *any* data within the first 1600 results (800 is our default size for how many results we'll process before we break up a query), we would exit without finding any results.

TipzCM:
closes #6372

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6372-lucene-and-jpa-search-combine-issues.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6372-lucene-and-jpa-search-combine-issues.yaml
@@ -1,10 +1,14 @@
 ---
 type: fix
 issue: 6372
-title: "When Lucene (fulltext search) is combined with jpa search,
-        where Lucene would match more results and require jpa search
-        to narrow these down based on additional filters, sometimes
-        no results would be returned due to a bug in how we combined
-        search results from jpa and search results from Lucene.
+title: "Sometimes when combining fulltext search filters
+        with jpa search filters, no data would be returned,
+        even though data matching these filters definitely exists.
+        
+        This happened because we break up large queries
+        into sets of 800 and check these for results.
+        If none are found in the first one, or the second,
+        we would abandon the search claiming it's a failure.
+        
         This has now been fixed.
 "
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6372-lucene-and-jpa-search-combine-issues.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6372-lucene-and-jpa-search-combine-issues.yaml
@@ -1,0 +1,10 @@
+---
+type: fix
+issue: 6372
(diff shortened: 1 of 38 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6372-lucene-and-jpa-search-combine-issues.yaml
```diff
@@ -0,0 +1,14 @@
...
+title: "Sometimes when combining fulltext search filters
+        with jpa search filters, no data would be returned,
+        even though data matching these filters definitely exists.
+        
+        This happened because we break up large queries
+        into sets of 800 and check these for results.
+        If none are found in the first one, or the second,
+        we would abandon the search claiming it's a failure.
+        
+        This has now been fixed.
```
michaelabuckley, line 13: minor: can we tighten this up a bit?  When would we fail?  Is it like this?
```suggestion
title: "Searches that combined full-text searching (i.e. `_text` or `_content`) with other search parameters
         could fail to return all results if we encountered 800 matches against the full-text index where none of them match the rest of the query.
        This has now been fixed.
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -2343,15 +2344,23 @@ private void fetchNext() {
...
+						/*
+						 * Because we combine our DB searches with Lucene
+						 * sometimes we can have multiple results iterators
+						 * (with only some having data in them to extract).
+						 *
+						 * We'll iterate our results iterators until we
+						 * either run out of results iterators, or we
+						 * have one that actually has data in it.
+						 */
+						while (!myResultsIterator.hasNext() && !myQueryList.isEmpty()) {
```
michaelabuckley, line 2356: This looks ok.  I have to think there's a better way to do this, but this will backport well.

michaelabuckley (approved):
Nice.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6372-lucene-and-jpa-search-combine-issues.yaml
TipzCM: Updated;

With the minor change that i said "1600" (we used to check. the first 2 queries only, so first 1600 would be viewed)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
TipzCM: The big thing i was thinking about is that we might only have to do this for the combined queries.

But such a change might add levels of complication i wasn't comfortable with.

PR #6375, opened 2024-10-16
Add hash identity token mode

jamesagnew:
This PR adds an experimental new feature to include the HASH_IDENTITY column as a predicate whenever searches against the SPIDX_TOKEN table are performed (for positive searches - does not cover :not searches, etc)


michaelabuckley (approved):
Looks good.

Issue #6394, opened 2024-10-22
Decrease delay caused by async processing of Subscriptions


SevcikMichal:
**Describe the issue**
In HAPI FHIR v6.10.0 there was a change where subscriptions are processed asynchronously. This is causing a delay that is based on the configured schedule time of the scheduled job. The default is 5s so if you hit the window just right you can wait the whole 5s before your subscription is activated and also to receive any messages on the subscription. I can confirm this by creating a simple code that:  

1. Creates subscription in state Requested.  
2. Wait for the subscription to get in state Active.  
3. Bind the subscription (web socket in my case). 
4. Submit resources relevant to the subscription.  
5.  Wait to receive ping from the web socket.  

With the default delay it can take up to 10s to get the whole round trip. 5s to get the subscription to state active, 5s to receive notification on the web socket.

**Environment:**
 - HAPI FHIR Version v6.10.0 and above
 - OS: Docker (gcr.io/distroless/java17-debian12:nonroot) run on MacBook Air M1
 - Browser N/A

**Additional context**

I am testing this using the hapi-fhir-jpaserver-starter.
I still have limited understanding of the whole code base but so far I came to two possible solutions.
Both of them improved the performance from 10s to around 2-3s.

1. Override the bean in the server-starter and configure the delay to be lower (5s -> 1s).
Pro: Simple to do and already supported.
Con: Increased load on DB.

2. Try to submit the processed messaged synchronously if it fails store it to the DB and let the job take over.
Pro: No extra load on DB generated.
Con: Due to my lack of overall understanding of code base may have unforeseen side effects.

I created a PR  for the point 2. and would like to hear some feedback if it would be possible to submit the message sooner.
Also if there are any solutions that I am missing at the moment, everything is wellcome.

PR #6395, opened 2024-10-22
Configurable submit ResourceModifiedMessage immediately if of type Subscription, store in DB on failure

SevcikMichal:
Trying to address #6394.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6395-allow-submit-subscription-synchronously.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6395-allow-submit-subscription-synchronously.yaml
@@ -1,0 +1,7 @@
+---
+type: perf
+issue: 6395
+title: "A new configuration option has been added to `SubsciptionSubmitterConfig` which
+  causes Subscription resources to be submitted to the processing queue synchronously
+  instead of asynchronously as all other resources are. This is useful for cases where
+  subscriptions need to be activated quickly. Thanks to Michal Sevcik for the contribution!"
--- a/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/config/BaseSubscriptionSettings.java
+++ b/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/config/BaseSubscriptionSettings.java
@@ -50,6 +50,14 @@ public abstract class BaseSubscriptionSettings {
 	private boolean myAllowOnlyInMemorySubscriptions = false;
 
 	/**
+	 * If this is enabled (default is {@literal false}), changes to Subscription resource would be put on queue immediately.
+	 * Reducing delay between creation of the Subscription and Activation.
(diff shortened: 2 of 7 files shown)


jamesagnew:
The async subscription processing was added for several reasons:

- It avoids issues where the database commit fails but the message has already been placed on the subscription processing queue, so subscription processing proceeds on a resource that was never actually written to the DB successfully
- It improves scalability by batching submissions to the processing queue, which is critical for very large throughput systems

This PR undoes both of those benefits, and can't be merged as such.

Looking at your ticket, I guess part of the issue is that subscriptions take a long time to activate. Certainly I could see a world where we put changes to the Subscription resource itself on the queue immediately since subscriptions themselves are typically pretty low volume (and the activator loads from the DB I'm pretty sure, so stale data shouldn't be an issue) - this would solve that issue, although it would still leave the delays associated with resources.

I'd assume something like that, combined with a smaller polling delay (1s?) could largely remove the delays you're experiencing. If you wanted to proceed with implementing such a change, please make sure that it's configuable (via JpaStorageSettings) and includes comprehensive unit tests.

SevcikMichal:
> The async subscription processing was added for several reasons:
> 
> * It avoids issues where the database commit fails but the message has already been placed on the subscription processing queue, so subscription processing proceeds on a resource that was never actually written to the DB successfully
> * It improves scalability by batching submissions to the processing queue, which is critical for very large throughput systems
> 
> This PR undoes both of those benefits, and can't be merged as such.
> 
> Looking at your ticket, I guess part of the issue is that subscriptions take a long time to activate. Certainly I could see a world where we put changes to the Subscription resource itself on the queue immediately since subscriptions themselves are typically pretty low volume (and the activator loads from the DB I'm pretty sure, so stale data shouldn't be an issue) - this would solve that issue, although it would still leave the delays associated with resources.
> 
> I'd assume something like that, combined with a smaller polling delay (1s?) could largely remove the delays you're experiencing. If you wanted to proceed with implementing such a change, please make sure that it's configuable (via JpaStorageSettings) and includes comprehensive unit tests.

Thanks for the explanation.
And yes even shaving off the few seconds from the subscription to get to activated would help.
Do you mean something like checking for resource type there and only submit it immediately if it's of type subscription? Condition also with a configuration in JpaStorageSettings?

I will try to implement this and provide test as well.

SevcikMichal:
@jamesagnew I reworked it so that only subscription resource would be submitted immediately and also only if configured that way.
I moved the configuration to the SubscriptionConfiguration as this is all relevant to subscriptions does that make sense?
Would you be able to briefly look at the changes if this makes sense?
I will start implementing the tests for this.

SevcikMichal:
Hi @jamesagnew,
I added a unit test testing the logic to submit Subscription immediately if configured to do so.

Let me know if there is anything else missing.

jamesagnew (approved):
Looks good to me! Thanks for the contribution!

jamesagnew:
This is now merged, thanks again for the contribution. FYI this has unfortunately missed the code freeze for our November 7.6.0 release, so it will be available as GA in the February release (probably called 8.0.0). Snapshot builds should start to appear by tomorrow though.

PR #6399, opened 2024-10-24
Fix tests about elementsSubset

SijieMei:
### Issue
Following Tests all used method "elementsSubset" of class "GenericClient". ElementsSubset used a HashSet to store URL queries. Since HashSet doesn't guarantee the orders of entries. The URL GenericClient prints will have different String in different times.

### Change
I found some tests have the handler for different output String but some test doesn't. I added the handler for these tests missing it.


jamesagnew (approved):
Thanks for the contribution! Test stability is always a good thing.

fil512:
Yes thank you so much! Intermittent test failures slow down everybody. So happy for this contribution!

PR #6402, opened 2024-10-24
Contained bug

### LINKED ISSUE #6403: Terser has odd behaviour
Currently, if you create a bundle with two resources, each with an identical contained resource, the JSON Parser will produce a garbled response, such as the following: 
```
{
	"resourceType": "Observation",
	"id": "1348",
	"contained": [
		{
			"resourceType": "Practitioner",
			"id": "1"
		}
	],
	"status": "final",
	"effectiveDateTime": "2024-10-11T15:24:00-06:00",
	"issued": "2024-10-11T15:24:00-07:00",
	"performer": [
		{
			"reference": "#1"
		}
	],
	"valueString": "0.032",
	"specimen": {
		"reference": "#1"
	}
}
```

tadgh:
- Test
- modify algorithm for server-assigned contained IDs.
- Process contained resources which have a client-assigned ID before those that do not in a resource. 
- moved ContainedResources from the Parser, to the EncodeContext object, so we get one per resource processed. 
Closes #6403

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
@@ -84,20 +84,22 @@ public class FhirTerser {
 
 	private static final Pattern COMPARTMENT_MATCHER_PATH =
 			Pattern.compile("([a-zA-Z.]+)\\.where\\(resolve\\(\\) is ([a-zA-Z]+)\\)");
+
 	private static final String USER_DATA_KEY_CONTAIN_RESOURCES_COMPLETED =
 			FhirTerser.class.getName() + "_CONTAIN_RESOURCES_COMPLETED";
+
 	private final FhirContext myContext;
 
 	/**
 	 * This comparator sorts IBaseReferences, and places any that are missing an ID at the end. Those with an ID go to the front.
 	 */
 	private static final Comparator<IBaseReference> REFERENCES_WITH_IDS_FIRST =
-		Comparator.nullsLast(Comparator.comparing(ref -> {
-			if (ref.getResource() == null) return true;
-			if (ref.getResource().getIdElement() == null) return true;
-			if (ref.getResource().getIdElement().getValue() == null) return true;
-			return false;
-		}));
+			Comparator.nullsLast(Comparator.comparing(ref -> {
+				if (ref.getResource() == null) return true;
+				if (ref.getResource().getIdElement() == null) return true;
+				if (ref.getResource().getIdElement().getValue() == null) return true;
+				return false;
+			}));
 
 	public FhirTerser(FhirContext theContext) {
 		super();
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
@@ -1803,7 +1803,6 @@ public class FhirTerser {
 
 	public static class ContainedResources {
 		private long myNextContainedId = 1;
-
 		private List<IBaseResource> myResourceList;
 		private IdentityHashMap<IBaseResource, IIdType> myResourceToIdMap;
 		private Map<String, IBaseResource> myExistingIdToContainedResourceMap;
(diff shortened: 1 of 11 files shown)


hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/parser/JsonParserR4Test.java
```diff
@@ -261,9 +265,83 @@ public void testDuplicateContainedResourcesNotOutputtedTwice() {
 
 		idx = encoded.indexOf("\"Medication\"", idx + 1);
 		assertEquals(-1, idx);
+	}
+
```
epeartree, line 269: ouch! 
what a great test.  once u get it, u also understand that the implementation is correct

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/parser/JsonParserR4Test.java
```diff
@@ -261,9 +265,83 @@ public void testDuplicateContainedResourcesNotOutputtedTwice() {
...
+			Practitioner practitioner = new Practitioner();
+			DiagnosticReport report = new DiagnosticReport();
+			report.addSpecimen(new Reference(specimen));
+
+			Observation obs = new Observation();
+			//When: The practitioner (which is parsed first, has an assigned id that will collide with auto-assigned
+			practitioner.setId("#1");
+			obs.addPerformer(new Reference(practitioner));
+			obs.setSpecimen(new Reference(specimen));
+
```
epeartree, line 314: extra line

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/parser/JsonParserR4Test.java
```diff
@@ -261,9 +265,83 @@ public void testDuplicateContainedResourcesNotOutputtedTwice() {
...
+			DiagnosticReport report = new DiagnosticReport();
+			report.addSpecimen(new Reference(specimen));
+
+			Observation obs = new Observation();
+
+			//When: The specimen (which is parsed second, has an assigned id that will collide with auto-assigned practitioner
+			specimen.setId("#1");
+			obs.addPerformer(new Reference(practitioner));
+			obs.setSpecimen(new Reference(specimen));
+
```
epeartree, line 335: extra line

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ContainedTest.java
```diff
@@ -46,6 +49,33 @@ public void after() throws Exception {
...
+		observation.getPerformerFirstRep().setReference("#1");
+		observation.getSpecimen().setReference("#2");
+		observation.setSubject(new Reference("Patient/test-patient"));
+
+		observation.getCode().getCodingFirstRep().setSystem("LN").setCode("NV");
+		observation.setValue(new StringType("abc123"));
+
+		System.out.println(myFhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(observation));
+		DaoMethodOutcome daoMethodOutcome = myObservationDao.create(observation, mySrd);
+
```
epeartree, line 76: this test has no assertion unless the test is about being able to create or encodeToString an Observation without raising an exception.

epeartree (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
```diff
@@ -1415,33 +1416,41 @@ public boolean acceptUndeclaredExtension(
 		});
 	}
 
+	private Integer hasId(IBaseReference ref) {
```
jamesagnew, line 1419: It's kinda confusing having a method in java called `hasXXX` that returns an int.... Recommend perhaps renaming it to `hasIdForComparison` or something like that. Also it can presumably be a static method, which is a tiny performance boost (important for fundamental stuff like parser code).

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
```diff
@@ -1415,33 +1416,41 @@ public boolean acceptUndeclaredExtension(
...
-						theContained.addContained(next.getReferenceElement(), potentialTarget);
-						containResourcesForEncoding(theContained, potentialTarget, theModifyResource);
-					}
-				}
-			}
-		}
 
+		//Note that we process all contained resources that have arrived here with an ID contained resources first, so that we dont accidentally auto-assign an ID
+		//which may collide with a resource we have yet to process.
+		allReferences = allReferences.stream()
```
jamesagnew, line 1432: Two more suggested performance tweaks (this is such heavily used code that we should be as performant as possible):

* Just sort the `allReferences` list directly instead of doing this stream thing - no sense creating a new list object every time when the list we're already working on is fresh for every call to this method
* Extract the comparator out into a static constant instead of creating a new one each time - Probably the JIT does this automatically, but it doesn't hurt to make it explicit

jamesagnew (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
jamesagnew: Actually - just looked it up and boolean is comparable just like int is. So you could probably actually just make this return boolean.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ContainedTest.java
tadgh: Ah this whole test should be removed , it was a first try before I found a better spot

tadgh: Ah this whole test should be removed , it was a first try before I found a better spot

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
tadgh: Both good calls, both done!

PR #6416, opened 2024-10-29
Update years on changelog

jkiddo:


dotasek (approved):
Looks good to me. Screenshots are from my local build and browser. 

One thing to note, which is out of scope for this PR, is that the `changelog('2025', '')` function appears to be catching any stray pre-release: HAPI FHIR 6.10.5 (Prerelease),
HAPI FHIR 5.5.0 (Prerelease)

This appears to happen in the previous version as well, so I think this should be treated by a different PR. 

Screenshot 2025-03-05 at 1 07 21 PM
Screenshot 2025-03-05 at 1 07 31 PM
Screenshot 2025-03-05 at 1 07 39 PM

Previous version:

Screenshot 2025-03-05 at 1 08 13 PM

jmarchionatto (approved):

epeartree (approved):

PR #6360, opened 2024-10-08
Add support for response parameter issues for `validate-code` operation

codeforgreen:
Fixes #6359 
Workaround for https://github.com/hapifhir/org.hl7.fhir.core/issues/1766

The fix only includes changes for Remote Terminology validator. 
I checked all the other validators we have and we seem to have implemented a similar approach by explicitly adding issue to the response of validateCode call (e.g. `InMemoryTerminologyServerValidationSupport`, `CommonCodeSystemsTerminologyService`, `TermReadSvcImpl`) which was implemented when the upgrade of the `org.hl7.fhir.core` library was done in #5772.

You will notice test `testValidateResourceContainingLoincCode` has boolean parameter added which controls the output of the mocked `validateCode` response. If the error response has a message, the FHIR Validator will not act on it unless it has issues populated as well. For Remote Terminology specifically a workaround has been implemented.

PR #6423, opened 2024-10-30
Fixes for the translation of parameter issues as part of the output for $validate-code operation.

### LINKED ISSUE #6422: The binding strength is not considered when validating resources against a profile
**Describe the bug**
The changes in #6360 introduced a bug because the issue detail code for Remote Terminology `$validate-code` calls was set to always be `invalid-code`. 

The org.hl7.fhir.core, more specifically org.hl7.fhir.validation library expects `invalid-code` to be used only for `CodeSystem` `$validate-code` and is expecting to see codes like `not-in-vs` for `ValueSet` `$validate-code` in cases when a code is not found in an existing ValueSet. This happens in `org.hl7.fhir.validation.instance.InstanceValidator#processTxIssues`.
When trying to use a profile where a binding for an element against a `ValueSet` is extensible, the validate operation will return an error which is not expected in this case because the code does exist in the extended profile binding.

See: https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java#L1718

**To Reproduce**
1. Create a custom profile for Encounter resources and store it to a hapi-fhir repository:
[profile-encounter.json](https://github.com/user-attachments/files/17577230/profile-encounter.json)
2. Setup a hapi-fhir client with a validator that includes an instance of `RemoteTerminologyValidationSupportService` with `https://r4.ontoserver.csiro.au/fhir/`.

e.g.
```
String remoteTerminologyServerUrl = "https://r4.ontoserver.csiro.au/fhir/";
FhirContext fhirCtx = FhirContext.forR4Cached();

FhirInstanceValidator fhirInstanceValidator = new FhirInstanceValidator(fhirCtx);
ValidationSupportChain validationSupport = new ValidationSupportChain(
	new DefaultProfileValidationSupport(fhirCtx),
	new RemoteTerminologyValidationSupportService(fhirCtx, remoteTerminologyServerUrl);
);

validator.setValidationSupport(validationSupport);
FhirValidator fhirValidator = fhirCtx.newValidator();
fhirValidator.registerValidatorModule(fhirInstanceValidator);
```
4. Create an encounter resource instance with an identifier that should be valid according to the profile.
e.g.
```
Encounter encounter = new Encounter();
encounter.getMeta().addProfile("http://example.ca/fhir/StructureDefinition/profile-encounter");
encounter.setStatus(Encounter.EncounterStatus.PLANNED);
encounter.addIdentifier()
	.getType().addCoding()
	.setSystem("http://terminology.hl7.org/CodeSystem/v2-0203")
	.setCode("VN")
	.setDisplay("Visit number");
```
6. Validate the resource (runtime resource, no need to store it in the hapi-fhir repository) and check the errors which relate to code `VN`
```
IGenericClient fhirClient = fhirCtx.newRestfulGenericClient("http://localhost:8000");

MethodOutcome resultProcedure = fhirClient.validate().resource(theResource).execute();
OperationOutcome operationOutcome = (OperationOutcome) resultProcedure.getOperationOutcome();
String output = fhirCtx.newJsonParser().setPrettyPrint(true).encodeResourceToString(operationOutcome);
```
7. You will notice the following issue with severity error:
`The provided code 'http://terminology.hl7.org/CodeSystem/v2-0203#VN' was not found in the value set 'http://hl7.org/fhir/ValueSet/identifier-type|5.0.0-ballot'`

**Expected behaviour**
No error should be returned since the profile `http://hl7.org/fhir/StructureDefinition/Identifier` which is in the hierarchy states that element `Identifier.type` has an extensible binding and the code `VN` exists in `CodeSystem`  `http://terminology.hl7.org/CodeSystem/v2-0203` which is the extended binding in the profile created at step 1. 

 - HAPI FHIR Version 7.4.4

codeforgreen:
Fixes #6422

* The main fix was to avoid the hard-coded translation of issue detail codes for `IValidationSupport.CodeValidationIssue`. These changes are more of a part 2 for changes made in #6360. The problematic code there was:
https://github.com/hapifhir/hapi-fhir/pull/6360/files#diff-003fe663a54000163c7b18239f3c2be278ed03548006676e595ab5d1d932c80dR739
where code 'invalid-code' was always being used. It seems that it is not he only code which can be passed.

* The changes in this PR are to ensure (1) the structure defined in `IValidationSupport.CodeValidationIssue` looks exactly like the FHIR spec and (2) the codes within that are translated in a generic way.

* The main changes for this PR are in `RemoteTerminologyServiceValidationSupport` and `VersionSpecificWorkerContextWrapper`. 

* Made some changes to enable test setup for hapi-fhir validation ($validate resource) use-cases with Remote Terminology which was not possible with other test step. Previous tests were only testing one single `$validate-code` call while the `$validate` operation against a resource can imply multiple such calls, as many or similar to the amount of elements which appear in the linked profile. See `ValidateWithRemoteTerminologyTest`.

* While I was writing/debugging new tests, I found a bug in `VersionSpecificWorkerContextWrapper#validateCodeInValueSet`. This method will invoke `validateCode` for `CodeSystem` even in cases where `system` is null. That will happen for cases where the entry point method into this class does not have `system` and method `ValidationSupportUtils.extractCodeSystemForCode` is unable to figure it out. That can happen for standard bindings such as `Observation.status` which are declared as enums and don't require explicit mention of the `system` in the resource. I've added a null check to prevent that call since it provides no value. I've added a comment to method `ValidationSupportUtils.extractCodeSystemForCode` as I believe it should not be widely used and as it will return incorrect results in some cases.


hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/IValidateCodeTest.java
```diff
@@ -316,14 +317,22 @@ default void validateCode_withValueSetErrorWithIssues_returnsCorrectly() {
...
+			assertEquals(expectedIssue.getDetails().getText(), actualIssue.getDetails().getText());
+			assertEquals(expectedIssue.getDetails().getCodings().size(), actualIssue.getDetails().getCodings().size());
+			for (int index = 0; index < expectedIssue.getDetails().getCodings().size(); index++) {
+				IValidationSupport.CodeValidationIssueCoding expectedCoding = expectedIssue.getDetails().getCodings().get(index);
+				IValidationSupport.CodeValidationIssueCoding actualCoding = actualIssue.getDetails().getCodings().get(index);
+				assertEquals(expectedCoding.getSystem(), actualCoding.getSystem());
+				assertEquals(expectedCoding.getCode(), actualCoding.getCode());
+			}
+			assertEquals(expectedIssue.getDetails().getText(), actualIssue.getDetails().getText());
+			assertEquals(expectedIssue.getDiagnostics(), actualIssue.getDiagnostics());
```
dotasek, line 335: I put a breakpoint here with the condition `expectedIssue.getDiagnostics() != null` to verify that we're checking that diagnostics are set in some test. My point wasn't hit.

dotasek (changes requested):
I encountered a NullPointer when executing RemoteTerminologyValidateCodeDstu3Test and RemoteTerminologyValidateCodeR4Test. I believe this code is missing:

```java
class CodeValidationIssueCodeableConcept {
		//...
		
		// intentionally not exposed
		void addCoding(CodeValidationIssueCoding theCoding) {
			if (myCodings == null) {
				myCodings = new ArrayList<>();
			}
			getCodings().add(theCoding);
		}
```

Changes look good otherwise.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -440,74 +440,207 @@ default String getName() {
...
-	enum CodeValidationIssueCode {
-		NOT_FOUND,
-		CODE_INVALID,
-		INVALID,
-		OTHER
+	/**
+	 * Defines codes in system <a href="http://hl7.org/fhir/issue-type">http://hl7.org/fhir/issue-type</a>.
+	 */
+	/* this enum would not be needed if we design/refactor to use org.hl7.fhir.r5.terminologies.utilities.ValidationResult */
+	class CodeValidationIssueTypeCode {
```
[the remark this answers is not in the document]
codeforgreen: I left the ones that are related to validation which were there in the code before as well. I can add a comment with that. I guess these are also the only codes we internally use to build issues for internal validators. If it's remote terminology any code will work because it will be using the generic constructor. 

We are currently permissive when it comes to issue type (required binding) and details codes (example binding), not severity code (required binding) because that is small enough. The part I am not sure about is checking nulls for severity and issue code, which have cardinality of 1 according to the profile.

When the validateCode call returns the response  to the `InstanceValidator` we can find this in method  `InstanceValidator#txIssue`:

```
ValidationMessage.IssueType code = IssueType.fromCode(issue.getCode().toCode());
ValidationMessage.IssueSeverity severity = IssueSeverity.fromCode(issue.getSeverity().toCode());
```

This will throw exceptions if the codes are not in those enums which seem to have all the elements.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/IValidateCodeTest.java
codeforgreen: Agree. We currently have it in our internal validators but we should not be replying on tests for those. I will add more for that here (diagnostics populated but not issue detail text).

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/IValidateCodeTest.java
dotasek, line 1: While reviewing other parts of this, I noticed that there are a few places where validation ignores the outcome code:

```java
// assertEquals(CODE, outcome.getCode());
```

This isn't part of this review, but is related. Is this a byproduct of the testing utilities, or something that will show up in validation everywhere?

dotasek (approved):
See comment in file.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/IValidateCodeTest.java
dotasek: On second review, I can see a few places we actually check the diagnostics field:

https://github.com/hapifhir/hapi-fhir/blob/a35383442cd93853b83657aff46099086ec01b53/hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/IValidateCodeTest.java#L217
https://github.com/hapifhir/hapi-fhir/blob/a35383442cd93853b83657aff46099086ec01b53/hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/IValidateCodeTest.java#L269

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/IValidateCodeTest.java
codeforgreen: It is because the boolean result outcome for validateCode (succes or not) in object `IValidationSupport.CodeValidationResult` is dictated by `isOk` method which has condition that code != null to determine success. I mentioned that in the previous PR as well. https://github.com/hapifhir/hapi-fhir/pull/6360/files#diff-003fe663a54000163c7b18239f3c2be278ed03548006676e595ab5d1d932c80dR652
That would require some more refactoring changes to update.
I also mentioned it in this defect (which I also found) so it's not forgotten: https://gitlab.com/simpatico.ai/cdr/-/issues/6708#solution-design

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/IValidateCodeTest.java
codeforgreen: Yes these checks were there but after I introduced the renamed getter (`getDiagnostics` in addition to `getMessage`) it became more obvious, because getMessage existed in the parent `CodeValidationResult` object as well. I kept all old methods and marked them as deprecated after introducing the alternative to avoid any breakage.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -440,74 +440,236 @@ default String getName() {
...
+		 */
+		public String getCode() {
+			return myCode;
+		}
+		/**
+		 * Creates a {@link IssueSeverity} object from the given code.
+		 * @return the {@link IssueSeverity}
+		 */
+		public static IssueSeverity fromCode(String theCode) {
+			return Arrays.stream(values())
```
jamesagnew, line 487: This seems kinda inefficient for something that is potentially called quite a bit... I doubt we'll ever add any more values to this enum, could we just put a simple switch statement here instead of looping through the enum values??

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -440,74 +440,236 @@ default String getName() {
...
 		public CodeValidationIssueCode getCode() {
+			return getType();
+		}
+
+		public CodeValidationIssueCode getType() {
 			return myCode;
 		}
 
+		/**
+		 * Deprecated. Please use {@link #getDetails()} instead. That has support for multiple codings.
```
jamesagnew, line 666:
```suggestion
		 * @deprecated Please use {@link #getDetails()} instead. That has support for multiple codings.
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -771,23 +937,48 @@ public CodeValidationResult setSeverity(IssueSeverity theSeverity) {
 			return this;
 		}
 
+		/**
+		 * Deprecated. Please use method {@link #getIssues()} instead.
```
jamesagnew, line 954:
```suggestion
		 * @deprecated Please use method {@link #getIssues()} instead.
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -771,23 +937,48 @@ public CodeValidationResult setSeverity(IssueSeverity theSeverity) {
...
 		public List<CodeValidationIssue> getCodeValidationIssues() {
-			if (myCodeValidationIssues == null) {
-				myCodeValidationIssues = new ArrayList<>();
-			}
-			return myCodeValidationIssues;
+			return getIssues();
 		}
 
+		/**
+		 * Deprecated. Please use method {@link #setIssues(List)} instead.
```
jamesagnew, line 962:
```suggestion
		 * @deprecated Please use method {@link #setIssues(List)} instead.
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -771,23 +937,48 @@ public CodeValidationResult setSeverity(IssueSeverity theSeverity) {
...
+		 */
+		@Deprecated(since = "7.4.6")
 		public CodeValidationResult setCodeValidationIssues(List<CodeValidationIssue> theCodeValidationIssues) {
-			myCodeValidationIssues = new ArrayList<>(theCodeValidationIssues);
-			return this;
+			return setIssues(theCodeValidationIssues);
 		}
 
+		/**
+		 * Deprecated. Please use method {@link #addIssue(CodeValidationIssue)} instead.
```
jamesagnew, line 970:
```suggestion
		 * @deprecated Please use method {@link #addIssue(CodeValidationIssue)} instead.
```

jamesagnew (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
codeforgreen: Oh, I think this is unused, I'll get rid of it.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -440,74 +440,207 @@ default String getName() {
...
 	}
 
-	enum CodeValidationIssueCoding {
-		VS_INVALID,
-		NOT_FOUND,
-		NOT_IN_VS,
+	/**
+	 * Defines codes in system <a href="http://hl7.org/fhir/tools/CodeSystem/tx-issue-type">http://hl7.org/fhir/tools/CodeSystem/tx-issue-type</a>.
+	 */
+	class CodeValidationIssueCoding {
```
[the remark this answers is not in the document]
codeforgreen: Because the binding strength is example in the spec and `http://hl7.org/fhir/tools/CodeSystem/tx-issue-type` is not required or even preferred at the moment. Even a custom system can be used which I have seen happening in case of a hapi-fhir client.

PR #6432, opened 2024-11-04
Fix wording of _include in docs

vadi2:
The `_include` parameter was incorrectly documented compared to the spec.


jamesagnew (approved):
Thanks for the contribution!!

PR #6435, opened 2024-11-04
Jd 20241030 fix fulltext searches with accurate total

### LINKED ISSUE #6434: Total is incorrect when using fulltext in combination with regular SearchParameter
When performing a search that uses `_total=accurate` that intersects a JPA search and a fulltext search, the total returned is incorrect. 

It turns out that we actually don't support this. We should block \`\_total=accurate\` on queries that intersect JPA and Fulltext, as it is impossible to calculate the true total. It should return a 400 with an explanation error.

jdar8:
closes #6434


jdar8:
Closing due to more design work needed. Delayed to next release.

PR #6436, opened 2024-11-04
 fix interceptor hooks from requestDetails not getting called for STORAGE_PRECHECK_FOR_CACHED_SEARCH

mrdnctrk:
In hapi fhir server, there are 2 InterceptorBroadcaster instances that could be used to call hooks when handling a request. The first Broadcaster is the one that could be associated JPA module, and the second broadcaster is the one that is optionally set in RequestDetails. 

- One issue I am fixing in this MR is that for pointcuts that return boolean types, the return value of the hooks associated with the Broadcaster set in RequestDetails were not taken into account. 

- The second issue I am fixing is specifically for `STORAGE_PRECHECK_FOR_CACHED_SEARCH` pointcut. This Pointcut returns boolean type. For pointcuts that return boolean type , we should be calling `doCallHooks` method instead of calling `doCallHooksAndReturnObject`, but that wasn't the case for `STORAGE_PRECHECK_FOR_CACHED_SEARCH`. The problem with calling `doCallHooksAndReturnObject` instead of `doCallHooks` is that   `doCallHooksAndReturnObject` function doesn't call the hooks for the Broadcaster set in the RequestDetails if a hook associated with the JPA module returns a non-null value: [see null check here](https://github.com/hapifhir/hapi-fhir/blob/rel_7_6/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/util/CompositeInterceptorBroadcaster.java#L97-L97). Since a boolean returning hook always returns a non-null value (true or false is not equal to null), then the Broadcaster set in the requestDetails is never called. 
So in this MR, I made a change to use doCallHooks when invoking `STORAGE_PRECHECK_FOR_CACHED_SEARCH` pointcut, so that the broadcaster set in the request details is invoked as well. 

closes: #6440

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6440-fix-hooks-not-called-for-precheck-for-cached-search.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6440-fix-hooks-not-called-for-precheck-for-cached-search.yaml
@@ -1,0 +1,8 @@
+---
+type: fix
+issue: 6440
+title: "Previously, if an IInterceptorBroadcaster was set in a RequestDetails object, 
+`STORAGE_PRECHECK_FOR_CACHED_SEARCH` hooks that were registered to that IInterceptorBroadcaster were not 
+called. Also, if an IInterceptorBroadcaster was set in the RequestDetails object, the boolean return value of the hooks 
+registered to InterceptorBroadcaster were not taken into account. This second issue existed for all pointcuts that 
+returned a boolean type, not just for `STORAGE_PRECHECK_FOR_CACHED_SEARCH`. These issues have now been fixed."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6440-fix-hooks-not-called-for-precheck-for-cached-search.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6440-fix-hooks-not-called-for-precheck-for-cached-search.yaml
@@ -1,8 +1,8 @@
 ---
 type: fix
 issue: 6440
-title: "Previously, if an IInterceptorBroadcaster was set in a RequestDetails object, 
-`STORAGE_PRECHECK_FOR_CACHED_SEARCH` hooks that were registered to that IInterceptorBroadcaster were not 
-called. Also, if an IInterceptorBroadcaster was set in the RequestDetails object, the boolean return value of the hooks 
-registered to InterceptorBroadcaster were not taken into account. This second issue existed for all pointcuts that 
-returned a boolean type, not just for `STORAGE_PRECHECK_FOR_CACHED_SEARCH`. These issues have now been fixed."
+title: "Previously, if an `IInterceptorBroadcaster` was set in a `RequestDetails` object,
+`STORAGE_PRECHECK_FOR_CACHED_SEARCH` hooks that were registered to that `IInterceptorBroadcaster` were not
+called. Also, if an `IInterceptorBroadcaster` was set in the `RequestDetails` object, the boolean return value of the hooks
+registered to that `IInterceptorBroadcaster` were not taken into account. This second issue existed for all pointcuts 
+that returned a boolean type, not just for `STORAGE_PRECHECK_FOR_CACHED_SEARCH`. These issues have now been fixed."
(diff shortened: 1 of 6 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6440-fix-hooks-not-called-for-precheck-for-cached-search.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: fix
+issue: 6440
+title: "Previously, if an IInterceptorBroadcaster was set in a RequestDetails object, 
+`STORAGE_PRECHECK_FOR_CACHED_SEARCH` hooks that were registered to that IInterceptorBroadcaster were not 
+called. Also, if an IInterceptorBroadcaster was set in the RequestDetails object, the boolean return value of the hooks 
+registered to InterceptorBroadcaster were not taken into account. This second issue existed for all pointcuts that
```
jdar8, line 7: nitpick:
```suggestion
title: "Previously, if an `IInterceptorBroadcaster` was set in a `RequestDetails` object, 
`STORAGE_PRECHECK_FOR_CACHED_SEARCH` hooks that were registered to that `IInterceptorBroadcaster` were not 
called. Also, if an `IInterceptorBroadcaster` was set in the `RequestDetails` object, the boolean return value of the hooks 
registered to `InterceptorBroadcaster` were not taken into account. This second issue existed for all pointcuts that 
```

jdar8 (commented):
Don't have approval powers so someone else will need to approve - but looks good to me. Just left one tiny nitpick on the changelog.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/util/CompositeInterceptorBroadcaster.java
```diff
@@ -81,7 +81,7 @@ public boolean callHooks(Pointcut thePointcut, HookParams theParams) {
 				}
 				if (theRequestDetails != null && theRequestDetails.getInterceptorBroadcaster() != null && retVal) {
 					IInterceptorBroadcaster interceptorBroadcaster = theRequestDetails.getInterceptorBroadcaster();
-					interceptorBroadcaster.callHooks(thePointcut, theParams);
+					retVal = interceptorBroadcaster.callHooks(thePointcut, theParams);
```
codeforgreen, line 84: Good catch. I see this is not unit tested and wondering if you can add a unit test for this.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
```diff
@@ -600,12 +600,12 @@ private PersistedJpaBundleProvider findCachedQuery(
 							.add(SearchParameterMap.class, theParams)
 							.add(RequestDetails.class, theRequestDetails)
 							.addIfMatchesType(ServletRequestDetails.class, theRequestDetails);
-					Object outcome = CompositeInterceptorBroadcaster.doCallHooksAndReturnObject(
+					boolean canUseCache = CompositeInterceptorBroadcaster.doCallHooks(
```
codeforgreen, line 603: I wonder if CompositeInterceptorBroadcaster is doing the right thing for `doCallHooksAndReturnObject`. Upon reading the code, I am thinking that `doCallHooks` is a sub-set of `doCallHooksAndReturnObject` for boolean return types, to simplify the call. So I presume `doCallHooksAndReturnObject` should work for boolean return types. Could it be that the default value is incorrect and it should be null? 
I do also suggest we add javadoc to these two methods once we understand what the difference is between them.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6440-fix-hooks-not-called-for-precheck-for-cached-search.yaml
mrdnctrk: thanks, applied.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
mrdnctrk: I see that there is a [javadoc](https://github.com/hapifhir/hapi-fhir/blob/e4eb805f4f033425bee6e4faed890da2bdf85ca4/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/IBaseInterceptorBroadcaster.java#L49-L49) doCallHooksAndReturnObject that `this method  should only be called for pointcuts that return a type other than void or boolean`. That is how it seems it was intended. Also the current default value of doCallHooksAndReturnObject is null. 

But i see your point about  doCallHooks being a sub-set of doCallHooksAndReturnObject, I guess to make it work that way doCallHooksAndReturnObject should keep calling hooks until either a non-null value is seen or a false boolean is seen. but that is not a change I am comfortable doing in this MR.

tadgh (approved):

tadgh (approved):

codeforgreen (approved):

jamesagnew (approved):

PR #6447, opened 2024-11-06
Patient/$match: Add remaining parameters

felixwiemuth:
Even if they are not used, it simplifies using the parameters when overriding the `match` function.

### DIFF
--- a/hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/provider/MdmProviderCrossPartitionR4Test.java
+++ b/hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/provider/MdmProviderCrossPartitionR4Test.java
@@ -54,7 +54,7 @@ public class MdmProviderCrossPartitionR4Test extends BaseProviderR4Test{
 		myMdmSettings.setSearchAllPartitionForMatch(true);
 		createPatientOnPartition(buildJanePatient(), RequestPartitionId.fromPartitionId(1));
 
-		Bundle result = (Bundle) myMdmProvider.match(buildJanePatient(), new SystemRequestDetails().setRequestPartitionId(RequestPartitionId.fromPartitionId(2)));
+		Bundle result = (Bundle) myMdmProvider.match(buildJanePatient(), null, null, new SystemRequestDetails().setRequestPartitionId(RequestPartitionId.fromPartitionId(2)));
 		assertThat(result.getEntry()).hasSize(1);
 	}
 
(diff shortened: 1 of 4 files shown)


felixwiemuth:
Hi @tadgh, thanks for your detailed comment, and sure I can provide further information why I think this is important.

#### Use case and reason for suggested change
Our use case is the following: We use `MdmProviderDstu3Plus`, as it provides useful functionality. However, we do have our own implementation of the `Patient/$match` operation. So we extend the `MdmProviderDstu3Plus` and override the `match` method. This usually works fine, however, one cannot change the method signature when overriding, so if HAPI's implementation is not declaring all parameters, we have to retrieve the parameters manually from the `requestDetails`, plus we have to manually add validation which would have been done by the `@OperationParam` annotation (e.g. `@OperationParam(name = ProviderConstants.MDM_MATCH_COUNT, min = 0, max = 1, typeName = "integer")`).

#### Regarding adding parameters which are not used
You are right regarding that providing parameters that are not used is confusing. On the other hand, as the standard specifies them, it's also fair that they are present.
I think it is a good idea to return an error when a parameter is provided that is not being considered by the implementation. Alternatively, it could be an `OperationOutcome` with a warning or error issue.

Probably eventually HAPI's example implementation should also support the parameters, even though the implementation as a whole is quite rudimentary, as it is based only on a parts of what MDM matching does (as opposed to an actual scoring, which is complex).

Regarding
> 1. Implement all the params from the spec, meaning onlySingleMatch as well.

I can see that `onlySingleMatch` is added in the current draft of FHIR R6 (but it is not present in R4 or R5) - do you want to add it already anyway?

#### Regarding annotating another method with the same `@Operation`
I actually had tried that approach, but it seems that always the parent class's method would be called.

Also, it's probably not a good idea for HAPI to decide dynamically based on the provided parameters which overloaded method to route to.

One could of course consider changing the routing so that if there are multiple `@Operation` annotations for the same operation in a class hierarchy, the lowest in the hierarchy takes precedence. This would add a robust mechanism to deal with use cases like the one described above, for any operation in general. At the same time, it should probably be forbidden to have multiple `@Operation` for the same operation in one class.

Apart from that, it is probably still preferable to have all parameters as according to the spec, because it saves work and error-sources due to manually replicating the spec with cardinalities etc.

#### Additional discussion regarding `match` being part of `MdmProviderDstu3Plus`
Generally, it is actually also not optimal that `match` lives in a provider that is strongly related to MDM, and probably should only be loaded when MDM is enabled, as the `Patient/$match` operation also makes sense without MDM.

To provide it in both cases, we have to redirect to the actual implementation both from the `MdmProviderDstu3Plus`-extension (enabled when MDM is enabled) and from an extra provider added specifically for this (loaded when MDM is disabled).

tadgh:
I'm a bit confused. If you are talking about `Patient/$match`, that is now implemented in [PatientMatchProvider](https://github.com/hapifhir/hapi-fhir/blob/main/hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/provider/PatientMatchProvider.java#L45-L45), which is divorced from MdmDstu3PlusProvider,  as of this pull request: https://github.com/hapifhir/hapi-fhir/pull/6503. Is your fork perhaps out of date? This was ~45 days ago we changed this. 

Given that, and the fact that PatientMatchProvider.java is only 50 lines long, it's probably easier for you to just write your own and load that? Alternatively, rework your pr to: 

1. Add the new parameters to this PatientMatchProvider.
2. Throw 400 if an unsupported param is used. 

And yeah, you can omit the R6-only parameter, that's fine.

felixwiemuth:
Good to hear that `Patient/$match` was separated into a different provider and yes, that will allow us to just have our own provider. (That update simply wasn't there yet when I worked on this.)

I'll update the PR as discussed.

I think I'll open a separate issue regarding `@Operation` routing for the case where there are multiple methods with the same annotation.

tadgh:
Sounds great, thanks! @me when its ready for review, i may not see it otherwise.

PR #6470, opened 2024-11-14
Update points.json

bartemb:


jamesagnew:
Thanks for the submission! Should appear on the map in the next few days.

PR #6479, opened 2024-11-18
6478 conditional updates should batch resourcesearchurl deletes

### LINKED ISSUE #6478: Updating multiple resources with conditional urls is slow due to multiple deletes of ResourceSearchUrlEntity
When updating multiple resources with conditional url, multiple deletes occur, resulting in a slow down.

TipzCM:
closes #6478


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6478-transaction-bundle-updates-with-multiple-conditional-urls.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: perf
+issue: 6478
+jira: SMILE-8955
+title: "Transactions with multiple saved search urls will have the saved search urls
+        deleted in a batch, instead of 1 at a time.
```
epeartree, line 6: Missing a period at the end of the sentence.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
```diff
@@ -192,6 +197,16 @@ private void preFetch(
 		systemDao.preFetchResources(JpaPid.fromLongList(idsToPreFetch), true);
 	}
 
+	@SuppressWarnings("rawtypes")
+	protected void postTransactionProcess(TransactionDetails theTransactionDetails) {
+		Set<IResourcePersistentId> resourceIds = theTransactionDetails.getUpdatedResourceIds();
```
epeartree, line 202: Might be incorrect but i think we have the possibility for a NPE when evaluating resourceIds.isEmpty() due to the way getUpdatedResourceIds() is implemented.  see comment below

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
```diff
@@ -118,9 +119,40 @@ public void clearRollbackUndoActions() {
 		}
 	}
 
+	/**
+	 * @since 7.6.0
+	 */
+	@SuppressWarnings("rawtypes")
+	public void addUpdatedResourceId(@Nonnull IResourcePersistentId theResourceId) {
+		Validate.notNull(theResourceId, "theResourceId must not be null");
```
epeartree, line 127: i would put conditional create of myUpdatedREsourceIds in the getUpdatedResourceIds and invoke the method here.  that way, we do not end up returning null in getUpdatedResourceIds if addUpdatedResourceId was not invoked.

epeartree (approved):
Approve pending modifications

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
TipzCM: It's initialized as an empty map, so we will never return an empty map.

We re-initialize it to a hashmap here so we can populate it.

This is inline with the rest of the maps in this class (which is why it's done here).

PR #6501, opened 2024-11-22
Profile reference param

### LINKED ISSUE #6502: _profile can't be used as a reference param
**Describe the bug**
`_profile` is a Reference Parameter [as per R4](https://www.hl7.org/fhir/r4/search.html#all) and in the CQL processor is used that way. It makes more sense to model it as a Uri Parameter but we should be true to the specification. 

Currently we throw a `ClassCastException` if it's a `ReferenceParam`.

**To Reproduce**
1. Perform `HapiFhirRepository.search(...)` with a `ReferenceParam` named `_profile` with any value

**Expected behavior**
The search should be executed without. issue

 - HAPI FHIR Version : latest

TahaAttari:
Closes #6502

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6502-profile-can-be-reference-param.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6502-profile-can-be-reference-param.yaml
@@ -1,0 +1,4 @@
+---
+type: fix
+issue: 6502
+title: "Allow `_profile` to be used as a ReferenceParam as per the spec"
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6502-profile-can-be-reference-param.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6502-profile-can-be-reference-param.yaml
@@ -1,4 +1,4 @@
 ---
 type: fix
 issue: 6502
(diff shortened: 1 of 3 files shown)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6502-profile-can-be-reference-param.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 6502
+title: "Allow `_profile` to be used as a ReferenceParam as per the spec"
```
michaelabuckley, line 4:
```suggestion
title: "Support ReferenceParam for `_profile` in queries using the SearchParameterMap."
```

michaelabuckley (changes requested):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6502-profile-can-be-reference-param.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 6502
+title: "Support ReferenceParam for `_profile` in queries using the SearchParameterMap."
```
michaelabuckley, line 4:
```suggestion
title: "Support ReferenceParam in addition to UriParam for `_profile` in queries using the SearchParameterMap to match the change in the specification from DSTU3 to R4."
```

michaelabuckley (approved):
Looks ok.

PR #6506, opened 2024-11-23
Small changes related to `VersionSpecificWorkerContextWrapper` and new tests

### LINKED ISSUE #6505: Unexpected validation errors occur when the profile resource cache entry expires within a $validate call
**Describe the bug**
`VersionSpecificWorkerContextWrapper` holds an internal cache used to store validation resources in the canonical version (R5). That is because the FHIR Core Validation needs to be invoked using canonical resource in R5.
When we validate a resource using the `$validate` operation against a profile we can end up in a situation with unexpected validation errors, which suggest the profile was not applied properly during validation. 

**Root Cause**
That happens when the same profile is fetched by the validation logic in the FHIR Core Validation library multiple times and the cache expires in between. Currently, the FHIR Core Validation library expects the same (object) instance of a validation resource to be returned within the same validate call. If that doesn't happen, some of the logic which checks the profile does not work.
See: https://github.com/hapifhir/org.hl7.fhir.core/blob/0642e123a1f72da241c325f0e5e65d84aef3a029/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L518
```
List<ElementDefinition> elements = profile.getSnapshot().getElement();
String path = element.getPath();
for (int index = elements.indexOf(element) + 1; index < elements.size(); index++) {
  ElementDefinition e = elements.get(index);
  if (e.getPath().startsWith(path + ".")) {
    // We only want direct children, not all descendants
    if (!e.getPath().substring(path.length()+1).contains("."))
      res.add(e);
  } else
    break;
}
```

**To Reproduce**
1. To reproduce this behaviour with a running hapi-fhir server, you would need a large Bundle document resource (at step 3), for which the validation takes more than 10s, ideally at least 30s because the expiry after write timeout for the cache was set at 10s. In this absence of such an example, reproduce the problem in a test where you manipulate the cache expiry for less for e.g. 10m by calling:
`HapiSystemProperties.setValidationResourceCacheTimeoutMillis(10);`
2. Store a profile with a constraint more permissive than the base standard profile: e.g. [profile-practitionerrole.json](https://github.com/user-attachments/files/17877580/profile-practitionerrole.json)
3. Validate a bundle document that references the profile multiple times: [bundle-with-medicationdispense.json](https://github.com/user-attachments/files/17877581/bundle-with-medicationdispense.json)
5. Observe the returned errors by the `$validate` call. For the above example it will show `PractitionerRole.telecom:TelecomPhone: max allowed = 1, but found 2`

Please refer to test `ValidateWithCacheNegativeTest` in the linked pull request to reproduce the issue.

**Expected behaviour**
No unexpected errors should be returned.

 - HAPI FHIR Version 7.4

**Additional context**
The example given in this ticket is not reproducible in hapi-fhir 7.2. There was a major version bump of FHIR Core libraries in starting hapi-fhir 7.4 and that brought up this problem. The version bump changes seem unrelated to how the FHIR Core Validation relies on object identity for fetched profiles. At the moment it is not known exactly what change in the FHIR Core libraries made this surface but nevertheless it does require a fix on hapi-fhir side.

codeforgreen:
Related to #6505

What was done:
* updated all implementors of `IValidationSupport` to use `Logs.getTerminologyTroubleshootingLog()`
* added a bunch of new tests on `VersionSpecificWorkerContextWrapper` related to the cache and also testing the methods to fetch validation resources at different levels (unit and integration).

Note that the initial scope of this PR was changed and it now does not include any functional change.


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_6_0/6505-validation-errors-profile-cache-expiry.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: fix
+issue: 6505
+title: "Previously, when invoking a $validate operation that took more than 10s some unexpected errors 
+would be returned which suggest the profile was not applied properly or used. 
+That happens because the validation cache entry expired within the $validate call 
+and the FHIR Core Validation library expects the same object instance for the same profile to be returned.
+This has been fixed by increasing the cache expiry to 10min."
```
michaelabuckley, line 8: question: Does this mean we are still vulnerable to rare failures if a cache entry expires in the middle of processing?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/validation/ValidatorResourceFetcher.java
```diff
@@ -105,10 +108,9 @@ private IBaseResource fetchByUrl(String url, IFhirResourceDao<?> dao, RequestDet
 		} catch (InvalidRequestException e) {
 			ourLog.info("Resource does not support 'url' or 'version' Search Parameters");
 		}
-		if (results != null && results.size() > 0) {
-			if (results.size() > 1) {
-				ourLog.warn(
-						String.format("Multiple results found for URL '%s', only the first will be considered.", url));
+		if (results != null && !results.isEmpty()) {
+			if (results.size() > 1 && ourLog.isWarnEnabled()) {
```
michaelabuckley, line 112: trivial: the ourLog.isWarnEnabled() check seems redundant since the call to warn uses a format string.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/DefaultProfileValidationSupportNpmStrategy.java
```diff
@@ -4,16 +4,16 @@
...
 import jakarta.annotation.Nonnull;
 import org.apache.commons.lang3.Validate;
 import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 
 public class DefaultProfileValidationSupportNpmStrategy extends NpmPackageValidationSupport {
-	private static final Logger ourLog = LoggerFactory.getLogger(DefaultProfileValidationSupportNpmStrategy.class);
+	private static final Logger ourLog = Logs.getTerminologyTroubleshootingLog();
```
michaelabuckley, line 16: Nice to see a new troubleshooting log.

michaelabuckley (approved):

codeforgreen:
> FYI @codeforgreen I believe this PR should be redundant once this merges (it also fixes the same issue by just never expiring StructureDefinions): #6508

I updated the scope of this PR to much smaller. I updated the description accordingly. It now just include the small compile warning or refactoring changes and additional tests. There is no change to the caches or validato, no changelog in this PR.

Issue #6040, opened 2024-06-24
Validation of resources with contained profile with version do not validate correctly


TipzCM:
Validating a resource using a contained profile with an appended version will not validate using that profile.

Eg:

```
"meta": {
   "profile": [
      "http://example.com/StructureDefinition|1.0.0"
   ]
}
```

PR #6507, opened 2024-11-23
[6463] FHIR resource validation breaks when server have multiple versions of same profile

MangalaEkanayake:
- version in canonical url is used to search StructureDefinition
- remain the version of a Resource to identify it properly

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupport.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupport.java
@@ -235,7 +235,13 @@ public class JpaPersistedResourceValidationSupport implements IValidationSupport
 				}
 				SearchParameterMap params = new SearchParameterMap();
 				params.setLoadSynchronousUpTo(1);
-				params.add(StructureDefinition.SP_URL, new UriParam(theUri));
+				int versionSeparator = theUri.lastIndexOf('|');
+				if (versionSeparator != -1) {
+					params.add(StructureDefinition.SP_VERSION, new TokenParam(theUri.substring(versionSeparator + 1)));
+					params.add(StructureDefinition.SP_URL, new UriParam(theUri.substring(0, versionSeparator)));
+				} else {
+					params.add(StructureDefinition.SP_URL, new UriParam(theUri));
+				}
 				search = myDaoRegistry.getResourceDao("StructureDefinition").search(params);
 				break;
 			}
--- a/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
+++ b/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
@@ -88,7 +88,7 @@ public class VersionSpecificWorkerContextWrapper extends I18nBase implements IWo
 	 * the converted version gets cached too.
 	 */
 	private static final String CANONICAL_USERDATA_KEY =
-			VersionSpecificWorkerContextWrapper.class.getName() + "_CANONICAL_USERDATA_KEY";
+		VersionSpecificWorkerContextWrapper.class.getName() + "_CANONICAL_USERDATA_KEY";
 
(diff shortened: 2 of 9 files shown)


hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
```diff
@@ -502,18 +502,14 @@ public <T extends Resource> T fetchResource(Class<T> class_, String theUri) {
 			return null;
 		}
 
-		String uri = theUri;
-		// handle profile version, if present
 		if (theUri.contains("|")) {
 			String[] parts = theUri.split("\\|");
```
jamesagnew, line 506: Since we don't actually need the split values anymore, just to count them it might make sense to replace this with something like `StringUtils.countMatches`

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
```diff
@@ -502,18 +502,14 @@ public <T extends Resource> T fetchResource(Class<T> class_, String theUri) {
...
-			if (parts.length == 2) {
-				uri = parts[0];
-			} else {
+			if (parts.length != 2) {
 				ourLog.warn("Unrecognized profile uri: {}", theUri);
 			}
 		}
 
-		ResourceKey key = new ResourceKey(class_.getSimpleName(), uri);
+		ResourceKey key = new ResourceKey(class_.getSimpleName(), theUri);
```
jamesagnew, line 512: This change means that we'll be passing canonical URLs containing versions to a number of other implementations of `fetchResource`, `fetchStructureDefinition`, `fetchCodeSystem`, and `fetchValueSet` in the various other implementations of `IValidationSupport`. 

In cases like `CachingValidationSupport` and `ValidationSupportChain` it doesn't matter since they just pass the URL through, and others like `InMemoryTerminologyServerValidationSupport` just don't implement these methods so they'll be fine too.

But others like `NpmJpaValidationSupport`, `PrePopulatedValidationSupport` and `DefaultProfileValidationSupport` all implement these methods. These will all need to be checked to ensure that they support both versionless and versioned URLs correctly. Ideally using unit tests.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaPersistedResourceValidationSupport.java
```diff
@@ -273,7 +273,13 @@ private <T extends IBaseResource> IBaseResource doFetchResource(@Nullable Class<
 				}
 				SearchParameterMap params = new SearchParameterMap();
 				params.setLoadSynchronousUpTo(1);
-				params.add(StructureDefinition.SP_URL, new UriParam(theUri));
+				int versionSeparator = theUri.lastIndexOf('|');
+				if (versionSeparator != -1) {
```
jamesagnew, line 277: This change should be unit tested.

jamesagnew (changes requested):
Hi @MangalaEkanayake - Thanks for the contribution! This seems like a useful change.

I've made some suggestions inline. This change will affect several other modules that you haven't modified, so these would need to be updated and tested, and unit tests written for the existing change as well.

With those changes, we can definitely get this merged.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
MangalaEkanayake: Those are old implementations. this change has done recently by https://github.com/hapifhir/hapi-fhir/issues/6040.
So I guess those should works as expected.
meantime no any existing unit tests are failing. 
we have an IValidationSupport implementation based on version and it works fine. I don't expect to write any unit tests to these existing modules.

MangalaEkanayake:
> Hi @MangalaEkanayake - Thanks for the contribution! This seems like a useful change.
> 
> I've made some suggestions inline. This change will affect several other modules that you haven't modified, so these would need to be updated and tested, and unit tests written for the existing change as well.
> 
> With those changes, we can definitely get this merged.

Hi @jamesagnew - can we merge this?

PR #6508, opened 2024-11-23
Streamline ValidationSupportChain

### LINKED ISSUE #6424: Validation cardinality error on Extensions
Since the update to hapi-fhir-validation 7.4.4 (version 6.3.23 of org.hl7.fhir.validation), I have random errors in my integration tests. Multiple tests randomly fail on incorrect Validation_VAL_Profile_Minimum errors. This always happens on Extensions.

I debugged the issue and found the problem. I'm using the "org.hl7.fhir.common.hapi.validation.validator.VersionSpecificWorkerContextWrapper". This uses a caching with a default of 10s. Whenever the cache has timed out, it will create a new canonical structure definition. This means when the cache times out during the validation of f.e. multiple resources in a Bundle, the "InstanceValidator.checkExtension" function will have a new StructureDefinition object compared to which was used before. This gives issues when "InstanceValidator.checkCardinalities" function tries to get a list of relevent slices by calling "profileUtilities.getSliceList". This function uses an indexOf with the given ElementDefinition to find the same element in the snapshot element list of the structure definition. When a new structure definition is used no element will be found and this will result in a cardinality error.

https://github.com/hapifhir/org.hl7.fhir.core/blob/cde5d86fe008ccb01fc037aea0332ade1bd4f743/org.hl7.fhir.r5/src/main/java/org/hl7/fhir/r5/conformance/profile/ProfileUtilities.java#L553

The following test will succeed after the first call to getSliceList (using the same element definition object as used in the snapshot as argument), but will fail on the results of the second call to getSliceList (using a newly created element definition as argument)
```
class ProfileUtilitiesTest {

  @Test
  void getSliceList() {
    ElementDefinition element1 = (ElementDefinition) new ElementDefinition().setPath("Extension").setId("Extension");
    ElementDefinition element2 = (ElementDefinition) new ElementDefinition().setPath("Extension.id").setId("Extension.id");
    ElementDefinition element3 = (ElementDefinition) new ElementDefinition().setPath("Extension.value[x]")
      .setSlicing(new ElementDefinition.ElementDefinitionSlicingComponent().setDescription("slice")).setId("Extension.value[x]");
    ElementDefinition element4 = (ElementDefinition) new ElementDefinition().setPath("Extension.value[x]").setSliceName("valueCodeableConcept").setId("Extension.value[x]:valueCodeableConcept");

    StructureDefinition.StructureDefinitionSnapshotComponent snapshot = new StructureDefinition.StructureDefinitionSnapshotComponent()
      .setElement(List.of(element1, element2, element3, element4));
    StructureDefinition profile = new StructureDefinition().setSnapshot(snapshot);

    // using same element from snapshot in element
    List<ElementDefinition> elements1 = new ProfileUtilities(null, null, null).getSliceList(profile, element3);
    Assertions.assertEquals(1, elements1.size());
    Assertions.assertEquals(element4, elements1.get(0));

    // using different element as in snapshot with same values
    List<ElementDefinition> elements2 = new ProfileUtilities(null, null, null).getSliceList(profile,
      (ElementDefinition) new ElementDefinition().setPath("Extension.value[x]")
        .setSlicing(new ElementDefinition.ElementDefinitionSlicingComponent().setDescription("slice")).setId("Extension.value[x]"));
    Assertions.assertEquals(1, elements2.size());
    Assertions.assertEquals(element4, elements2.get(0));
  }

}
```

A compare on id to find the right element in the list would solve this issue. A workaround is to increase the cache timeout, so this issue has a much smaller chance to occur. I can make a pull request for this issue if that helps.

### LINKED ISSUE #6505: Unexpected validation errors occur when the profile resource cache entry expires within a $validate call

jamesagnew:
This PR:

* Adds new caching capabilities to ValidationSupportChain. This is an improvement over the previous separate caching module because the chain can now remember which entries in the cache responded affirmative to `isValueSetSupported()` and will therefore be more efficient about trying entries in the chain. It also makes debugging much less confusing as there is less recursion and the caches don't use loadingCache.
* Importantly, the caching in ValidationSupportChain caches negative lookups (i.e. items that could not be found by URL) as well as positive lookups. This is a change from the historical caching behaviour.
* Changes ValidationSupportChain to never expire StructureDefinition entries in the cache, which is needed because the validator makes assumptions about structuredefinitions never changing. Fixes #6424. Also fixes #6505.
* Modifies `VersionSpecificWorkerContextWrapper` so that it doesn't use a separate cache and instead relies on the caching provided by ValidationSupportChain. This class previously used a cache because it converts arbitrary versions of FHIR StructureDefinitions into the canonical version required by the validator (R5), but these converted versions are now stored in the userdata map of objects returned by and cached by ValidationSupportChain. This makes the caching more predictable since there is only one cache to track.
* Adds OpenTelemetry support to ValidationSupportChain, with metrics for tracking the cache size.
* Deprecates CachingValidationSupport since caching is now provided by ValidationSupportChain. CachingValidationSupport is now just a passthrough and should be removed from applications. It will be removed from the library in a future release.

These changes result in very significant performance improvements when performing validation in the JPA server. Throughput improvements of 1000% have been recorded in benchmarking for some use cases.


hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChainMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+	/*
+	 * See ValidationSupportChainTest#testMetrics for a unit test
+	 * which exercises the functionality in this class.
+	 */
+
+	public static final String EXPIRING_CACHE_MAXIMUM_SIZE = "expiring_cache.maximum_size";
+	public static final String EXPIRING_CACHE_CURRENT_ENTRIES = "expiring_cache.current_entries";
+	public static final String NON_EXPIRING_CACHE_CURRENT_ENTRIES = "non_expiring_cache.current_entries";
+	private static final String INSTRUMENTATION_NAME = "io.hapifhir.ValidationSupportChain";
+	private static final AttributeKey<String> INSTANCE_NAME = stringKey("instance.name");
```
mrdnctrk, line 30: OpenTelemetry also suggests using namespaces for attribute names to prevent clashes. Could this be attribute prefixed with  `hapifhir.validation_support_chain` so that the full name is `hapifhir.validation_support_chain.instance_name` or something similar you find appropriate?

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChainMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+ * This class provides OpenTelemetry metrics for the {@link ValidationSupportChain} cache.
+ */
+public class ValidationSupportChainMetrics {
+
+	/*
+	 * See ValidationSupportChainTest#testMetrics for a unit test
+	 * which exercises the functionality in this class.
+	 */
+
+	public static final String EXPIRING_CACHE_MAXIMUM_SIZE = "expiring_cache.maximum_size";
```
mrdnctrk, line 26: I would prefix the name of these metrics with namespaces, like `hapifhir.validation_support_chain.expiring_cache.maximum_size`

pom.xml
```diff
@@ -1034,6 +1034,7 @@
 		<maven_license_plugin_version>1.8</maven_license_plugin_version>
 		<okhttp_version>4.12.0</okhttp_version>
 		<otel_instrumentation.version>2.8.0</otel_instrumentation.version>
+		<otel_agent_for_testing.version>2.10.0-alpha</otel_agent_for_testing.version>
```
mrdnctrk, line 1037: I see this exists only as alpha but can we make this use the same version as the `otel_instrumentation`? Maybe we could use`${otel_instrumentation.version}-alpha` as the version string whenever I need an alpha version.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChainMetrics.java
```diff
@@ -0,0 +1,85 @@
...
+					expiringCacheCurrentEntries.record(expiringCacheEntries, baseAttribute);
+					nonExpiringCacheCurrentEntries.record(nonExpiringCacheEntries, baseAttribute);
+				},
+				expiringCacheMaxSize,
+				expiringCacheCurrentEntries,
+				nonExpiringCacheCurrentEntries);
+	}
+
+	public void stop() {
+		myBatchCallback.close();
```
mrdnctrk, line 83: A null check on myBatchCallback could be added before calling close, just in case.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChainTest.java
```diff
@@ -29,54 +83,644 @@ public void testVersionCheck() {
...
+
+		prepareMock(myValidationSupport0, myValidationSupport1, myValidationSupport2);
+		when(myValidationSupport0.fetchStructureDefinition("http://foo")).thenReturn(new StructureDefinition().setUrl("http://foo"));
+		ValidationSupportChain chain = new ValidationSupportChain(newCacheConfiguration(theUseCache), myValidationSupport0, myValidationSupport1, myValidationSupport2);
+		chain.setName("FOO_NAME");
+
+		chain.start();
+		try {
+			if (theUseCache) {
+				assertEquals(5000L, getLastMetricValue(libraryTestRunner, ValidationSupportChainMetrics.EXPIRING_CACHE_MAXIMUM_SIZE));
```
mrdnctrk, line 664: I think we shouldn't use production constants in test code. If some one changes the production constant to another  value by mistake, this test wouldn't catch it since it uses the same constant.

mrdnctrk (commented):
I have reviewed the otel related changes, and added some minor suggestions.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -189,58 +483,132 @@ public void removeValidationSupport(IValidationSupport theValidationSupport) {
...
-		if (myChain != null) {
-			Optional<IValidationSupport> remoteTerminologyService = myChain.stream()
-					.filter(RemoteTerminologyServiceValidationSupport.class::isInstance)
-					.findFirst();
-			if (remoteTerminologyService.isPresent()) {
-				return true;
-			}
-		}
-		return false;
+		return myChain.stream().anyMatch(RemoteTerminologyServiceValidationSupport.class::isInstance);
```
fil512, line 553: heh

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -189,58 +483,132 @@ public void removeValidationSupport(IValidationSupport theValidationSupport) {
...
+						String url = terser.getSinglePrimitiveValueOrNull(structureDefinition, "url");
+						url = defaultIfBlank(url, UUID.randomUUID().toString());
+						if (myStructureDefinitionsByUrl.putIfAbsent(url, structureDefinition) == null) {
+							myStructureDefinitionsAsList.add(structureDefinition);
+						}
+					}
+				}
+			}
+		}
+		return Collections.unmodifiableList(new ArrayList<>(myStructureDefinitionsAsList));
```
fil512, line 592:
```suggestion
		return List.copyOf(myStructureDefinitionsAsList);
```

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -267,84 +635,98 @@ private List<IBaseResource> doFetchStructureDefinitions(
...
-							"CodeSystem {} with System {} fetched by {}",
-							retVal.getIdElement(),
-							theSystem,
-							next.getName());
+		Function<IValidationSupport, IBaseResource> invoker = v -> v.fetchCodeSystem(theSystem);
+		ResourceByUrlKey<IBaseResource> key = new ResourceByUrlKey<>(ResourceByUrlKey.TypeEnum.CODESYSTEM, theSystem);
+		return fetchResource(key, invoker, theSystem);
+	}
+
+	private <T> T fetchResource(ResourceByUrlKey<T> theKey, Function<IValidationSupport, T> theInvoker, String theUrl) {
```
fil512, line 643: Resource has special meaning in FHIR. Consider `fetchValue` or perhaps `fetchCachedValue`

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -393,56 +803,520 @@ public CodeValidationResult validateCodeInValueSet(
...
+			if (retVal != null) {
+				Runnable loaderTask = () -> {
+					List<IBaseResource> loadedItem = theLoader.get();
+					CacheValue<List<IBaseResource>> value = new CacheValue<>(loadedItem);
+					myNonExpiringCache.put(theKey, value);
+					putInCache(theKey, value);
+				};
+				myBackgroundExecutor.execute(loaderTask);
+
+				return retVal.getValue();
```
fil512, line 933: Is there a race condition here? If so, consider guarding with

```suggestion
				returnValue = retVal.getValue();
				
				myBackgroundExecutor.execute(loaderTask);

				return returnValue;
```

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -393,56 +803,520 @@ public CodeValidationResult validateCodeInValueSet(
...
+		public static CacheConfiguration disabled() {
+			return new CacheConfiguration().setCacheSize(0).setCacheTimeout(Duration.ofMillis(0));
+		}
+	}
+
+	/**
+	 * @param <V> The value type associated with this key
+	 */
+	@SuppressWarnings("unused")
+	abstract static class BaseKey<V> {
```
fil512, line 1025: Strange to have an unused generic, but I like it. Sort of a tag interface...

fil512 (approved):
Fantastic optimization

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChainMetrics.java
jamesagnew: Ah thanks, good suggestion. I'll do that.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChainMetrics.java
jamesagnew: Done

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChainMetrics.java
jamesagnew: Done

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChainTest.java
jamesagnew: That's an interesting point. Changed as suggested.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
```diff
@@ -230,11 +232,26 @@ public class JpaConfig {
...
 
+	@Autowired
+	private FhirContext myFhirContext;
+
+	@Bean
+	public ValidationSupportChain.CacheConfiguration validationSupportChainCacheConfiguration() {
+		return ValidationSupportChain.CacheConfiguration.defaultValues();
+	}
+
+	@Bean(name = JpaConfig.JPA_VALIDATION_SUPPORT_CHAIN)
```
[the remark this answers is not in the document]
jamesagnew: Yeah, agreed. The intent here originally was that there were a bunch of beans that all implemented the IValidationSupport interface... But in the end it probably would be better to depend on the concrete subclass in that case.

But yeah, I'm not going to change that here... Already enough change in this ticket.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -10,80 +11,345 @@
...
+ *     and {@link #fetchAllStructureDefinitions()} will call every method in the chain in
+ *     order, and aggregate the results into a single list to return.
+ * </li>
+ * <li>
+ *     Calls to fetch or validate codes, such as {@link #validateCode(ValidationSupportContext, ConceptValidationOptions, String, String, String, String)}
+ *     and {@link #lookupCode(ValidationSupportContext, LookupCodeRequest)} will first test
+ *     each module in the chain using the {@link #isCodeSystemSupported(ValidationSupportContext, String)}
+ *     or {@link #isValueSetSupported(ValidationSupportContext, String)}
+ *     methods (depending on whether a ValueSet URL is present in the method parameters)
+ *     and will any methods in the chain which return that they can handle the given
```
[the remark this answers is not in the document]
jamesagnew: hehe the missing word is "invoke". added.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -10,80 +11,345 @@
...
+					CacheFactory.build(theCacheConfiguration.getCacheTimeout(), theCacheConfiguration.getCacheSize());
+			myNonExpiringCache = Collections.synchronizedMap(new HashMap<>());
+
+			LinkedBlockingQueue<Runnable> executorQueue = new LinkedBlockingQueue<>(1000);
+			BasicThreadFactory threadFactory = new BasicThreadFactory.Builder()
+					.namingPattern("CachingValidationSupport-%d")
+					.daemon(false)
+					.priority(Thread.NORM_PRIORITY)
+					.build();
+			myBackgroundExecutor = new ThreadPoolExecutor(
```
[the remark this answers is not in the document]
jamesagnew: I thought about it but the problem is that class needs Spring, and we want the validator module to run without spring. I'll add a comment to that effect.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -10,80 +11,345 @@
...
+	}
+
+	@Override
+	public boolean isEnabledValidationForCodingsLogicalAnd() {
+		return myEnabledValidationForCodingsLogicalAnd;
+	}
+
+	/**
+	 * <p
+	 * Warning: This method's behaviour and naming is preserved for backwards compatibility, BUT the actual naming and
```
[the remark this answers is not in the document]
jamesagnew: Good call. Done.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -10,80 +11,345 @@
...
 	public TranslateConceptResults translateConcept(TranslateCodeRequest theRequest) {
-		TranslateConceptResults retVal = null;
-		for (IValidationSupport next : myChain) {
-			TranslateConceptResults translations = next.translateConcept(theRequest);
-			if (translations != null) {
-				if (retVal == null) {
-					retVal = new TranslateConceptResults();
-				}
+		TranslateConceptKey key = new TranslateConceptKey(theRequest);
+		CacheValue<TranslateConceptResults> retVal = getFromCache(key);
```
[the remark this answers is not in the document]
jamesagnew: I considered this - originally `CacheValue` WAS `Optional`, but we pass these around as parameters and it generates endless warnings. Ultimately I opted to just use null because I think in this case it made the code easier to follow.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -10,80 +11,345 @@
...
+					}
+
+					if (outcome.getMessage() == null) {
+						outcome.setMessage(translations.getMessage());
+					}
 
-				if (!translations.isEmpty()) {
-					if (ourLog.isDebugEnabled()) {
+					if (translations.getResult() && !outcome.getResult()) {
+						outcome.setResult(translations.getResult());
```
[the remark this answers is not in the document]
jamesagnew: I don't disagree... However, I have no idea who is depending on it being this way. It's a very old API.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -92,36 +358,63 @@ public void invalidateCaches() {
...
+		 */
+
 		for (IValidationSupport next : myChain) {
 			IBaseResource retVal =
 					next.generateSnapshot(theValidationSupportContext, theInput, theUrl, theWebUrl, theProfileName);
 			if (retVal != null) {
-				if (ourLog.isDebugEnabled()) {
-					ourLog.debug("Profile snapshot for {} generated by {}", theInput.getIdElement(), next.getName());
-				}
+				ourLog.debug("Profile snapshot for {} generated by {}", theInput.getIdElement(), next.getName());
```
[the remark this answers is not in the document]
jamesagnew: I guess.. I can't imagine that side effect actually affecting anything, but I'll make it conditional again.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -160,6 +453,7 @@ public void addValidationSupport(IValidationSupport theValidationSupport) {
 	 */
 	public void addValidationSupport(int theIndex, IValidationSupport theValidationSupport) {
 		Validate.notNull(theValidationSupport, "theValidationSupport must not be null");
+		invalidateCaches();
```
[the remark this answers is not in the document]
jamesagnew: Added

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -189,58 +483,132 @@ public void removeValidationSupport(IValidationSupport theValidationSupport) {
...
+							next.expandValueSet(theValidationSupportContext, expansionOptions, theValueSetUrlToExpand);
+					if (expanded != null) {
+						ourLog.debug("ValueSet {} expanded by URL by {}", theValueSetUrlToExpand, next.getName());
+						retVal = new CacheValue<>(expanded);
+						break;
+					}
+				}
+			}
+
+			putInCache(key, retVal);
```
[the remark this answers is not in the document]
jamesagnew: I mostly have - the thing is the behaviour isn't identical for each method. Some aggregate, some return the first non-null, some refresh asynchronously. So it isn't perfectly clean.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -189,58 +483,132 @@ public void removeValidationSupport(IValidationSupport theValidationSupport) {
...
-					next.expandValueSet(theValidationSupportContext, theExpansionOptions, theValueSetToExpand);
-			if (expanded != null) {
-				if (ourLog.isDebugEnabled()) {
+
+		ValueSetExpansionOptions expansionOptions = defaultIfNull(theExpansionOptions, EMPTY_EXPANSION_OPTIONS);
+		String id = theValueSetToExpand.getIdElement().getValue();
+		ExpandValueSetKey key = null;
+		CacheValue<ValueSetExpansionOutcome> retVal = null;
+		if (isNotBlank(id)) {
+			key = new ExpandValueSetKey(expansionOptions, id, null);
```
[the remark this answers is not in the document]
jamesagnew: I thought about it and originally actually did that.. but we have lots of tests asserting the current behaviour. I'm not actually sure that the current way is optimal but it's well tested and.. maybe people depend on it?

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
jamesagnew: This isn't allowed in the current android version we're trying to support.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
jamesagnew: Changed

PR #6512, opened 2024-11-25
Start removing dependency from FhirVersionEnum to FhirContext

michaelabuckley:
Our model classes use FhirVersionEnum, and the path from there to FhirContext brings in the whole world.
Snip this path to start disentangling the core libraries from hapi-fhir.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirVersionEnum.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirVersionEnum.java
@@ -135,15 +135,19 @@ public enum FhirVersionEnum {
 
 	/**
 	 * Creates a new FhirContext for this FHIR version
+	 * @deprecated since 7.7.  Use  {@link FhirContext#forVersion(FhirVersionEnum)} instead
 	 */
+	@Deprecated(forRemoval = true, since = "7.7")
 	public FhirContext newContext() {
-		return new FhirContext(this);
+		return FhirContext.forVersion(this);
 	}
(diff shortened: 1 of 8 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirVersionEnum.java
```diff
@@ -135,15 +135,19 @@ public boolean isRi() {
...
-		return new FhirContext(this);
+		return FhirContext.forVersion(this);
 	}
 
 	/**
 	 * Creates a new FhirContext for this FHIR version, or returns a previously created one if one exists. This
 	 * method uses {@link FhirContext#forCached(FhirVersionEnum)} to return a cached instance.
+	 * @deprecated since 7.7.  Use  {@link FhirContext#forCached(FhirVersionEnum)} instead
 	 */
+	@Deprecated(forRemoval = true, since = "7.7")
```
jamesagnew, line 150:
```suggestion
	@Deprecated(forRemoval = true, since = "8.0.0")
```

jamesagnew (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/FhirVersionEnum.java
michaelabuckley: TBH, I'm thinking about going the other way - 7.6, and backporting this.  If we really plan to remove this in 8, we need to give warning now.

PR #6517, opened 2024-11-26
6404 lucene search fulltext fix

### LINKED ISSUE #6404: Any search using fulltext (lucene) search that combines _lastUpdated and any supported search parameter will always fail
Search using _lastUpdated and a lucene-supported SP (say, _tags).

This will always return nothing, even if something should match.

TipzCM:
closes #6404


jmarchionatto (approved):
Looks good! Approved

PR #6521, opened 2024-11-27
Jd 20241126 fix versioned refs in transaction bundles

### LINKED ISSUE #6519: Versions still stripped when posting in transaction bundle
**Describe the bug**
When creating resources with a transaction with `DontStripVersionsFromReferences` enabled, references are still stripped from submitted resources.

**To Reproduce**
1. Enable auto-create placeholder
2. Set Client ID - mode to any 
3. Set values in version-references to allow all 
4. Create a transaction bundle that contains a resource referencing another resource where the reference contains a version 
5. Post the bundle
6. Perform a read on the resource you created and look at the reference -- the version has been stripped 

**Expected behavior**
The version in the reference should not be stripped

jdar8:
Closes #6519


jamesagnew (approved):
Nice enhancement!

PR #6523, opened 2024-11-28
tests for fix ref integrity with client assigned ids

### LINKED ISSUE #6476: Bug with Referential Integrity on Write when Referencing Resources with Client Assigned IDs by PID
**Describe the bug**
If you POST/PUT a Resource with a reference to another Resource, using the `res_id` (PID) value of the target instead of the `fhir_id` (forced ID), the Resource is accepted.

Some additional context:
The FHIR specification [here](https://hl7.org/fhir/R4/references.html#literal) defines a literal reference as a relative or absolute URL pointing to a resource. Based on my understanding, the relative or absolute URL should be following the same requirements as the URL used for a RESTful read request, which is defined [here](https://hl7.org/fhir/R4/http.html#read) as [base]/[type]/[id] where [id] is the logical id that appears in the resource.

When we create a resource using a client-assigned ID, the client-assigned ID appears as the logical id in the resource and as such should be required when referencing the resource elsewhere. In other words, we should be able to use the reference to look up the referenced resource. This does not appear to always be the case. See my test findings below.

Additionally, the correct behaviour can be observed when `JpaStorageSettings.ClientIdStrategyEnum = ANY`

**To Reproduce**
1. Ensure referential integrity on write is enabled
2. Create a resource using PUT with a client assigned ID
3. Query the database for the `res_id` (PID) of this resource 
4. Try reading the resource by PID vs by client assigned ID. Note that we get 404 when referencing by PID, and 200 when referencing by client assigned ID
5. Create a second resource that references the first using it's PID --> This is successful (when it shouldn't be)

**Expected behavior**
It should fail to create a resource with a PID reference to another resource created with a client assigned ID

jdar8:
This PR only contains unit tests for #6476. My original PR contained these tests + a fix,  however, IdHelperService was recently refactored as part of #6460. This refactor fixed the bug, but I thought it would still be good to include these unit tests. 

Closes #6476


hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/index/IdHelperServiceTest.java
```diff
@@ -158,5 +168,49 @@ public void testResolveResourceIdentity_defaultFunctionality(){
 		assertEquals(tuple[3], result.getDeleted());
 	}
 
+	@Test
+	public void testResolveResourceResourceIdentity_withPersistentIdOfResourceWithForcedIdAndDefaultClientIdStrategy_returnsNotFound(){
```
TipzCM, line 172: is this misnamed?

is "Resource" meant to be duplicated here?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ReferentialIntegrityTest.java
```diff
@@ -190,6 +198,29 @@ public void testRefIntegrityOnWrite_withReferenceIdOfDeletedResource(boolean the
...
+
+		Organization o = new Organization();
+		o.setName("FOO");
+		o.setId("O1");
+		DaoMethodOutcome outcome = myOrganizationDao.update(o);
+		Long organizationPid = (Long) outcome.getEntity().getPersistentId().getId();
+
+		Patient p = new Patient();
+		p.setManagingOrganization(new Reference("Organization/" + organizationPid));
+//110
```
TipzCM, line 215: ?

TipzCM (approved):

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/dao/index/IdHelperServiceTest.java
jdar8: Yup, misnamed. Good catch

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ReferentialIntegrityTest.java
jdar8: Not sure how that got there... removed. Thanks for the review!

PR #6527, opened 2024-11-28
Expose ConsentInterceptor#getConsentService for integration test purposes

codeforgreen:


nathandoef (approved):

dotasek (changes requested):
A little nit-pick:

There are explicit methods to add and remove consent services; exposing the whole list via this new method offers a way to bypass those.

Could this be wrapped in a unmodifiableList, or return just an Iterable, or does all the list functionality really need to be exposed?

codeforgreen:
> A little nit-pick:
> 
> There are explicit methods to add and remove consent services; exposing the whole list via this new method offers a way to bypass those.
> 
> Could this be wrapped in a unmodifiableList, or return just an Iterable, or does all the list functionality really need to be exposed?

Makes sense. Resolved.

jamesagnew (approved):

PR #6534, opened 2024-12-04
[6463] use version in canonical url of StructureDefinition to identify it

MangalaEkanayake:


jamesagnew (approved):
Thanks for the contribution, and for adding the additional testing needed for merging.

PR #6551, opened 2024-12-11
Concurrent batch reduction

fil512:
A batch2 reduction step likely holds state. This state is currently shared between jobs which is incorrect.

Proposed solution: reduction step in job definition should be a factory that creates a reducer rather than thes same stateful reducer instance shared by all runs.


fil512:
abandoning this approach. the reducer worker needs to be kafka driven like all the other workers.

PR #6566, opened 2024-12-18
Issue with combo index search parameter when using date

epeartree:
**the problem:**
Given a combo index search parameters configured with a date search parameter components, combo index searching is skipped when a date query parameter value is submitted to the search stack.

**root cause:**
The root cause of the problem is that when we marshal query string 'birthdate=2021-01-01' into a DateRangeParam before submission to the search layer, a default prefix of 'EQUAL' is assigned to both DateParam objects encapsulated by the dateRangeParam.  Due to the presence of the prefix, the search layer skips combo indexes inspection.  

The issue described above evolved into harmonising the search logic behaviour in favour of allowing combo indexes inspection even with the 'eq' prefix since date=2021-01-01 is equivalent to date=eq2021-01-01 searches.

**what was done:**
- Modified class `DateRangeParam` to not default the prefix to 'EQUAL' as doing so would make it impossible for indexes created by a combo search index parameters to be queried during searches;
- added utility method and comment to return a prefix of 'equal' when performing computation based on prefix value;
- Modified class SearchBuilder to allow combo indexes inspection when a date query string is prefixed with 'eq';
- Added a unit test to class FhirResourceDaoR4ComboNonUniqueParamTest;
- added changelog;


jmarchionatto (dismissed):
Fix looks good, but there is no test for the specific point of the fix, which is that the combo search is now considered. Please add test for that case.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6603-combo-index-sp-not-invoked-with-dateparam.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 6603
+jira: SMILE-8048
+title: "Previously, searches where processing was optimised with a combo index search parameters would skip searching the 
+optimised indexes when the query string included date parameters. This issue is fixed."
```
jamesagnew, line 6: I'm assuming based on what was fixed that the issue is actually when the query string contained a date range (i.e. multiple repetitions of the same date parameter). Might be helpful to note that.

```suggestion
optimised indexes when the query string included a date range (i.e. multiple instances of the same date parameter). This issue is fixed."
```

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/rest/param/DateRangeParamR4Test.java
```diff
@@ -114,8 +114,8 @@ public void testSearchForOneUnqualifiedDate() throws Exception {
 
 		assertEquals(parseLowerForDatePrecision("2012-01-01 00:00:00.0000"), ourLastDateRange.getLowerBoundAsInstant());
 		assertEquals(parseUpperForDatePrecision("2012-01-03 00:00:00.0000"), ourLastDateRange.getUpperBoundAsInstant());
-		assertEquals(ParamPrefixEnum.EQUAL, ourLastDateRange.getLowerBound().getPrefix());
-		assertEquals(ParamPrefixEnum.EQUAL, ourLastDateRange.getUpperBound().getPrefix());
+		assertNull(ourLastDateRange.getLowerBound().getPrefix());
+		assertNull(ourLastDateRange.getUpperBound().getPrefix());
```
jamesagnew, line 118: I'd recommend adding a comment here indicating why it's important that this be null - Also should probably include a test in `FhirResourceDaoR4ComboUniqueParamTest` that verifies the actual overall functionality being fixed here.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6603-combo-index-sp-not-invoked-with-dateparam.yaml
epeartree: updated the changelog to include the recommended syntax.

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/rest/param/DateRangeParamR4Test.java
epeartree: Added a comment to the test;
Added a new test to class FhirResourceDaoR4ComboNonUniqueParamTest since 'birthdate' is of type DATE  which will make the search no valid for unique combo searching.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6603-combo-index-sp-not-invoked-with-dateparam.yaml
```diff
@@ -0,0 +1,7 @@
+---
+type: fix
+issue: 6603
+jira: SMILE-8048
+title: "Previously, searches where processing was optimised with a combo index search parameters would skip searching the 
+optimised indexes when the query string included a date ragne (i.e. multiple instances of the same date parameter). This
```
jamesagnew, line 6:
```suggestion
optimised indexes when the query string included a date range (i.e. multiple instances of the same date parameter). This
```

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ComboNonUniqueParamTest.java
```diff
@@ -907,6 +909,44 @@ public void testDateModifier(boolean theUseComparator) {
 
 		}
 
+		@ParameterizedTest
+		@CsvSource(value = {
+			"2025-01-01, true",
+			"eq2025-01-01, false"
```
jamesagnew, line 915: I'm not sure I follow the test. If those two query strings are considered equivalent, why are we treating them differently? I guess it's because we turn this into a date range, but that logic is kinda confusing. Recommend at least adding comments here to explain what's going on with this test and why we should expect a different result for these two values.

jamesagnew (changes requested):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ComboNonUniqueParamTest.java
epeartree: NP at all.  to avoid a regression, this test and its purpose needs be crystal clear on the first read.

jamesagnew (approved):

PR #6578, opened 2024-12-22
Startup and Testpage Overlay tweaks

jamesagnew:
This PR:

* Tweaks the startup process to avoid overloading the database when starting up under heavy load
* Fixes a bunch of display issues in the Testpage Overlay module related to the recent upgrade to Bootstrap 5

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/graphql/GraphQLProviderWithIntrospection.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/graphql/GraphQLProviderWithIntrospection.java
@@ -31,6 +31,8 @@ import ca.uhn.fhir.rest.server.servlet.ServletRequestDetails;
 import ca.uhn.fhir.rest.server.util.ISearchParamRegistry;
 import ca.uhn.fhir.util.StringUtil;
 import ca.uhn.fhir.util.VersionUtil;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonArray;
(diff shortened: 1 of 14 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/graphql/GraphQLProviderWithIntrospection.java
```diff
@@ -153,37 +152,23 @@ private String generateSchema(
 			Collection<String> theResourceTypes,
 			EnumSet<GraphQLSchemaGenerator.FHIROperationType> theOperations) {
 
+		GraphQLSchemaGenerator generator = myGenerator;
```
jmarchionatto, line 155: Why not simply: 
if (myGenerator == null) {
   myGenerator = new GraphQLSchemaGenerator(myContext, VersionUtil.getVersion());
}
and using myGenerator in l:160 ?

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/topic/SubscriptionTopicLoader.java
```diff
@@ -59,6 +61,7 @@ public SubscriptionTopicLoader() {
 	}
 
 	@Override
+	@EventListener(classes = ContextStartedEvent.class)
```
jmarchionatto, line 64: Question: couldn't the overridden method be called twice, once by the super call and one by its annotation?

jmarchionatto (approved):
Looks good. Approved

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/graphql/GraphQLProviderWithIntrospection.java
jamesagnew: The intent here was to do a [threadsafe double-check](https://medium.com/android-news/lazy-initialisation-whats-a-correct-implementation-64c4638561e) initialization, but it wasn't synchronized which it probably should be.. I'm going to just replace this with a call to Guava's memoizer.

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/topic/SubscriptionTopicLoader.java
jamesagnew: Nah, have tested it and it only gets fired once.

PR #6581, opened 2025-01-02
Improve automatic placeholder creation

jamesagnew:
This PR makes 2 changes:

- Optimize the creation of placeholder reference target resources when they are happening within a FHIR transaction. When FHIR transactions will result in the creation of placeholder resources, the number of SELECT queries issued to the database is significantly decreased.
- If placeholder resources are automatically created, details about this are now included in the response OperationOutcome


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
```diff
@@ -2741,6 +2742,51 @@ public void testTransactionWithCreateClientAssignedIdAndReference() {
...
+		bb.addTransactionUpdateEntry(supplier.apply("Patient/P5", "Organization/O2"));
+		Bundle input = bb.getBundleTyped();
+
+		// Test
+		myCaptureQueriesListener.clear();
+		Bundle output = mySystemDao.transaction(mySrd, input);
+		ourLog.info(myFhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(output));
+
+		// Verify
+		assertEquals(1, myCaptureQueriesListener.countSelectQueries());
```
jamesagnew, line 2771: Review note: This action produced 7 select queries before this change, now it's 1. That's the main point here.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
```diff
@@ -442,4 +461,21 @@ public boolean isFhirTransaction() {
...
+	public void addAutoCreatedPlaceholderResource(IIdType theResource) {
+		if (myAutoCreatedPlaceholderResources == null) {
+			myAutoCreatedPlaceholderResources = new ArrayList<>();
+		}
+		myAutoCreatedPlaceholderResources.add(theResource);
+	}
+
+	@Nonnull
+	public List<IIdType> getAutoCreatedPlaceholderResourcesAndClear() {
+		List<IIdType> retVal = myAutoCreatedPlaceholderResources;
```
jmarchionatto, line 474: Wondering if it is worth adding this complexity to avoid initializing the field with an empty list

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java
```diff
@@ -1945,9 +1946,18 @@ private void resolveReferencesThenSaveAndIndexResource(
 
 			theIdSubstitutions.updateTargets(newId);
 
-			if (theDaoMethodOutcome.getOperationOutcome() != null) {
+			IBaseOperationOutcome operationOutcome = theDaoMethodOutcome.getOperationOutcome();
+			if (operationOutcome != null) {
```
jmarchionatto, line 1950: If the outcome could be null, shouldn't be created at this point?

jmarchionatto (approved):
Looks good! Approved with a couple of questions

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
jamesagnew: Yeah, I guess you're right. Reworking this.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java
jamesagnew: It'll only be null if we didn't create one because we don't intend to return one (the client can request not to receive an OO in the response). I'll add a comment.

PR #6583, opened 2025-01-03
Fix batch2 intermittent failure

jamesagnew:
When processing a batch2 job under heavy load, a race condition meant that the final reducer step would occasionally fail if it started immediately following a batch2 maintenance task.


jmarchionatto (approved):
Looks good. Approved

PR #6592, opened 2025-01-06
fix change log

fil512:


fil512:
forgot to update changelog when making enum changes in review feedback

PR #6598, opened 2025-01-07
Fix NPE in Storage with Mass Ingestion and Compartment Interceptor

jdar8:


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -2581,7 +2586,8 @@ protected DaoMethodOutcome doUpdateForUpdateOrPatch(
 				theResource,
 				theEntity,
 				theOperationType,
-				theTransactionDetails);
+				theTransactionDetails,
+				theShouldForcePopulateOldResourceForProcessing);
```
TipzCM, line 2590: We're at 11 parameters here.

We probably should've wrapped these in a parameters object 9 parameters ago.

What would be the impact of doing so now?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseStorageResourceDao.java
```diff
@@ -226,6 +226,30 @@ protected DaoMethodOutcome doUpdateForUpdateOrPatch(
...
+				thePerformIndexing,
+				theForceUpdateVersion,
+				theResource,
+				theEntity,
+				theOperationType,
+				theTransactionDetails,
+				false);
+	}
+
+	protected DaoMethodOutcome doUpdateForUpdateOrPatch(
```
TipzCM, line 242: This multiple method thing is exactly why i think we should make a new parameters object and have that API be the 'main' one.

Multiple overrides and multiple methods will lead to funny bugs if people make methods with different 'combinations' of parameters.

TipzCM (approved):
minor changes

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
jdar8: Refactored so that it now takes in a parameters object

PR #6605, opened 2025-01-09
Refactor RuleBuilder to allow for extension of Rules

jbonzohln:
I was trying to add a custom rule to our project, where we could allow exporting of all Groups instead of individual Groups, and found it difficult to implement. This was because most of the classes were in package or private scope, so I would need to re-implement (copy) existing HAPI classes into my project. By making  the classes protected, I'm able to easily create a custom rule. There may be more that could be refactored, but this got me to where I needed the code.

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/AppliesTypeEnum.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/AppliesTypeEnum.java
@@ -19,7 +19,7 @@
  */
 package ca.uhn.fhir.rest.server.interceptor.auth;
 
-enum AppliesTypeEnum {
+public enum AppliesTypeEnum {
 	ALL_RESOURCES,
 	TYPES,
 	INSTANCES
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/BaseRule.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/BaseRule.java
@@ -33,12 +33,12 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
-abstract class BaseRule implements IAuthRule {
-	private String myName;
+public abstract class BaseRule implements IAuthRule {
+	private final String myName;
 	private PolicyEnum myMode;
 	private List<IAuthRuleTester> myTesters;
 
-	BaseRule(String theRuleName) {
+	protected BaseRule(String theRuleName) {
 		myName = theRuleName;
 	}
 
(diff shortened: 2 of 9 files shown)


jamesagnew:
This is a neat idea, we didn't really think much about extensibility when we designed this framework.

Would you be willing to add an example of such a custom rule? I think this would be greatly enhanced if we had documentation to show how someone would use this.

jbonzohln:
The rule I'd like to add is one that allows/disallows bulk export of groups. The currently available rules only allow for creating a rule for specific groups, so I'd need to create a rule for every Group defined in our system. 

The rule would be something like this:
```
private static class CustomRuleBulkExportImpl implements IAuthRule
    {
        private final String name;
        private final PolicyEnum mode;
        private final Collection<String> resourceTypes;

        private CustomRuleBulkExportImpl(String name, PolicyEnum mode, Collection<String> resourceTypes)
        {
            this.name = name;
            this.mode = mode;
            this.resourceTypes = resourceTypes;
        }

        @Override
        public String getName()
        {
            return name;
        }

        // see ca.uhn.fhir.rest.server.interceptor.auth.RuleBulkExportImpl
        @Override
        public AuthorizationInterceptor.Verdict applyRule(RestOperationTypeEnum theOperation, RequestDetails theRequestDetails,
                IBaseResource theInputResource, IIdType theInputResourceId, IBaseResource theOutputResource,
                IRuleApplier theRuleApplier, Set<AuthorizationFlagsEnum> theFlags, Pointcut thePointcut)
        {
            if (theRequestDetails == null)
            {
                return null;
            }

            BulkExportJobParameters inboundBulkExportRequestOptions = (BulkExportJobParameters) theRequestDetails.getAttribute(
                    AuthorizationInterceptor.REQUEST_ATTRIBUTE_BULK_DATA_EXPORT_OPTIONS);
            if (inboundBulkExportRequestOptions == null)
            {
                return null;
            }

            // if style doesn't match - abstain
            if (inboundBulkExportRequestOptions.getExportStyle() != BulkExportJobParameters.ExportStyle.GROUP)
            {
                return null;
            }

            if (isNotEmpty(resourceTypes))
            {
                if (isEmpty(inboundBulkExportRequestOptions.getResourceTypes()))
                {
                    return new AuthorizationInterceptor.Verdict(PolicyEnum.DENY, this);
                }
                if (!resourceTypes.containsAll(inboundBulkExportRequestOptions.getResourceTypes()))
                {
                    return new AuthorizationInterceptor.Verdict(PolicyEnum.DENY, this);
                }
            }
            else
            {
                // if no types were specified, we're going to consider that a deny unlike HAPI's implementation
                // this way we need to specifically allow the types to be exported
                return new AuthorizationInterceptor.Verdict(PolicyEnum.DENY, this);
            }

            return new AuthorizationInterceptor.Verdict(mode, this);
        }
    }
```

It would be nice to be able to use the current builder interfaces to create this rule as well. I could currently just create this rule and put it in the list of rules created by RuleBuilder, but that seems hacky.
Something like this, where `groupExportOnAnyGroup` is a new method I added:
                        `ruleBuilder.allow().bulkExport().groupExportOnAnyGroup().withResourceTypes(resourceTypesAllowedToRead).andThen();`

PR #6606, opened 2025-01-09
Moving checks from azure to gh actions

markiantorno:
This commit adds a workflow, with associated actions to build and test HAPI, then upload the coverage results to CODECOV. 

This was previously done on Azure Pipelines.

Please do not set this to automerge.


.github/actions/build-module/action.yml
```diff
@@ -0,0 +1,95 @@
...
+      with:
+        distribution: 'temurin'
+        java-version: '11'
+
+    - name: Install Docker
+      shell: bash
+      run: |
+        curl -fsSL https://get.docker.com -o get-docker.sh
+        sudo sh get-docker.sh
+        docker --version
```
tadgh, line 33: note: seems weird there is no default task for getting docker. Is there a reason you're home-rolling adding docker>?

.github/actions/build-module/action.yml
```diff
@@ -0,0 +1,95 @@
...
+    
+    - name: Create and List Maven Cache Directory
+      shell: bash
+      run: |
+        mkdir -p $HOME/.m2/repository
+        pwd
+        ls -al $HOME/.m2/repository
+      env:
+        MAVEN_CACHE_FOLDER: $HOME/.m2/repository
```
tadgh, line 42: Question: Is the goal of this to just ensure the folder is empty?

.github/actions/build-module/action.yml
```diff
@@ -0,0 +1,95 @@
...
+        MAVEN_CACHE_FOLDER: $HOME/.m2/repository 
+
+    - name: Restore Maven Cache
+      uses: actions/cache@v4
+      with:
+        path: $HOME/.m2/repository
+        enableCrossOsArchive: true  
+        key: maven-${{ runner.os }}-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+            maven-${{ runner.os }}-
```
tadgh, line 51: question: What is a restore-key vs a key?

.github/actions/build-module/action.yml
```diff
@@ -0,0 +1,95 @@
...
+
+    - name: Verify cache was restored
+      shell: bash
+      run: |
+        if [ -d "$HOME/.m2/repository" ]; then
+          echo "Cache exists"
+          ls -R $HOME/.m2/repository
+        else
+          echo "Cache not found"
+        fi
```
tadgh, line 61: Note: This will always be true, because in the Create and List , you create the empty folder. Is this intentional?

.github/actions/build-module/action.yml
```diff
@@ -0,0 +1,95 @@
...
+      with:
+        name: ${{ inputs.name }}_full_logs_${{ github.run_id }}_${{ github.run_number }}
+        path: $GITHUB_WORKSPACE/reports
+
+    - name: Upload Target Artifacts
+      uses: actions/upload-artifact@v3
+      if: always()
+      with:
+        name: ${{ inputs.name }}_target
+        path: ${{ github.workspace }}/${{ inputs.module }}/target
```
tadgh, line 95: doesnt the whole target folder include the logs? WHy two separate uploads?

.github/actions/cache-maven/action.yml
```diff
@@ -0,0 +1,64 @@
+name: "Build Cache Maven Dependencies"
+description: "Caches Maven dependencies, resolves plugins, and builds if the cache is not hit."
+inputs:
+  java-version:
+    description: "The Java version to use"
+    required: true
+    default: "17"
```
tadgh, line 7: issue: Here we use 17, but [above](https://github.com/hapifhir/hapi-fhir/pull/6606/files#diff-428a7177b54ca6b2b4028db835b68f961fa03230d5387fc455284013b13f8497R26), we are pinned to 11. Recommend 17 across the board.

.github/actions/cache-maven/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        distribution: "temurin"
+
+    - name: Cache Maven dependencies
+      id: cache-maven
+      uses: actions/cache@v4
+      with:
+        path: ${{ inputs.cache-path }}
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
```
tadgh, line 29: Seems this is a duplicate of an earlier action. Is there any way to reuse the logic across both spots? or at least  the key definition?

.github/actions/cache-maven/action.yml
```diff
@@ -0,0 +1,64 @@
...
+
+    - name: Populate Cache with Remote Dependencies
+      shell: bash      
+      if: steps.cache-maven.outputs.cache-hit != 'true'
+      env:
+        MAVEN_CACHE_FOLDER: ${{ inputs.cache-path }}
+      run: |
+        mvn dependency:resolve-plugins dependency:resolve test-compile \
+          -Dmaven.repo.local=$MAVEN_CACHE_FOLDER \
+          -Dstyle.color=always -Djansi.force=true
```
tadgh, line 39: wow this is nice. Does it actually pull failsafe and surefire _correctly_? if so, i have to steal this code for gitlab-ci...

.github/actions/cache-maven/action.yml
```diff
@@ -0,0 +1,64 @@
...
+        MAVEN_CACHE_FOLDER: ${{ inputs.cache-path }}
+        MAVEN_OPTS: '-Xmx1024m -Dorg.slf4j.simpleLogger.showDateTime=true -Dorg.slf4j.simpleLogger.dateTimeFormat=HH:mm:ss,SSS -Duser.timezone=America/Toronto'
+      run: |
+        mvn clean install -P CI,CHECKSTYLE \
+          -Dmaven.test.skip=true -e -B \
+          -Dmaven.javadoc.skip=true \
+          -Dmaven.wagon.http.pool=false \
+          -Dmaven.repo.local=$MAVEN_CACHE_FOLDER \
+          -Dhttp.keepAlive=false \
+          -Dstyle.color=always -Djansi.force=true
```
tadgh, line 54: question: Are you confident that all deps are stored in the cache at this point? If so, try running with `-o` to force offline mode, and see if it explodes.

.github/actions/build-module/action.yml
dotasek, line 1: This doesn't appear to be referred to anywhere, and does very similar things to run-tests-and-archive-results. Is this here intentionally for later re-use or just left over work?

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,86 @@
...
+    runs-on: ubuntu-latest
+    needs: 
+      - generate_module_list
+      - test-module
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      
+      - uses: ./.github/actions/assemble-test-reports
```
dotasek, line 80: Nitpick: a name here would help.

.github/actions/compile-test-results/action.yml
```diff
@@ -0,0 +1,68 @@
...
+          fi
+
+          # Create target directory
+          mkdir -p "${GITHUB_WORKSPACE}/${module}/target/"
+          
+          # Move all contents from the artifact directory to the target directory
+          echo "Moving contents of $artifact_dir to $target_dir..."
+          mv "${GITHUB_WORKSPACE}/artifacts/${cleaned_name}_target"/* "${GITHUB_WORKSPACE}/${module}/target/"
+          echo "Moved contents for $module."
+        done
```
tadgh, line 68: I'm a bit confused, what does this do? forces the `target/` dir artifacts to move to where they are meant to live? If that's the case, why rename them in the first place, e.g. `whatever_target`. By default, if you upload an artifact, does it not retain its existing path? e.g. if i mark `**/target` as an artifact, when i go to unzip that, won't it resolve to
```
hapi-fhir-jpaserverbase/target/blah
hapi-fhir-storage/target/blah
etc
```

At least that's the way artifacts work in gitlab CI.

.github/actions/generate-module-list/action.yml
```diff
@@ -0,0 +1,91 @@
...
+  using: 'composite'
+  steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set up JDK 11
+      uses: actions/setup-java@v4
+      with:
+        distribution: 'adopt'
+        java-version: '11'
```
tadgh, line 22: 17 again

dotasek (changes requested):
Two requests:

* Is there any place where a run of this is visible?
* See the comment regarding unreferenced build-module

.github/actions/generate-module-list/action.yml
```diff
@@ -0,0 +1,91 @@
...
+        ignored_modules="${{ inputs.ignored_modules }}"  # The ignored modules list
+        ignored_modules_array=(${ignored_modules//,/ })   # Convert to an array
+        modules_list=""
+    
+        # Function to recursively find modules and maintain the directory structure
+        find_modules() {
+          local current_dir=$1
+          for dir in "$current_dir"/*/; do
+            if [[ -f "$dir/pom.xml" ]]; then
+              module_name=$(basename "$dir" | tr -d '\n')  # Ensure no newline characters
```
tadgh, line 45: Question: Could you not just use the `<modules>` xml tag in the top-level pom.xml, as it lists all modules in the reactor?

.github/actions/run-tests-and-archive-results/action.yml
```diff
@@ -0,0 +1,74 @@
...
+        find . -wholename '**/target/*-reports/*.txt' -exec cp {} ${{ github.workspace }}/artifacts \;
+
+    - name: Clean Module Name
+      id: clean_module_name
+      shell: bash
+      if: always()
+      run: |
+        module_name="${{ inputs.module-name }}"
+        cleaned_name=$(basename "$module_name")
+        echo "cleaned_name=$cleaned_name" >> $GITHUB_OUTPUT
```
tadgh, line 60: question: what in tarnation is this doin

.github/workflows/pull-request.yml
```diff
@@ -0,0 +1,86 @@
...
+        run: |
+          # Use jq to format the modules list as a JSON array
+          echo "modules_list=$(jq -R -s -c 'split(",") | map(gsub("[\\n]"; ""))' <<< '${{ steps.generate_list.outputs.modules_list }}')" >> $GITHUB_OUTPUT
+          
+      - name: Print formatted modules as JSON
+        run: |
+          echo "Formatted JSON Array:"
+          echo "${{ steps.format_modules.outputs.modules_list }}"
+
+  test-module:
```
tadgh, line 47: issue:  alternating between `snake_case` and `kebab-case` for job names. recommend unifying to `kebab-case`

tadgh (approved):

.github/actions/generate-module-list/action.yml
markiantorno: I wanted a fullproof way to detect added modules, that may or not be added to the base pom?

Might be overkill...i dunno.

PR #6617, opened 2025-01-14
renamed $replace-references operation

mrdnctrk:
This MR renames `$replace-references` operation to `$hapi.fhir.replace-references`


tadgh (approved):
Approved.  Are there any docs in HAPI that need to be updated due to this?

mrdnctrk:
> Approved. Are there any docs in HAPI that need to be updated due to this?

no, there is no documentation on the HAPI.

PR #6632, opened 2025-01-17
Resolve deadlock in subscription registry

jamesagnew:


tadgh (approved):
Looks like you moved the synchronizing over  into the map,  and removed synchronization on the cache itself? I don't understand why this fixes it, but I assume you do!

fil512 (approved):
Nice!

PR #6645, opened 2025-01-23
allow deleting resources if references are all versioned only

### LINKED ISSUE #6644: Allow deleting of resources that are only referenced with versioned references
Allow deleting of resources if the only contained references are versioned references.

Eg:

An Encounter resource can contain references to an Observation resource under `Encounter.reason`.

If the reference is versioned (`Encounter.reason = Observation/1/_history/1`), then deleting the observation (`Observation/1`) should be allowed.

But, if the reference is versionless (`Encounter.reason = Observation/1`), deleting the observation should not be allowed.

TipzCM:
closes #6644

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6644-allow-deleting-resources-if-only-references-are-versioned.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6644-allow-deleting-resources-if-only-references-are-versioned.yaml
@@ -2,7 +2,10 @@
 type: add
 issue: 6644
 jira: SMILE-9604
-title: "If references are allowed to contain versioned references,
-        and a resource is only referenced using versioned references,
-        DELETE actions will be permitted.
+title: "By default, referential integrity is enforced on deletes.
+        This change introduces support for an exceptional case such
+        that deletion of a given resource will not be blocked if all
+        references to that resource are versioned. If there is at
+        least one unversioned reference to the resource, deletion will
+        still be blocked.
 "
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/DeleteConflictFinderService.java
```diff
@@ -35,10 +35,12 @@ public class DeleteConflictFinderService {
...
 
 	List<ResourceLink> findConflicts(ResourceTable theEntity, int maxResults) {
-		TypedQuery<ResourceLink> query = myEntityManager.createQuery(
-				"SELECT l FROM ResourceLink l WHERE l.myTargetResource.myPid = :target_pid", ResourceLink.class);
+		String queryStr = "SELECT l FROM ResourceLink l WHERE l.myTargetResource.myPid = :target_pid AND (l.myTargetResourceVersion IS NULL)";
+		TypedQuery<ResourceLink> query = myEntityManager.createQuery(queryStr, ResourceLink.class);
 		query.setParameter("target_pid", theEntity.getId());
 		query.setMaxResults(maxResults);
-		return query.getResultList();
+		List<ResourceLink> resourceLinks = query.getResultList();
```
jmarchionatto, line 42: nitpick: return directly to avoid warning

jmarchionatto (approved):
Looks good! Approved

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6644-allow-deleting-resources-if-only-references-are-versioned.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: add
+issue: 6644
+jira: SMILE-9604
+title: "If references are allowed to contain versioned references,
+        and a resource is only referenced using versioned references,
+        DELETE actions will be permitted.
+"
```
dmuylwyk, line 8: **suggestion:** Clarify the change and new behaviour.

```suggestion
title: "By default, referential integrity is enforced on deletes. This change introduces
        support for an exceptional case such that deletion of a given resource will not be
        blocked if all references to that resource are versioned. If there is at least one
        unversioned reference to the resource, deletion will still be blocked.
"
```

TipzCM: ### updated

Issue #6554, opened 2024-12-11
Concurrency issue in WorkerContextValidationSupportAdapter


Boereck:
Hi, 
after upgrading to HAPI 7.6.0 we got `ConcurrentModificationException`s in our validation test runs, which uses the same `FhirValidator` instance (which the documentation claims is thread safe) to validate many example resources in parallel. It was quite tricky to find the source of the error, due to the concurrent nature, but we are certain there is a problem in the `VersionSpecificWorkerContextWrapper`.

**Describe the bug**
The problematic function is `validateCodeInValueSet`:
https://github.com/hapifhir/hapi-fhir/blob/a999cb32c5b916d9911762656198a32d1cd24199/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java#L803-L834

In this function `.getRootValidationSupport().validateCodeInValueSet(...)` is called to get a `CodeValidationResult`. Since our root ValidationSupport is a `CachingValidationSupport`, this may return a cached object. Later it is checked if the code is valid in the CodeSystem. If this is not the case, the issues will be added to the result from the ValueSet validation, which may be a cached object!
In this case multiple parallel validations may write to the issues list of the result, causing `ConcurrentModificationException`s when reading the issue list later.

In future versions the `ValidationSupportChain` will do the caching and provoke the same problem.

Fortunately, the fix is pretty staight forward. Since the `validateCodeInValueSet` returns a result object, in the 
```java
if (codeSystemResult != null) {
	//...
}
```
block, `result` can be assigned to a copy of `result`. This can be implemented in `VersionSpecificWorkerContextWrapper` for the moment, but ideally `IValidationSupport.CodeValidationResult` could provide a copy constructor, or a copy instance method in future. However this class is in the `org.hl7.fhir.core` project, so this would have to be added upstream.
Since all objects referenced from `IValidationSupport.CodeValidationResult` (apart from the list-fields) are immutable, it would suffice to do a shallow copy, and just create list copies of the list-fields.

If wanted, we can provide a PR to fix this issue. We already tried the change and we did not face any `ConcurrentModificationException` after the change. We could also back-port this issue to the 7.6.0 stream, we just need to know against which branch the PR should go.

**To Reproduce**
[...]

PR #6648, opened 2025-01-24
Fixing concurrency issue in VersionSpecificWorkerContextWrapper (#6554)

Boereck:
When validating if a code is valid against a ValueSet, a potentially cached (and therefore shared) result object was mutated. This mutation is now done on a copy of the result object.

Without this change, there might be a concurrent modification on a cached result of `validateCodeInValueSet` (item added to issue list). We got `ConcurrentModificationException`s when validating resources with the same problem in parallel.

Please let me know if there is an issue with this PR. Unfortunately, due to the concurrent nature of the problem, it seems impossible to test for the absence of the fixed error deterministically.


Boereck:
I have added a test case to `org.hl7.fhir.common.hapi.validation.validator.VersionSpecificWorkerContextWrapperTest` that actually tests for the problem. Since this is a concurrency bug, it uses concurrent access to trigger the problem. This, unfortunately, makes the test case non-deterministic (the test case could theoretically pass, despite the bug being present), even though it works every time on my machine. I did split the change into two commits. The first commit contains the test case, and if this commit is checked out (without the second commit), the test case will fail. The second commit contains the fix, so the test case will pass.

robogary:
Formatting check succeeded!

Boereck:
Updated PR to adjust to current master.

Boereck:
I would appreciate feedback on this PR, if there are adjustments needed, please let me know. The change is fairly limited, and we run into this issue frequently.

jamesagnew (approved):
@Boereck Apologies for missing this. This looks good to me, thanks for the contribution!

PR #6651, opened 2025-01-24
Feature/openapi spec generation improvements 2

nathanloyer:
My team needs to add custom extensions to the generated openapi spec, so I've put together some changes that will allow us to override the OpenApiInterceptor and customize what gets generated. We're also going to be using this to append more detailed json schema definitions into the spec so developers can quickly see what the FHIR objects look like without jumping out to read the FHIR Docs.

Let me know what you think and if I need to make any adjustments before we can merge. I have been mostly testing with direct integration with my server so far, but I will add more rigorous testing if you agree that this is a useful feature that we can merge.


jamesagnew (approved):
This looks good to me, thanks for the contribution! Will merge if is passes CI.

jamesagnew:
Hmm, why is Github not running CI.. I think I screwed this up by pushing the credit to the wrong place. I'm going to resubmit your changes as a fresh PR (will make sure you get credit though!)

jamesagnew:
The resubmitted version has merged.

PR #6684, opened 2025-02-05
made $hapi.fhir.replace-references not replace versioned references

mrdnctrk:
closes: #6689  

The versioned references are usually kept for historical record keeping purposes such as in Provenance.target .  For that reason, it is important to keep them as they are. This MR adds a changes so that versioned references are not replaced by the $hapi.fhir.replace-references operation.


mrdnctrk:
> Looks good, but needs a supporting issue and changelog

thanks, created issue #6689, and  added the changelog.

jmarchionatto (approved):
Approved

PR #6687, opened 2025-02-06
Fixing Failure when creating a cross-partition subscription if the default partition id is not null

### LINKED ISSUE #6686: Failure when creating a cross-partition subscription if the default partition id is not null.
Attempting to create a cross-partition subscription fails when the ID of the default partition was configured to a value that is different of the default value (null).

epeartree:
**Background:**
- The default value of the ID for the default partition is null;
- clients can provide another value for the default partition id through the partitionSettings.setDefaultPartitionId

**Root cause of the issue:**
invocation of RequestPartitionId.isDefaultParttition() only considers that the default partition id is and should be null;

**What was done:**
- added failing test;
- added api to an existing helper service to perform default partition check against the client configured defaultPartition Id (PartitionSettings);

Closes #6686

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/model/RequestPartitionId.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/model/RequestPartitionId.java
@@ -243,7 +243,7 @@ public class RequestPartitionId implements IModelJson {
 				&& Objects.equals(getPartitionIds().get(0), theDefaultPartitionId);
 	}
 
-	public boolean hasPartitionId(@Nullable Integer thePartitionId) {
+	public boolean hasPartitionId(Integer thePartitionId) {
 		Validate.notNull(myPartitionIds, "Partition IDs not set");
 		return myPartitionIds.contains(thePartitionId);
 	}
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/model/RequestPartitionId.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/model/RequestPartitionId.java
@@ -222,15 +222,28 @@ public class RequestPartitionId implements IModelJson {
 	 * Returns true if this request partition contains only one partition ID and it is the DEFAULT partition ID (null)
 	 */
 	public boolean isDefaultPartition() {
+		return isDefaultPartition(null);
+	}
+
+	/**
+	 * Test whether this request partition is for a given default partition ID.
+	 *
+	 * @param theDefaultPartitionId is the ID that was given to the default partition.  The default partition ID can be
+	 *                              NULL as per default or specifically assigned another value.
+	 *                              See PartitionSettings#setDefaultPartitionId.
+	 * @return <code>true</code> if the request partition contains only one partition ID and the partition ID is
+	 *         <code>theDefaultPartitionId</code>.
+	 */
+	public boolean isDefaultPartition(@Nullable Integer theDefaultPartitionId) {
 		if (isAllPartitions()) {
 			return false;
 		}
 		return hasPartitionIds()
 				&& getPartitionIds().size() == 1
-				&& getPartitionIds().get(0) == null;
+				&& Objects.equals(getPartitionIds().get(0), theDefaultPartitionId);
 	}
 
-	public boolean hasPartitionId(Integer thePartitionId) {
+	public boolean hasPartitionId(@Nullable Integer thePartitionId) {
 		Validate.notNull(myPartitionIds, "Partition IDs not set");
 		return myPartitionIds.contains(thePartitionId);
 	}
@@ -244,7 +257,11 @@ public class RequestPartitionId implements IModelJson {
 	}
 
 	public boolean hasDefaultPartitionId() {
-		return getPartitionIds().contains(null);
+		return hasDefaultPartitionId(null);
+	}
+
+	public boolean hasDefaultPartitionId(@Nullable Integer theDefaultPartitionId) {
+		return getPartitionIds().contains(theDefaultPartitionId);
 	}
 
 	public List<Integer> getPartitionIdsWithoutDefault() {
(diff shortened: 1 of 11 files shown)


hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/model/RequestPartitionId.java
```diff
@@ -222,15 +222,28 @@ public String getFirstPartitionNameOrNull() {
...
 		}
 		return hasPartitionIds()
 				&& getPartitionIds().size() == 1
-				&& getPartitionIds().get(0) == null;
+				&& Objects.equals(getPartitionIds().get(0), theDefaultPartitionId);
 	}
 
-	public boolean hasPartitionId(Integer thePartitionId) {
+	public boolean hasPartitionId(@Nullable Integer thePartitionId) {
 		Validate.notNull(myPartitionIds, "Partition IDs not set");
```
TipzCM, line 247: you might want to remove this if we allow nulls

TipzCM:
Request: can you put a logline in SubscriptionCanonicalizer to log something if no subscriptions match?

Just to make tracking bugs down here a bit easier next time?

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/model/RequestPartitionId.java
```diff
@@ -244,7 +257,11 @@ public boolean hasPartitionNames() {
 	}
 
 	public boolean hasDefaultPartitionId() {
-		return getPartitionIds().contains(null);
+		return hasDefaultPartitionId(null);
```
TipzCM, line 260: this isn't the same as what `isDefaultParition` was doing; presumably (i assume) one can have multiple partition ids and one of them is the default one (but the others are not).

Perhaps this method should take the default partition id as well;

EDIT: i see you have such a method below. I think this method needs to be removed because it's not going to return anything usefulf.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6686-failure-creating-cross-partition-subscription.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 6686
+title: "Previously, attempting to create a cross-partition subscription would fail when the ID of the default partition 
+was configured to a value that is different of the default value (null). This issue is fixed."
```
TipzCM, line 5: I think i would make this more generalized, considering all the issues we found:

`Previously, assigning a default partition id different than the default value (null) would result in issues with cross-partition subscriptions being not respected. This has been fixed`

but i'm just nitpicking now

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/partition/IRequestPartitionHelperSvc.java
```diff
@@ -182,4 +182,15 @@ RequestPartitionId determineCreatePartitionForRequest(
...
+
+	/**
+	 * Test whether <code>theRequestPartitionId</code> is targeting the default partition.  Implementers should overwrite
+	 * this method to enhance the default behavior which is to invoke <code>theRequestPartitionId#isDefaultPartition</code>
+	 *
+	 * @param theRequestPartitionId to perform the evaluation upon.
+	 * @return true if the <code>theRequestPartitionId</code> is for the default partition.
+	 */
+	default boolean isDefaultPartition(@Nonnull RequestPartitionId theRequestPartitionId) {
+		return theRequestPartitionId.isDefaultPartition();
```
TipzCM, line 194: As a suggestion, since i think you're doing this to avoid having to change whoever the implementer of this interface is, i would instead suggest adding yet *another* default implementation method here that returns the default partition id (defaults to null)

Then you can have this method instead look like this:

```
default Integer getDefaultPartitionId() {
    return null;
}

default boolean isDefaultPartition(@Nonnull RequestPartitionId theRequestPartitionId) {
    return theRequestPartitionId.isDefaultPartition(getDefaultPartitionId());
}
```

This way at least the implementer has to provide the correct defualt partition id; whereas if they do not, they will never know that the underlying (default) implementation here is going to almost always be wrong.

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/model/RequestPartitionId.java
```diff
@@ -222,12 +222,25 @@ public String getFirstPartitionNameOrNull() {
 	 * Returns true if this request partition contains only one partition ID and it is the DEFAULT partition ID (null)
 	 */
 	public boolean isDefaultPartition() {
+		return isDefaultPartition(null);
```
TipzCM, line 225: Can you slap an `@Deprecated` on here?

I get you're doing this to avoid changing hundreds of lines and/or lines in dependent projects.

But it's basically incorrect to use this. So we should flag that as not a valid method.

TipzCM (approved):
Also, please add a log line in `SubscriptionMatchingSubscriber` so that when no subscriptions match, we have some loglines (warnings is my suggestion, but can ask in design)

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/7_8_0/6686-failure-creating-cross-partition-subscription.yaml
epeartree: massaged the description a bit but i'll keep it in the same lines as this PR provides the constructions to fix issues related to subscriptions and default partition but only fixes a specific scenario.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/partition/IRequestPartitionHelperSvc.java
epeartree: oooo, love it

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/model/RequestPartitionId.java
epeartree: done

PR #6691, opened 2025-02-07
Fix _security label filtering with :not operator in the in-memory matcher

nathandoef:
Previously, when the in-memory matcher was used to match resources with a `_security` label filter 
and a `:not` operator (i.e. `_security:not=http://terminology.hl7.org/CodeSystem/v3-ActCode|NODSCLCD`), resources with no security labels at all were not matched. This has been fixed.


jmarchionatto (approved):
Looks good! Approved

PR #6698, opened 2025-02-10
Account for case differences in DBPM contributor

jamesagnew:
This change fixes 2 regressions caused by Database Partitioning Mode:

* The mapping contributor was accidentally case sensitive when comparing table and column names, which meant that it could miss columns when running in an environment using lower-case identifiers (this is the case in the JPA Starter project)
* Several partitioned foreign keys in the terminology tables were not set as `insertable=false` which caused them to fail when used in non-DBPM mode.

### DIFF
--- a/hapi-fhir-jpa-hibernate-services/src/main/java/ca/uhn/hapi/fhir/sql/hibernatesvc/DatabasePartitionModeIdFilteringMappingContributor.java
+++ b/hapi-fhir-jpa-hibernate-services/src/main/java/ca/uhn/hapi/fhir/sql/hibernatesvc/DatabasePartitionModeIdFilteringMappingContributor.java
@@ -272,15 +272,16 @@ public class DatabasePartitionModeIdFilteringMappingContributor
 			if (remove != null) {
 				iter.remove();
 				idRemovedColumns.addAll(property.getColumns());
-				idRemovedColumnNames.addAll(
-						property.getColumns().stream().map(Column::getName).collect(Collectors.toSet()));
+				idRemovedColumnNames.addAll(property.getColumns().stream()
+						.map(c -> c.getName().toUpperCase(Locale.ROOT))
+						.collect(Collectors.toSet()));
 				property.getColumns().stream()
 						.map(theColumn -> new TableAndColumnName(table.getName(), theColumn.getName()))
 						.forEach(myQualifiedIdRemovedColumnNames::add);
 				idRemovedProperties.add(property.getName());
 
 				for (Column next : entityPersistentClass.getTable().getColumns()) {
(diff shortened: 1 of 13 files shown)


dotasek (changes requested):
This passes all integration tests and smoke tests on hapi-fhir-jpaserver-starter.

However, there is an exception that appears during initialization.

To replicate, build hapi-fhir-jpaserver-starter, and then run it with the following command:

` mvn -P jetty spring-boot:run`

I can see the following in the startup log:

```log
Error executing DDL "create index IDX_IDXCMPSTRUNIQ_RESOURCE on hfj_idx_cmp_string_uniq ()" via JDBC [Syntax error in SQL statement "create index IDX_IDXCMPSTRUNIQ_RESOURCE on hfj_idx_cmp_string_uniq ([*])"; expected "identifier";]

org.hibernate.tool.schema.spi.CommandAcceptanceException: Error executing DDL "create index IDX_IDXCMPSTRUNIQ_RESOURCE on hfj_idx_cmp_string_uniq ()" via JDBC [Syntax error in SQL statement "create index IDX_IDXCMPSTRUNIQ_RESOURCE on hfj_idx_cmp_string_uniq ([*])"; expected "identifier";]
```

This doesn't appear to be fatal, as the server still starts, and smoke tests run successfully, but there could be some other undiagnosed breakage that this causes.

hapi-fhir-jpa-hibernate-services/src/main/java/ca/uhn/hapi/fhir/sql/hibernatesvc/DatabasePartitionModeIdFilteringMappingContributor.java
```diff
@@ -552,4 +584,33 @@ private static Class<?> getType(ClassLoaderService theClassLoaderService, String
 		Validate.notNull(entityType, "Could not load type: %s", theEntityTypeName);
 		return entityType;
 	}
+
+	private static class TableAndColumnName {
```
michaelabuckley, line 588: changelog?

michaelabuckley (approved):

hapi-fhir-jpa-hibernate-services/src/main/java/ca/uhn/hapi/fhir/sql/hibernatesvc/DatabasePartitionModeIdFilteringMappingContributor.java
jamesagnew: This is addressing a regression introduced in the same version, so I don't think it makes sense in this case

dotasek (approved):
Subsequent commits have fixed the exceptions on jpa startup. Looks good to me!

PR #6678, opened 2025-02-04
5368-Lenient-error-handler-doesn't-always-include-location-information

elavy-harris:
Additional logging info, per #5368

Issue #5368, opened 2023-10-17
Lenient error handler doesn't always include location information


elavy-harris:
**Describe the issue**
Warning messages created without context. Example:
`2023-10-17 09:40:44.629 | 10:40:44.629 [hl7v2-receiver-endpoint_hl7v2_in_adt_siu-19] WARN  c.u.fhir.parser.LenientErrorHandler - Resource is missing required element: value`

**Environment (please complete the following information):**
 - HAPI FHIR Version: 6.8.0, build f33627087f 2023-08-17
 - OS: Linux

elavy-harris:
Recommended mod to https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/LenientErrorHandler.java:

```
public void missingRequiredElement(IParseLocation theLocation, String theElementName) {
	if (myLogErrors) {
		ourLog.warn("{}Resource is missing required element: {}", describeLocation(theLocation), theElementName);
	}
}
```

There are other calls to `ourLog.warn()` that would benefit similarly.

PR #6705, opened 2025-02-12
Typo fix from 5368

elavy-harris:


elavy-harris:
Typo in PR #6678 (fix for #5368).

PR #6748, opened 2025-02-18
With partitioning selection mode set to request tenant, creating a cross partition subscription will fail if the default partition id is not null

### LINKED ISSUE #6747: With partitioning selection mode set to REQUEST_TENANT, creating a cross-partition subscription will fail if the default partition id is not null.
With partitioning enabled and partition selection mode is set to REQUEST_TENANT,

attempting to create a cross-partition subscription fails when the ID of the default partition was configured to a value that is different of the default value (null).

epeartree:
**Background:**

- The default value of the ID for the default partition is null;
- clients can provide another value for the default partition id through the partitionSettings.setDefaultPartitionId

**Root cause of the issue:**
Issuing POST /DEFAULT/Subscription {} would result in  the RequestPartitionHelperSvc normalising the DEFAULT partition name to the default partition id of null.

**What was done:**

- added failing test;
- resolving the defaultPartitionId to PartitionSettings.getDefaultPartitionId instead of null;
- added changelog;
- updated changlog for #6686

Closes #6747

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/RequestPartitionHelperSvc.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/RequestPartitionHelperSvc.java
@@ -144,7 +144,7 @@ public class RequestPartitionHelperSvc extends BaseRequestPartitionHelperSvc {
 				if (partition != null) {
 					ids.add(partition.getId());
 				} else {
-					ids.add(null);
+					ids.add(myPartitionSettings.getDefaultPartitionId());
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/partition/RequestPartitionHelperSvc.java
```diff
@@ -144,7 +144,7 @@ public RequestPartitionId validateAndNormalizePartitionNames(RequestPartitionId
 				if (partition != null) {
 					ids.add(partition.getId());
 				} else {
-					ids.add(null);
+					ids.add(myPartitionSettings.getDefaultPartitionId());
```
tadgh, line 147: man this stuff is brutal. we need to once over the repo

tadgh (approved):

PR #6754, opened 2025-02-21
Ignore contained resources in IPA generation reference check so they are not removed

### LINKED ISSUE #6708: IPA generator removes references to contained resources
**Describe the bug**
In the IPS generator, the references [are cleared out](https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-jpaserver-ips/src/main/java/ca/uhn/fhir/jpa/ips/generator/IpsGeneratorSvcImpl.java#L228) when it's pointing to resources that are not included in the bundle. However, this clears out resources that are contained as well. This creates a problem when Medication[x].medicationReference points to a contained Medication resource (which I assume is a common approach?).

**To Reproduce**
1. Turn on IPA generation in HAPI JPA Server.
2. Add a Medication[x] resource eg. MedicationStatement with medicationReference point to a contained Medication resource to a Patient.
3. Call $summary operation on the patient.
4. In the response, the reference to the contained Medication resource has been removed. This also made them inaccessible in narrative generation.

**Expected behavior**
The reference to contained resource should not be removed, as they are included in the bundle.

**Environment:**
 - Current master of HAPI JPA STARTER project and HAPI FHIR

xmhorsehead:
Closes #6708 
Also fixed composition code display to exactly match LOINC, see https://loinc.org/60591-5/
Apparently FHIR validator is case sensitive...

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_2_0/6754-preserve-contained-resources-in-ips.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_2_0/6754-preserve-contained-resources-in-ips.yaml
@@ -1,0 +1,5 @@
+---
+type: fix
+issue: 6754
+title: "The IPS generator will no longer strip references to contained resources when
+  assembling data to create an IPS document. Thanks to Xun Ma for the contribution!"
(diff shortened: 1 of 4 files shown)


jamesagnew:
Hi @xmhorsehead - Thanks for the contribution! Would you be able to add a test (probably to `IpsGeneratorSvcImplTest`) which verifies the behaviour you are adding?

PR #6757, opened 2025-02-24
Load SPs by search during startup for faster boot.

michaelabuckley:


hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
```diff
@@ -459,15 +463,23 @@ private String unqualified(List<IIdType> theIds) {
...
-				IBaseResource searchParam = mySearchParamProvider.read(id);
-				searchParams.add(searchParam);
-			} catch (ResourceNotFoundException e) {
-				ourLog.warn("SearchParameter {} not found.  Excluding from list of active search params.", id);
-			}
-		}
+
+		Stream<ReferenceParam> paramStream = theResourceIds.stream().map(ReferenceParam::new);
+
+		TaskChunker.chunk(paramStream, QUERY_CHUNK_SIZE)
```
jmarchionatto, line 471: Loosing te warnings for not found params but I guess that is not too important, and if it were, it could be obtained back by a collection comparison at the end.

jmarchionatto (approved):
Looks good. Approved.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
michaelabuckley: Good catch.  That would be caused by an odd data-race, and I'm not worried if they delete an SP in the middle of boot as long as we don't fail.

Issue #5834, opened 2024-04-09
ClassCast exception arises when loading in implementation guide.


jkiddo:
at https://github.com/hapifhir/hapi-fhir/blob/107de2cfba04691ec085453062ee98f7200d7a4e/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java#L407 - the following bug occur with stacktrace (from HAPI FHIR JPA starter):

```
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java)
	at org.apache.tomcat.util.threads.InlineExecutorService.execute(InlineExecutorService.java:75)
	at java.base/java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:145)
	at org.apache.catalina.core.ContainerBase.startInternal(ContainerBase.java:866)
	at org.apache.catalina.core.StandardHost.startInternal(StandardHost.java:845)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:171)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1332)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1322)
	at java.base/java.util.concurrent.FutureTask.run$$$capture(FutureTask.java:317)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java)
	at org.apache.tomcat.util.threads.InlineExecutorService.execute(InlineExecutorService.java:75)
	at java.base/java.util.concurrent.AbstractExecutorService.submit(AbstractExecutorService.java:145)
	at org.apache.catalina.core.ContainerBase.startInternal(ContainerBase.java:866)
	at org.apache.catalina.core.StandardEngine.startInternal(StandardEngine.java:240)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:171)
	at org.apache.catalina.core.StandardService.startInternal(StandardService.java:433)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:171)
	at org.apache.catalina.core.StandardServer.startInternal(StandardServer.java:917)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:171)
	at org.apache.catalina.startup.Tomcat.start(Tomcat.java:488)
	at org.springframework.boot.web.embedded.tomcat.TomcatWebServer.initialize(TomcatWebServer.java:123)
[...]

PR #6762, opened 2025-02-27
Update JpaPackageCache.java

jkiddo:
Fix for https://github.com/hapifhir/hapi-fhir/issues/5834


dotasek (changes requested):
This wouldn't initially compile for me. I needed to add the following imports:

```java
import org.hl7.fhir.instance.model.api.IBase;
import org.hl7.fhir.instance.model.api.IBaseBackboneElement;
```

If I understand the issue correctly, I think it would be possible to replicate the original failure in JpaPackageCacheTest with something similar to this:

```java
myPackageCacheManager.addPackageToCache("hl7.fhir.uv.shorthand", "0.12.0", stream, "hl7.fhir.uv.shorthand");
```

 I was about to do so myself, but the IG in the original issue is not available: https://build.fhir.org/ig/WorldHealthOrganization/smart-anc/package.tgz

try (InputStream stream = ClasspathUtil.loadResourceAsStream("/packages/hl7.fhir.uv.shorthand-0.12.0.tgz")) {
				
			}

jkiddo:
@dotasek all three bullets should now be adressed.

dotasek (changes requested):
Excellent. The test correctly targets the bug (it does fail on the master branch). 

One more nitpick: could you add a comment linking the test to the original issue? Though correct, the test doesn't really explain why a profiled device should be tested.

dotasek (approved):

PR #6766, opened 2025-03-03
Fixing concurrent modification problem in VersionSpecificContextWrapper 

### LINKED ISSUE #6661: Bug in VersionSpecificWorkerContextWrapper.allStructureDefinitions()
**Describe the bug**
The following code on the current (at the time of writing) master:

https://github.com/hapifhir/hapi-fhir/blob/e412b38dfeb2bffee747b570a42d617caf5bace3/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java#L232-L246

Starting from the `try` block, the cached StructureDefinitions are snapshotted. However, instead of creating a new `List` for the resulting snapshotted StructureDefinitions, the resulting Resources are added to the list of previously fetched and converted StructureDefinitions. I am sure this is not correct, because both the non-snapshotted and the snapshotted version are stored to the same list. 

In addition to that, the issue causes `ConcurrentModificationException`s, since the `List` being modified in the try block was stored to the `myAllStructures` field before. When a second thread calls `allStructureDefinitions`, before the first one enters the `try` block, it will get the list and can read it, while the first thread adds elements to that list.

However, this issue is easy to fix: Before entering the try, add the line 
```java
retVal = new ArrayList<>();`
```
@fil512 it seems this was introduced in commit 362dc09 by you. At least you are attributed as the author. Maybe you want to fix the issue, since this is a pretty trivial fix, I can open a PR as well, if this is preferred.

**To Reproduce**
Validate several resources in parallel.

Boereck:
This change fixes #6661. The test case added only checks the synchronous error case. 
The concurrency issue could be tested, but not deterministically. And on my machine this
needs many iterations to trigger the problem. On my machine it takes around 5 seconds
to trigger the issue.

### DIFF
--- a/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
+++ b/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/VersionSpecificWorkerContextWrapper.java
@@ -235,6 +235,7 @@ public class VersionSpecificWorkerContextWrapper extends I18nBase implements IWo
 			myAllStructures = retVal;
 
 			try {
+				retVal = new ArrayList<>();
 				for (IBaseResource next : allStructureDefinitions) {
 					Resource converted = convertToCanonicalVersionAndGenerateSnapshot(next, false);
 					retVal.add((StructureDefinition) converted);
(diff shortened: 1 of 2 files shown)


Boereck:
Closing this PR, as fix for [#6963](https://github.com/hapifhir/hapi-fhir/pull/6963) was included in [#6963](https://github.com/hapifhir/hapi-fhir/pull/6963)

PR #6795, opened 2025-03-13
Don't stream a Binary content when a FHIR content is requested (#6630)

### LINKED ISSUE #6630: Binary read: wrong extraction of the contained data when accepting a FHIR MIME type
**Describe the bug**
When reading a Binary that contains a FHIR resource and setting the Accept header to the value of `Binary.contentType`, HAPI returns the contained resource instead of the Binary resource. It should return the Binary resource instead of extracting the contained resource.

**To Reproduce**
An example with the following Binary resource, containing a JSON Patient resource, created on https://hapi.fhir.org/baseR4/ (running HAPI FHIR 7.7.15-SNAPSHOT):
```json
{
  "resourceType": "Binary",
  "contentType": "application/fhir+json",
  "data": "[Patient resource]"
}
```

When reading the resource with `Accept: application/fhir+json`, the server returns the contained resource (Patient):
```http
GET /baseR4/Binary/45848035 HTTP/1.1
Accept: application/fhir+json
 
HTTP/1.1 200 OK
Content-Type: application/fhir+json
 
{
  "resourceType": "Patient",
  // ...
}
```

When requesting the resource with the XML MIME type, it returns the BInary instead:
```http
GET /baseR4/Binary/45848035 HTTP/1.1
Accept: application/fhir+xml
 
HTTP/1.1 200 OK
Content-Type: application/fhir+xml;charset=utf-8
 
<Binary xmlns="http://hl7.org/fhir">
  <contentType value="application/fhir+json"/>
  <data value="..."/>
</Binary>
```

**Expected behavior**
In both cases, the server shall return the Binary resource, as specified in the [FHIR specification](https://www.hl7.org/fhir/R4/binary.html):

> When a read request is made with a FHIR type in the Accept header (e.g. "application/fhir+xml" or "application/fhir+json") the Binary resource is returned in the requested FHIR format. This applies even when the binary data itself has a FHIR mime type

It seems HAPI checks whether the requested MIME type is the same as the `Binary.contentType` before processing the special case with "application/fhir+xml" and "application/fhir+json".

 - HAPI FHIR 7.7.15-SNAPSHOT/3594770bf6/2024-12-21 (deployed on https://hapi.fhir.org/baseR4/)

**Additional context**
I checked the behaviour when no Accept header is specified; in that case, the contained resource is returned. While that case is not explicitly described in the spec, it seems coherent with the stated intent:

> However, the intent is that unless specifically requested, the FHIR XML/JSON representation is not returned

qligier:
Fixes #6630

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/RestfulServerUtils.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/RestfulServerUtils.java
@@ -1078,30 +1078,29 @@ public class RestfulServerUtils {
 	 * Determines whether we should stream out Binary resource content based on the content-type. Logic is:
 	 * - If the binary was externalized and has not been reinflated upstream, return false.
 	 * - If they request octet-stream, return true;
+	 * - If they request a FHIR content-type, return false (as per the specification);
 	 * - If the content-type happens to be a match, return true.
-	 * <p>
-	 * - Construct an EncodingEnum out of the contentType. If this matches the responseEncoding, return true.
 	 * - Otherwise, return false.
 	 *
 	 * @param theResponseEncoding the requested {@link EncodingEnum} determined by the incoming Content-Type header.
 	 * @param theBinary           the {@link IBaseBinary} resource to be streamed out.
 	 * @return True if response can be streamed as the requested encoding type, false otherwise.
 	 */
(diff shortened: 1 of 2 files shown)


qligier:
Added an entry to the changelog.

qligier:
@jamesagnew @tadgh @michaelabuckley
As previous committers to that file/that feature, could you please have a look at it (or point me to the right person for this task)? Thanks!

jamesagnew (approved):

robogary:
Formatting check succeeded!

PR #6862, opened 2025-04-10
fix: nested modifierExtensions are not allowed

xytis:
This is a proposed fix to an observed bug.

We are using JSON to communicate with our FHIR server, and we received the following output:
```json
{
  "resourceType": "MedicationRequest",
  ...
  "modifierExtension": [
    {
      "url": "http://esveikata.lt/Profile/lt-med-prescription#prescriptionStatus",
      "modifierExtension": [
        {
          "url": "http://esveikata.lt/Profile/lt-med-prescription#prescriptionStatusCode",
          "valueCode": "active"
        },
        ...
```

According to FHIR spec (irrelevant of version) [link](https://www.hl7.org/fhir/R4B/extensibility.html#modifierExtension):
> Note that complex extensions are allowed to have elements in the complex extension that are marked [Is-Modifier = true](https://www.hl7.org/fhir/R4B/conformance-rules.html#isModifier), which means that these elements modify the extension value itself. Internal extensions like this marked "Is-Modifier" are still represented using the extension element, not modifierExtension because the impact of the modifier element is expected to be known by applications that understand the containing extension.

Also, the same resource when using xml is emitted as:
```xml
<MedicationRequest xmlns="http://hl7.org/fhir">
...
    <modifierExtension url="http://esveikata.lt/Profile/lt-med-prescription#prescriptionStatus">
        <extension url="http://esveikata.lt/Profile/lt-med-prescription#prescriptionStatusCode">
            <valueCode value="active"></valueCode>
        </extension>
        ...
```

The proposed fix simply removes the seemingly incorrect condition. I was unable to build and test this locally, if you could provide me with a snapshot build of this commit, I would be able to plug this into our production system and validate.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
@@ -2006,11 +2006,7 @@ public class JsonParser extends BaseParser implements IJsonLikeParser {
 				// Write child extensions
 				if (!ext.getExtension().isEmpty()) {
 
-					if (myModifier) {
-						beginArray(theEventWriter, "modifierExtension");
-					} else {
-						beginArray(theEventWriter, "extension");
-					}
+					beginArray(theEventWriter, "extension");
 
 					for (Object next : ext.getExtension()) {
 						writeUndeclaredExtension(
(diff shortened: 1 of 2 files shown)


jamesagnew:
Hi @xytis - Thanks for the submission.

Could you please add appropriate unit tests verifying the behaviour you are change?

xytis:
I would love to, could you point me to the location where this parser has any unit tests? I was unable to find them on my own.

jamesagnew:
@xytis this would be a good test to add to: https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/parser/JsonParserR4Test.java

xytis:
@jamesagnew Introduced a failing test. My change fixes it.

robogary:
Formatting check succeeded!

PR #6904, opened 2025-04-25
feat: Add full support for ValueSet filter operators in InMemoryTerminologyServerValidationSupport

iyt-trifork:
### 💬 **Pull Request Description** 
This PR enhances `InMemoryTerminologyServerValidationSupport` to support all ValueSet filter operators defined in the FHIR spec, including structural and existence-based filters. This allows for more accurate behavior when running validation/expansion using preloaded in-memory terminology, such as during unit testing, CLI tools, or bootstrapped deployments.

#### 🔧 Features Implemented 
* `=`, `in`, `not-in`, `regex` filters on code & display 
* `is-a`, `descendent-of`, `generalizes`, `is-not-a`, `child-of`, `descendent-leaf` filters on code 
* `exists` filter support (only for `code` and `display` properties) 
* Case sensitivity respected per CodeSystem definition

#### 🧪 Test Coverage
A complete parameterized test suite validates all filter combinations using realistic hierarchical CodeSystems. Tests are version-aware and toggle R4 vs. R5 model usage accordingly.

#### 💡 Motivation
The FHIR ValueSet filter operator descendent-of is currently not handled by HAPI FHIR’s in-memory validation/expansion chain. When a ValueSet compose includes a filter like property = "concept", op = "descendent-of", value = X, the existing logic fails to apply it. This leads to incorrect expansions – notably, the ancestor concept X itself ends up included in the expansion (or an error is thrown) instead of only its descendants. In other words, the in-memory expander does not distinguish descendent-of from is-a, causing validation tests to incorrectly include the parent code in the expanded results. This issue has been observed in unit tests for custom profiles that define ValueSets using descendent-of filters. Those tests currently either fail or require workarounds (like forcing JPA-based expansion via $reindex) because the in-memory expander doesn’t support descendent-of. This also causes confusing behavior in validation and made it hard to rely on FHIR profile tests without spinning up full persistence layers.

This change enables full correctness out-of-the-box using only PrePopulatedValidationSupport.

#### ⚙️ Performance Considerations
Concept tree indexing is lazily initialized and cached internally per expansion, making repeated filtering efficient. Matching is short-circuiting and avoids unnecessary traversal for filters like in and equal.

#### 🤔 Optional vs. Default Behavior
The new logic applies automatically to in-memory expansions. If needed, this could be guarded via ValueSetExpansionOptions, but current design keeps behavior spec-compliant by default.

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ValueSetTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ValueSetTest.java
@@ -141,17 +141,10 @@ public class FhirResourceDaoR4ValueSetTest extends BaseJpaR4Test {
 		ValidationSupportContext ctx = new ValidationSupportContext(myValidationSupport);
 		ConceptValidationOptions options = new ConceptValidationOptions();
 
-		// In memory - Hierarchy in existing CS
-
-		outcome = myValidationSupport.validateCode(ctx, options, "http://cs", "child10", null, "http://vs");
-		assertNotNull(outcome);
-		assertTrue(outcome.isOk());
-		assertEquals("Code was validated against in-memory expansion of ValueSet: http://vs", outcome.getSourceDetails());
-
 		outcome = myValidationSupport.validateCode(ctx, options, "http://cs", "childX", null, "http://vs");
 		assertNotNull(outcome);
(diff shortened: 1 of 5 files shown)


robogary:

robogary:
Formatting check succeeded!

robogary:
Formatting check succeeded!

robogary:
Formatting check succeeded!

robogary:
Formatting check succeeded!

robogary:

jamesagnew (approved):
This is super cool! Thanks for the contribution!

PR #6916, opened 2025-04-30
Allow updates in transaction when updating deleted resource is in transaction

SevcikMichal:


jamesagnew:
Superceded by #6922 - Thanks for submitting this, you've been credited with the solution design.

PR #6923, opened 2025-05-01
6911 mdm max field matching 2

### LINKED ISSUE #6911: MDM vector calculations cannot handle large number of fields
Because of the way Vector is calculated (using the binary shift operator << by the index of the rule in the matchRules set), there is a maximum number of fields that can be used before we hit overflow.

Currently it should allow up to 63 fields (ie, the max number of digits in a binary representation of a Long).

But overflow issues are appearing with lower numbers (31 digits)

TipzCM:
closes #6911


robogary:
Formatting check succeeded!

jmarchionatto (approved):
Approving replaced MR

PR #6924, opened 2025-05-01
removed fhircontext param from CdsPrefetchService constructor

mrdnctrk:
I had recently added the FhirContext as a parameter to the CDSPrefetchSvc constructor, but I noticed that broke some projects like [hapi-fhir-jpaserver-starter](https://github.com/hapifhir/hapi-fhir-jpaserver-starter/blob/master/src/main/java/ca/uhn/fhir/jpa/starter/cdshooks/ModuleConfigurationPrefetchSvc.java#L78) and [clinical reasoning](https://github.com/cqframework/clinical-reasoning/blob/710efbe154cd2239a70cc63ea2ef54802db4a27a/cqf-fhir-cr-hapi/src/main/java/org/opencds/cqf/fhir/cr/hapi/config/test/TestCdsHooksConfig.java#L101). 

So rather than updating all the projects that use the constructor, in this MR I removed the parameter from the constructor and  get the fhirContext through the CdsPrefetchDaoSvc that is already passed in to the constructor. I noticed that the class is already using this approach when it needs to access the FhirContext.


robogary:
Formatting check succeeded!

robogary:
Formatting check succeeded!

dotasek (approved):
This allows jpaserver-starter to compile, pass all java tests, and pass all smoke tests. From my perspective, this looks good.

robogary:
Formatting check succeeded!

Issue #6939, opened 2025-05-06
BaseResourceCacheSynchronizer requestRefresh uses null SystemRequestDetails if registerListener is not called yet


mrdnctrk:
[`BaseResourceCacheSynhronizer`](https://github.com/hapifhir/hapi-fhir/blob/5f18274f90a595f0b83dafc1cce542678e08f95e/hapi-fhir-storage/src/main/java/ca/uhn/fhir/cache/BaseResourceCacheSynchronizer.java#L57-L57) has a public `requestRefresh` method which eventually makes use of the `mySystemRequestDetails` field in the same class to make a db call. The problem is `mySystemRequestDetails` is only initialized in the same class' `registerListener` method. It means that if `requestRefresh` is called before `registerListener` then  `mySystemRequestDetails` is null, which is not the expected behaviour `mySystemRequestDetails` should be initialized earlier.

mrdnctrk:
Closed. Fixed by #6940.

PR #6940, opened 2025-05-06
initialize requestdetails earlier in cache synchronizer

mrdnctrk:
issue: #6939 

[BaseResourceCacheSynhronizer](https://github.com/hapifhir/hapi-fhir/blob/5f18274f90a595f0b83dafc1cce542678e08f95e/hapi-fhir-storage/src/main/java/ca/uhn/fhir/cache/BaseResourceCacheSynchronizer.java#L57-L57) has a public `requestRefresh` method which eventually makes use of the `mySystemRequestDetails` field in the same class to make a db call. The problem is `mySystemRequestDetails` is only initialized in the same class' `registerListener` method. It means that if `requestRefresh` is called before `registerListener` then `mySystemRequestDetails` is null, which is not the expected behaviour. 

This MR initializes `mySystemRequestDetails` field earlier.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_2_0/6939-initialize-RequestDetails-BaseResourceCacheSynhronizer.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_2_0/6939-initialize-RequestDetails-BaseResourceCacheSynhronizer.yaml
@@ -1,0 +1,6 @@
+---
+type: fix
+issue: 6939
+title: "BaseResourceCacheSynchronizer class, which is mainly used for caching Subscription and 
+SubscriptionTopic resources, was using a null RequestDetails object when querying the database, if its 
+`requestRefresh` method is called before the `registerListener` method during initialization. This has now been fixed."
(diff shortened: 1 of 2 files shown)


robogary:
Formatting check succeeded!

epeartree (changes requested):
u will need a changelog

robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_2_0/6939-initialize-RequestDetails-BaseResourceCacheSynhronizer.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 6939
```
epeartree, line 3: this description is difficult to read.  to understand it, someone needs quite the knowledge of the class structure and initialisation process.   it's about finding the balance between conciseness, implementation details and relevance.  i suggest the following:

Previously, the Subscription module could fail to boot due to a race condition happening during its initialisation sequence.  The issue has been fixed.

mrdnctrk: changed as suggested

robogary:
Formatting check succeeded!

PR #6941, opened 2025-05-06
Add index for ResourceHistoryTable's sourceUri field 

jmarchionatto:


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceHistoryTable.java
```diff
@@ -67,7 +67,8 @@
 		indexes = {
 			@Index(name = "IDX_RESVER_TYPE_DATE", columnList = "RES_TYPE,RES_UPDATED,RES_ID"),
 			@Index(name = "IDX_RESVER_ID_DATE", columnList = "RES_ID,RES_UPDATED"),
-			@Index(name = "IDX_RESVER_DATE", columnList = "RES_UPDATED,RES_ID")
+			@Index(name = "IDX_RESVER_DATE", columnList = "RES_UPDATED,RES_ID"),
```
epeartree, line 70: this may be correct but please verify.  
the index defined on the entity is comprised of 3 columns (partition_id, res_id, source_uri) while the index added on the migration has only 2.  this may be ok and the way to do it post roll out of 'partition_mode' .

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_2_0/7393-add-res-ver-source-index.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
```
epeartree, line 2: 7393 is a gitlab number.  please open a github issue and provide the issue number here.

epeartree (approved):

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceHistoryTable.java
```diff
@@ -67,7 +67,8 @@
...
+			@Index(name = "IDX_RESVER_ID_SRC_URI", columnList = "PARTITION_ID,RES_ID,SOURCE_URI")
```
jamesagnew, line 71: Adding an index on this table has a potentially significant performance impact - Is this really just needed in order to support the `_source:contains=` that feels like a really niche case to be driving such a change.

I don't know if there is a critical need to do this in 8.2.0 but unless it's customer critical I think this is too risky a change to be introducing at the 11th hour.

jmarchionatto:
Implementation cancelled

PR #6954, opened 2025-05-09
Revert "Prevent spurious circular-dependency warnings in SnapshotGene…

tadgh:
This reverts commit 5ca9c6ce94da9ac16a6480e86bcff1059196d014.

In practice, this causes a relatively large memory explosion, both locally, and in CI, about 80-85% of the time. Here's a heap dump from such a failure. 

https://drive.google.com/file/d/1cOrghPgHVwzoRBI22z_jfgTTreB_nXs2/view?usp=drive_link

Reverting this for now until we can discover the reason. @patrick-werner, apologies for the version, but we need to stabilize CI. 

it's possible to replicate this OOM issue locally by calling:
 
```bash
mvn -P FASTINSTALL,NOJAVADOCS install -T 10; mvn clean install -pl hapi-fhir-jpaserver-test-dstu3
```

Happy to take the PR back in once we have the OOMs resolved.


robogary:
Formatting check succeeded!

jmarchionatto (approved):

PR #7004, opened 2025-05-28
[7003] Fix Narrative not being added to resources when a Narrative Module is enabled and properly configured

### LINKED ISSUE #7003: Narrative not being added to resources when a Narrative Module is enabled and properly configured
**Problem Statement**

Narrative elements are not added to resources even when a Narrative Module is enabled and properly configured.
The root cause of the problem is with the INarrativeGenerator bean created by the NarrativeTemplateLibraryCtxConfig not making it to the fhirContext injected in the BaseFhirEndpoint Servlet. As a result, the jsonParser (or xmlParser) will never be able to add narrative since its reference to an iNarrativeGenerator is null

**Implementation**

Inject Narrative Generator to FhirContext
Added test to verify that the resource created with a custom narrative generator contains the narrative block

YalingPeiS:
**Problem Statement**

Narrative elements are not added to resources even when a Narrative Module is enabled and properly configured.
The root cause of the problem is with the INarrativeGenerator bean created by the NarrativeTemplateLibraryCtxConfig not making it to the fhirContext injected in the BaseFhirEndpoint Servlet.  As a result, the jsonParser (or xmlParser) will never be able to add narrative since its reference to an iNarrativeGenerator is null 

**Implementation**

- Inject Narrative Generator to FhirContext
- Added test to verify the resource created with a custom narrative generator contains the <text> narrative block

Closes #7003


robogary:
Formatting check succeeded!

PR #7010, opened 2025-06-02
First run at updating release pipeline.

markiantorno:
Release pipeline needs to be updated. I am going to do this in two passes. This first one add the modular tools to run the build in parallel, and updates the MR pipeline to use the re-usable workflow.


robogary:
Formatting check succeeded!

.github/actions/normalize-project-version/action.yml
```diff
@@ -0,0 +1,33 @@
...
+      shell: bash
+      run: |
+        echo "Original project version: ${{ inputs.version }}"
+        VERSION_FROM_POM="${{ inputs.version }}"
+        BRANCH_NAME="${GITHUB_REF##*/}"
+        
+        # First, strip "-SNAPSHOT"
+        STRIPPED_VERSION="${VERSION_FROM_POM%-SNAPSHOT}"
+        
+        # Then, if on master, strip "-PRE*"
```
tadgh: question: do we use -PRE anywhere in HAPI-FHIR? I think that is solely a CDR construct.

.github/workflows/release.yml
```diff
@@ -0,0 +1,71 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install yq
+        run: |
+          sudo wget https://github.com/mikefarah/yq/releases/download/v4.40.5/yq_linux_amd64 -O /usr/bin/yq
+          sudo chmod +x /usr/bin/yq
```
tadgh: question: Is it possible to use the official yq action? https://github.com/marketplace/actions/yq-portable-yaml-processor

tadgh (approved):

.github/workflows/release.yml
markiantorno: good catch, will update to use the official action

.github/actions/normalize-project-version/action.yml
markiantorno: We do not. It can be removed.

PR #7018, opened 2025-06-04
Mark match query spans so interceptors can distinguish.

michaelabuckley:


robogary:
Formatting check succeeded!

AbayomiE-T:
P.S.: @michaelabuckley  I had to undo those nonNull requirements. It was causing test failures in a bunch of places and debugging would've added more time to spend on the overall ticket than necessary. Might be beyond the scope of the ticket.

PR #7027, opened 2025-06-06
Rewriting a deleted resource causes history and resource table desync

jdar8:


robogary:
Formatting check succeeded!

PR #7035, opened 2025-06-10
fix transaction boundary inside reduction step process chunk

### LINKED ISSUE #7036: Fix transaction boundary inside reduction step process chunk
**Problem Statement**
- When a final step is a reduction step, the code runs in batch2 transaction boundary which doesn't allow the batch2 step to use its own transaction boundary.
- Thus, any reduction step needs to use `transactionNested`, which is not ideal because the transaction open inside reduction step is for chunks and not the actual code inside step.

AD1306:
closes #7036


robogary:
Formatting check succeeded!

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/ReductionStepExecutorServiceImpl.java
```diff
@@ -350,6 +349,13 @@ private <PT extends IModelJson, IT extends IModelJson, OT extends IModelJson> vo
 		}
 	}
 
+	private void executeInNoTransaction(Runnable theRunnable) {
```
michaelabuckley, line 367: Needs some javadoc.
```suggestion
/**
 * We need to keep a long transaction here to keep our ResultSet Stream open for the duration.  But we don't want to expose this transaction to the user.  We use a null tx here to reset all our thread-locals so the actual chunk work runs cleanly.
 */
	private void executeInNoTransaction(Runnable theRunnable) {
```

michaelabuckley (commented):
I shouldn't review since I wrote the code.

PR #7041, opened 2025-06-11
making the parser maintain a reference to the contained resource

### LINKED ISSUE #7040: Parser should add a contained reference link when parsing/serializing resources that contain contained resources
When a resource has a contained resource, that contained resource's id does not get prepended with a "#".

So if this resource's contained resource is referenced, not by a Reference, but by the actual resource itself, it cannot be processed by the TransactionProcessor (because the contained resource does not contain a contained resource id!)

We should add a contained resource reference (that does contain the contained resource's id prepended by an #) so that transactions afterwards can store the resource.

This is to fix a regression

TipzCM:
closes #7040

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
@@ -193,7 +193,7 @@ public abstract class BaseParser implements IParser {
 	 * process the resource (since the reference will not be set).
 	 */
 	private void setReference(IBaseReference theReference, String theText) {
-		myContext.newTerser().setElement(theReference, "reference", theText);
+		theReference.setReference(theText);
 	}
 
 	private String determineReferenceText(


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7040-maintain-contained-resource-reference-ids.yaml
```diff
@@ -0,0 +1,15 @@
...
+      will no longer have their ids prepended with a '#'.
+      This caused a regression where contained resources were
+      referenced, not by an Id Reference, but by the actual
+      full embedded resource, would no longer be able to
+      be processed in a transaction.
+      To fix this, the parser/serializer (which is responsible
+      for creating these contained reference ids) will also
+      create a contained resource id reference, which does
+      contain the prepended '#', thereby allowing these
+      transactions to work as before.
```
jdar8, line 14:
```suggestion
issue: 7040
jira: SMILE-10538
title: "Due to a change in Release 8.2.0, contained resources
      no longer have their ids prepended with a '#'.
      This caused a regression where contained resources that were
      referenced, not by an Id Reference, but by the actual
      full embedded resource, would no longer be able to
      be processed in a transaction.
      To fix this, the parser/serializer (which is responsible
      for creating these contained reference ids) will also
      create a contained resource id reference, which does
      contain the prepended '#', thereby allowing these
      transactions to work as before.
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
```diff
@@ -183,6 +183,19 @@ protected Iterable<CompositeChildElement> compositeChildIterator(
...
+	/**
+	 * We add the reference to the input resources to ensure it doesn't get
+	 * overwritten if it wasn't there.
+	 * The resource is being mutated anyways, so this is fine.
+	 *
+	 * We need to maintain the reference as an internal reference (ie,
+	 * preceded by an #) because otherwise later transactions cannot
+	 * process the resource (since the reference will not be set).
+	 */
+	private void setReference(IBaseReference theReference, String theText) {
```
jdar8, line 195: suggestion, non-blocking: change variable name `theText` to something like `theReferenceId`

hapi-fhir-structures-r4/src/test/java/ca/uhn/fhir/parser/JsonParserR4Test.java
```diff
@@ -1490,6 +1491,38 @@ public void testNestedModifierExtensions() {
...
+		IParser parser = ourCtx.newJsonParser();
+
+		// test
+		String encoded = parser.encodeResourceToString(patient);
+		ourLog.info(encoded);
+
+		// verify
+		assertNotNull(encoded);
+		assertTrue(isNotBlank(encoded));
+		assertFalse(patient.getGeneralPractitioner().isEmpty());
```
jdar8, line 1519: Nitpick, non-blocking: asserting false on isEmpty() is kind of hard to read
```suggestion
		assertThat(patient.getGeneralPractitioner()).isNotEmpty();
```

jdar8 (commented):
Requires additional approval by someone with approval powers.

Looks good - just left a few non-blocking comments.

PR #7045, opened 2025-06-12
Initial commit of test matrix

tadgh:
- Test matrix for Java Bundle with contained resources -> Transaction ingestion, including
- Matrix dimensions: 
    - Do we call run it through the parser before submitting?
    - Do we reference the contained resource by Object, or by ID?
    - Do we set the contained resource's ID manually?

Before any code changes, here are the results: 
image

After production code changes, here are the results: 
image

Changes include so far: 
- Mark contained resources with userData on ingestion, as we cannot use a `#`-prefixed id as an indicator, (since that is not permitted). 
- During parse, ensure that any references that start with `#` are _actually_ contained in the resource. This prevents a situation where you have a resource that looks like this (which cleanly ingests!)
```
{
  "resourceType": "Bundle",
  "type": "transaction",
  "entry": [ {
    "resource": {
      "resourceType": "Patient",
      "managingOrganization": {
        "reference": "#2a43f5a2-1b23-485c-a3d8-6b31243fc8a3"
      }
    },
    "request": {
      "method": "POST",
      "url": "Patient"
    }
  } ]
}
```

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
@@ -227,6 +227,10 @@ public abstract class BaseParser implements IParser {
 				}
 			}
 			return reference;
+		} else {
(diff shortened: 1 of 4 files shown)


robogary:

PR #7047, opened 2025-06-12
Fix issue with fhir batch failing if running on parallel threads with shared resource creation

AD1306:


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoCreatePlaceholdersR4Test.java
```diff
@@ -7,14 +7,17 @@
 import ca.uhn.fhir.jpa.model.entity.ResourceTable;
```
jamesagnew, line 7: Meta: This change needs a changelog.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java
```diff
@@ -422,13 +423,20 @@ private IBaseBundle batch(final RequestDetails theRequestDetails, IBaseBundle th
 
 		IBaseBundle response;
 		for (int i = 1; ; i++) {
+			// We're resetting the request here because processBatch/processTransaction mutates
+			// the original request, and it fails with InvalidRequestException on retry. Processing
+			// FHIR batch and transaction retry seems to follow different paths. FHIR transaction
+			// doesn't seem to use the header X-Transaction-Semantics for retries vs batch uses it.
+			// The deeper fix might be to actually reconcile how that's handled. Cloning is costly
+			// on performance.
+			final IBaseBundle request = TerserUtil.clone(myContext, theRequest);
```
jamesagnew: Cloning the bundle is potentially quite expensive, so I don't love this fix. It is preferable to specify a "rollback action" which undoes any mutations to the model in the case of a transaction rollback. You can seen an example of these here: https://github.com/hapifhir/hapi-fhir/blob/e8adcda93ec3fe46f27eaf0d2d29bc3c6e4b6565/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java#L1342-L1345

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoCreatePlaceholdersR4Test.java
```diff
@@ -860,4 +865,63 @@ public void testMultipleReferencesToAutocreatedPlaceholder() {
 		assertThat(outcome.getAllResources()).hasSize(3);
 	}
 
+	@Test
+	public void runBundleBatchInParallelThreads() throws Exception {
+		// setup
+		myStorageSettings.setAutoCreatePlaceholderReferenceTargets(true);
+		myStorageSettings.setAllowInlineMatchUrlReferences(true);
+		final IParser parser = myFhirContext.newJsonParser();
+		final SystemRequestDetails requestDetails = new SystemRequestDetails();
```
jamesagnew: I'm assuming the two threads sharing a single SystemRequestDetails isn't actually a part of the cause of the issue you're fixing, but the fact that we're doing that kind of jumps out at me. It'd recommend creating a new SRD for each thread so that future readers don't make the same assumption (plus, it's more realistic since we'd never actually share an SRD in that way)

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java
michaelabuckley: The problem is that HapiTransactionService does ONLY applies the rollback when doing an internal retry.  The rollback actions are NOT applied with this external retry loop.  The deeper fix is to probably to change this code 						

```
						if (i < maxRetries) {
							if (theExecutionBuilder.myTransactionDetails != null) {
								theExecutionBuilder
										.myTransactionDetails
										.getRollbackUndoActions()
										.forEach(Runnable::run);
```

to always apply the rollback, but we wanted to ask you why it doesn't apply rollback on the last retry.

jamesagnew: @michaelabuckley I don't think that is a deliberate design decision, jsut that we haven't previously had any use cases where it made a difference to not apply the rollback after the last retry so we never realized we didn't. No objections at all to us fixing that.

michaelabuckley: Cool.  Done.  There are a couple of FIXMEs with questions for you in my changes.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoCreatePlaceholdersR4Test.java
AD1306: done

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoCreatePlaceholdersR4Test.java
AD1306: done

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
```diff
@@ -430,6 +370,37 @@ private <T> T doExecuteInTransaction(
...
+	/**
+	 * Sleep a bit more each time, with 0 sleep on first retry and some random dither.
+	 * @param theAttemptIndex 0-index for the first attempt, 1 for second, etc.
+	 */
+	private void sleepForRetry(int theAttemptIndex) {
+		double sleepAmount = (250.0d * theAttemptIndex) * Math.random();
+		long sleepAmountLong = (long) sleepAmount;
+		mySleepUtil.sleepAtLeast(sleepAmountLong, false);
+
+		ourLog.info(
```
jamesagnew, line 397: Nitpick: This log line should presumably be move one line up, before the sleep actually happens

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
```diff
@@ -565,6 +571,27 @@ protected RequestPartitionId getEffectiveRequestPartitionId() {
...
 		}
+
+		/**
+		 * This method is called when a transaction has failed, and we need to rollback any changes made to the
+		 * state of our objects in RAM.
+		 * <p>
+		 * This is used to undo any changes made during transaction resolution, such as conditional references,
+		 * placeholders, etc.
+		 */
+		void rollbackTransactionProcessingChanges() {
```
jamesagnew, line 581: This is such a nice quality-of-life change, putting all the rollback stuff into its own method.

jamesagnew (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java
AD1306: done

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/UserRequestRetryVersionConflictsInterceptor.java
```diff
@@ -55,6 +56,7 @@ public class UserRequestRetryVersionConflictsInterceptor {
 	public ResourceVersionConflictResolutionStrategy check(RequestDetails theRequestDetails) {
 		ResourceVersionConflictResolutionStrategy retVal = new ResourceVersionConflictResolutionStrategy();
 		boolean shouldSetRetries = theRequestDetails != null && theRequestDetails.isRetry();
+		// FIXME REVIEW QUESTION: Shouldn't this be in HapiTransactionService directly?
```
jamesagnew: I don't personally think it should be... IMO keeping it here is good separation of concerns. 

The HapiTransactionService is all about database logic and doesn't deal with anything at the HTTP/client layer. Technically it does take a RequestDetails, but that's only used for interceptor broadcasting.

I dunno, I wouldn't mind if this got moved, but I wouldn't personally move it

Issue #6953, opened 2025-05-09
Validation error for resources that contain the same resource type


dotasek:
Originally reported [here](https://github.com/hapifhir/org.hl7.fhir.core/issues/1889).

**Describe the bug**

Validating an R4 resource containing a resource of the same resource type fails validation. (for example, a Medication resource containing another Medication resource that it is referencing as an ingredient)

**To Reproduce**

Try validating the following using R4:

```json
{
    "resourceType": "Medication",
    "contained": [ {
      "resourceType": "Medication",
      "id": "Contained1",
      "ingredient": [ {
        "itemCodeableConcept": {
          "text": "Ingredient Text"
        }
      } ]
    } ],
    "ingredient": [ {
      "itemReference": {
        "reference": "#Contained1"
      }
    } ]
  }
```

It will produce 3 errors:

```
SingleValidationMessage[col=8,row=13,locationString=Medication.ingredient[0],message=Unrecognized property 'itemReference',severity=error],
    SingleValidationMessage[col=4,row=1,locationString=Medication,message=Constraint failed: dom-6: 'A resource should have narrative for robust management' (defined in http://hl7.org/fhir/StructureDefinition/DomainResource) (Best Practice Recommendation),http://hl7.org/fhir/StructureDefinition/DomainResource#dom-6,severity=warning],
    SingleValidationMessage[col=6,row=3,locationString=Medication.contained[0],message=The contained resource 'Contained1' is not referenced to from elsewhere in the containing resource nor does it refer to the containing resource (dom-3),CONTAINED_ORPHAN_DOM3,severity=error]]
```

The following branch contains this in the form of a failing test: https://github.com/hapifhir/hapi-fhir/tree/do-20250509-contained-resource-same-type-errors

**Expected behavior**

Using the FHIR core validator (download [here](https://github.com/hapifhir/org.hl7.fhir.core/releases/tag/6.5.21), this will produce no errors, only a warning regarding narrative.

execution:
```sh
java -jar validator.jar -version R4 resource.json
```

output:
```sh
Success: 0 errors, 1 warnings, 0 notes
  Warning @ Medication (line 1, col4): Constraint failed: dom-6: 'A resource should have narrative for robust management' (defined in http://hl7.org/fhir/StructureDefinition/DomainResource) (Best Practice Recommendation)
Done. Times: Loading: 00:40.995, validation: 00:00.480. Max Memory = 4Gb
```

**Additional context**

[...]

PR #7050, opened 2025-06-13
DRAFT: #6953 VersionSpecificWorkerContextWrapper: using cache to return stru…

alexey-tschudnowsky:
fix for the described problem. The main change was to use the cache of canonicals with snapshots in fetchResource method instead of the original structuredefinitions from the validationsupportchain


alexey-tschudnowsky:
the error is already fixed in master

PR #7051, opened 2025-06-13
7040 contained resources plusplus

### LINKED ISSUE #7040: Parser should add a contained reference link when parsing/serializing resources that contain contained resources

TipzCM:
closes #7040


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
```diff
@@ -227,6 +242,12 @@ private String determineReferenceText(
 				}
 			}
 			return reference;
+		} else {
+			if (ref.isLocal() && !theContext.getContainedResources().referenceMatchesAContainedResource(ref)) {
+				throw new DataFormatException(
```
jamesagnew: Reminder: This will need an error code

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
```diff
@@ -1829,7 +1829,14 @@ public Map<String, IBaseResource> getExistingIdToContainedResource() {
 			return myExistingIdToContainedResourceMap;
 		}
 
+		public boolean referenceMatchesAContainedResource(IIdType theRefId) {
```
jamesagnew, line 1832: Suggestion: Given how tricky this all is, I'd suggest adding the following line at the top of this method:

```java
assert theRefId.getIdPart().startsWith("#");
```

..just to catch any weirdness in unit tests. Given the calling context of the code here, I expect that will always be true, but it never hurts to add assertions about what you expect to always be true

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
```diff
@@ -1829,7 +1829,14 @@ public Map<String, IBaseResource> getExistingIdToContainedResource() {
...
+			String expectedResourceId = theRefId.getIdPart().substring(1);
+			return this.getContainedResources().stream()
+					.anyMatch(res -> res.getIdElement().getIdPart().equals(expectedResourceId));
+		}
+
 		public IIdType addContained(IBaseResource theResource) {
+			theResource.setUserData("IS_CONTAINED", true);
```
jamesagnew: Question: I don't see this getting used anywhere. Is it needed? If so, should it be extracted into a constant?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4Test.java
```diff
@@ -1246,7 +1249,7 @@ public void testCreateAndReadBackResourceWithContainedReferenceToContainer() {
 		ourLog.info(output);
 
 		Location loc = (Location) org.getContained().get(0);
-		assertEquals("#", loc.getManagingOrganization().getReference());
+		assertEquals("1", loc.getManagingOrganization().getReference());
```
jamesagnew, line 1252: Oh this is so interesting. It took me a while of reading to figure out why you had changed `#` to `1` this in the first place, but I see: this was a bug before that your fix has actually corrected

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
TipzCM: added

hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/BaseParser.java
TipzCM: added

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
TipzCM: i'll verify with gary, but i think it can be removed

Issue #6993, opened 2025-05-26
Registering the `IInterceptorService` bean on `RestfulServer` results in duplicate interceptor calls


MartinBernstorff:
## The problem
If we use the bean registered in `hapi-fhir-jpaserver-base` [here](https://github.com/hapifhir/hapi-fhir/blob/5ca9c6ce94da9ac16a6480e86bcff1059196d014/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java#L468) to construct the `RestfulServer`, any REST-request calls each interceptor twice. 

## The cause
I think this is because of the use of the composite broadcaster, e.g. [here](https://github.com/hapifhir/hapi-fhir/blob/5ca9c6ce94da9ac16a6480e86bcff1059196d014/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseStorageDao.java#L356).

It receives the interceptors from:
* The request, which is populated in `RestfulServer.newRequestDetails`
* The DAO itself, which is autowired in `BaseHapiFhirDao.myInterceptorBroadcaster`

However, in this case, they end up being the same one!

## Workaround
My understanding is that Interceptors are triggered like this:

#### Plain pointcuts

|                 | Registered on |                     |
| --------------- | ------------- | ------------------- |
| **Call source** | RestfulServer | IInterceptorService |
| REST            | ✅             | 🔴                  |
| DAO             | 🔴            | 🔴                  |

#### Storage pointcuts
|                 | Registered on |                     |
| --------------- | ------------- | ------------------- |
| **Call source** | RestfulServer | IInterceptorService |
| REST            | ✅             | ✅                   |
| DAO             | 🔴  | ✅ |

✅: Triggered
🔴: Not triggered

Therefor, we maintain two different `InterceptorService`s, one for REST and one for non-REST. 

Interceptors registered on `RestfulServer`s `InterceptorService` are:
* Triggered on REST calls (non-storage pointcuts)
  * Because they are populated on the `RequestDetails` and `BaseHapiFhirDao.myInterceptorBroadcaster` combines both sources.
* Not triggered on storage pointcuts (DAO calls)
  * E.g. `dao.update(myResource, new SystemRequestDetails())`

Interceptors registered on the JPA `InterceptorService` bean are:
* Not triggered on non-storage pointcuts
  * Because we never hit the `RestfulServer`, the `RequestDetails`' interceptorservice is not populated
[...]

PR #7060, opened 2025-06-17
docs: document differential interceptor registration depending on pointcut

MartinBernstorff:
Based on this issue: https://github.com/hapifhir/hapi-fhir/issues/6993#issuecomment-2946125671

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/interceptors/server_interceptors.md
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/interceptors/server_interceptors.md
@@ -8,29 +8,32 @@ In the case of a JPA Server, HAPI FHIR itself performs both roles. This means th
 
 # Registering Interceptors
 
-How to register an interceptor differs depending on the type of pointcut it registers. This is especially important if you modify resources through a DAO.
+How to register an interceptor differs depending on the type of pointcut it registers. This is especially important if you modify resources without going through the `RestfulServer`.
 
-For example, a call through a DAO will not trigger an interceptor registered on the RestfulServer, even if this interceptor hooks onto a **STORAGE_xxx** pointcut.
+For example, if you are listening to a message queue, you can have a call:
 
-### **SERVER_xxx**
+`Message Queue -> Event listener -> DAO`
 
-|                 | Registered on |                     |
-| --------------- | ------------- | ------------------- |
-| **Call source** | RestfulServer | IInterceptorService |
-| REST            | ✅            | ❌                  |
-| DAO             | ❌             | ❌                   |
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/interceptors/server_interceptors.md
```diff
@@ -6,6 +6,32 @@ In the case of a Plain Server, HAPI FHIR itself performs the role of the Server
...
+How to register an interceptor differs depending on the type of pointcut it registers. This is especially important if you modify resources through a DAO.
+
+For example, a call through a DAO will not trigger an interceptor registered on the RestfulServer, even if this interceptor hooks onto a **STORAGE_xxx** pointcut.
+
+### **SERVER_xxx**
+
+|                 | Registered on |                     |
+| --------------- | ------------- | ------------------- |
+| **Call source** | RestfulServer | IInterceptorService |
+| REST            | ✅            | 🔴                 |
```
fil512: suggest using :x: instead of the red circle

MartinBernstorff: Done 👍

jamesagnew:
@MartinBernstorff : I'm seconding @tadgh 's comment above. There shouldn't be any cases where an interceptor registered against the RestfulServer don't fire within DAO methods, provided the DAO method is actually being called by the server.

So, this documentation is presumably not actually true.

If there are cases where this isn't working, can you please provide details, or even better, a unit test?

MartinBernstorff:
> @MartinBernstorff : I'm seconding @tadgh 's comment above. There shouldn't be any cases where an interceptor registered against the RestfulServer don't fire within DAO methods, provided the DAO method is actually being called by the server.
> 
> So, this documentation is presumably not actually true.
> 
> If there are cases where this isn't working, can you please provide details, or even better, a unit test?

Absolutely! I've had a few things come up, but I'll get back to this ASAP.

MartinBernstorff:
Good catch!

We have a slightly different case (which was not highlighted in the docs): we inject the DAOs into an event listener. The flow is:

`Message queue -> Event listener -> DAO`

The call does not come through the `RestfulServer`, which does not trigger the interceptors. However, if I register the interceptors on the `IInterceptorService` bean, the interceptor with storage pointcuts is triggered. 

I've pushed a new commit to reflect this.

I have reproduced this behaviour in proprietary code. If this does not match the behaviour you guys would expect, I'll of course reproduce in a unit test. Would appreciate any pointers to where in the HAPI-FHIR code such a repro could live.

PR #7066, opened 2025-06-18
Group resource without a patient member, when partitioning is enabled, can now be successfully created.

AbayomiE-T:
In patient id partitioning, Group resources are now placed in the default partition.


robogary:
Formatting check succeeded!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -112,8 +112,9 @@ public RequestPartitionId identifyForCreate(IBaseResource theResource, RequestDe
 				return provideCompartmentMemberInstanceResponse(theRequestDetails, oCompartmentIdentity.get());
 			} else {
 				return getPartitionViaPartiallyProcessedReference(theResource)
-						// or give up and fail
-						.orElseGet(() -> throwNonCompartmentMemberInstanceFailureResponse(theResource));
+						// Put into the default partition
```
michaelabuckley: problem: we should still fail on this path.  Please remove this new code.

suggestion: add explicit handling to ALWAYS put Group resources in the default compartment, just as if it had no compartment SPs defined.

michaelabuckley (changes requested):
Change it to always put Group in the default compartment, and make sure search matches.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
AbayomiE-T: Noted. But what about other resources in the Patient compartment that might be created without a value for the patient reference? Like `Observation.subject` for example. The create would fail in that instance. Unless it's unlikely that a customer would create an observation without a subject field?

michaelabuckley: Yes.  That is an explicit limitation of patient id mode.  Every resource in the patient compartment must have exactly ONE patient attached.  It is also illegal to change the patient once assigned.  That would change the partition id, and we don't support that.

That's why Group is such a bad fit, and we should have a special rule to treat it like it isn't.  Groups usually have many patients, but a resource can only be assigned to one partition.

AbayomiE-T: Aah gotcha. Thanks!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -104,6 +104,9 @@ public RequestPartitionId identifyForCreate(IBaseResource theResource, RequestDe
 								+ "Patient resource IDs must be client-assigned in patient compartment mode, or server id strategy must be UUID");
 			}
 			return provideCompartmentMemberInstanceResponse(theRequestDetails, idElement.getIdPart());
+		}
+		if (resourceDef.getName().equals("Group")) {
+			return RequestPartitionId.fromPartitionId(myPartitionSettings.getDefaultPartitionId());
```
TipzCM: Q: why does group need special case handling (and no other resource does)?

AbayomiE-T: Michael explained that a Group can have many patients but a resource can only be assigned to one partition. So this makes the Group resource a bad fit hence why it's given this special treatment of being assigned to the default partition.

michaelabuckley: suggestion: Please combine this branch with line 95 and call provideNonCompartmentMemberTypeResponse for consistency.

Remember that what we are doing is treating Group as if it were not a member of the Patient compartment.

Also add some docs to explain the exception.  Maybe link to the ticket.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7065-cant-create-group-resource-without-a-patient-member-when-partitioning-is-enabled.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 7065
+title: "Previously, if partitioning mode was enabled and a Group resource didn't have a Patient member,
+the PatientIdPartitionInterceptor would throw an error during a CREATE/PUT, leading to an unsuccessful group creation or update. This has been fixed and the Group resource in the request will be placed in the default partition."
```
michaelabuckley: Problem:The solution has changed, so the changelog must change.

[asdfas](url)
```suggestion
title: "In Patient Id partitioning (PatientIdPartitionInterceptor), Group resources are now treated as if they are not part of the Patient compartment and will be created in the default partition in all cases.  This avoids problems errors creating Groups in no Patient compartment, or many.  Previously, patient id partitioning mode was enabled and a Group resource didn't have a Patient member,
the PatientIdPartitionInterceptor would throw an error during a CREATE/PUT, leading to an unsuccessful group creation or update."
```

AbayomiE-T: Done!

PR #7070, opened 2025-06-19
Do not swallow errors

MartinBernstorff:
Currently, the `RepositoryValidatingInterceptor` outputs only the **first** issue, whether or not that is the issue that triggers the operation to fail validation. E.g. if the first issue is `INFO`, the user will not be informed why the validation failed.

This is a simple improvement that outputs diagnostics for all issues. 

Ideally, I think we'd sort the issues by severity, but I am not certain whether this information is available in `OperationOutcomeUtil.getIssueStringPart`.


jamesagnew:
This is a good idea. Would you mind to add a test of the new behaviour? Presumably RepositoryValidatingInterceptorR4Test would be a good place for it.

PR #7073, opened 2025-06-19
Return when trying to set validationSupport to the existing value

dotasek:
During configuration in JPA Server Starter, setValidationSupport is called twice, which triggers the "Can not set the validation support after it has already been set" error, even for working configurations.

However, in that instance, `myValidationSupport` does not actually change, as it is the same instance as the passed `theValidationSupport`. 

In this case, we should just return early.


robogary:
Formatting check succeeded!

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapter.java
```diff
@@ -169,6 +169,9 @@ public WorkerContextValidationSupportAdapter(@Nullable IValidationSupport theVal
 	 * Provides the {@link IValidationSupport} module that backs this adapter.
 	 */
 	public void setValidationSupport(IValidationSupport theValidationSupport) {
+		if (myValidationSupport == theValidationSupport) {
```
jamesagnew, line 172: Nitpick: A simple unit test for this would be a good thing

jamesagnew (approved):

PR #7075, opened 2025-06-19
move to new plugin

tadgh:
- Replace nexus plugin with central publishing plugin 
- Remove distribution management as that is taken care of by the plugin 
- Remove snapshots because that is also taken care of by the plugin 
- Remove repositories from certain projects as they can now all use central which is default in maven. 
- Version bump to ensure downstream projects are forced to move onto central (we will not publish 8.3.12-SNAPSHOT on oss.sonatype, only on central)

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/getting_started/downloading_and_importing.md
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/getting_started/downloading_and_importing.md
@@ -65,18 +65,22 @@ Generally speaking, it is a good idea to use a stable build. However, FHIR is a 
 
 Snapshot builds of HAPI are pre-release builds which can contain fixes and new features not yet released in a formal release. As of HAPI-FHIR 8.3.11-SNAPSHOT, snapshots are automatically pulled from central. For older versions, you may use the following snippet to use snapshot builds of HAPI which adds a reference to the OSS snapshot repository to your project build file.
 
-Using a snapshot build generally involves appending *-SNAPSHOT* to the version number, e.g. `4.1.0-SNAPSHOT`. In order to automatically download snapshot builds. If you are attempting to use 8.3.11-SNAPSHOT or earlier,  you will need to add a snapshot repository to your build file as shown below:
+Using a snapshot build generally involves appending *-SNAPSHOT* to the version number, e.g. `8.3.12-SNAPSHOT`. In order to automatically download snapshot builds. If you are attempting to use 8.3.11-SNAPSHOT or later, you will need to add a snapshot repository to your build file as shown below:
 
 ### Using Maven:
 
 ```xml
 <repositories>
     <repository>
-        <id>oss-snapshots</id>
+        <name>Central Portal Snapshots</name>
+        <id>central-portal-snapshots</id>
+        <url>https://central.sonatype.com/repository/maven-snapshots/</url>
+        <releases>
+            <enabled>false</enabled>
+        </releases>
         <snapshots>
             <enabled>true</enabled>
         </snapshots>
-        <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
     </repository>
 </repositories>
 ```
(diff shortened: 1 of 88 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/getting_started/downloading_and_importing.md
```diff
@@ -65,7 +65,7 @@ Generally speaking, it is a good idea to use a stable build. However, FHIR is a
 
 Snapshot builds of HAPI are pre-release builds which can contain fixes and new features not yet released in a formal release. As of HAPI-FHIR 8.3.11-SNAPSHOT, snapshots are automatically pulled from central. For older versions, you may use the following snippet to use snapshot builds of HAPI which adds a reference to the OSS snapshot repository to your project build file.
 
-Using a snapshot build generally involves appending *-SNAPSHOT* to the version number, e.g. `4.1.0-SNAPSHOT`. In order to automatically download snapshot builds, you may also need to add a snapshot repository to your build file as shown below:
+Using a snapshot build generally involves appending *-SNAPSHOT* to the version number, e.g. `4.1.0-SNAPSHOT`. In order to automatically download snapshot builds. If you are attempting to use 8.3.11-SNAPSHOT or earlier,  you will need to add a snapshot repository to your build file as shown below:
 
 ### Using Maven:
```
dotasek, line 70: The snapshot repository will need to point to maven central now:

```xml
 <repository>
                <name>Central Portal Snapshots</name>
                <id>central-portal-snapshots</id>
                <url>https://central.sonatype.com/repository/maven-snapshots/</url>
                <releases>
                    <enabled>false</enabled>
                </releases>
                <snapshots>
                    <enabled>true</enabled>
                </snapshots>
            </repository>
```

tests/hapi-fhir-base-test-jaxrsserver-kotlin/pom.xml
```diff
@@ -17,17 +17,6 @@
 
 	<name>HAPI FHIR JAX-RS Server Kotlin test</name>
 
-	<repositories>
```
dotasek, line 20: Same change as above:

```xml
 <repository>
                <name>Central Portal Snapshots</name>
                <id>central-portal-snapshots</id>
                <url>https://central.sonatype.com/repository/maven-snapshots/</url>
                <releases>
                    <enabled>false</enabled>
                </releases>
                <snapshots>
                    <enabled>true</enabled>
                </snapshots>
            </repository>
```

dotasek (changes requested):
Didn't realize the docs contained the old oss sonatype server. That should be fixed.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/getting_started/downloading_and_importing.md
dotasek: I just checked the sonatype snapshots and it appears as though our past SNAPSHOTs are not present on central.sonatype.com. That could trip a few people up; either we deploy every previous SNAPSHOT, or we suggest they build locally.

tadgh: I dont think we want to do the former, and people should not really be relying on old snapshots. They can upgrade to PRE-12 imo.

tadgh: Ah yeah gotta manually define the snapshot repo. Fixing

tests/hapi-fhir-base-test-jaxrsserver-kotlin/pom.xml
tadgh: Why do we even need this at all? the kotlin server test has a local parent, why does it need a snapshot repo defined?

dotasek: Good point. If it built, it should be fine without it. Leave it out.

PR #7076, opened 2025-06-19
Add hapi-fhir-storage-batch2-test-utilities to bom.

lukedegruchy:
- Expose hapi-fhir-storage-batch2-test-utilities in the bom
- This means that consumers of the bom can now make use of test utilities, such as InlineJobCoordinator to test batch2 JobDefinitions in a streamlined way


robogary:
Formatting check succeeded!

jmarchionatto (approved):

PR #7077, opened 2025-06-19
67121: Stop sorting the RuntimeSearchParams based on name.

another-luis:
They are in declared order, which should match the positional order of the request parameters.

Also added a test to replicate the issue and verify the fix.

### DIFF
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/util/JpaParamUtil.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/util/JpaParamUtil.java
@@ -146,13 +146,7 @@ public enum JpaParamUtil {
 
 	public static List<RuntimeSearchParam> resolveComponentParameters(
 			ISearchParamRegistry theSearchParamRegistry, RuntimeSearchParam theParamDef) {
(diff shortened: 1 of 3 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/67121-stop-using-runtime-search-params-sorted-by-name.yaml
```diff
@@ -0,0 +1,5 @@
+type: fix
+issue: 67121
+title: "Context-value-type searches were using search-param lists sorted alphabetically. 
+They now use positional sorting"
```
michaelabuckley: minor: This is hard to understand from a user perspective.
```suggestion
title: "Composite SearchParameters used the query parameter components in the wrong order when the component order did not match alphabetical order.  This affected Library.context-value-type searches. 
This has been corrected."
```

another-luis: Modified the message to make it more user-oriented.

PR #7080, opened 2025-06-23
Add interceptor hook to customize auto-created placeholder reference targets

jamesagnew:
This PR adds a new interceptor hook which can be used to prevent specific auto-created placeholder reference targets from being created, or to modify those resources.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
@@ -1087,6 +1087,44 @@ public enum Pointcut implements IPointcut {
 
 	/**
 	 * <b>Storage Hook:</b>
+	 * Invoked when we are about to <a href="https://smilecdr.com/docs/fhir_repository/creating_data.html#auto-create-placeholder-reference-targets">Auto-Create a Placeholder Reference</a>.
+	 * Hooks may modify/enhance the placeholder reference target that is about to be created, or
+	 * reject the creation of the resource, which generally means that the transaction will be
+	 * rejected instead (because of the invalid reference).
+	 * <p>
+	 * Hooks may accept the following parameters:
+	 * </p>
+	 * <ul>
+	 * <li>
+	 * ca.uhn.fhir.storage.interceptor.AutoCreatePlaceholderReferenceTargetRequest - Contains details about the placeholder that is about to be created, including the source resource whose reference is being fulfilled, as well as the candidate placeholder resource that will be created.
+	 * </li>
+	 * <li>
+	 * ca.uhn.fhir.rest.api.server.RequestDetails - A bean containing details about the request that is about to be processed, including details such as the
+	 * resource type and logical ID (if any) and other FHIR-specific aspects of the request which have been
+	 * pulled out of the servlet request. Note that the bean
+	 * properties are not all guaranteed to be populated, depending on how early during processing the
+	 * exception occurred. <b>Note that this parameter may be null in contexts where the request is not
+	 * known, such as while processing searches</b>
+	 * </li>
+	 * <li>
+	 * ca.uhn.fhir.rest.server.servlet.ServletRequestDetails - A bean containing details about the request that is about to be processed, including details such as the
+	 * resource type and logical ID (if any) and other FHIR-specific aspects of the request which have been
+	 * pulled out of the servlet request. This parameter is identical to the RequestDetails parameter above but will
+	 * only be populated when operating in a RestfulServer implementation. It is provided as a convenience.
+	 * </li>
+	 * </ul>
+	 * <p>
+	 * Hooks may return <code>void</code>, an object of type <code>ca.uhn.fhir.storage.interceptor.AutoCreatePlaceholderReferenceTargetResponse</code>, and can throw exceptions.
+	 * </p>
+	 */
+	STORAGE_PRE_AUTO_CREATE_PLACEHOLDER_REFERENCE(
+		"ca.uhn.fhir.storage.interceptor.AutoCreatePlaceholderReferenceTargetResponse",
+		"ca.uhn.fhir.storage.interceptor.AutoCreatePlaceholderReferenceTargetRequest",
+		"ca.uhn.fhir.rest.api.server.RequestDetails",
+		"ca.uhn.fhir.rest.server.servlet.ServletRequestDetails"),
+
+	/**
+	 * <b>Storage Hook:</b>
 	 * Invoked when a Bulk Export job is being kicked off, but before any permission checks
 	 * have been done.
 	 * This hook can be used to modify or update parameters as need be before
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoCreatePlaceholdersR4Test.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoCreatePlaceholdersR4Test.java
@@ -1,25 +1,30 @@
 package ca.uhn.fhir.jpa.dao.r4;
 
 import ca.uhn.fhir.i18n.Msg;
+import ca.uhn.fhir.interceptor.api.Hook;
+import ca.uhn.fhir.interceptor.api.Pointcut;
 import ca.uhn.fhir.jpa.api.config.JpaStorageSettings;
 import ca.uhn.fhir.jpa.api.model.DaoMethodOutcome;
 import ca.uhn.fhir.jpa.dao.TransactionUtil;
+import ca.uhn.fhir.jpa.model.dao.JpaPid;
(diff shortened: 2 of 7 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoCreatePlaceholdersR4Test.java
```diff
@@ -1,25 +1,31 @@
 package ca.uhn.fhir.jpa.dao.r4;
```
jamesagnew, line 1: Note to reviewer: Most of the changes in this file are just fixing compiler warnings from deprecated method calls that are old and crufty. The actual meaningful changes are the `testInterceptor_***` methods towards the bottom of this class.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoResourceLinkResolver.java
```diff
@@ -302,6 +315,41 @@ public <T extends IBaseResource> Optional<IBasePersistedResource> createPlacehol
 					theTransactionDetails.addRollbackUndoAction(() -> newResource.setId(existingId));
 				}
 				newResource.setId(resName + "/" + theIdToAssignToPlaceholder);
+			}
```
fil512, line 304: I know not in this MR, but line 214 above doesn't make any sense:
```
	if (!searchParam.hasTargets() && searchParam.getTargets().contains(resourceType)) {
			return null;
		}
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
```diff
@@ -1085,6 +1085,44 @@ public enum Pointcut implements IPointcut {
...
+	 * Invoked when we are about to <a href="https://smilecdr.com/docs/fhir_repository/creating_data.html#auto-create-placeholder-reference-targets">Auto-Create a Placeholder Reference</a>.
+	 * Hooks may modify/enhance the placeholder reference target that is about to be created, or
+	 * reject the creation of the resource, which generally means that the transaction will be
+	 * rejected instead (because of the invalid reference).
+	 * <p>
+	 * Hooks may accept the following parameters:
+	 * </p>
+	 * <ul>
+	 * <li>
+	 * ca.uhn.fhir.storage.interceptor.AutoCreatePlaceholderReferenceTargetRequest - Contains details about the placeholder that is about to be created, including the source resource whose reference is being fulfilled, as well as the candidate placeholder resource that will be created.
```
fil512, line 1099: Nice. request objects are the way to go for pointcuts.

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
```diff
@@ -1085,6 +1085,44 @@ public enum Pointcut implements IPointcut {
...
+	 * </li>
+	 * <li>
+	 * ca.uhn.fhir.rest.api.server.RequestDetails - A bean containing details about the request that is about to be processed, including details such as the
+	 * resource type and logical ID (if any) and other FHIR-specific aspects of the request which have been
+	 * pulled out of the servlet request. Note that the bean
+	 * properties are not all guaranteed to be populated, depending on how early during processing the
+	 * exception occurred. <b>Note that this parameter may be null in contexts where the request is not
+	 * known, such as while processing searches</b>
```
fil512: I don't believe RequestDetails could ever be null when creating a placeholder reference

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
```diff
@@ -1085,6 +1085,44 @@ public enum Pointcut implements IPointcut {
...
+	 * </li>
+	 * <li>
+	 * ca.uhn.fhir.rest.server.servlet.ServletRequestDetails - A bean containing details about the request that is about to be processed, including details such as the
+	 * resource type and logical ID (if any) and other FHIR-specific aspects of the request which have been
+	 * pulled out of the servlet request. This parameter is identical to the RequestDetails parameter above but will
+	 * only be populated when operating in a RestfulServer implementation. It is provided as a convenience.
+	 * </li>
+	 * </ul>
+	 * <p>
+	 * Hooks may return <code>void</code>, an object of type <code>ca.uhn.fhir.storage.interceptor.AutoCreatePlaceholderReferenceTargetResponse</code>, and can throw exceptions.
```
fil512: Is it true these hooks can return void? I thought we validated return type...?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoResourceLinkResolver.java
```diff
@@ -302,6 +315,41 @@ public <T extends IBaseResource> Optional<IBasePersistedResource> createPlacehol
...
+								Pointcut.STORAGE_PRE_AUTO_CREATE_PLACEHOLDER_REFERENCE, params);
+				if (response != null) {
+					if (response.isDoNotCreateTarget()) {
+						return Optional.empty();
+					}
+				}
+
+				// Sanity check: Make sure that interceptors haven't changed the ID
+				if (theIdToAssignToPlaceholder != null) {
+					Validate.isTrue(
```
fil512, line 327: What about identifiers?  I think when we create a placeholder based on an identifier we also set that identifier?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/storage/interceptor/AutoCreatePlaceholderReferenceTargetResponse.java
```diff
@@ -0,0 +1,45 @@
...
+	public boolean isDoNotCreateTarget() {
+		return myDoNotCreateTarget;
+	}
+
+	/**
+	 * Create response: The placeholder reference target should <b>not</b> be created.
+	 * Under most circumstances this will cause the operation to be aborted due to
+	 * the invalid reference.
+	 */
+	public static AutoCreatePlaceholderReferenceTargetResponse doNotCreateTarget() {
```
fil512, line 35: nitpick: This follows the Collections.emptyList() convention. Personally, I find all-caps static final fields a bit easier to follow. I.e.
```
public static final DO_NOT_CREATE_TARGET = new AutoCreatePlaceholderReferenceTargetResponse(false);
```
but admiteldly it's a style preference.

When I see a method with a name `doNotCreateTarget()` I imagine that method is going to go do something...

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoResourceLinkResolver.java
jamesagnew: Wow, good catch! I just did some spelunking, and it looks like oddly enough, you wrote the original version of this line (which had more sensible meaning) and I messed it up as a part of cleanup in this commit: 9c93d5d7f7e2c8d3b4a9177a12765733b1823a7f

I added a test, and it looks like this line is redundant. We have the same check (but working correctly) in SearchParamExtractorService. I'm gonna remove it (which feels zero-risk) given that the line was clearly wrong.

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
jamesagnew: Ah good call. That whole blurb is copy-pasted and doesn't actually make sense. Fixed.

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
jamesagnew: Oh yeah, null is fine. Just added a test and beefed up the docs.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoResourceLinkResolver.java
jamesagnew: We do - I dont see a reason to restrict this though. Changing the ID is just going to cause broken semantics, but I can think of valid reasons for rewriting identifiers.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/storage/interceptor/AutoCreatePlaceholderReferenceTargetResponse.java
jamesagnew: I feel like we use the camelCase convention pretty universally though in HAPI.... and the JDK uses it too....

PR #7083, opened 2025-06-24
7082 add ability to filter everything

### LINKED ISSUE #7082: Add ability to filter everything/export results
Everything and export operations follow a spec.

But there's a desire to have an ability to filter out specific resources from these operations all the same.

TipzCM:
closes #7082

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
@@ -928,7 +928,7 @@ public class SearchBuilder implements ISearchBuilder<JpaPid> {
 
 		if (myParams.getEverythingMode().isPatient()) {
 			Collection<String> resourcesToOmit =
-				mySearchLimiterSvc.getResourcesToOmitForOperationSearches(JpaConstants.OPERATION_EVERYTHING);
+					mySearchLimiterSvc.getResourcesToOmitForOperationSearches(JpaConstants.OPERATION_EVERYTHING);
 			sqlBuilder.excludeResourceTypesPredicate(resourcesToOmit);
 		}
 
--- a/hapi-fhir-docs/src/main/java/ca/uhn/hapi/fhir/docs/ServerOperations.java
+++ b/hapi-fhir-docs/src/main/java/ca/uhn/hapi/fhir/docs/ServerOperations.java
@@ -123,7 +123,6 @@ public class ServerOperations {
 	}
 	// END SNIPPET: serverOperation
 
-
 	private ISearchLimiterSvc mySearchLimiterSvc;
 
 	// START SNIPPET: resourceTypeFiltering
--- a/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/FetchResourceIdsStep.java
+++ b/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/FetchResourceIdsStep.java
@@ -90,10 +90,11 @@ public class FetchResourceIdsStep implements IFirstJobStepWorker<BulkExportJobPa
 		 */
 		List<String> resourceTypesToFetch;
 		if (theStepExecutionDetails.getParameters().getExportStyle() == BulkExportJobParameters.ExportStyle.PATIENT) {
-			Collection<String> omitted = mySearchLimiterSvc.getResourcesToOmitForOperationSearches(ProviderConstants.OPERATION_EXPORT);
+			Collection<String> omitted =
+					mySearchLimiterSvc.getResourcesToOmitForOperationSearches(ProviderConstants.OPERATION_EXPORT);
 			resourceTypesToFetch = params.getResourceTypes().stream()
-				.filter(r -> !omitted.contains(r))
-				.toList();
+					.filter(r -> !omitted.contains(r))
+					.toList();
 		} else {
 			resourceTypesToFetch = params.getResourceTypes().stream().toList();
 		}


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7082-add-ability-to-filter-everything-export.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: add
+issue: 7082
+title: "Add an ability to filter resources out of
+        `$everything` and `$export` operations.
+"
```
tadgh, line 13: issue(blocking): This changelog requires more information. How? Where? For what purpose?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -844,6 +851,7 @@ private void createChunkedQueryForEverythingSearch(
 			List<JpaPid> thePidList,
 			List<ISearchQueryExecutor> theSearchQueryExecutors) {
 
+		// EVERYTHING
```
tadgh: issue: remove comment

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/models/DoLoadPidsParams.java
```diff
@@ -0,0 +1,76 @@
...
+	/**
+	 * Whether or not this is for a history operation
+	 */
+	private boolean myForHistoryOperation;
+	/**
+	 * The position map
+	 */
+	private Map<Long, Integer> myPosition;
+
+	public SearchParameterMap getParameterMap() {
```
tadgh: issue: parameterMap -> SearchParameterMap

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/models/DoLoadPidsParams.java
```diff
@@ -0,0 +1,76 @@
+package ca.uhn.fhir.jpa.search.builder.models;
+
+import ca.uhn.fhir.jpa.model.dao.JpaPid;
+import ca.uhn.fhir.jpa.searchparam.SearchParameterMap;
+
+import java.util.Collection;
+import java.util.Map;
+
+public class DoLoadPidsParams {
```
tadgh: issue: absence of a javadoc

tadgh: suggestion: I would think a bit more about the name here. I see its the params for `doLoadPids` method, but the name of this class itself is a bit sharp. Maybe `PidLoadingParams` or `PidInflationParameters`? Since what this is doing is taking the pid list, and being used to actually return resource bodies. Not a huge blocker, but a name that would be meaningful _without_ the context of the method would be helpful.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -1431,9 +1442,15 @@ public void loadResourcesByPid(
 		}
 
 		// We only chunk because some jdbc drivers can't handle long param lists.
-		QueryChunker.chunk(
-				thePids,
-				t -> doLoadPids(t, theIncludedPids, theResourceListToPopulate, theForHistoryOperation, position));
+		DoLoadPidsParams params = new DoLoadPidsParams();
```
tadgh: issue: If we are going to be using this parameters object, why not propagate it up the stack?

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
```diff
@@ -699,6 +699,12 @@ private void extractResourceLinks(
 		}
 
 		IIdType referenceElement = thePathAndRef.getRef().getReferenceElement();
+		if (isBlank(referenceElement.getValue())) {
+			// it's an embedded element maybe;
+			// we need a valid referenceElement, becuase we
+			// resolve the resource by this value (and if we use "null", we can't resolve multiple values)
+			referenceElement = thePathAndRef.getRef().getResource().getIdElement();
+		}
```
tadgh, line 707: question: Why is this code here? It feels unrelated to the top-level resource omitting logic.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/SearchLimiterSvc.java
```diff
@@ -0,0 +1,46 @@
...
+					theResourceType);
+		}
+		myOperationToOmittedResourceTypes.put(theOperationName, theResourceType);
+	}
+
+	@Override
+	public Set<String> getResourcesToOmitForOperationSearches(@Nonnull String theOperationName) {
+		Collection<String> omittedResourceNames = myOperationToOmittedResourceTypes.get(theOperationName);
+		return Set.copyOf(omittedResourceNames);
+	}
```
tadgh, line 35: Question: This code is on the extremely hot path, do we know the performance impact of cloning the set on every get call?

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/SearchLimiterSvc.java
```diff
@@ -0,0 +1,46 @@
...
+
+	@Override
+	public void removeOmittedResourceType(@Nonnull String theOperationName, String theResourceType) {
+		if (isBlank(theResourceType)) {
```
tadgh: issue: this blank behaviour is not documented anywhere, and I think blank should be an error or a no-op. If you really want to keep it, i would make two different signatures, one which only takes theOperationName and removes all.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportTest.java
```diff
@@ -320,6 +330,67 @@ public void testGroupBulkExportWithTypeFilter_OnTags_InlineTagMode() {
...
+		myGroupDao.update(group, rd);
+
+		BulkDataExportSupport support = new BulkDataExportSupport(
+			myFhirContext,
+			myDaoRegistry,
+			myRequestPartitionHelperSvc
+		);
+
+		// test
+		StorageSettings.IndexEnabledEnum currentIndexSetting = myStorageSettings.getIndexMissingFields();
```
tadgh, line 362: issue: even with the try-finally reset, surefire will parallelize tests, and its possible another test will run with this swapped value. Not the end of the world though, I thikn its pretty low impat.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportTest.java
```diff
@@ -320,6 +330,67 @@ public void testGroupBulkExportWithTypeFilter_OnTags_InlineTagMode() {
...
+		try {
+			// setup for test
+			mySearchLimiterSvc.addOmittedResourceType(ProviderConstants.OPERATION_EXPORT, "Group");
+			myStorageSettings.setIndexMissingFields(StorageSettings.IndexEnabledEnum.ENABLED);
```
tadgh, line 368: question: Why do we have to index missing to test the behaviour, especially given that the resources are already created and indexed by the time we get here?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportTest.java
```diff
@@ -320,6 +330,67 @@ public void testGroupBulkExportWithTypeFilter_OnTags_InlineTagMode() {
...
+			// verify
+			awaitExportComplete(results.getInstanceId());
+
+			BulkExportJobResults response = getBulkExportJobResults(results.getInstanceId());
+
+			Map<String, List<String>> outputMap = response.getResourceTypeToBinaryIds();
+			ourLog.info("Resource Types returned: " + String.join(", ", outputMap.keySet()));
+			assertTrue(outputMap.containsKey("Observation"));
+			assertTrue(outputMap.containsKey("Patient"));
+			assertFalse(outputMap.containsKey("Group"));
```
tadgh: nitpick: swap for assertJ single-line style assert, e.g. 

```java
		assertThat(theMap).containsOnlyKeys("Patient", "Observation").doesNotContainKey("Group")

```

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportTest.java
```diff
@@ -320,6 +330,67 @@ public void testGroupBulkExportWithTypeFilter_OnTags_InlineTagMode() {
...
+		} finally {
+			// reset
+			mySearchLimiterSvc.removeOmittedResourceType(ProviderConstants.OPERATION_EXPORT, null);
```
tadgh: actually, i wouldn't be shocked if this broke other tests. Consider extracting this behaviour into its own class for test runner isolation

tadgh (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7082-add-ability-to-filter-everything-export.yaml
TipzCM: I added more details.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
TipzCM: removed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/models/DoLoadPidsParams.java
TipzCM: Changed to PidLoadingParams

TipzCM: added some more docs (but this is a private method params object, so i'm not sure how useful it'll be)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/models/DoLoadPidsParams.java
TipzCM: updated

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
TipzCM: A bug i found while writing my test.

The test requres ensuring that embedded resources can be filtered out, even if they are embedded as references holding the entire resource itself.

Problem is, if we have a reference holding an entire resource, the server oesn't work (it can't actually create the references because of this bug).

Because it's a 1 line fix, i thought it ok to put here.

Should i instead make an entirely new bug?

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/SearchLimiterSvc.java
TipzCM: Removed

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/SearchLimiterSvc.java
TipzCM: Split into separate method

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportTest.java
TipzCM: Required for bulkexports at all.

I've added some lines documenting this.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportTest.java
TipzCM: ok...

TipzCM:
Re: general comment

1. I did not do any SQL analysis. But for $everything , HFJ_Resource table has an index with RES_TYPE as the first column; the filtering added adds a where on this type, so it's more (not less) likely to use the index. $export removes the resource (and querying for it) entirely, so it's not applicable here. I suppose, there might be a concern with the where getting too long. But that's a general issue and i doubt even if they put every resource into the omission list the query would be too long because of this feature only

2. should i add an entire doc page for this new service, or do you think it's ok to just add a line about it in rest_operations_operations (the only relevant place in hapi-fhir docs that references $everything operation) . Or is there a different place i should put it (search.md maybe?)

3. This is by design. We only intended to add it (here) for $everything and $export, but wanted to leave it open ended enough that future operations could use it (even if they don't right now). Still, if you want, i can scale the service back to something a bit more rigidly defined for operation names (although, i might do it with an enum, so any future changes are as minimal as possible; thoughts?)

4. this is a thing i guess we agree to disagree on; ironically with the exact same reasoning (i find parameters objects far less mentally taxing than lists of parameters that require me to count or scroll; i can also more easily find where it's created by finding the setters use, instead of chaining up the tree until it no longer has a 'the'). But either way, i've removed the object since it's not needed in this specific case anyways

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
TipzCM: removed params so this is no longer applicable

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportTest.java
TipzCM: I'll leave the test here for now. It runs in 2 sec and i feel like creating a new full test class (including set up and teardown) is going to be overkill for such a small test.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportTest.java
TipzCM: Hmm. So should i extract this single test into its own class then?

I'm not sure how surefire's parallelizing works. Does it use the same exact build and code and just run the tests at the same time?

If so, how does any set of tests not interfere with itself (we create resources, etc, but other tests frequently search for resources as well and ensure only the ones they created exist - see any mdm test for instance).

I just wonder if this is truly the case, it means we *can't* run tests in parallel (at least not ITs, only true Unit Tests (with mocking) where there is no persistent state at all.

tadgh: we dont run ITs in parallel. ITs always run single-file (via failsafe, not surefire). Do me a favour, and just loop run the mvn verify command over this subproject like 50-60 times. If it never fails, then we can go ahead and keep it.

tadgh (approved):

PR #7094, opened 2025-06-30
Bump to core 6.5.27

dotasek:
This removes several long deprecated methods that were finally deleted from org.hl7.fhir.core

It also introduces an unimplemented validateTxResource() method now present in org.hl7.fhir.core and accounts for some new error messaging regarding base64 validation.


robogary:
Formatting check succeeded!

volodymyr-korzh (approved):

PR #7107, opened 2025-07-06
7031 valueset expand operation does not support equl filter operator

fil512:
Evaluating how Claude performs to resolve this issue


robogary:
Formatting check succeeded!

PR #7113, opened 2025-07-09
improve FhirAutoConfiguration

plchldr:
I'm not sure whether we're doing something wrong, but `HapiEntityManagerFactoryUtil.newEntityManagerFactory(ConfigurableListableBeanFactory, FhirContext, JpaStorageSettings)` requires `JpaStorageSettings` built in  `FhirJpaStorageSettingsConfiguration`, leading to the construction of the configuration requiring the `EntityManagerFactory` that is currently being constructed by `HapiEntityManagerFactoryUtil`. So it's a dependency circle and Spring Boot aborts.

Our current workaround is duplicating the `FhirJpaStorageSettingsConfiguration::storageSettings` method - so no big deal, but this seems cleaner either way.


robogary:
Formatting check succeeded!

PR #7123, opened 2025-07-14
Export specification - '_until' Query parameter

MadsSwensson:
Added the '_until' Query parameter to the bulk data export endpoints according to the specs:
https://build.fhir.org/ig/HL7/bulk-data/export.html#query-parameters

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportProviderR4Test.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportProviderR4Test.java
@@ -193,6 +193,7 @@ public class BulkDataExportProviderR4Test {
 		input.addParameter(JpaConstants.PARAM_EXPORT_OUTPUT_FORMAT, new StringType(Constants.CT_FHIR_NDJSON));
 		input.addParameter(JpaConstants.PARAM_EXPORT_TYPE, new StringType(patientResource + ", " + practitionerResource));
 		input.addParameter(JpaConstants.PARAM_EXPORT_SINCE, now);
+		input.addParameter(JpaConstants.PARAM_EXPORT_UNTIL, now);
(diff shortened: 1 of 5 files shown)


hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkDataExportProviderR4Test.java
```diff
@@ -193,6 +193,7 @@ public void testSuccessfulInitiateBulkRequest_Post_WithFixedBaseURLAndPartitioni
 		input.addParameter(JpaConstants.PARAM_EXPORT_OUTPUT_FORMAT, new StringType(Constants.CT_FHIR_NDJSON));
 		input.addParameter(JpaConstants.PARAM_EXPORT_TYPE, new StringType(patientResource + ", " + practitionerResource));
 		input.addParameter(JpaConstants.PARAM_EXPORT_SINCE, now);
+		input.addParameter(JpaConstants.PARAM_EXPORT_UNTIL, now);
```
jamesagnew: Suggestion: This would probably be a more robust test if we used different dates for the `_since` and `_until` params.

jamesagnew:
@MadsSwensson This looks like a useful addition. It looks like the `_until` param is supported in the backend, but it doesn't have any good tests that I can see. If we're adding it to the API for calling the job we should probably have a test that verifies it actually works in the backend too.

Would you be able to add a test to `BulkDataExportTest` which verifies that `_until` actually works? You can search that test for `_since` to see an example of that parameter being tested.

MadsSwensson:
@jamesagnew I've added a couple of tests, one that contains multiple resources which are in the export and one where some are filtered out based on the ```_until``` parameter. 

Feel free to rename or give suggestions to the tests, since I am not familiarized with your test case naming conventions.

robogary:
Formatting check succeeded!

jamesagnew:
Thanks for the contribution! This will be in the upcoming 8.4.0 release.

PR #7125, opened 2025-07-14
Case-sensitivity in Conditional Update based on tokens

### LINKED ISSUE #7124: Case-sensitivity in Conditional Update based on tokens
The condition in a Conditional Update when based on a token does not behave properly if the case of the condition doesn’t match the case of the Resource.

**Steps to reproduce**

Do a PUT to /Patient?identifier=[http://tempuri.org|http://tempuri.org/]|ABC, with the following body:
```
{
    "resourceType": "Patient",
    "identifier": [
        {
            "system": "http://tempuri.org",
            "value": "abc"
        }
    ]
}
```

notice how the resource gets created even though ABC does not match abc. It should fail with "HAPI-0929: Failed to process conditional create. The supplied resource did not satisfy the conditional URL." because the the case of the search string doesn't match.

Do a GET /Patient?identifier=[http://tempuri.org|http://tempuri.org/]|ABC, and notice that no patient resource is in the response. **GET request proves that the value is to be case sensitive as it does not match when searching the upper case value**.

YalingPeiS:
The condition in a Conditional Update when based on a token does not behave properly if the case of the condition doesn’t match the case of the Resource.

**Steps to reproduce**

Do a PUT to /Patient?identifier=[http://tempuri.org|http://tempuri.org/]|ABC, with the following body:
```
{
    "resourceType": "Patient",
    "identifier": [
        {
            "system": "http://tempuri.org",
            "value": "abc"
        }
    ]
}
```

notice how the resource gets created even though ABC does not match abc. It should fail with "HAPI-0929: Failed to process conditional create. The supplied resource did not satisfy the conditional URL." because the the case of the search string doesn't match.

Do a GET /Patient?identifier=[http://tempuri.org|http://tempuri.org/]|ABC, and notice that no patient resource is in the response. **GET request proves that the value is to be case sensitive as it does not match when searching the upper case value**.

Closes #7124

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7125-case-sensitivity-in-conditional-update-by-token.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7125-case-sensitivity-in-conditional-update-by-token.yaml
@@ -1,0 +1,7 @@
+---
+type: fix
+issue: 7125
+jira: SMILE-10427
+title: "Previously, conditional updates based on tokens were case-insensitive. As a result, a new resource would be 
+created if the token value in the URL used a different case than the one in the request body, leading to incorrect 
+search results when using token-based search parameters. This issue has now been resolved."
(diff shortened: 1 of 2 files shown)


robogary:
Formatting check succeeded!

PR #7131, opened 2025-07-15
Fix backward incompatible method signature change in batch2 JobDefinitionStartRequest

lukedegruchy:
- Restore the old method signature returning void
- Add a new method similar to the updated returning the request, but with a different name.


robogary:
Formatting check succeeded!

tadgh (commented):
Waiting to see if we actually need to do this

tadgh (changes requested):
blocking merge until we figure it out.

tadgh (approved):

PR #7135, opened 2025-07-16
Update fhir terser

AD1306:
- small util to fhir terser to check if the field exists.


robogary:
Formatting check succeeded!

jmarchionatto (approved):
Approved

PR #7142, opened 2025-07-17
enforce_referential_integrity_on_delete.disable_for_paths ignored for expunge

### LINKED ISSUE #7139: enforce_referential_integrity_on_delete.disable_for_paths ignored for expunge
Delete_Expunge does not work when placing referential integrity exceptions using Disable Deletion Ref Checks for Paths

**To Reproduce**
1. Create any Resource
2. Create an AuditEvent Resource, where "entity.what" is a reference to the Resource created in 1.
3. Do a DELETE on {base}/{resourceType}/_id={id from 1}&_expunge=true

A batch job will be created and will fail because of the reference from AuditEvent. Note that issuing the DELETE without _expunge=true succeeds.

YalingPeiS:
Delete_Expunge does not work when placing referential integrity exceptions using Disable Deletion Ref Checks for Paths

### To Reproduce
1. Create any Resource
2. Create an AuditEvent Resource, where "entity.what" is a reference to the Resource created in 1.
3. Do a DELETE on {base}/{resourceType}/_id={id from 1}&_expunge=true

A batch job will be created and will fail because of the reference from AuditEvent. Note that issuing the DELETE without _expunge=true succeeds.

### Cause
No property `EnforceReferentialIntegrityOnDeleteDisableForPaths` is defined when running the job of  Delete + Expunge

### Solution
Define the property `EnforceReferentialIntegrityOnDeleteDisableForPaths` and updated Delete Expunge batch job process with disabled path checks

Closes #7139

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/batch2/DeleteExpungeSqlBuilder.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/batch2/DeleteExpungeSqlBuilder.java
@@ -132,6 +132,12 @@ public class DeleteExpungeSqlBuilder {
 					break;
 				}
 			}
+		} else {
+			// check if the user has configured any paths to ignore
+			Set<String> pathsToIgnore = myStorageSettings.getEnforceReferentialIntegrityOnDeleteDisableForPaths();
+			if (conflictResourceLinks.stream().anyMatch(link -> pathsToIgnore.contains(link.getSourcePath()))) {
+				return;
+			}
 		}
 
 		ResourceLink firstConflict = conflictResourceLinks.get(0);
--- a/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/batch2/DeleteExpungeSqlBuilderTest.java
+++ b/hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/batch2/DeleteExpungeSqlBuilderTest.java
@@ -7,40 +7,41 @@ import ca.uhn.fhir.jpa.dao.expunge.ResourceTableFKProvider;
 import ca.uhn.fhir.jpa.delete.batch2.DeleteExpungeSqlBuilder;
 import ca.uhn.fhir.jpa.model.config.PartitionSettings;
 import ca.uhn.fhir.jpa.model.dao.JpaPid;
+import ca.uhn.fhir.jpa.model.entity.ResourceLink;
+import ca.uhn.fhir.jpa.model.entity.ResourceTable;
 import ca.uhn.fhir.jpa.util.QueryChunker;
-import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.mockito.MockedStatic;
 import org.mockito.Mockito;
-import org.springframework.beans.factory.annotation.Autowired;
+import org.mockito.junit.jupiter.MockitoExtension;
 
 import java.util.List;
+import java.util.Set;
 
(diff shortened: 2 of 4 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/batch2/DeleteExpungeSqlBuilder.java
```diff
@@ -132,6 +132,12 @@ public void validateOkToDeleteAndExpunge(Set<JpaPid> thePids, boolean theCascade
 					break;
 				}
 			}
+		} else {
+			// check if the user has configured any paths to ignore
+			Set<String> pathsToIgnore = myStorageSettings.getEnforceReferentialIntegrityOnDeleteDisableForPaths();
+			if (conflictResourceLinks.stream().anyMatch(link -> pathsToIgnore.contains(link.getSourcePath()))) {
```
jmarchionatto, line 138: Shouldn't check that ALL match? Otherwise any match would cover for all the conflicts. Looks to me that each conflict should be matched.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/batch2/DeleteExpungeSqlBuilderTest.java
```diff
@@ -57,4 +58,29 @@ public void testQueryChunkingWhenFindingChildResources(){
 			queryChunker.verify(() -> QueryChunker.chunk(anyCollection(), any()), times(1));
 		}
 	}
+
+	@Test
+	void testValidateOkToDeleteAndExpunge_whenEnforceReferentialIntegrityDisableForPaths_noException() {
```
jmarchionatto, line 63: Please add test with multiple links and not all matching, to cover for the previously commented use case

jmarchionatto (changes requested):
Added a question

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/delete/batch2/DeleteExpungeSqlBuilder.java
YalingPeiS: I thought about this when working on it and determined to use `anyMatch` based on the following observations:

- We have similar checks for DELETE and it adds the resource to the ignore list when a single path matches one in the configuration paths. See 
  -  https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/OverridePathBasedReferentialIntegrityForDeletesInterceptor.java#L122-L131 and  
  - https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/delete/DeleteConflictUtil.java#L41
- When I test this and create an encounter with subject referencing the patient, two `SRC_PATH` values of this reference are created in `HFJ_RES_LINK` table. One is `Encounter.subject`, the other is `Encounter.subject.where(resolve() is Patient`. When deleting/expunge the patient, the conflict list contains both SRC_PATH since it is loaded from `HFJ_RES_LINK` table. If we check each of the paths against the configured path `Encounter.subject` when delete/expunge the patient, it will fail.

jmarchionatto: I see. OK. Sounds right.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/batch2/DeleteExpungeSqlBuilderTest.java
jmarchionatto: Obsolete by main subject reply

PR #7143, opened 2025-07-18
7002 supporting reindex of deleted resources

### LINKED ISSUE #7002: Supporting reindex of deleted resources

jdar8:
- added the new `_includeDeleted` parameter for search urls to allow reindexing of deleted resources
- allow for some meta related search params to be performed without resource type for batch jobs (this was done so that reindex of deleted resources can be done on all resources)
- tests
- changelog 

Closes #7002


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/Batch2DaoSvcImpl.java
```diff
@@ -72,6 +78,8 @@ public class Batch2DaoSvcImpl implements IBatch2DaoSvc {
 
 	private final PartitionSettings myPartitionSettings;
 
+	private SearchBuilderFactory<JpaPid> mySearchBuilderFactory;
```
volodymyr-korzh:
```suggestion
	private final SearchBuilderFactory<JpaPid> mySearchBuilderFactory;
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/Batch2DaoSvcImpl.java
```diff
@@ -84,20 +92,24 @@ public Batch2DaoSvcImpl(
 			DaoRegistry theDaoRegistry,
 			FhirContext theFhirContext,
 			IHapiTransactionService theTransactionService,
-			PartitionSettings thePartitionSettings) {
+			PartitionSettings thePartitionSettings,
+			SearchBuilderFactory theSearchBuilderFactory) {
```
volodymyr-korzh:
```suggestion
			SearchBuilderFactory<JpaPid> theSearchBuilderFactory) {
```

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/Batch2JobHelper.java
```diff
@@ -51,7 +51,7 @@ public class Batch2JobHelper {
 	private static final Logger ourLog = LoggerFactory.getLogger(Batch2JobHelper.class);
 
 	private static final int BATCH_SIZE = 100;
-	public static final int DEFAULT_WAIT_DEADLINE = 30;
+	public static final int DEFAULT_WAIT_DEADLINE = 3000; //todo jdjd reset to 30
```
volodymyr-korzh: Question: was 3000 set for testing purposes? should it be reverted ?

hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/jpa/dao/r5/dbpartitionmode/DbpmDisabledPartitioningEnabledNullDefaultPartitionTest.java
```diff
@@ -2,6 +2,7 @@
 
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
```
volodymyr-korzh: trivial: unused import, can be deleted

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/ReindexTaskTest.java
```diff
@@ -22,29 +22,43 @@
...
 import java.util.stream.Stream;
 
 import static ca.uhn.fhir.batch2.jobs.reindex.ReindexUtils.JOB_REINDEX;
+import static ca.uhn.fhir.jpa.api.dao.ReindexParameters.OptimizeStorageModeEnum.ALL_VERSIONS;
+import static ca.uhn.fhir.util.TestUtil.sleepAtLeast;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
```
volodymyr-korzh: trivial: unused import

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/ReindexTaskTest.java
```diff
@@ -378,6 +392,409 @@ public void testReindex_ByUrl() {
 		myStorageSettings.setMarkResourcesForReindexingUponSearchParameterChange(reindexPropertyCache);
 	}
 
+	@Nested
+	class ReindexDeletedResourcesTest {
```
volodymyr-korzh, line 395: Great test coverage !

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/Batch2DaoSvcImplTest.java
```diff
@@ -225,4 +258,304 @@ public void fetchResourceIds_NoUrl_NoData(String theMissingUrl) {
...
+		IdType idDeletedBeforeDate = new IdType(patientMatchOutcomeDel.getPersistentId().getResourceType(), (Long) patientMatchOutcomeDel.getPersistentId().getId());
+		deleteResource(new IdType("Patient", "p-match-del"));
+
+		BaseDateTimeDt date = new DateTimeDt().setValue(new Date());
+		sleepAtLeast(1000);
+
+		doUpdateResource(observationAfterDate);
+
+		DaoMethodOutcome patientNoMatchOutcome = doUpdateResourceAndReturnOutcome(noMatchPatient);
+		IdType idAfterDate = new IdType(patientNoMatchOutcome.getPersistentId().getResourceType(), (Long) patientNoMatchOutcome.getPersistentId().getId());
```
volodymyr-korzh: trivial: can be removed, unused

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/Batch2DaoSvcImplTest.java
```diff
@@ -225,4 +258,304 @@ public void fetchResourceIds_NoUrl_NoData(String theMissingUrl) {
...
+
+		DaoMethodOutcome patientNoMatchOutcomeDel = doUpdateResourceAndReturnOutcome(noMatchPatientDeleted);
+		IdType idDeletedAfterDate = new IdType(patientNoMatchOutcomeDel.getPersistentId().getResourceType(), (Long) patientNoMatchOutcomeDel.getPersistentId().getId());
```
volodymyr-korzh: trivial: can be removed, unused

volodymyr-korzh (approved):

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/Batch2JobHelper.java
jdar8: Oops, yes it was for testing and debugging purposes. Reverted.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/ReindexTaskTest.java
jdar8: removed

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/reindex/ReindexTaskTest.java
jdar8: thanks!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/Batch2DaoSvcImplTest.java
jdar8: removed

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/batch2/Batch2DaoSvcImplTest.java
jdar8: removed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/Batch2DaoSvcImpl.java
```diff
@@ -116,17 +128,57 @@ private Stream<TypedResourcePid> streamResourceIdsWithUrl(
 			Date theStart, Date theEnd, String theUrl, RequestPartitionId theRequestPartitionId) {
 		validateUrl(theUrl);
 
-		SearchParameterMap searchParamMap = parseQuery(theUrl);
+		String resourceType = theUrl.substring(0, theUrl.indexOf('?'));
```
jamesagnew: I know you're only moving this line from elsewhere, but it stands out as risky since it'll fail with an unhelpful error message if the input doesn't have a `?` in it.

Recommend replacing this line with a call to `UrlUtil.determineResourceTypeInResourceUrl`

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
```diff
@@ -712,6 +722,14 @@ public BaseJoiningPredicateBuilder getOrCreateFirstPredicateBuilder(
 		return myFirstPredicateBuilder;
 	}
 
+	public BaseJoiningPredicateBuilder getOrCreateFirstPredicateBuilder(
+			SearchIncludeDeletedEnum theIncludeDeletedFlag) {
+		if (myFirstPredicateBuilder == null) {
+			getOrCreateResourceTablePredicateBuilder(theIncludeDeletedFlag);
+		}
+		return myFirstPredicateBuilder;
```
jamesagnew, line 738: In the case where we already have a `myFirstPredicateBuilder` we should probably have an assert to verify that `assert theIncludeDeletedFlag == myFirstPredicateBuilder.myIncludeDeletedFlag` just to make sure we never end up returning a predicate builder with a mismatched flag. I'm assuming this shouldn't actually be possible but an assert never hurts..

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
```diff
@@ -712,6 +722,14 @@ public BaseJoiningPredicateBuilder getOrCreateFirstPredicateBuilder(
 		return myFirstPredicateBuilder;
 	}
 
+	public BaseJoiningPredicateBuilder getOrCreateFirstPredicateBuilder(
```
jamesagnew: General thought: This is really cool, but it's really confusing after these changes how there are now 3 versions of each of `getOrCreateFirstPredicateBuilder` and `getOrCreateResourceTablePredicateBuilder`: One with no args, and then two different-but-similar-seeming params: `SearchIncludeDeletedEnum theIncludeDeletedFlag` and `boolean theIncludeResourceTypeAndNonDeletedFlag`.

Should we collapse these into a default one and a both-parameters one? I feel like the current setup will be very confusing to anyone who comes later trying to understand what's going on here

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/MatchUrlService.java
```diff
@@ -57,6 +61,20 @@ public class MatchUrlService {
 	@Autowired
 	private ISearchParamRegistry mySearchParamRegistry;
 
+	public MatchUrlService() {
+		super();
+	}
+
+	private static boolean isSupportedQueryForNoProvidedResourceType(Set<String> theParamNames) {
```
jamesagnew: Nitpick: Stylistically it's better not to put static methods in random spots in the file - This method should go with the other static methods at the bottom of the file, right before any static nested classes

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/MatchUrlService.java
```diff
@@ -185,6 +225,29 @@ public SearchParameterMap translateMatchUrl(
...
+	/**
+	 * The _includeDeleted parameter should only be supported with _lastUpdated, and _id iff resource type is given
+	 * This is because these are the common search parameter values that are still stored on the deleted resource version in DB
+	 * However, since resources are unique by type and id, only _lastUpdated is supported if no resource type is given
+	 * @param theParamsToCheck the list of parameters found in the URL
+	 * @param theHasNoResourceType whether the request is on the base URL (ie `?_param` - without resource type)
+	 */
+	private static void validateParamsAreCompatibleForDeleteOrThrow(
+			Set<String> theParamsToCheck, boolean theHasNoResourceType) {
+		Set<String> theCompatibleParams = theHasNoResourceType
```
jamesagnew: Nitpick: `theFoo` should be used for method parameters only, so `theCompatibleParams` should just be called `compatibleParams`. Also, the 2 sets below should probably be constants for better performance.

hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/Constants.java
```diff
@@ -220,7 +220,7 @@ public class Constants {
 	public static final String PARAM_CONTENT_URL = "http://hl7.org/fhir/SearchParameter/Resource-content";
 	public static final String PARAM_COUNT = "_count";
 	public static final String PARAM_OFFSET = "_offset";
-	public static final String PARAM_DELETE = "_delete";
+	public static final String PARAM_DELETED = "_includeDeleted";
```
jamesagnew: General thought 1: I find the syntax a bit confusing.. I would expect something which "includes deleted" to add deleted resources to what was already being returned. But `_includeDeleted=true` actually excludes non-deleted resources, and then there is `both` which does what I would have expected but isn't a boolean answer like the name suggests. We probably shouldn't mix the boolean and enum parameter patterns like this, or people will have trouble remembering the syntax.

How about something like:

* _includeDeleted=never (the default)
* _includeDeleted=exclusive (only include deleted resources)
* _includeDeleted=both (include both deleted and non-deleted resources)

General thought 2: The constant should probably be called `PARAM_INCLUDE_DELETED`

hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/SearchIncludeDeletedEnum.java
```diff
@@ -0,0 +1,75 @@
...
+			for (SearchIncludeDeletedEnum next : values()) {
+				codeToEnum.put(next.getCode(), next);
+			}
+			ourCodeToEnum = codeToEnum;
+		}
+
+		SearchIncludeDeletedEnum retVal = codeToEnum.get(theCode);
+		if (retVal == null) {
+			throw new InvalidRequestException(
+					Msg.code(2741) + "Invalid \"_deleted\" mode: " + UrlUtil.sanitizeUrlPart(theCode));
```
jamesagnew: The parameter name is wrong in this message. Recommend just using the constant.

```suggestion
					Msg.code(2741) + "Invalid \"" + Constants.PARAM_INCLUDE_DELETED + "\" mode: " + UrlUtil.sanitizeUrlPart(theCode));
```

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/ITestDataBuilder.java
```diff
@@ -282,6 +283,10 @@ default IIdType createPractitioner(ICreationArgument... theModifiers) {
 		return createResource("Practitioner", theModifiers);
 	}
 
+	default void deleteResource(IIdType theIIdType){
```
jamesagnew, line 286: Great idea adding this here!

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/MatchUrlService.java
```diff
@@ -86,6 +104,18 @@ public SearchParameterMap translateMatchUrl(
...
+		boolean hasNoResourceType = hasNoResourceTypeInUrl(theMatchUrl, theResourceDefinition);
+
+		if (hasNoResourceType && !isSupportedQueryForNoProvidedResourceType(nameToParamLists.keySet())) {
+			// Of all the general FHIR search parameters: https://hl7.org/fhir/R4/search.html#table
+			// We can _only_ process the parameters on resource.meta fields for server requests
+			// The following require a provided resource type because:
+			// - Both _text and _content requires the FullTextSearchSvc and can only be performed on DomainResources
+			// - _id since it is part of the unique constraint in the DB (see ResourceTableDao)
+			// - Both _list and _has allows complex chaining with other resource-specific search params
+			throw new IllegalArgumentException(Msg.code(2742) + "theResourceName must not be blank");
```
jamesagnew: Two things:

- [ ] This message happens when a user-supplied match URL isn't usable and will get bubbled back to the client, so mentioning internal variable names is probably not going to be meaningful to the people who will see the message. Recommend something like "Conditional URL does not include a resource type, but includes parameters which require a resource type"
- [ ] This message should be moved into `hapi-messages.properties` for localization since it's a client-facing message

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7002-support-reindex-of-deleted-resources.yaml
```diff
@@ -0,0 +1,7 @@
+---
+type: add
```
jamesagnew, line 2: Meta thought: This changelog describes this change in terms of adding `_includeDeleted` support for `$reindex` URLs, but I don't see any code anywhere which would block this from being used in other places. E.g. I'm assuming I could use this new parameter anywhere search URLs are used, e.g. conditional updates and even normal searches? At least I don't see anything blocking this.

I'd recommend adding tests for both of these scenarios and making sure that we do something useful, and don't just fail with a cryptic error... E.g. I haven't tried it or anything, but I'm kind of speculating that if you performed a standard search for `Patient?_includeDeleted=both` you'd probably get either a weird empty entry or maybe even a failure if the server found a deleted resource? No idea. If it works properly, we should document it on the [searching data](https://smilecdr.com/docs/fhir_repository/searching_data.html) page. If it doesn't, perhaps we should just find a way to block it and fail fast with an error saying that this parameter isn't supported for standard searches.

jamesagnew (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/Batch2DaoSvcImpl.java
jdar8: Changed it to `determineResourceTypeInResourceUrl`

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/MatchUrlService.java
jdar8: Thanks for the advice! Moved.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
jdar8: Question: if the assert fails, should this be considered an error or should we create a new predicate builder with the correct deleted flags?

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/MatchUrlService.java
jdar8: changed to `compatibleParams` and extracted the sets to constants

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
jdar8: Work done so far: Added the flag to `ResourceTablePredicateBuilder` so that we can assert that they are the same before returning `myFirstPredicateBuilder`

hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/Constants.java
jdar8: I agree, it does sound kind of confusing. 

I like the names you suggested. Changed!

Also changed the constant name.

hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/api/SearchIncludeDeletedEnum.java
jdar8: Changed

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/MatchUrlService.java
jdar8: Changed the error message to the suggested one and moved the message to hapi-messages.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/sql/SearchQueryBuilder.java
jdar8: refactored to a default one, and a both-parameters one

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7002-support-reindex-of-deleted-resources.yaml
jdar8: I just added a test for both scenarios, search and conditional update with `_includeDeleted`, and also confirmed these manually:

The search case successfully returns nothing. It never parses the  `_includeDeleted` parameter since it never passes through `MatchUrlService#translateMatchUrl()`

The conditional update case returns "HAPI-0518: Invalid match URL[Patient?_includeDeleted=exclusive] - URL has no search parameters". This does pass through `MatchUrlService#translateMatchUrl()`, but I guess `_includeDeleted` is not a recognized search parameter. 

That being said, I'm not 100% sure how widely used `translateMatchUrl()` is used, and it may throw cryptic errors where we do resolve URLs by passing through here. I could add a flag parameter in `translateMatchUrl()` so that it will fail outside of usage in Batch. Any thoughts?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -3265,5 +3266,26 @@ public SearchForIdsParams setRequestPartitionId(RequestPartitionId theRequestPar
 			myRequestPartitionId = theRequestPartitionId;
 			return this;
 		}
+
+		public SearchIncludeDeletedEnum getIncludeDeleted() {
```
jamesagnew, line 3272: A `@Nonnull` annotation would probably be a nice touch here given that this can't return null.

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
jdar8: added @Nonnull. thanks for the review :)

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7002-support-reindex-of-deleted-resources.yaml
jdar8: Committed the tests, which assert that the search case returns nothing and conditional update ignores the `_includeDeleted` search parameter. 

Discussed with James a bit, and given the existing behaviour with those two tests, it's probably ok for now without further error handling. 

We may create a new ticket to support `_includeDeleted` elsewhere (such as in general searches) in the future since it would be a nice to have.

PR #7146, opened 2025-07-18
add undo-merge operation

mrdnctrk:
closes: #7148 

This PR adds a new `hapi.fhir.undo-merge` operation for Patient types. This operation restores the 
resources that were updated by a Patient `$merge` operation to their previous versions based on the Provence resource 
created by the `$merge` operation.

There are certain limitations currently:
 - It fails if any resources to be restored have been subsequently changed since the `$merge` operation was performed.
 - It can only run synchronously.
 - It fails if the number of resources to restore exceeds a specified resource limit (currently set to same size as getInternalSynchronousSearchSize in JPAStorageSettings by the operation provider).

## Details:

The undo-merge operation can be called using similar parameters to the merge operation, meaning that it takes in a subset of the parameters of the merge: 
 - a single `source-patient` reference or 1 or many `source-patient-identifier`
 - a single `target-patient` reference or 1 or many `target-patient-identifier`

 
To implement this I made the following changes to the Provenance resource created by the merge operation: 
- The provenance wasn't containing a reference to the source Patient if the source patient was deleted as part of the $merge operation (delete-source was true) previously, now Provenance resource will always contain a reference to the source Patient, even if it was deleted by merge. Since the references added to the Provenance resource is versioned, the source resource is still allowed to be deleted after the Provenance is created, because the referential integrety check on delete ignores versioned references. Having a record of the the source id in the provenance helps find the Provenance resource by source id together with the target id. 
- The provenance resource will contain two contained resources: 
1. The first contained resource is the original input parameters to the $merge operation. Keeping the  original input parameter helps with the undo operation, in finding the provenance resource for the merge operation if the source resource is identified by the  `source-patient-identifier` parameter instead of a reference. Since the source patient might have been deleted by the merge operation, it wouldn't be possible to find the source resource if we didn't keep the identifiers we used to locate it during the merge.  
2. The second contained resource saved in the Provenance resource is the OperationOutcome of the update to the target resource invoked during the merge. This helps because there are certain cases where the target resource may not be updated by a merge operation so during undo we shouldn't change target either. These cases are when delete-source is true (in this case we don't add replaces link to the target), and the source resource doesn't have any identifiers to copy over to the target resource or a result-patient parameter was provided to the merge, that didn't actually change anything in the target resource. The undo operation looks at this outcome to see if the update to the target was a noop and in that case doesn't try to revert the target to a previous version.

Also internally I made  a change to the internal object we use to pass in operation parameters to the batch job for merge (for async merge operation). I've added the complete input parameters object to the job parameter object, which is needed to for storing it in the provenance. Since the complete input parameters also contain result-patient and delete-source parameters, I have deprecated those in the job parameters object, the reason I didn't remove them completely yet was for any in-flight job $merge during an upgrade to be able to complete without breaking because of this change.

### DIFF
--- a/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/merge/MergeUpdateTaskReducerStep.java
+++ b/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/merge/MergeUpdateTaskReducerStep.java
@@ -137,11 +137,11 @@ public class MergeUpdateTaskReducerStep extends ReplaceReferenceUpdateTaskReduce
 		return deleteSource;
 	}
 
-	private @Nullable Patient getResultResource(Parameters originalInputParameters) {
+	private @Nullable Patient getResultResource(Parameters theOriginalInputParameters) {
 		Patient resultResource = null;
 		String resultResourceParamName = myMergeOperationInputParameterNames.getResultResourceParameterName();
-		if (originalInputParameters.hasParameter(resultResourceParamName)) {
-			resultResource = (Patient) originalInputParameters
+		if (theOriginalInputParameters.hasParameter(resultResourceParamName)) {
+			resultResource = (Patient) theOriginalInputParameters
 					.getParameter(resultResourceParamName)
 					.getResource();
 		}


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7148-undo-merge-operation.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: add
+issue: 7148
+title: "An operation named `$hapi.fhir.undo-merge` has been added for the Patient type. This operation restores the
```
nathandoef:
```suggestion
title: "An operation named `$hapi.fhir.undo-merge` has been added for Patient resources. This operation restores the 
```

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7148-undo-merge-operation.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: add
+issue: 7148
+title: "An operation named `$hapi.fhir.undo-merge` has been added for the Patient type. This operation restores the 
+resources that were updated by a Patient `$merge` operation to their previous versions based on the Provence resource
```
nathandoef:
```suggestion
resources that were updated by a Patient `$merge` operation to their previous versions based on the Provenance resource 
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
```diff
@@ -1015,6 +1015,7 @@ public IReplaceReferencesSvc replaceReferencesSvc(
 				theProvenanceSvc);
 	}
 
+	@Primary
```
nathandoef, line 1018: Question: Is this needed? Doesn't look like there was a corresponding CDR change

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/merge/PatientMergeProvider.java
```diff
@@ -114,7 +118,8 @@ public IBaseParameters patientMerge(
 					theDeleteSource,
 					theResultPatient,
 					resourceLimit,
-					provenanceAgents);
+					provenanceAgents,
+					(Parameters) theRequestDetails.getResource());
```
nathandoef: Question: Is there any possibility that this isn't a Parameters resource?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/merge/PatientMergeProvider.java
```diff
@@ -157,32 +203,72 @@ private IBaseParameters buildMergeOperationOutputParameters(
...
 		if (theTargetPatientIdentifier != null) {
 			List<CanonicalIdentifier> targetResourceIdentifiers = theTargetPatientIdentifier.stream()
 					.map(CanonicalIdentifier::fromIdentifier)
 					.collect(Collectors.toList());
-			mergeOperationParameters.setTargetResourceIdentifiers(targetResourceIdentifiers);
+			theMergeOperationParameters.setTargetResourceIdentifiers(targetResourceIdentifiers);
 		}
-		mergeOperationParameters.setSourceResource(theSourcePatient);
-		mergeOperationParameters.setTargetResource(theTargetPatient);
+		theMergeOperationParameters.setSourceResource(theSourcePatient);
```
nathandoef, line 245: Looks like these two lines are duplicated:

```
theMergeOperationParameters.setSourceResource(theSourcePatient);
theMergeOperationParameters.setTargetResource(theTargetPatient);
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/merge/ResourceUndoMergeService.java
```diff
@@ -0,0 +1,239 @@
...
+					"Number of references to update (%d) exceeds the limit (%d)",
+					references.size(), inputParameters.getResourceLimit());
+			throw new InvalidRequestException(Msg.code(2748) + msg);
+		}
+
+		RequestPartitionId partitionId = myRequestPartitionHelperSvc.determineReadPartitionForRequest(
+				theRequestDetails, ReadPartitionIdRequestDetails.forRead(targetPatient.getIdElement()));
+
+		Set<Reference> allowedToUndelete = new HashSet<>();
+		Reference sourceReference = provenance.getTarget().get(1);
```
nathandoef: minor: would suggest moving this inside the if {} block to make things a little more intuitive

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/merge/ResourceUndoMergeService.java
```diff
@@ -0,0 +1,239 @@
...
+		if (wasSourceResourceDeletedByMergeOperation(provenance)) {
+			// If the source resource was deleted by the merge operation,
+			// let the version restorer know it can be undeleted.
+			allowedToUndelete.add(sourceReference);
+		}
+
+		List<Reference> referencesToRestore = references;
```
nathandoef, line 177: Was this intentional? Seems strange to create a new list that's just a reference to another list.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/merge/MergeBatchTest.java
```diff
@@ -54,9 +52,7 @@ public void before() throws Exception {
 		myTestData.createTestResources();
 		// keep the version on Provenance.target fields to verify that Provenance resources were saved
 		// with versioned target references
-		myFhirContext.getParserOptions()
-			.setDontStripVersionsFromReferencesAtPaths("Provenance.target");
-
+		myFhirContext.getParserOptions().setStripVersionsFromReferences(false);
```
nathandoef, line 55: Question: Does this need to be done anywhere in the production code?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/PatientUndoMergeR4Test.java
```diff
@@ -0,0 +1,469 @@
...
+import static ca.uhn.fhir.rest.server.provider.ProviderConstants.OPERATION_MERGE_OUTPUT_PARAM_OUTCOME;
+import static java.lang.String.format;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.catchException;
+
+/**
+ * Tests for the $hapi.fhir.undo-merge operation on Patient resources in R4.
+ */
+public class PatientUndoMergeR4Test extends BaseResourceProviderR4Test {
```
nathandoef, line 51: Excellent test!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
mrdnctrk: This is needed because MergeProvenanceSvc is a subclass of ReplaceReferencesProvenanceSvc, and in this PR I added a bean of type MergeProvenanceSvc. Previously, an instance of MergeProvenanceSvc was created in the constructor of the MergeService internally.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/merge/PatientMergeProvider.java
mrdnctrk: thanks, looks like I pasted it twice, removed the duplicates :)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/merge/PatientMergeProvider.java
mrdnctrk: very good question, thanks for catching this. yes, I tried posting a non-parameter resource as the request body and this cast failed because the resource isn't a Parameters. So I have removed this cast, started using the general Resource and IBaseResource classes. Now, when a non-parameter resource is posted, it returns a 400 because of missing required parameters. I am adding tests for this case to both merge and undo-merge operations.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/merge/ResourceUndoMergeService.java
mrdnctrk: definitely, fixing it.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/provider/merge/ResourceUndoMergeService.java
mrdnctrk: Yes, I did this intentionally. The reason is normally we restore all references, so it is initialized to all references initially, but there might be an exception for the target patient, which I check right below. If the target patient was not modified by merge, I re-set the referencesToRestore to everything except the target patient.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/merge/MergeBatchTest.java
mrdnctrk: good question, I was assuming no, but I will investigate this further.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/PatientUndoMergeR4Test.java
mrdnctrk: thank you

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7148-undo-merge-operation.yaml
mrdnctrk: thanks, fixed

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/merge/MergeUpdateTaskReducerStep.java
```diff
@@ -60,48 +66,85 @@ public MergeUpdateTaskReducerStep(
...
+	private boolean isDeleteSource(Parameters originalInputParameters) {
+		boolean deleteSource = false;
+		String deleteSourceParamName = myMergeOperationInputParameterNames.getDeleteSourceParameterName();
+		if (originalInputParameters.hasParameter(deleteSourceParamName)) {
+			deleteSource = originalInputParameters.getParameterBool(deleteSourceParamName);
+		}
+		return deleteSource;
+	}
+
+	private @Nullable Patient getResultResource(Parameters originalInputParameters) {
```
tadgh: nitpick: `the` prefix

tadgh (approved):

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/merge/MergeUpdateTaskReducerStep.java
mrdnctrk: thanks. fixed

PR #7149, opened 2025-07-22
Update to org.hl7.fhir.core 6.6.1

### LINKED ISSUE #6316: Error when using chained parameters in conditional references
**Describe the bug**
We are attempting to use chained parameters in match URLs within a FHIR resource (e.g., `PractitionerRole?practitioner.identifier=myValue`), but HAPI FHIR is returning an error indicating that the query part of the conditional reference is not valid.

**To Reproduce**

1. Create a FHIR resource, such as a Communication, that includes a reference with a chained parameter for conditional matching.
2. Use a reference in the format `PractitionerRole?practitioner.identifier=myValue`.
3. POST the resource to the HAPI FHIR server.
4. Observe the error returned by the server.

_Minimal Example_

This is a minimal `Communication` resource that demonstrates the issue:

```
{
  "resourceType": "Communication",
  "status": "completed",
  "sender": { "reference": "PractitionerRole?practitioner.identifier=myValue" }
}
```

**Expected behavior**
We expected the server to correctly process the chained parameter in the match URL, resolving the reference as `PractitionerRole` where the associated `practitioner.identifier` matches the given value (`myValue`).

**Error Response**

The server returns the following error:

```
Error
Line: 4, Col: 78
The query part of the conditional reference is not a valid query string (PractitionerRole?practitioner.identifier=myValue)
```

 - HAPI FHIR Version 7.0.2
 - OS: Microsoft Windows 10
 - Browser: Not applicable

**Additional context**
We are trying to use chained parameters to reference resources based on identifiers, which should be valid according to FHIR chaining rules. The issue may be related to how the query part of the reference is parsed and validated.

dotasek:
This increments the org.hl7.fhir.core to version 6.6.1.

Note that several deprecated classes and methods in org.hl7.fhir.core have changed, as has the API for FHIRPath.

There may be more breaking changes to org.hl7.fhir.core on the way. I believe this PR should stay open until all of those are resolved, but be reviewed anew when new core changes come in. When changes are no longer expected, this can be merged, and HAPI can use the new stable version.

Closes #6316


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/fhirpath/IFhirPathEvaluationContext.java
```diff
@@ -39,6 +46,17 @@ default IBase resolveReference(@Nonnull IIdType theReference, @Nullable IBase th
...
 	}
 
+	/**
+	 *
+	 * @param appContext
+	 * @param name The name of the constant(s) to be resolved
+	 * @param mode
+	 * @return
+	 */
+	default List<IBase> resolveConstant(Object appContext, String name, ConstantEvaluationMode mode) {
```
dotasek: @jamesagnew @grahamegrieve this is the biggest change I needed to make. (edited)

This adds a new resolveConstant which uses ConstantEvaluationMode (FHIRPathConstantEvaluationMode in core), but also keeps the previous resolveConstant method that uses beforeContext.

This results in two methods, one which is use by r5 and one which is used by everything prior (actually, HAPI only uses beforeContext in r4b).

Or, if the old method with beforeContext could be removed if we were to translate beforeContext into a mode in R4B, like so:

```java
myEngine.setHostServices(new FHIRPathEngine.IEvaluationContext() {

			@Override
			public List<Base> resolveConstant(
					FHIRPathEngine engine,
					Object appContext,
					String name,
					boolean beforeContext,
					boolean explicitConstant)
					throws PathEngineException {
				return Collections.unmodifiableList(theEvaluationContext
						.resolveConstant(
								appContext,
								name,
								beforeContext
										? IFhirPathEvaluationContext.ConstantEvaluationMode.IMPLICIT_BEFORE
										: IFhirPathEvaluationContext.ConstantEvaluationMode.IMPLICIT_AFTER)
						.stream()
						.map(Base.class::cast)
						.collect(Collectors.toList()));
			}
```

But ideally, if the core library were to uniformly use FHIRPathConstantEvaluationMode across all versions, this would make things simpler.

grahamegrieve: I don't undesrstand that. Base is version dependent, so what's going on here?

dotasek: Addressed in https://github.com/hapifhir/org.hl7.fhir.core/commit/e7b1b77b95a227319f13d21a75548a6c4a7fc59a

dotasek:
This could be a reasonable stopping point, but the there are validation differences that need to be evaluated.

There are new errors being added to code validation. For example, the following test now has an error, where previously there was only a single warning:

https://github.com/hapifhir/hapi-fhir/blob/5a047cc4b2495fb6b735d8bf6197f6909e06a243/hapi-fhir-validation/src/test/java/org/hl7/fhir/dstu3/hapi/validation/FhirInstanceValidatorDstu3Test.java#L604

```
Unknown code 'http://dicom.nema.org/resources/ontology/DCM#BAR',Terminology_PassThrough_TX_Message,severity=error]
```

In all other cases, the message added added an error to a validation that already had errors in place:

https://github.com/hapifhir/hapi-fhir/blob/5a047cc4b2495fb6b735d8bf6197f6909e06a243/hapi-fhir-validation/src/test/java/org/hl7/fhir/r4b/validation/FhirInstanceValidatorR4BTest.java#L1216

```
Error processing unit 'Heck': The unit 'Heck' is unknown' at position 0 (for 'http://unitsofmeasure.org#Heck'),Terminology_PassThrough_TX_Message,severity=error]
```

hapi-fhir-base/src/main/java/ca/uhn/fhir/fhirpath/IFhirPathEvaluationContext.java
```diff
@@ -28,6 +28,13 @@
 
 public interface IFhirPathEvaluationContext {
 
+	public enum ConstantEvaluationMode {
+		EXPLICIT,
+		NOVALUE,
+		IMPLICIT_BEFORE,
+		IMPLICIT_AFTER
```
TipzCM: Can i request at least 1 line explaining what each of these means?

hapi-fhir-validation/src/test/java/org/hl7/fhir/dstu3/hapi/validation/StructureMapTest.java
```diff
@@ -296,14 +296,17 @@ public void testMappingTransform() throws Exception {
 		this.hapiContext = new HapiWorkerContext(this.myCtx, this.validationSupport);// Init the Hapi Work
 		StructureMap map = this.createTestStructuremap();
 		maps.put(map.getUrl(), map);
+		/*FIXME
```
TipzCM: what's this for?

Do we need it? should it be removed? if not, can we add some kind of description to explain why?

Also, there's a comentedo ut import

hapi-fhir-validation/src/test/java/org/hl7/fhir/dstu3/elementmodel/PropertyDstu3Test.java
```diff
@@ -25,7 +25,7 @@
  * Created by axemj on 14/07/2017.
  */
 public class PropertyDstu3Test extends BaseValidationTestWithInlineMocks {
-
+/*
```
TipzCM: Why not just annotate the entire class with @Disabled?

I'm confused why we want to disable all these tests generally (and i think such an action requires at least some documentation to explain why).

I they aren't needed, maybe just delete them?

hapi-fhir-validation/src/test/java/org/hl7/fhir/r4b/validation/FhirInstanceValidatorR4BTest.java
```diff
@@ -1227,19 +1227,19 @@ public void testValidateWithUcum() throws IOException {
 		input.getValueQuantity().setCode("Heck");
 		output = myFhirValidator.validateWithResult(input);
 		all = logResultsAndReturnNonInformationalOnes(output);
-		assertThat(all).hasSize(3);
+		assertThat(all).hasSize(4);
 		// validate first error, in R4B (as opposed to R4) Observation.value.ofType(Quantity) has ValueSet binding,
 		// so first error has `Unknown code for ValueSet` error message
-		assertThat(all.get(0).getMessage()).contains("The Coding provided (http://unitsofmeasure.org#Heck) was not found in the value set 'Vital Signs Units' " +
+		assertThat(all.get(1).getMessage()).contains("The Coding provided (http://unitsofmeasure.org#Heck) was not found in the value set 'Vital Signs Units' " +
```
TipzCM: Is the order of the errors really important?

Can we not make these tests more robust by checking the size and then using a bunch of:

```
assertTrue(all.stream().anyMatch(m -> m.getMessage().contains(....)));
```

?

If the order does matter, disregard this comment

hapi-fhir-validation/src/test/java/org/hl7/fhir/dstu3/hapi/validation/StructureMapTest.java
dotasek: Good question. It was answered outside of this PR, and should have been linked somewhere in this one, but wasn't:

https://github.com/hapifhir/org.hl7.fhir.core/pull/2091#discussion_r2205530886

From @jamesagnew 
> We can definitely safely remove those.. They're just long outdated tests.

hapi-fhir-validation/src/test/java/org/hl7/fhir/dstu3/elementmodel/PropertyDstu3Test.java
dotasek: Good question. It was answered outside of this PR, and should have been linked somewhere in this one, but wasn't:

https://github.com/hapifhir/org.hl7.fhir.core/pull/2091#discussion_r2205525250

From @jamesagnew 
> We can definitely safely remove those.. They're just long outdated tests.

hapi-fhir-validation/src/test/java/org/hl7/fhir/r4b/validation/FhirInstanceValidatorR4BTest.java
dotasek: Order doesn't matter, but the locationString bound to the message does: this would pass for the test, but be slightly incorrect. But it's partway there, and I'll come up with a more robust set of assertions.

dotasek: Fixed in https://github.com/hapifhir/hapi-fhir/pull/7149/commits/4c0a258329053d508eb2eda64dc0ab7c398a93c7

hapi-fhir-validation/src/test/java/org/hl7/fhir/dstu3/elementmodel/PropertyDstu3Test.java
dotasek: Fixed in https://github.com/hapifhir/hapi-fhir/pull/7149/commits/f33756ca31d995314332cc4b20cbd477221279ad

hapi-fhir-validation/src/test/java/org/hl7/fhir/dstu3/hapi/validation/StructureMapTest.java
dotasek: Fixed in https://github.com/hapifhir/hapi-fhir/pull/7149/commits/f33756ca31d995314332cc4b20cbd477221279ad

hapi-fhir-base/src/main/java/ca/uhn/fhir/fhirpath/IFhirPathEvaluationContext.java
dotasek: Fixed in https://github.com/hapifhir/hapi-fhir/pull/7149/commits/0385ba3654d3eaad1d3377a8ac1557e5c6b5566c

hapi-fhir-base/src/main/java/ca/uhn/fhir/fhirpath/IFhirPathEvaluationContext.java
```diff
@@ -43,10 +62,10 @@ default IBase resolveReference(@Nonnull IIdType theReference, @Nullable IBase th
 	 *
 	 * @param appContext
 	 * @param name The name of the constant(s) to be resolved
-	 * @param beforeContext
+	 * @param mode
 	 * @return
 	 */
-	default List<IBase> resolveConstant(Object appContext, String name, boolean beforeContext) {
+	default List<IBase> resolveConstant(Object appContext, String name, ConstantEvaluationMode mode) {
```
tadgh: issue: naming prefix convention

tadgh: hm, i see that we seem to generally ignore the naming convention in these classes. any idea why?

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapter.java
```diff
@@ -1006,6 +1004,12 @@ public <T extends Resource> List<T> fetchResourcesByType(Class<T> theClass) {
 		throw new UnsupportedOperationException(Msg.code(650) + "Unable to fetch resources of type: " + theClass);
 	}
 
+	@Override
+	public <T extends Resource> List<T> fetchResourceVersionsByTypeAndUrl(Class<T> class_, String url) {
+		throw new UnsupportedOperationException(
+				Msg.code(2758) + "Can't fetch all resources of type : " + class_ + " and url: " + url);
```
tadgh: this PR is quite stale, so double check you aren't stepping on anybody's msg code numbers

tadgh (approved):

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapter.java
dotasek: IFhirPathEvaluationContext.java was in error, and is fixed in https://github.com/hapifhir/hapi-fhir/pull/7149/commits/2ebebb8e132bbfac240a2b73bc53a86d4f8839e0

I re-reviewed the other methods introduced, and they all override methods in org.hl7.fhir.core; they all use the same param names as their interfaces as a result. I believe this maintains clarity, but can change to HAPI style if necessary.

hapi-fhir-base/src/main/java/ca/uhn/fhir/fhirpath/IFhirPathEvaluationContext.java
dotasek: IFhirPathEvaluationContext.java was in error, and is fixed in https://github.com/hapifhir/hapi-fhir/commit/2ebebb8e132bbfac240a2b73bc53a86d4f8839e0

I re-reviewed the other methods introduced, and they all override methods in org.hl7.fhir.core; they all use the same param names as their interfaces as a result. I believe this maintains clarity, but can change to HAPI style if necessary.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapter.java
dotasek: Resolved in https://github.com/hapifhir/hapi-fhir/pull/7149/commits/5aa2a3842765e6900645c8b92075cc13de8b5007

hapi-fhir-base/src/main/java/ca/uhn/fhir/fhirpath/IFhirPathEvaluationContext.java
tadgh: leave it for now, we can re-evaluate later if we so decide.

PR #7154, opened 2025-07-23
7152 - Add resource type to HFJ_RESOURCE_MODIFIED's PK to avoid conflicts

another-luis:


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7152-add-resource-type-to-primary-key-to-guarantee-uniqueness.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 7152
+jira: SMILE-10226
+title: "Modified resources could potentially be removed from the subscription list of they shared ID and Version with
```
tadgh, line 5: issue: rewrite this changelog. There's a typo in it, and it should be written from a user failure scenario point of view. Something like: 
```
Previously, subscriptions could occasionally fail to be delivered if a resource had the same ID as a resource of another type, (e.g. `Patient/abc-123` and `Observation/abc-123`), as well as the same version number. This has been corrected, and now the resource type is part of the primary key, ensuring uniqueness during clearing of the table.

File: hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java [hunk 2]
```diff
@@ -130,6 +130,15 @@ public HapiFhirJpaMigrationTasks(Set<String> theFlags) {
...
 		init840();
+		init860();
+	}
+
+	protected void init860() {
+		Builder version = forVersion(VersionEnum.V8_6_0);
+		{
+			version.onTable("HFJ_RESOURCE_MODIFIED").dropPrimaryKey("20250729.1");
+			version.onTable("HFJ_RESOURCE_MODIFIED").addPrimaryKey("20250729.2", "RES_ID", "RES_VER", "RESOURCE_TYPE");
+		}
```
tadgh, line 141: question: is this not making it into 8.4.0? Check with the support person/the ticket to ensure that this is known to them, and it isn't expected for 8.4.0

tadgh (approved):
Approved pending minor stuff. Probably good to get a second set of eyes on this too.

PR #7155, opened 2025-07-23
7153 total should not include non match resources

### LINKED ISSUE #7153: Resource entries in bundles that are OPERATION_OUTCOME or INCLUDE should not be part of the total
Bundle total should not include resources that are marked as OPERATION_OUTCOME or INCLUDE.

See https://hl7.org/fhir/R4/bundle-definitions.html#Bundle.total for explanation

TipzCM:
closes #7153

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/BundleUtil.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/BundleUtil.java
@@ -25,7 +25,9 @@ import ca.uhn.fhir.context.BaseRuntimeElementDefinition;
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.context.RuntimeResourceDefinition;
 import ca.uhn.fhir.i18n.Msg;
+import ca.uhn.fhir.model.api.ResourceMetadataKeyEnum;
 import ca.uhn.fhir.model.primitive.IdDt;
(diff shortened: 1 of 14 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/BundleUtil.java
```diff
@@ -206,6 +208,20 @@ public static List<Pair<String, IBaseResource>> getBundleEntryFullUrlsAndResourc
...
 
+	/**
+	 * Returns true if the resource provided is *not* a BundleEntrySearchModeEnum OUTCOME or INCLUDE
+	 * (ie, if MATCH or null, this will return true)
+	 * @param theResource
+	 * @return
+	 */
+	public static boolean isMatchResource(IBaseResource theResource) {
+		BundleEntrySearchModeEnum matchType = ResourceMetadataKeyEnum.ENTRY_SEARCH_MODE.get(theResource);
+		if (matchType == null || matchType == BundleEntrySearchModeEnum.MATCH) {
```
michaelabuckley, line 219: minor: I think we have code elsewhere that assumes if the resource type is OperationOutcome, and the matchType is null, that it is not a match.

Hmm.  Maybe not.  Can you take a look for other users of this enum and make sure this is consistent?  E.g. ca.cdr.endpoint.fhirgw.server.provider.GatewayBundleProvider#addResource

PR #7157, opened 2025-07-24
7118 rel group list hapi

### LINKED ISSUE #7118: change how resource filtering from certain operations is done
Revert existing MR.

        
      Make extremely targeted fix which simply excludes List and Group resources from $everything, $export.

        
      Indicate this change in the Upgrade.md notes as a breaking change in behaviour.

        
      Somebody with the patient compartment permission. e.g. A user with Patient/READ_ALL_IN_COMPARTMENT  Patient/123  should not be able to read the group that is in.

        
      Inidicate this in the upgrade.md notes as a breaking change in behaviour.

TipzCM:
closes #7118

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBuilder.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBuilder.java
@@ -1003,7 +1003,7 @@ public class RuleBuilder implements IAuthRuleBuilder {
 		}
 	}
 
-	private static String toTypeName(Class<? extends IBaseResource> theType) {
+	private String toTypeName(Class<? extends IBaseResource> theType) {
 		String retVal = ourTypeToName.get(theType);
 		if (retVal == null) {
 			ResourceDef resourceDef = theType.getAnnotation(ResourceDef.class);


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
```diff
@@ -0,0 +1,133 @@
...
+		myAdditionalResourceTypeToParameterCodeMap
+				.computeIfAbsent(theResourceType.toLowerCase(), (key) -> new HashSet<>())
+				.add(theSPName);
+	}
+
+	/**
+	 * add a SP to include into the compartment
+	 * @param theQualifiedSearchParameters - the additional sp. Must be of the format: {resourceType}:{fhirPath}
+	 */
+	@Deprecated
```
jamesagnew: I don't get why would a class that's only being newly added by this PR would have Deprecated methods? Why not just not add the method? (Also, deprecated methods should include a javadoc `@deprecated` tag explaining why the method is deprecated and what should be called instead, and the annotation should have the `since` attribute populated)

jamesagnew: Ah, in reviewing further I see it's moved from elsewhere. Can we just remove the method since we're already making a breaking change to this API?

TipzCM: removed (along with a test that was testing this api specifically)

PR #7158, opened 2025-07-24
fix delete conflict on allowed versioned patient ref on Provenance.ta…

mrdnctrk:
…rget


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
```diff
@@ -1090,6 +1085,31 @@ public void extractSearchParamComboNonUnique(ResourceTable theEntity, ResourceIn
...
+			return false;
+		}
+
+		// there are some search parameters using a where clause to index the element for a specific resource type, such
+		// as
+		// "Provenance.target.where(resolve() is Patient)". We insert these in the ResourceLink table as well.
+		// Such entries in the ResourceLink table should remain versioned if the element is allowed to be versioned.
+		return pathsAllowedToHaveVersionedRefs.stream()
+				.noneMatch(pathToKeepVersioned ->
+						theSearchParamPath.startsWith(pathToKeepVersioned + ".where(resolve() is"));
```
TahaAttari: is there a chance using `startsWith` will cause problems with a chained SP?

mrdnctrk: The path here originates from an SP definition's expression element. So what's happening here is if the configuration was to keep versions at "MyResourceType.element" then any SP that has an expression that starts with  "MyResourceType.element where (resolve() is" would also have a versioned record in the ResourceLink table. 

Now, I am not very knowledgable with  chained SP's. Is there an example you can think where this would create a problem?

TahaAttari: I was wondering if there was an SP with something like:
```
where(resolve() is Patient).resolve().name.family
```
but I don't think we actually support that for SPs. Nvm 👍🏾

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
```diff
@@ -1090,6 +1085,30 @@ public void extractSearchParamComboNonUnique(ResourceTable theEntity, ResourceIn
 		populateResourceTableForComboParams(theParams.myComboTokenNonUnique, theEntity);
 	}
 
+	private boolean shouldStripVersionFromReferenceAtPath(String theSearchParamPath) {
```
nathandoef, line 1088: Nice refactor and thanks for adding these comments

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/PatientMergeR4Test.java
```diff
@@ -75,13 +77,16 @@ public class PatientMergeR4Test extends BaseResourceProviderR4Test {
 
 	ReplaceReferencesLargeTestData myLargeTestData;
 
+	private Set<String> originalDontStripVersionPaths;
```
nathandoef: `myOriginalDontStripVersionPaths`

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/PatientMergeR4Test.java
```diff
@@ -91,13 +96,17 @@ public void before() throws Exception {
...
 		myStorageSettings.setAllowMultipleDelete(true);
 		myFhirContext.setParserErrorHandler(new StrictErrorHandler());
-		// we need to keep the version on Provenance.target fields to
-		// verify that Provenance resources were saved with versioned target references
-		myFhirContext.getParserOptions().setStripVersionsFromReferences(false);
+		// we need to keep the version on Provenance.target fields to verify that Provenance resources were saved
+		// with versioned target references, and delete-source on merge works correctly.
+		// For some reason, if this value is not restored to original at the end, it interferes with other suites,
+		// so keep the original value and restore it at the end
+		originalDontStripVersionPaths = myFhirContext.getParserOptions().getDontStripVersionsFromReferencesAtPaths();
```
nathandoef: Question: Do you know what value of `originalDontStripVersionPaths` should be used in `@AfterEach`? Can it just be hardcoded there instead?

nathandoef (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/PatientMergeR4Test.java
mrdnctrk: sure, I found the default value to be defined [here](https://github.com/hapifhir/hapi-fhir/blob/fccd5b189125e400eb98fe74525e3720edf3adbc/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/r4/FhirContextR4Config.java#L35-L35). I am updating my code to use it instead.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/PatientMergeR4Test.java
mrdnctrk: removed this variable.

Issue #7163, opened 2025-07-29
$hapi.fhir.replace-references throws NPE when a referencing resource contains a logical reference


mrdnctrk:
To reproduce: 

1. Create a source resource, for example a Patient, let's call it  'src-pat' 
2. Create a target resource, another Patient, let's call it 'trg-pat'
3. Create a resource that references the src-pat but also contains some logical reference, for a example an observation 

```json
{
  "resourceType": "Observation",
  "id": "minimal-example-identifier-ref",
  "subject": {
    "reference": "Patient/src-pat"
  },
  "performer": [
    {
      "identifier": {
        "system": "http://example.org/fhir/sid/practitioner-ids",
        "value": "PRC12345"
      }
    }
  ]
}
```

4. now call replace-reference on `src-pat` and `trg-pat`.

```json
{
    "resourceType": "Parameters"
    "parameter": [
        {
            "valueString": "Patient/src-pat",
            "name": "source-reference-id"
        },
        {
            "valueString": "Patient/trg-pat",
            "name": "target-reference-id"
        }
    ]
}
```

Actual outcome is a NPE:
[HAPI-0389](https://simpaticois.atlassian.net/browse/HAPI-0389?atlOrigin=eyJpIjoiY2QyZTJiZDRkNGZhNGZlMWI3NzRkNTBmZmVlNzNiZTkiLCJwIjoianN3LWdpdGxhYi1pbnQifQ): Failed to call access method: java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because the return value of "org.hl7.fhir.instance.model.api.IIdType.getValueAsString()" is null

Expected outcome is success.

mrdnctrk:
Closed. Fixed by #7164.

PR #7164, opened 2025-07-29
fix replace references logical reference

mrdnctrk:
closes: #7163

this PR fixes the NPE on `hapi.fhir.replace-references` operation by adding a check to ensure that a reference it is trying to replace has referenceElement and is not just a logical reference (which is a reference that doesn't have a reference element but an identifier instead)


hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7163-fix-replace-references-NPE.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 7163
+title: "$hapi.fhir.replace-references operation was failing with a NPE when a referencing resource had a logical
```
nathandoef:
```suggestion
title: "Previously, the `$hapi.fhir.replace-references` operation was failing with a NPE when a referencing resource had a logical 
```

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/ReplaceReferencesPatchBundleSvc.java
```diff
@@ -111,11 +111,15 @@ private Bundle buildPatchBundle(
 	}
 
 	private static boolean matches(ResourceReferenceInfo refInfo, IIdType theSourceId) {
-		return refInfo.getResourceReference()
-				.getReferenceElement()
-				.toUnqualified()
-				.getValueAsString()
-				.equals(theSourceId.getValueAsString());
+
+		Reference ref = (Reference) refInfo.getResourceReference();
```
nathandoef: Would recommend using `instanceof` here to guard against invalid casting and another potential null pointer (i.e. `refInfo.getResourceReference()`:

```
		IBaseReference iBaseRef = refInfo.getResourceReference();
		if (iBaseRef instanceof Reference ref && !ref.hasReferenceElement()) {
			// the reference doesn't have a reference element, it is probably just a logical reference (using an
			// identifier).
			// so this is not a match.
			return false;
		}
		return iBaseRef.getReferenceElement().toUnqualified().getValueAsString().equals(theSourceId.getValueAsString());
```

nathandoef (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_4_0/7163-fix-replace-references-NPE.yaml
mrdnctrk: thanks, fixed as suggested.

robogary:
Formatting check succeeded!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/ReplaceReferencesPatchBundleSvc.java
mrdnctrk: thanks, I applied this change, but I also added an explicit check if IBaseRef is null, since `iBaseRef instanceof Reference ref` would return false if IBaseRef was null.

nathandoef: Ahh good call

PR #7171, opened 2025-07-31
Add resource type to the HFJ_RESOURCE_TYPE primary key to avoid collisions

### LINKED ISSUE #7152: Non-unique PK in HFJ_RESOURCE_MODIFIED
Unexpected HTTP 409 may occur if different resource types share the same ID and version. This may happen while processing subscriptions.

closes: #7495

CDR: [7495](https://gitlab.com/simpatico.ai/cdr/-/issues/7495)
JIRA: [SMILE-10226](https://simpaticois.atlassian.net/browse/SMILE-10226)

another-luis:
The original issue occurs when more than one resource type uses the same external ID and version. This can happen during POSTs, which can lead to UPSERTS. Since the primary key used to contain only ID and Version, inserting a resource of a different type could overwrite an existing resource that has the same ID and version. Adding the resource type to the primary key enforces the necessary uniqueness.

Closes #7152

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
@@ -134,7 +134,7 @@ public class HapiFhirJpaMigrationTasks extends BaseMigrationTasks<VersionEnum> {
 	}
 
 	protected void init860() {
-		Builder version = forVersion(VersionEnum.V8_6_0);
+		Builder version = forVersion(VersionEnum.V8_4_0);
 		{
 			version.onTable("HFJ_RESOURCE_MODIFIED").dropPrimaryKey("20250729.1");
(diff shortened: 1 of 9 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/migrate/tasks/HapiFhirJpaMigrationTasks.java
```diff
@@ -130,6 +130,15 @@ public HapiFhirJpaMigrationTasks(Set<String> theFlags) {
 		init780();
 		init820();
 		init840();
+		init860();
+	}
+
+	protected void init860() {
+		Builder version = forVersion(VersionEnum.V8_4_0);
+		{
+			version.onTable("HFJ_RESOURCE_MODIFIED").dropPrimaryKey("20250729.1");
```
michaelabuckley, line 139: This is going to require careful testing on our target databases: pg, sql server, and oracle.

another-luis:
Merging these changes into 8.4.

PR #7174, opened 2025-08-05
Handle Lowercase PK Column Names

jamesagnew:
Per a report here: https://groups.google.com/d/msgid/hapi-fhir/114f7097-dff9-494b-ab44-cbc128bc5557n%40googlegroups.com?utm_medium=email&utm_source=footer

The database partition mode detector fails if the database uses lowercased identifiers, even though we are case-insensitive pretty much everywhere else.


robogary:
Formatting check succeeded!

TipzCM (approved):

PR #7177, opened 2025-08-06
Correctly unescape token system

jamesagnew:
Identifier systems containing characters needing FHIR's backslash escaping in the identifier system (e.g. `"identifier": [{"system":"http://foo,123", "ABC"}]`) can not be searched for using standard escaping (`Patient?identifier=http://foo\,123|ABC`) due to a bug in the search unescaping logic. This 10-year old bug has been corrected.


robogary:
Formatting check succeeded!

volodymyr-korzh (approved):

PR #7188, opened 2025-08-13
7790: move 8_4 db migration steps to appropriate migration method

another-luis:


robogary:
Formatting check succeeded!

PR #7197, opened 2025-08-14
Server with Spring Boot autoconfiguration should not depend on AbstractJaxRsProvider

adamzkover:
Autoconfiguration with Spring Boot depends on `AbstractJaxRsProvider`, which is in dependency `hapi-fhir-jaxrsserver-base`.

For simpler server implementations it should be enough to have `hapi-fhir-server` as dependency.

### DIFF
--- a/hapi-fhir-spring-boot/hapi-fhir-spring-boot-samples/hapi-fhir-spring-boot-sample-client-apache/src/main/java/sample/fhir/client/SampleApacheRestfulClientApplication.java
+++ b/hapi-fhir-spring-boot/hapi-fhir-spring-boot-samples/hapi-fhir-spring-boot-sample-client-apache/src/main/java/sample/fhir/client/SampleApacheRestfulClientApplication.java
@@ -49,7 +49,7 @@ public class SampleApacheRestfulClientApplication {
 
 			@Override
 			public void run(String... args) throws Exception {
-				fhirClient.capabilities().ofType(CapabilityStatement.class) /* .execute() */;
+				fhirClient.capabilities().ofType(CapabilityStatement.class).execute();
 			}
 		};
 	}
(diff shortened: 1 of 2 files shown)


hapi-fhir-spring-boot/hapi-fhir-spring-boot-samples/hapi-fhir-spring-boot-sample-client-apache/src/main/java/sample/fhir/client/SampleApacheRestfulClientApplication.java
```diff
@@ -45,7 +49,7 @@ public CommandLineRunner runner(final IGenericClient fhirClient) {
 
 			@Override
 			public void run(String... args) throws Exception {
-				fhirClient.capabilities().ofType(CapabilityStatement.class).execute();
+				fhirClient.capabilities().ofType(CapabilityStatement.class) /* .execute() */;
```
jkiddo: Why wouldn't you do the `execute` invocation here?

adamzkover: Calling `execute` would actually execute the request, and fail if there isn't a FHIR server running on `localhost:8080`.

adamzkover: Enabled execute again, and added a mock CommandLineRunner in the test class.

PR #7202, opened 2025-08-19
ensure validation error level issues are generated for unknown codesy…

### LINKED ISSUE #7196: validation setNonExistentCodeSystemSeverity is not working as expected when set to error
when setNonExistentCodeSystemSeverity=ERROR and an unknown CodeSystem is used on a FHIR resource, the validation doesn't return an ERROR as expected, it completes with no issues. 

## To Reproduce 
1. setNonExistentCodeSystemSeverity=ERROR in UnknownCodeSystemWarningValidationSupport
2. Try to validate the following Medication resource, which contains an unknown codesystem, using the Medication/$validate operation. 

```json 
{
    "resourceType": "Medication",
	"text": {
	    "status": "generated",
	    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">empty</div>"
	},
        "code": {
        "coding":  [
            {
		"system": "http://test123.org",
		"code": "00559407",
		"display": "TYLENOL EXTRA STRENGTH"
            }
        ]
    }
}
``` 

Actual: The validation completes without any issues. 

Expected: An error should be generated. 

Note: when setNonExistentCodeSystemSeverity=WARNING, a warning for the unknown code system is generated as expected.

mrdnctrk:
#closes #7196

### Description

This PR addresses a bug where validating a resource with an unknown **CodeSystem** fails to produce any validation issues, even when the **UnknownCodeSystem** issue severity is set to **ERROR**.

---

### Root Cause

The main issue was that none of **ValidationSupport** classes were returning a validation result for unknown code systems. The **UnknownCodeSystemWarningValidationSupport** class, whose purpose is to generate these issues, was skipping the creation of an issue if the severity was set to **ERROR**, but would correctly produce an issue if the severity was set to **WARNING**.

It appears this feature may have never worked as intended. The existing validation tests were also not expecting any **ERROR** level issues and were, in fact, incorrectly asserting that no issues would be produced when the severity was set to **ERROR**.

---

### Changes in this PR

- This PR deprecates and replaces the UnknownCodeSystemWarningValidationSupport and moves the logic of creating issues for unknown CodeSystems to the validator chain. In the new code path, a validation issue for an  unknown CodeSystem is produced regardless of the desired severity. 

- The desired code severity is now handled by an interceptor that runs after the validation is complete. The new `ValidationMessageUnknownCodeSystemPostProcessingInterceptor`, which extends `ValidationMessagePostProcessingInterceptor` changes the severity of the unknown code system related error to the configured level.   

* This PR also changes the behaviour for the **INFORMATION** severity setting for unknown CodeSystems. Previously UnknownCodeSystemWarningValidationSupport was not producing a validation issue if desired severity is INFORMATION, we now correctly generate **INFORMATION** level issues as configured. If the goal is to suppress issues, a mechanism like [ValidationMessageSuppressingInterceptor.java](https://github.com/hapifhir/hapi-fhir/blob/e7180f29855f3751f69d076c7d9b8514cc007864/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/validation/ValidationMessageSuppressingInterceptor.java#L36-L36) would be a more appropriate approach.

- A NPE bug in `ValidationMessagePostProcessingInterceptor` is fixed by ensuring a validation message has in id before trying to access it. 

- Updated tests to reflect the new reality. 

#### Why deprecate/replace UnknownCodeSystemWarningValidationSupport

- The `UnknownCodeSystemWarningValidationSupport` class is not a true **ValidationSupport** class because it cannot actually validate a code system or a code. However, to ensure it was called at the end of the validation chain, it "abuses" the **`ValidationSupport`** interface, for example by returning `true` for **`isSupportedCodeSystem`**, which would mislead the core validator.  This has the unintended effect of causing the core library to follow a different code path for the same resource when validated through the CLI versus HAPI-FHIR because core validator also calls  isSupportedCodeSystem itself in several places through `context.supportsSystem`, see for example [src1](https://github.com/hapifhir/org.hl7.fhir.core/blob/644d3d097ed6655e11549ccf5ef2be58564aa72b/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java#L1180-L1180), [src2](https://github.com/hapifhir/org.hl7.fhir.core/blob/644d3d097ed6655e11549ccf5ef2be58564aa72b/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java#L1699-L1699)

- Also even if we make the  UnknownCodeSystemWarningValidationSupport return a validation issue with desired severity, there are cases where the severity set by UnknownCodeSystemWarningValidationSupport would be overriden by the core validator.  The core validator has its own logic ([src](https://github.com/hapifhir/org.hl7.fhir.core/blob/6.4.0/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java#L1722)) for determining the severity of an issue for an unknown CodeSystem, which is based on the binding strength. During validation, the core ignores the error severity set by HAPI-FHIR. It produces errors only when binding strength is REQUIRED. For any other binding strength, it  assigns a warning severity be default. See the question on the core validator: https://github.com/hapifhir/org.hl7.fhir.core/issues/2129.

### DIFF
--- a/hapi-deployable-pom/pom.xml
+++ b/hapi-deployable-pom/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		 <groupId>ca.uhn.hapi.fhir</groupId>
 		 <artifactId>hapi-fhir</artifactId>
-		 <version>8.5.3-SNAPSHOT</version>
+		 <version>8.5.4-SNAPSHOT</version>
 
 		 <relativePath>../pom.xml</relativePath>
 	 </parent>
--- a/hapi-fhir-android/pom.xml
+++ b/hapi-fhir-android/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-base/pom.xml
+++ b/hapi-fhir-base/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-bom/pom.xml
+++ b/hapi-fhir-bom/pom.xml
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>ca.uhn.hapi.fhir</groupId>
 	<artifactId>hapi-fhir-bom</artifactId>
-	<version>8.5.3-SNAPSHOT</version>
+	<version>8.5.4-SNAPSHOT</version>
 
 	<packaging>pom</packaging>
 	<name>HAPI FHIR BOM</name>
@@ -12,7 +12,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-checkstyle/pom.xml
+++ b/hapi-fhir-checkstyle/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-fhir</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-cli/hapi-fhir-cli-api/pom.xml
+++ b/hapi-fhir-cli/hapi-fhir-cli-api/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-cli/hapi-fhir-cli-app/pom.xml
+++ b/hapi-fhir-cli/hapi-fhir-cli-app/pom.xml
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-fhir-cli</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-cli/pom.xml
+++ b/hapi-fhir-cli/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-fhir</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-client-apache-http5/pom.xml
+++ b/hapi-fhir-client-apache-http5/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-client-okhttp/pom.xml
+++ b/hapi-fhir-client-okhttp/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-client/pom.xml
+++ b/hapi-fhir-client/pom.xml
@@ -4,7 +4,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.3-SNAPSHOT</version>
+		<version>8.5.4-SNAPSHOT</version>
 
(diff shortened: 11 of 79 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -811,12 +817,64 @@ && isValueSetSupported(theValidationSupportContext, next, theValueSetUrl))) {
...
+		}
+		IBaseResource codeSystem = fetchCodeSystem(theCodeSystem);
+		if (codeSystem != null) {
+			return null;
+		}
+
+		CodeValidationResult result = new CodeValidationResult();
+		result.setSeverity(IssueSeverity.ERROR);
+		String theMessage = "CodeSystem is unknown and can't be validated: %s for '%s#%s'"
+				.formatted(theCodeSystem, theCodeSystem, theCode);
```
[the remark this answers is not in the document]
mrdnctrk: True, but that's an existing error message, it is ok to have it the way it is.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/validation/ValidateCodeWithRemoteTerminologyR4Test.java
```diff
@@ -140,8 +140,7 @@ public void validateCodeOperationOnCodeSystem_byCodingAndUrlWhereCodeSystemIsUnk
 		ourLog.info(resp);
 
 		assertFalse(((BooleanType) respParam.getParameterValue("result")).booleanValue());
-		assertThat(respParam.getParameterValue("message").toString()).isEqualTo("Terminology service was unable to provide validation for " + INVALID_CODE_SYSTEM_URI +
-			"#P");
+		assertThat(respParam.getParameterValue("message").toString()).isEqualTo("CodeSystem is unknown and can't be validated: %s for '%s%s'", INVALID_CODE_SYSTEM_URI, INVALID_CODE_SYSTEM_URI, "#P");
```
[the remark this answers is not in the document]
mrdnctrk: isEqualTo has an override that takes in format parameters, so this is a false positive.

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/R4ValidationTestUtil.java
```diff
@@ -56,11 +58,21 @@ public static String toString(OperationOutcome theOperationOutcome) {
 		return ourFhirContext.newJsonParser().setPrettyPrint(true).encodeResourceToString(theOperationOutcome);
 	}
 
+	public static List<OperationOutcome.OperationOutcomeIssueComponent> getIssuesDiagnosticContainsString(OperationOutcome theOo, String theExpectedDiagnosticSubstring) {
```
jamesagnew: Nitpick: I think `getIssuesDiagnosticContainingString` would be a more intuitive name

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/validation/ValidationMessageUnknownCodeSystemPostProcessingInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+import ca.uhn.fhir.context.support.IValidationSupport;
+import ca.uhn.fhir.validation.ResultSeverityEnum;
+
+import java.util.List;
+
+/**
+ * This interceptor is for changing the severity of the Unknown Code System validation issues to a desired severity
+ * after the validation has been performed.
+ */
+public class ValidationMessageUnknownCodeSystemPostProcessingInterceptor
```
jamesagnew, line 39: A few comments on this class:

- [x] Please add a javadoc tag `@since 8.6.0`
- [x] Please add marker annotation `@Interceptor` to the class (this doesn't actually do anything, but is a marker we put on all the HAPI FHIR built-in interceptors by convention)
- [x] The [UnknownCodeSystemWarningValidationSupport](https://hapifhir.io/hapi-fhir/docs/validation/validation_support_modules.html#unknowncodesystemwarningvalidationsupport) section of the docs should probably be modified to indicate that this module is deprecated, and link to the new way of doing this.
- [x] We should probably include a section on the [Instance Validator](https://hapifhir.io/hapi-fhir/docs/validation/instance_validator.html) page explaining how to use this new interceptor.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportChain.java
```diff
@@ -811,12 +817,64 @@ && isValueSetSupported(theValidationSupportContext, next, theValueSetUrl))) {
...
+			return null;
+		}
+		IBaseResource codeSystem = fetchCodeSystem(theCodeSystem);
+		if (codeSystem != null) {
+			return null;
+		}
+
+		CodeValidationResult result = new CodeValidationResult();
+		result.setSeverity(IssueSeverity.ERROR);
+		String theMessage = "CodeSystem is unknown and can't be validated: %s for '%s#%s'"
```
jamesagnew: nitpick: rename `theMessage` to `message` since it's not a parameter

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/validation/performance/ValidationCanonicalizationTest.java
```diff
@@ -245,15 +251,21 @@ private void assertValidationErrors(ValidationResult theValidationResult) {
...
 
 			SingleValidationMessage message3 = theValidationResult.getMessages().get(2);
-			assertEquals(ResultSeverityEnum.INFORMATION, message3.getSeverity());
-			assertEquals("Procedure.code.coding[2]", message3.getLocationString());
-			String expectedMessage3 = "This element does not match any known slice defined in the profile http://example.org/fhir/StructureDefinition/TestProcedure|1.0.0 (this may not be a problem, but you should check that it's not intended to match a slice)";
-			assertEquals(expectedMessage3, message3.getMessage());
+			assertEquals(ResultSeverityEnum.ERROR, message3.getSeverity());
+			assertEquals("Procedure.code", message3.getLocationString());
+			Map<String, String> formatValues = Map.of("conceptNumber", String.valueOf(NUM_CONCEPTS));
+			String expectedMessage3 = "None of the codings provided are in the value set 'Value Set Combined' (http://acme.org/ValueSet/valueset-combined|1), and a coding from this value set is required) (codes = http://acme.org/CodeSystem/codesystem-1#codesystem-1-concept-${conceptNumber}, http://acme.org/CodeSystem/codesystem-2#codesystem-2-concept-${conceptNumber}, http://acme.org/invalid#invalid)";
```
dotasek: Nitpick: These variable names could better indicate the nature of the message, rather than just the index. Example: requiredCodingNotProvidedMessage

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/validation/ValidationMessageUnknownCodeSystemPostProcessingInterceptor.java
```diff
@@ -0,0 +1,78 @@
...
+	public ValidationMessageUnknownCodeSystemPostProcessingInterceptor(
+			IValidationSupport.IssueSeverity theDesiredSeverityForUnknownCodeSystem) {
+
+		ResultSeverityEnum desiredResultSeverity =
+				mapIssueSeverityToResultSeverityEnum(theDesiredSeverityForUnknownCodeSystem);
+
+		super.addPostProcessingPatterns(
+				// this is an unknown code system error produced by the validator
+				new ValidationPostProcessingRuleJson(
+						"Terminology_TX_System_Unknown",
```
dotasek: I see this as matching the constant in org.hl7.fhir.core:

https://github.com/hapifhir/org.hl7.fhir.core/blob/43ac3d440258740eef867998bb5247c4cf574b72/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/i18n/I18nConstants.java#L890

We should either link the constant, so this breaks at compile time if it changes, or at least a comment indicating where this value came from.

PR #7205, opened 2025-08-21
Add History to bulk export

jmarchionatto:


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_6_0/7204-add-history-to-bulk-export.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: add
+issue: 7204
+title: "A new request parameter called `_includeHistory` was added to the `$export` operation,
+  allowing to export all historical versions of exported resources."
```
TipzCM, line 5: allowing the export of all historical versions..

hapi-fhir-jpaserver-elastic-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchWithElasticSearchIT.java
```diff
@@ -1952,14 +1942,14 @@ public void offsetNoCount() {
...
-			IIdType idCode2 = myTestDataBuilder.createObservation(asArray(myTestDataBuilder.withObservationCode("http://example.com/", "code-2")));
+			myTestDataBuilder.createObservation(asArray(myTestDataBuilder.withObservationCode("http://example.com/", "code-2")));
 			myTestDataBuilder.createObservation(asArray(myTestDataBuilder.withObservationCode("http://example.com/", "code-3")));
 
 			myCaptureQueriesListener.clear();
 			List<String> resultIds = myTestDaoSearch.searchForIds("Observation?code=code-1,code-2,code-3&_offset=1&_count=1");
 			myCaptureQueriesListener.logSelectQueriesForCurrentThread();
 
-			assertThat(resultIds).containsExactlyInAnyOrder(idCode2.getIdPart());
+			assertThat(resultIds).hasSize(1);
```
TipzCM, line 1952: this looks like a weaker test.

Why do we no longger expect to find the id we were searching for?

hapi-fhir-jpaserver-elastic-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchWithElasticSearchIT.java
```diff
@@ -1936,14 +1926,14 @@ public void enableResourceStorage() {
...
-			IIdType idCode3 = myTestDataBuilder.createObservation(asArray(myTestDataBuilder.withObservationCode("http://example.com/", "code-3")));
+			myTestDataBuilder.createObservation(asArray(myTestDataBuilder.withObservationCode("http://example.com/", "code-2")));
+			myTestDataBuilder.createObservation(asArray(myTestDataBuilder.withObservationCode("http://example.com/", "code-3")));
 
 			myCaptureQueriesListener.clear();
 			List<String> resultIds = myTestDaoSearch.searchForIds("Observation?code=code-1,code-2,code-3&_offset=1");
 			myCaptureQueriesListener.logSelectQueriesForCurrentThread();
 
-			assertThat(resultIds).containsExactlyInAnyOrder(idCode2.getIdPart(), idCode3.getIdPart());
+			assertThat(resultIds).hasSize(2);
```
TipzCM, line 1936: same as below; why are we verifying the weaker count and not the actual results.

Especially in this case because the actual search is for specific codes. I'd expect those specific codes to be returned.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -1421,7 +1432,419 @@ public void testGroupExportPatientOnly() {
...
+			HashSet<String> resourceTypes = Sets.newHashSet("Observation", "Coverage");
+			BulkExportJobResults bulkExportJobResults = startGroupBulkExportJobAndAwaitCompletion(resourceTypes, new HashSet<>(), "G2", true);
+
+			Map<String, Map<String, Set<String>>> typeToResourceVersionsMap = convertJobResultsToResourceVersionMap(bulkExportJobResults);
+			assertThat(typeToResourceVersionsMap.get("Observation")).isEqualTo(observationVersionsMap);
+			assertThat(typeToResourceVersionsMap.get("Coverage")).isEqualTo(coverageVersionsMap);
+		}
+
+		@Test
+		public void testSystemBulkExport() {
```
TipzCM, line 1520: this test name seems far too generic to me.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -1421,7 +1432,419 @@ public void testGroupExportPatientOnly() {
...
+		void test_includeHistoryParameter_defaultsToFalse() {
+			// Given
+			BulkExportJobParametersBuilder builder = new BulkExportJobParametersBuilder();
+
+			// When
+			BulkExportJobParameters parameters = builder.build();
+
+			// Then
+			assertFalse(parameters.isIncludeHistory());
+		}
```
TipzCM, line 1637: These tests only about the parameter builder should be in a unit test area named just after the bulkexportparameters builder.

This class is already way way too big.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -1421,7 +1432,419 @@ public void testGroupExportPatientOnly() {
...
+	private Parameters getCoveragePatch(int theVersion) {
+		Parameters patch = new Parameters();
+		Parameters.ParametersParameterComponent op = patch.addParameter().setName("operation");
+		op.addPart().setName("type").setValue(new CodeType("replace"));
+		op.addPart().setName("path").setValue(new CodeType("Coverage.dependent"));
+		op.addPart().setName("value").setValue(new StringType("version-" + theVersion + 1));
+		return patch;
+	}
+
+	private void waitForCompletion(Batch2JobStartResponse job) {
```
TipzCM, line 1774: Why not use `myBatch2JobHelper.awaitJobCompletion(job.getInstanceId(), 60);`? You do so elsewhere in this test.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -1421,7 +1432,419 @@ public void testGroupExportPatientOnly() {
...
+			.until(() -> {
+				StatusEnum status = myJobCoordinator.getInstance(job.getInstanceId()).getStatus();
+				if (!StatusEnum.COMPLETED.equals(status)) {
+					fail("Job status was changed from COMPLETE to " + status);
+				}
+				return myJobCoordinator.getInstance(job.getInstanceId()).getReport() != null;
+			});
+	}
+
+		private Map<String, Map<String, Set<String>>> extractExportedResourceVersionsByTypeMap(Batch2JobStartResponse theJob) {
```
TipzCM, line 1786: the formatting 😱

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -1421,7 +1432,419 @@ public void testGroupExportPatientOnly() {
...
+		op.addPart().setName("path").setValue(new CodeType("Patient.name.given"));
+		op.addPart().setName("value").setValue(new StringType("given-v" + theVersion + 1));
+		return patch;
+	}
+
+	/**
+	 * Create the indicated number of patients with the indicated number of total versions each
+	 */
+	private Map<String, Set<String>> createPatientsWithHistory(
+		@SuppressWarnings("SameParameterValue") int thePatientCount, @SuppressWarnings("SameParameterValue") int theVersionCount) {
```
TipzCM, line 1827: I wonder if you can just put one suppression at the top of the method? i'm just nitpicking now though

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -1421,7 +1432,419 @@ public void testGroupExportPatientOnly() {
...
+				lastVersonId = patchOutcome.getId();
+				patientVersionIds.add(lastVersonId.getValueAsString());
+			}
+		}
+		return retVal;
+	}
+
+	private IIdType createPatient() {
+		Patient p = new Patient();
+		p.addName()
```
TipzCM, line 1860: they don't need to be active?

I guess it's fine if they are all the exact same patient, but it is a big confusing (usually i'd append a passed-in value like a number or something to give some kind of distinction if things go wrong).

But again, it' sa nitpick

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/bulk/BulkExportJobParameters.java
```diff
@@ -124,6 +124,9 @@ public class BulkExportJobParameters extends BaseBatchJobParameters {
 	@JsonProperty("binarySecurityContextIdentifierValue")
 	private String myBinarySecurityContextIdentifierValue;
 
+	@JsonProperty("includeHistory")
+	private boolean myIncludeHistory;
```
TipzCM, line 128: Can you javadoc this please?

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStep.java
```diff
@@ -222,8 +329,72 @@ private void fetchResourcesByIdAndConsumeThem(
...
+	@SuppressWarnings("rawtypes")
+	private @NotNull List<String> convertToStringIds(
+			RequestPartitionId theRequestPartitionId, String theResourceType, List<TypedPidJson> theQueryBatch) {
+
+		Set<IResourcePersistentId> queryBatchPids = theQueryBatch.stream()
+				.map(t -> myIdHelperService.newPidFromStringIdAndResourceName(
+						t.getPartitionId(), t.getPid(), theResourceType))
+				.collect(Collectors.toSet());
+
+		@SuppressWarnings("unchecked")
```
TipzCM, line 367: why not add it to the method one?

Then you won't need it twice here (and below)

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStepIncludeHistoryTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+	private ExpandResourceAndWriteBinaryStep myTestedStep;
+
+	@Captor
+	private ArgumentCaptor<List<IBaseResource>> resourcesConsumedCaptor;
+
+	@BeforeEach
+	void setUp() {
+		myTestedStep = spy(myJobStep);
+		doReturn(MAXIMUM_BATCH_SIZE_DEFAULT).when(myTestedStep).getMaxSizeBatchDefault();
```
TipzCM: I'm not fond of this because now you have 2 ExpandResourceAndWriteBinaryStep objects in this class; it's not clear which one should be used for testing (even though you label one 'for testing').

I think for what you want to do (override a single method), there's 2 better options:

1
Do not use @InjectMocks and instead create a spy object and inject the mocks yourself (`MockitoExtension.openMocks(myObj)`)

2
Create a test class that extends the actual class under test and have that test class localized only to this file and use that as your test object.

```java
class TestStep extends ExpandResourceAndWriteBinaryStep {

@Override
int getMaxSizeBatchdefault() {
  return MAXIUMUM_BATCH_SIZE_DEFAULT;
}

}
```

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStepIncludeHistoryTest.java
```diff
@@ -0,0 +1,202 @@
...
+			default -> throw new InvalidParameterException("Unexpected resource type: " + theResourceType);
+		};
+	}
+
+	private List<TypedPidJson> getResourcesTypedPidJsonList(String theResourceType, int theCount) {
+		return IntStream.range(1, theCount+1).mapToObj(i -> new TypedPidJson(theResourceType, 1, String.valueOf(i))).toList();
+	}
+
+	// setup mock as a Spy for required methods as module doesn't have access to JPA module implementations
+	private void setupIdHelperServiceMockAsASpy() {
```
TipzCM: This name is confusing

You're not actually setting up any spies, you're just mocking out myIdHelperService.

As an aside, even though it's "more verbose", i almost always prefer that the tests call this method directly, because future tests might not want to. And then we're stuck slapping "lenient" onto these methods.

TipzCM (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_6_0/7204-add-history-to-bulk-export.yaml
jmarchionatto: Fixed

hapi-fhir-jpaserver-elastic-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchWithElasticSearchIT.java
jmarchionatto: Test had an intermittency caused by checking IDs without a sort parameter, and given the test purpose I considered that checking the size met the purpose.

hapi-fhir-jpaserver-elastic-test-utilities/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchWithElasticSearchIT.java
jmarchionatto: Same reply as before

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
jmarchionatto: Test is to check the includeHistory parameter for system export. The wrapping nesteed class indicates the parameter name and the thest class the fact that is for system export. What else would you add?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
jmarchionatto: Good point! Moved to BulkExportJobParametersBuilderTest

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
jmarchionatto: Done

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
jmarchionatto: fixed

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
jmarchionatto: It can. Done

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
jmarchionatto: In tests which matters, versions are distinguished by the name

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/bulk/BulkExportJobParameters.java
jmarchionatto: Done

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStep.java
jmarchionatto: Because there are other generics in the method which are all right, so I prefer to be specific for those which don't.

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStepIncludeHistoryTest.java
jmarchionatto: I did a third, which was easier and I hope solved your concern: moved the spy definition into the test which required it.

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStepIncludeHistoryTest.java
jmarchionatto: I think you missed the method top comment. The "as spy" name is because I am moving the original functionality in a mock (virtually converting it into a spy) because the test doesn't have access to the real class. 
If you see a more elegant way to solve this, I'm all ears.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStep.java
```diff
@@ -169,50 +175,151 @@ private void expandResourcesFromList(
...
+			}
+
+			resourcesToConsume.addAll(page);
+
+			// process complete batches
+			while (resourcesToConsume.size() >= fileLimitBatchSize) {
+				List<IBaseResource> batch = new ArrayList<>(resourcesToConsume.subList(0, fileLimitBatchSize));
+				theResourceListConsumer.accept(batch);
+				ourLog.debug("Sent batch of {} history resources to consumer", batch.size());
+				resourcesToConsume.subList(0, fileLimitBatchSize).clear();
```
jamesagnew: Wow, I had no idea you could do this. I actually started writing up a comment saying I didn't think this would work, then checked if you had a test which would exercize it and you do, so I looked at the List JavaDoc and sure enough, it explicitly describes this contruct! Neat! :)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/HistoryBuilder.java
```diff
@@ -94,10 +98,22 @@ public HistoryBuilder(
...
 		myResourceId = theResourceId;
+		myResourceIds = null;
 		myRangeStartInclusive = theRangeStartInclusive;
 		myRangeEndInclusive = theRangeEndInclusive;
 	}
 
+	/**
+	 * Constructor for multiple resource IDs
+	 */
+	public HistoryBuilder(@Nullable String theResourceType, @Nullable List<String> theResourceIds) {
```
jamesagnew: Overall thought: I worry that this new history will not be stable if resources are added. E.g. suppose I request an export and it loads the first N versions of a resource, then a new one gets added before the next page is loaded? I'd assume in that case you're going to end up with duplicates in the export. Probably not the biggest deal, but it should be easy enough to defend against.

IMO it'd make sense to require an `myRangeEndInclusive` date passed into this constructor, with the value being the job creation time by default.

An additional enhancement that would be useful if you wanted to (although this could easily be done later in another ticket) would be to allow the person submitting the job to supply a start and end date. I think we'll end up needing those options in the real world for anyone using this functionality as a part of a data migration between databases, since you'll need a way of migrating "most of the data" in the background with a live system running, but then also a way of migrating "just the most recent data" in the foreground at the end, in order to minimize downtime.

jamesagnew: Update: I actually see that `BulkExportJobParameters` already has date range parameters. We should respect them for historical versions (and add a test to ensure that we do), and assume a rangeEnd of the job creation time if no range end is supplied.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStep.java
```diff
@@ -169,50 +175,151 @@ private void expandResourcesFromList(
...
+		final int fileLimitBatchSize = myStorageSettings.getBulkExportFileMaximumCapacity();
+		final int pageSize = Math.min(getMaxSizeBatchDefault(), fileLimitBatchSize);
+
+		IBundleProvider resHistoryProvider =
+				searchForResourcesHistory(theResourceType, theIdList, theRequestPartitionId);
+
+		int currentIndex = 0;
+		List<IBaseResource> resourcesToConsume = new ArrayList<>();
+
+		while (true) {
```
jamesagnew, line 267: I worry about blowing out the memory on this loop. Most resources won't have too many versions or have huge versions, but it's guaranteed that some customers will have huge versions, and I worry that they'll experience OOMs.

A few thoughts thoughts about this method:

- [ ] You're respecting `myStorageSettings.getBulkExportFileMaximumCapacity()` (which limits the number of resources per file) but you aren't respecting `myStorageSettings.getBulkExportFileMaximumSize()` (which limits the maximum file size). You should add checks in the loop that the file isn't about to exceed that threshold for any given file, and add tests verifying this.  See examples in `ExpandResourcesAndWriteBinaryStepJpaTest`, which test this setting for normal export.
-  [ ] The page size should be small, honestly I'd just set it to something like 10 so that you never hold more than 10 resources in memory in cases where the serialized versions are big enough to hit the maximum file size
- [ ] I also worry that the hibernate L2 cache may be holding all of your previously looked up entities in memory for the lifetime of the FHIR transaction that wraps this entire method. I'm not sure that there is one though.... Maybe just add a `HapiTransactionService.noTransactionAllowed()` at the top of this method just to ensure that it's never called transactionally (meaning the individual calls to `resHistoryProvider` would each open and close a TX, and the L2 cache wouldn't be a factor. Thinking about this more, I suspect that's the case.. Adding the noTransactionAllowed would guarantee it.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/bulk/BulkExportJobParameters.java
```diff
@@ -284,6 +291,13 @@ public String getBinarySecurityContextIdentifierValue() {
 		return myBinarySecurityContextIdentifierValue;
 	}
 
+	/**
+	 * Indicates if all history must be exported for exported resources
+	 */
+	public void setIncludeHistory(boolean theIncludeHistory) {
```
jamesagnew, line 297: Maybe I'm missing it, but I don't see anywhere that this can actually be set by a client? I see unit tests setting it manually, but how would an end user use this feature?

jamesagnew (changes requested):

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStep.java
jmarchionatto: Thanks

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/HistoryBuilder.java
jmarchionatto: OK. Will add that and tests

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStep.java
jmarchionatto: 1_ OK. Will add the file size limit. 
2_ I thought those page size limits were enough. I'll get the page size down to 10
3_ Not an issue anymore after 1_ and 2_, but I think the app doesn't use hibernate L2 cache. Does it?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/bulk/BulkExportJobParameters.java
jmarchionatto: It is a new parameter in the provider. User adds the url parameter to obtan the history.
See BulkDataExportProvider lines 137, 193, 265 and 317

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStep.java
jamesagnew: Oops, my apologies - You're correct that we don't use the L2 cache. It's the L1 cache I was worrying about in my head, I'm not sure why I wrote L2. If there is an outer transaction surrounding this method, then the L1 cache could be collecting an ever-growing collection of `ResourceTable` and/or `ResourceHistoryTable` entities even if were using smaller page sizes..

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/HistoryBuilder.java
jmarchionatto: Fixed

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/ExpandResourceAndWriteBinaryStep.java
jmarchionatto: File size limit was already considered when consuming entities. Added test verifying it works for history entities.
Limited page size to 10 entries.
Also changed the history retrieval sort when exporting history to get output in sensible order when filtering by resource IDs.

PR #7218, opened 2025-08-26
Non-partitionable resource searches could be executed against a partition that is not the default partition

epeartree:
fixing issue where non-partitionable resource searches could be executed against a partition that is not the default partition.


robogary:

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/partition/BaseRequestPartitionHelperSvc.java
```diff
@@ -492,10 +489,19 @@ private static void logTroubleshootingResult(
 				theResult);
 	}
 
-	private void logSystemRequestDetailsResolution(SystemRequestDetails theRequest) {
-		ourLog.trace(
-				"Partitioning: request is a SystemRequestDetails, with RequestPartitionId={}.",
-				theRequest.getRequestPartitionId());
+	private void logRequestDetailsResolution(RequestDetails theRequest) {
```
jamesagnew, line 517: This robust logging method warms my heart :)

PR #7225, opened 2025-08-28
added eid based expansion in match-only mode

mrdnctrk:
closes: #7247 

The PR adds MDM expansion support when using MATCH_ONLY mode and MDM rules definition contains eidSystems
The expansion is based on having the same EID as the resource being operated on (rather than full MDM golden resource relationships). It is supported with the existing mdm expansion parameters on search, group bulk export, and $everything operation.

###  Key Changes
- For search expansion, created a new implementation of IMdmLinkExpandSvc, so this interface now has two implementations: 
      1. `MdmLinkExpandSvc` existing implementation used in the full MDM mode
      2. `MdmEidMatchOnlyExpandSvc` the new implementation introduced in this PR, which expands a given resource by finding other resources that have the same eid. 

- For group bulk export expansion, there is  a need for similar functionality but this time to expand every member on the Group. I created an interface called `IBulkExportMdmResourceExpander`, this also have two implementations: 
        1. `BulkExportMdmResourceExpander` contains the existing implementation for the full MDM mode, that makes use of the MDM links in the db for expansion. I extracted this class out of the `JpaBulkExportProcessor`, by moving logic specifically used for group expansion to this new class. 
       2. `BulkExportMdmEidMatchOnlyResourceExpander` this is the new implementation used for expansion in MATCH_ONLY mode. 

- To overcome some dependency injection challenges, I  created class called `MdmExpandersHolder` which holds the references to both pairs of implementation and returns the right implementation based on the MDM mode and  whether eid systems defined in mdm rules. 

  
 ### Notes about the refactoring JpaBulkExportProcessor

- In addition moving group expansion logic from JpaBulkExportProcessor to BulkExportMdmResourceExpander, I also refactored some related code in JpaBulkExportProcessor, which had two functionally very similar methods called `getExpandedPatientList` and `expandAllPatientPidsFromGroup`. The only meaningful difference between these methods was that they called `getMembersFromGroupWithFilter` function with a different boolean parameter. I unified calls to these methods by exposing that boolean parameter in `getExpandedPatientList` and removing the `expandAllPatientPidsFromGroup` one.

- Another thing to note is that  JpaBulkExportProcessor had a cache for storing the mappings between the source resource ids and golden resource ids that are processed to annotate the exported resources with the golden resource identifiers. I also moved the logic related to maintaining and using this cache to  BulkExportMdmResourceExpander. In the MATCH_ONLY mode this is not applicable because there is no concept of golden resource.

### DIFF
--- a/hapi-deployable-pom/pom.xml
+++ b/hapi-deployable-pom/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		 <groupId>ca.uhn.hapi.fhir</groupId>
 		 <artifactId>hapi-fhir</artifactId>
-		 <version>8.5.5-SNAPSHOT</version>
+		 <version>8.5.6-SNAPSHOT</version>
 
 		 <relativePath>../pom.xml</relativePath>
 	 </parent>
--- a/hapi-fhir-android/pom.xml
+++ b/hapi-fhir-android/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.5-SNAPSHOT</version>
+		<version>8.5.6-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-base/pom.xml
+++ b/hapi-fhir-base/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.5-SNAPSHOT</version>
+		<version>8.5.6-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-bom/pom.xml
+++ b/hapi-fhir-bom/pom.xml
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>ca.uhn.hapi.fhir</groupId>
 	<artifactId>hapi-fhir-bom</artifactId>
-	<version>8.5.5-SNAPSHOT</version>
+	<version>8.5.6-SNAPSHOT</version>
 
 	<packaging>pom</packaging>
 	<name>HAPI FHIR BOM</name>
@@ -12,7 +12,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.5.5-SNAPSHOT</version>
+		<version>8.5.6-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-checkstyle/pom.xml
+++ b/hapi-fhir-checkstyle/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-fhir</artifactId>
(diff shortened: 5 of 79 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/svc/BulkExportMdmEidMatchOnlyResourceExpander.java
```diff
@@ -0,0 +1,124 @@
...
+			FhirContext theFhirContext,
+			IIdHelperService<JpaPid> theIdHelperService) {
+		myDaoRegistry = theDaoRegistry;
+		myMdmEidMatchOnlyLinkExpandSvc = theMdmEidMatchOnlyLinkExpandSvc;
+		myFhirContext = theFhirContext;
+		myIdHelperService = theIdHelperService;
+	}
+
+	/**
+	 * Expands a Group resource and returns theirs persistent ids.
```
jamesagnew:
```suggestion
	 * Expands a Group resource and returns the Group members' resource persistent ids.
```

I'm assuming this is right? Great docs on this method either way!

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/svc/BulkExportMdmEidMatchOnlyResourceExpander.java
```diff
@@ -0,0 +1,124 @@
...
+		// Read the Group resource
+		SystemRequestDetails srd = SystemRequestDetails.forRequestPartitionId(requestPartitionId);
+		IIdType groupId = myFhirContext.getVersion().newIdType(groupResourceId);
+		IFhirResourceDao<?> groupDao = myDaoRegistry.getResourceDao("Group");
+		IBaseResource groupResource = groupDao.read(groupId, srd);
+
+		Set<String> allResourceIds = new HashSet<>();
+		FhirTerser terser = myFhirContext.newTerser();
+		// Extract all member.entity references from the Group resource
+		List<IBase> memberEntities = terser.getValues(groupResource, "Group.member.entity");
```
jamesagnew: Nitpick: You could just call `terser.getValues(groupResource, "Group.member.entity", IBaseReference.class)` to have the terser handle the checking and casting for you.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_6_0/7247-eid-based-mdm-expansion-in-match-only-mode.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: add
+issue: 7247
+title: "The MDM expansion is now supported when using MATCH_ONLY mode and MDM rules definition contains eidSystems.
```
tadgh:
```suggestion
title: "MDM expansion is now supported when using MATCH_ONLY mode and MDM rules definition contains eidSystems. 
```

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/bulk/BulkExportUseCaseTest.java
```diff
@@ -1422,8 +1439,91 @@ public void testGroupExportPatientOnly() {
...
 
+	private static void assertResourcesIds(List<IBaseResource> theResources, String... theExpectedResourceIds) {
+		assertThat(theResources).hasSize(theExpectedResourceIds.length);
+		assertThat(theResources)
+			.extracting(IBaseResource::getIdElement)
+			.extracting(IIdType::toUnqualifiedVersionless)
+			.extracting(IIdType::getValue)
+			.containsExactlyInAnyOrder(theExpectedResourceIds);
+	}
+
```
tadgh, line 1525: nice tests

tadgh (approved):

PR #7242, opened 2025-09-07
Check for null before removing 'max' parameter

jkiddo:
Based on some elasticsearch tweaking on the starter project, I encountered this missing null check.


robogary:
Formatting check succeeded!

tadgh (approved):

PR #7253, opened 2025-09-11
Add dry-run and delete options to bulk patch / bulk modification

jamesagnew:
This PR adds:

* A dry-run option to the bulk patch operation
* Delete capability to the bulk modification framework (not used by bulk patch but could be in the future)
* Improved responses to status polling


robogary:
Formatting check succeeded!

hapi-fhir-test-utilities/src/main/java/ca/uhn/test/util/ParsedHttpResponse.java
```diff
@@ -48,4 +53,27 @@ public String toString() {
...
+			.findFirst();
+	}
+
+	/**
+	 * Returns the {@literal Content-Type} header value with any additional parameters
+	 * (e.g. {@literal charset} parameter) stripped off.
+	 */
+	public String contentType() {
+		String retVal = headers().get(Constants.HEADER_CONTENT_TYPE).stream().findFirst().orElse(null);
+		if (retVal.contains(";")) {
```
TipzCM, line 80: should this be null checked?

You have `orElse(null)` above, which suggests that it could potentially be null.

jamesagnew: Oh yeah good call, added a check

PR #7249, opened 2025-09-10
fix(7248): IdHelperService.resolveResourceIdentities fails with PreconditionFailedException on duplicate id inputs

ttnTrifork:
closes https://github.com/hapifhir/hapi-fhir/issues/7248

PR #7254, opened 2025-09-11
fix(7248): IdHelperService.resolveResourceIdentities fails with PreconditionFailedException on duplicate id inputs

jamesagnew:
Supercedes #7249 which I couldn't modify due to branch permissions.


robogary:
Formatting check succeeded!

PR #7257, opened 2025-09-15
Speed up transaction processing with logical references.

michaelabuckley:
- Add a reverse cache from pid->FHIR id to TransactionDetails
- Resolve FHIR id at the same time we resolve match url to pid.
- Change match-url prefetch chunking so we chunk by url, not randomly by reference.

During bundle processing, we collect all references, and resolve them.  For simple local references, this is just the fhir id->pid lookup and that is cached.  But for logical references, we do a bulk query during pre-fetch to resolve match-urls->pids.  This is already cached.  But during processing, we replace logical references with local references.  To do that we use the match-url->pid cache, but then need to turn a pid into a FHIR id.  This was not cached.  We speed this up by adding a reverse-cache to TransactionDetails and using it during reference resolution.

Also, we were chunking match url resolution during prefetch.  For large bundles, we might have 100s of references, but far fewer distinct urls.  E.g. a synthea bundle might have 100s of references with only a handful of distinct Organization and Location references.  We change the chunking during pre-fetch to chunk by groups of urls, not groups of references.


robogary:
Formatting check succeeded!

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
```diff
@@ -258,8 +261,11 @@ public void addResolvedResourceId(IIdType theResourceId, @Nullable IResourcePers
 
 		if (myResolvedResourceIds.isEmpty()) {
 			myResolvedResourceIds = new HashMap<>();
+			myReverseResolvedResourceIds = new HashMap<>();
```
tadgh, line 264: thought (non-blocking): did you consider using a bidirectional map?  https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/BidiMap.html No significant difference really, just wondering why the two map approach was chosen.

tadgh (approved):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/api/server/storage/TransactionDetails.java
michaelabuckley: I'm not sure I even knew that existed.

I'm going to leave this as-is for now.  I'd need the FhirContext to turn the String back into an IIdType, and I don't want to do the plumbing for this.

PR #7268, opened 2025-09-19
Update packages

madduci:
Hi

I've performed an upgrade of many libraries to resolve important CVEs. 

I had to change some code to address deprecated / removed / unused code.

Current Status: some tests are still failing, mostly due to dirty context, i am trying to correct them

I've also applied `mvn spotless:apply`, but the GitHub Action is still failing

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/gclient/IEntityResult.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/rest/gclient/IEntityResult.java
@@ -19,7 +19,7 @@
  */
 package ca.uhn.fhir.rest.gclient;
 
-import org.checkerframework.checker.nullness.qual.NonNull;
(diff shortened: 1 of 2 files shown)


robogary:

jamesagnew:
Hi @madduci - Thanks for the contribution! This looks to be failing quite a few CI tests though, it's not currently in a state that could be merged.

madduci:
Hi @jamesagnew 

yes after the upgrade of packages, some tests are failing, currently i don't know if the expected wrong actual values in tests are there because of the internal processing of the tinder-plugin, which seems to behave differently after the upgrade of some plugins/libraries

jamesagnew:
I'm happy to weigh in on specific issues, but there are just so many CI failures in this PR. It might be worth breaking this into smaller changes with fewer dependency upgrades so that we could discuss what is breaking.

PR #7279, opened 2025-09-29
7273-generic-client -non-fhir-call-fix-custom-headers

ibaisi:
#7280


robogary:
Formatting check succeeded!

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/BaseClient.java
```diff
@@ -439,6 +439,8 @@ protected <T> T invokeClient(
 						keepResponseAndLogIt(theLogRequestAndResponse, response, responseString);
 						inputStreamToReturn = new ByteArrayInputStream(responseString.getBytes(Charsets.UTF_8));
 					}
+				} else {
+					inputStreamToReturn = new ByteArrayInputStream(inputStream.readAllBytes());
```
TipzCM: is the if statement below still needed? does a bytearrayinputstream with 0 bytes return null?

TipzCM (approved):

hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/impl/BaseClient.java
ibaisi: I restructured the null check to the top as yes, it can be null and it looks like we should always return an input stream.

PR #7292, opened 2025-10-08
[7291] Fix bulk export get after delete response

### LINKED ISSUE #7291: Bulk Export status returns 202 after delete request rather than 404
Following the spec: [HL7.FHIR.UV.BULKDATA\Bulk Data Export - FHIR v4.0.1](https://hl7.org/fhir/uv/bulkdata/STU2/export.html#bulk-data-delete-request)
> Following the delete request, when subsequent requests are made to the polling location, the server SHALL return a 404 Not Found error and an associated FHIR OperationOutcome in JSON format.

However, when querying the polling location right after the delete request is approved with 202, a 202 status is received despite the Delete request having been approved and processed.

YalingPeiS:
### Issue

With bulk export, when querying the polling location right after the delete request being approved with 202, a 202 status is received despite the Delete request having been approved and processed.

### Cause

This is a race condition of reading/writing Job status. When a delete job request is received, the web service thread marks the Job instance as cancelled but it does not change the Job status. The update of the Job status is done by the background task that runs every minute. On a subsequent GET after DELETE, the cancelled flag is set but the current status may still be QUEUED or IN_PROGRESS, resulting in success GET and 202 being returned.  

### Fix
Add the check of `isCancelled` for the Job instance in addition to the original Job Status check. This ensures that proper response is sent to user.

Closes #7291

### DIFF
--- a/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkDataExportProvider.java
+++ b/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkDataExportProvider.java
@@ -495,11 +495,11 @@ public class BulkDataExportProvider {
 			case IN_PROGRESS:
 				//noinspection deprecation - we need to support old jobs after upgrade.
 			case ERRORED:
-				if (theRequestDetails.getRequestType() == RequestTypeEnum.DELETE) {
-					handleDeleteRequest(theJobId, response, info.getStatus());
-				} else if (info.isCancelled()) {
-					ourLog.info("{} job instance <{}> is marked cancelled.", info.getStatus(), theJobId);
+				if (info.isCancelled()) {
+					ourLog.info("{} job instance <{}> was marked cancelled.", info.getStatus(), theJobId);
 					processCancelledJobResponse(response, theJobId);
+				} else if (theRequestDetails.getRequestType() == RequestTypeEnum.DELETE) {
+					handleDeleteRequest(theJobId, response, info.getStatus());
 				} else {
 					response.setStatus(Constants.STATUS_HTTP_202_ACCEPTED);
 					String dateString = getTransitionTimeOfJobInfo(info);


robogary:
Formatting check succeeded!

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkDataExportProvider.java
```diff
@@ -497,6 +497,9 @@ public void exportPollStatus(
 			case ERRORED:
 				if (theRequestDetails.getRequestType() == RequestTypeEnum.DELETE) {
 					handleDeleteRequest(theJobId, response, info.getStatus());
+				} else if (info.isCancelled()) {
+					ourLog.info("{} job instance <{}> is marked cancelled.", info.getStatus(), theJobId);
+					processCancelledJobResponse(response, theJobId);
```
robcaruso: The `info.isCancelled()` needs to be checked first.  There is a use case where the user could send in a delete twice and the second delete should return a 404

YalingPeiS: Updated.  Thank you.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_6_0/7291-fix-bulk-export-get-after-delete-response.yaml
```diff
@@ -0,0 +1,7 @@
+---
+type: fix
+issue: 7291
+jira: SMILE-11072, SMILE-11142
+title: "Previously, when a Bulk Export job was created, querying the polling location immediately after a successful
+delete request returned a 202 status instead of the expected 404. This has been fixed and the response now contains
+an `OperationOutcome` indicating that the job was cancelled to conform to the FHIR Bulk Data specification."
```
tadgh, line 7: Extremely well-written changelog!

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkDataExportProvider.java
```diff
@@ -495,7 +495,10 @@ public void exportPollStatus(
 			case IN_PROGRESS:
 				//noinspection deprecation - we need to support old jobs after upgrade.
 			case ERRORED:
-				if (theRequestDetails.getRequestType() == RequestTypeEnum.DELETE) {
+				if (info.isCancelled()) {
+					ourLog.info("{} job instance <{}> was marked cancelled.", info.getStatus(), theJobId);
+					processCancelledJobResponse(response, theJobId);
+				} else if (theRequestDetails.getRequestType() == RequestTypeEnum.DELETE) {
```
tadgh, line 501: question: I'm a bit confused here. It looks like we already had a cancelled state, which would return a 404. Why does this code branch need to exist for ERRORED? Is it a job that gets cancelled after it has ERROR state?

YalingPeiS: This code branch is for FINALIZE, QUEUED, IN_PROGRESS and ERRORED.
It looks that ERRORED is @deprecated(since = "6.6") so it might be there just for backward compatibility.

PR #7310, opened 2025-10-17
Feature - Permit Bulk Export on Compartment by FHIR Query 

### LINKED ISSUE #7342: Enhance RuleBuilder code to support compartment matching by FHIR query

**Describe the issue**
A clear and concise description of what the feature request is. Please include details about what problem you are
ultimately trying to solve (i.e. what are you building with HAPI FHIR) and how this feature would help.

- HAPI FHIR Version

**Additional context**

jdar8:
**What was done:** 
- Add new `AuthResourceResolver`
- Add 2 new rules to support these permissions. They need to match resources/compartments based on a FHIR query (see cases in issue comments)
   - These rules use a new `AuthResourceResolver` service to query the DB to apply SearchParameter matching using an `InMemoryResourceMatcher`
- Add new logic to rule builder, to allow building these new rules
- Tests and changelog

Closes #7342

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7457-restore-removed-method.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7457-restore-removed-method.yaml
@@ -1,0 +1,6 @@
+---
+type: fix
+issue: 7457
+title: "A method on `IResourceChangeListenerRegistry` now requires a partition
+   ID to be supplied with new listeners being registered. This previous method
+   was removed, but has been restored and deprecated."
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/cache/IResourceChangeListenerRegistry.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/cache/IResourceChangeListenerRegistry.java
@@ -36,6 +36,27 @@ import java.util.Set;
 public interface IResourceChangeListenerRegistry {
 
 	/**
+	 * @deprecated Use {@link #registerResourceResourceChangeListener(String, RequestPartitionId, SearchParameterMap, IResourceChangeListener, long)}
+	 * instead. This method assumes the default partition, but will be removed in the future.
+	 */
+	@Deprecated(since = "8.8.0", forRemoval = true)
+	default IResourceChangeListenerCache registerResourceResourceChangeListener(
+			String theResourceName,
+			SearchParameterMap theSearchParameterMap,
+			IResourceChangeListener theResourceChangeListener,
+			long theRemoteRefreshIntervalMs) {
+		// Note: We use RequestPartitionId.defaultPartition() in this deprecated method even though
+		// it's really not a safe object to use generally. In this case it's safe because we check
+		// for it downstream and replace it with RequestPartitionId.defaultPartition(PartitionSettings)
+		return registerResourceResourceChangeListener(
+				theResourceName,
+				RequestPartitionId.defaultPartition(),
+				theSearchParameterMap,
+				theResourceChangeListener,
+				theRemoteRefreshIntervalMs);
+	}
+
+	/**
 	 * Register a listener in order to be notified whenever a resource matching the provided SearchParameterMap
 	 * changes in any way.  If the change happened on the same jvm process where this registry resides, then the listener will be called
 	 * within {@link ResourceChangeListenerCacheRefresherImpl#LOCAL_REFRESH_INTERVAL_MS} of the change happening.  If the change happened
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/cache/ResourceChangeListenerRegistryImpl.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/cache/ResourceChangeListenerRegistryImpl.java
@@ -23,6 +23,7 @@ import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.context.RuntimeResourceDefinition;
 import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.interceptor.model.RequestPartitionId;
+import ca.uhn.fhir.jpa.model.config.PartitionSettings;
 import ca.uhn.fhir.jpa.searchparam.SearchParameterMap;
 import ca.uhn.fhir.jpa.searchparam.matcher.InMemoryMatchResult;
 import ca.uhn.fhir.jpa.searchparam.matcher.InMemoryResourceMatcher;
@@ -52,13 +53,16 @@ public class ResourceChangeListenerRegistryImpl implements IResourceChangeListen
 	private final Queue<ResourceChangeListenerCache> myListenerEntries = new ConcurrentLinkedQueue<>();
 	private final FhirContext myFhirContext;
 	private final ResourceChangeListenerCacheFactory myResourceChangeListenerCacheFactory;
+	private final PartitionSettings myPartitionSettings;
 	private InMemoryResourceMatcher myInMemoryResourceMatcher;
 
 	public ResourceChangeListenerRegistryImpl(
 			FhirContext theFhirContext,
+			PartitionSettings thePartitionSettings,
 			ResourceChangeListenerCacheFactory theResourceChangeListenerCacheFactory,
 			InMemoryResourceMatcher theInMemoryResourceMatcher) {
 		myFhirContext = theFhirContext;
+		myPartitionSettings = thePartitionSettings;
 		myResourceChangeListenerCacheFactory = theResourceChangeListenerCacheFactory;
 		myInMemoryResourceMatcher = theInMemoryResourceMatcher;
 	}
@@ -93,9 +97,15 @@ public class ResourceChangeListenerRegistryImpl implements IResourceChangeListen
 					+ " cannot be evaluated in-memory: " + inMemoryMatchResult.getUnsupportedReason()
 					+ ".  Only search parameter maps that can be evaluated in-memory may be registered.");
 		}
+
+		RequestPartitionId requestPartitionId = theRequestPartitionId;
+		if (requestPartitionId != null && requestPartitionId.isDefaultPartition()) {
+			requestPartitionId = RequestPartitionId.defaultPartition(myPartitionSettings);
+		}
+
 		return add(
 				theResourceName,
-				theRequestPartitionId,
+				requestPartitionId,
 				theResourceChangeListener,
 				theSearchParameterMap,
 				theRemoteRefreshIntervalMs);
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/config/SearchParamConfig.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/config/SearchParamConfig.java
@@ -29,6 +29,7 @@ import ca.uhn.fhir.jpa.cache.ResourceChangeListenerCache;
 import ca.uhn.fhir.jpa.cache.ResourceChangeListenerCacheFactory;
 import ca.uhn.fhir.jpa.cache.ResourceChangeListenerCacheRefresherImpl;
 import ca.uhn.fhir.jpa.cache.ResourceChangeListenerRegistryImpl;
+import ca.uhn.fhir.jpa.model.config.PartitionSettings;
 import ca.uhn.fhir.jpa.searchparam.MatchUrlService;
 import ca.uhn.fhir.jpa.searchparam.SearchParameterMap;
 import ca.uhn.fhir.jpa.searchparam.extractor.ISearchParamExtractor;
@@ -118,10 +119,14 @@ public class SearchParamConfig {
 	@Bean
 	IResourceChangeListenerRegistry resourceChangeListenerRegistry(
 			FhirContext theFhirContext,
+			PartitionSettings thePartitionSettings,
 			ResourceChangeListenerCacheFactory theResourceChangeListenerCacheFactory,
 			InMemoryResourceMatcher theInMemoryResourceMatcher) {
 		return new ResourceChangeListenerRegistryImpl(
-				theFhirContext, theResourceChangeListenerCacheFactory, theInMemoryResourceMatcher);
+				theFhirContext,
+				thePartitionSettings,
+				theResourceChangeListenerCacheFactory,
+				theInMemoryResourceMatcher);
 	}
 
 	@Bean
--- a/hapi-fhir-jpaserver-searchparam/src/test/java/ca/uhn/fhir/jpa/cache/ResourceChangeListenerRegistryImplTest.java
+++ b/hapi-fhir-jpaserver-searchparam/src/test/java/ca/uhn/fhir/jpa/cache/ResourceChangeListenerRegistryImplTest.java
@@ -4,6 +4,7 @@ import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.interceptor.model.RequestPartitionId;
 import ca.uhn.fhir.jpa.cache.config.RegisteredResourceListenerFactoryConfig;
+import ca.uhn.fhir.jpa.model.config.PartitionSettings;
 import ca.uhn.fhir.jpa.searchparam.SearchParameterMap;
 import ca.uhn.fhir.jpa.searchparam.matcher.InMemoryMatchResult;
 import ca.uhn.fhir.jpa.searchparam.matcher.InMemoryResourceMatcher;
@@ -66,7 +67,7 @@ class ResourceChangeListenerRegistryImplTest {
 
 		@Bean
 		public IResourceChangeListenerRegistry resourceChangeListenerRegistry(InMemoryResourceMatcher theInMemoryResourceMatcher) {
-			return new ResourceChangeListenerRegistryImpl(ourFhirContext, myResourceChangeListenerCacheFactory, theInMemoryResourceMatcher);
+			return new ResourceChangeListenerRegistryImpl(ourFhirContext, new PartitionSettings(), myResourceChangeListenerCacheFactory, theInMemoryResourceMatcher);
 		}
 
 		@Bean
@@ -112,6 +113,20 @@ class ResourceChangeListenerRegistryImplTest {
 	public void after() {
 		myResourceChangeListenerRegistry.clearListenersForUnitTest();
 			ResourceChangeListenerCache.setNowForUnitTests(null);
+	}
+
+	/**
+	 * Calling the deprecated registerResourceResourceChangeListener method should still
+	 * register a listener
+	 */
+	@Test
+	public void testRegisterListenerWithNoPartitionSpecified() {
+		IResourceChangeListener listener1 = mock(IResourceChangeListener.class);
+		myResourceChangeListenerRegistry.registerResourceResourceChangeListener(PATIENT_RESOURCE_NAME, ourMap, listener1, TEST_REFRESH_INTERVAL_MS);
+
+		when(mySearchParamMatcher.match(any(), any())).thenReturn(InMemoryMatchResult.successfulMatch());
+
+		assertEquals(1, myResourceChangeListenerRegistry.size());
 	}
 
 	@Test
--- a/hapi-fhir-jpaserver-searchparam/src/test/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImplTest.java
+++ b/hapi-fhir-jpaserver-searchparam/src/test/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImplTest.java
@@ -68,7 +68,6 @@ import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertSame;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.reset;
@@ -543,7 +542,7 @@ public class SearchParamRegistryImplTest {
 
 		@Bean
 		IResourceChangeListenerRegistry resourceChangeListenerRegistry(FhirContext theFhirContext, ResourceChangeListenerCacheFactory theResourceChangeListenerCacheFactory, InMemoryResourceMatcher theInMemoryResourceMatcher) {
-			return new ResourceChangeListenerRegistryImpl(theFhirContext, theResourceChangeListenerCacheFactory, theInMemoryResourceMatcher);
+			return new ResourceChangeListenerRegistryImpl(theFhirContext, new PartitionSettings(), theResourceChangeListenerCacheFactory, theInMemoryResourceMatcher);
 		}
 
 		@Bean
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/stresstest/GiantTransactionPerfTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/stresstest/GiantTransactionPerfTest.java
@@ -1,7 +1,6 @@
 package ca.uhn.fhir.jpa.stresstest;
 
 import ca.uhn.fhir.context.FhirContext;
-import ca.uhn.fhir.interceptor.api.IInterceptorBroadcaster;
 import ca.uhn.fhir.interceptor.api.IInterceptorService;
 import ca.uhn.fhir.interceptor.executor.InterceptorService;
 import ca.uhn.fhir.interceptor.model.ReadPartitionIdRequestDetails;
@@ -237,7 +236,7 @@ public class GiantTransactionPerfTest {
 			return retVal;
 		});
 
-		myResourceChangeListenerRegistry = new ResourceChangeListenerRegistryImpl(ourFhirContext, myResourceChangeListenerCacheFactory, myInMemoryResourceMatcher);
+		myResourceChangeListenerRegistry = new ResourceChangeListenerRegistryImpl(ourFhirContext, new PartitionSettings(), myResourceChangeListenerCacheFactory, myInMemoryResourceMatcher);
 		myResourceChangeListenerCacheRefresher.setResourceChangeListenerRegistry(myResourceChangeListenerRegistry);
 
 		mySearchParamRegistry = new SearchParamRegistryImpl();
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/repository/SearchConverter.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/repository/SearchConverter.java
@@ -1,0 +1,146 @@
+/*-
+ * #%L
+ * HAPI FHIR Storage api
+ * %%
+ * Copyright (C) 2014 - 2026 Smile CDR, Inc.
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
+package ca.uhn.fhir.jpa.repository;
+
+import ca.uhn.fhir.context.FhirContext;
+import ca.uhn.fhir.jpa.searchparam.SearchParameterMap;
+import ca.uhn.fhir.model.api.IQueryParameterAnd;
+import ca.uhn.fhir.model.api.IQueryParameterOr;
+import ca.uhn.fhir.model.api.IQueryParameterType;
+import ca.uhn.fhir.rest.param.TokenOrListParam;
+import ca.uhn.fhir.rest.param.TokenParam;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+import jakarta.annotation.Nonnull;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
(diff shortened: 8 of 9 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBuilder.java
```diff
@@ -1001,6 +1019,93 @@ public IAuthRuleBuilderRuleBulkExportWithTarget withResourceTypes(Collection<Str
...
+			}
+		}
+
+		private class RuleBuilderPatientMatcherBulkExport implements IAuthRuleBuilderRulePatientMatcherBulkExport {
+			private RulePatientBulkExportByCompartmentMatcherImpl myRulePatientBulkExportByCompartmentMatcher;
+
+			@Override
+			public IAuthRuleBuilderRuleBulkExportWithTarget patientExportOnPatient(
+					@Nonnull String theCompartmentFilterMatcher) {
+				if (myRulePatientBulkExportByCompartmentMatcher == null) {
```
TipzCM, line 1015: This code looks suspiciously similar to the code above (but for the RulePatientBulkExportByCompartmentMatcher obj).

It's screaming to have a common base class (even if they have their own separate interfaces) just to minimize the code duplication

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
```diff
@@ -0,0 +1,212 @@
...
+
+		if (theRequestDetails == null) {
+			return null;
+		}
+
+		BulkExportJobParameters inboundBulkExportRequestOptions = (BulkExportJobParameters)
+				theRequestDetails.getUserData().get(REQUEST_ATTRIBUTE_BULK_DATA_EXPORT_OPTIONS);
+
+		if (inboundBulkExportRequestOptions.getExportStyle() != OUR_EXPORT_STYLE) {
+			// If the requested export style is not for a GROUP, then abstain
```
TipzCM: Copy-pasted code is often a sign you need to refactor this to have a common base class that handles this shared logic.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
```diff
@@ -0,0 +1,212 @@
...
+		}
+
+		List<IBaseResource> thePatientResources =
+				theRuleApplier.getAuthResourceResolver().resolveCompartmentByIds(patientIdOptions, "Patient");
+
+		// Apply the FhirQueryTester (which contains a inMemoryResourceMatcher) to the found Patient compartment
+		// resource,
+		// and return the verdict
+		// All requested Patient IDs must be permitted to return an ALLOW verdict.
+		List<Boolean> verdicts = thePatientResources.stream()
```
TipzCM: So i think this is one of those cases where the stream is less efficient than a standard for-loop.

After all, we don't really need to evaluate *every* patient to make our conclusion in every case.

And i assume "doesn't apply" is the most likely outcome (as opposed to ALLOW/DENY)

```
Map<Boolean, Integer> counts = new HashMap();
counts.put(true, 0);
counts.put(false, 0);

for (IBaseResource patient : thePatientResources) {
   boolean applies = applyTestersAtLeastOneMatch(theOperation, theRequestDetails, patient, theRuleApplier);

   counts.put(apples, counts.get(applies) + 1);

   if (counts.get(applies) > 0 && counts.get(!applies) > 0) {
    // the second we have a single discrepency, we dont need to check the rest - we know we're going to return null
     return null;
  }
}

return counts.get(true) > 0 ? ALLOW : DENY;
```

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
```diff
@@ -0,0 +1,212 @@
...
+		}
+		return true;
+	}
+
+	/**
+	 * Remove the resource type and "?" prefix, if present
+	 * since resource type is implied for the rule based on the permission (Patient in this case)
+	 */
+	private static String sanitizeQueryFilter(String theFilter) {
+		if (theFilter.contains("?")) {
```
TipzCM: You have this code above too. Maybe put this into a FilterUtil class or something.

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBuilderTest.java
```diff
@@ -84,6 +84,70 @@ public void testBulkExportPermitsIfASingleGroupMatches() {
...
+	}
+
+	private static Stream<Arguments> patientMatcherBulkExportParams() {
+		return Stream.of(
+			Arguments.of(List.of("identifier=foo|bar"), List.of()),
+			Arguments.of(List.of("identifier=foo|bar"), List.of("Patient", "Observation")),
+			// Multiple arguments may be added to the filter when multiple FHIR_OP_INITIATE_BULK_DATA_EXPORT_PATIENTS_MATCHING permissions
+			// are added to the same user, even when the permission does not accept multiple (a list of) arguments by itself.
+			Arguments.of(List.of("identifier=foo|bar", "name=Doe"), List.of()),
+			Arguments.of(List.of("identifier=foo|bar", "name=Doe&active=true"), List.of("Patient", "Observation"))
```
TipzCM: since parameter ordering doesn't matter, why not have the same thing, but with flipped parameters (`active=true&name=Doe`)? just for testing fun.

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleGroupBulkExportByCompartmentMatcherImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+import static org.mockito.Mockito.when;
+
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+class RuleGroupBulkExportByCompartmentMatcherImplTest {
+
+	private final RestOperationTypeEnum myOperation = RestOperationTypeEnum.EXTENDED_OPERATION_SERVER;
+	private final Pointcut myPointcut = Pointcut.STORAGE_INITIATE_BULK_EXPORT;
```
TipzCM: `private static final` for constants.

And you don't need to name them 'my...' since they are both constants and are never updated.

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleGroupBulkExportByCompartmentMatcherImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+	@Mock
+	private IAuthResourceResolver myAuthResourceResolver;
+	@Mock
+	private IBaseResource myResource;
+	@Mock
+	private IAuthorizationSearchParamMatcher mySearchParamMatcher;
+	@Mock
+	private RequestDetails myRequestDetails;
+	@Mock
+	private Set<AuthorizationFlagsEnum> myFlags;
```
TipzCM: I'm confused why some of these need to be mocks.

Mocking is usually only needed for services or (if we're in some generic case) interfaces we cannot construct real objects of.

But this is a set. And you're never mocking it, so why does it need to be a mock at all?

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleGroupBulkExportByCompartmentMatcherImplTest.java
```diff
@@ -0,0 +1,109 @@
...
+		options.setResourceTypes(theRequestedResourceTypes);
+		options.setGroupId("Group/G1");
+
+		when(myRequestDetails.getUserData()).thenReturn(Map.of(AuthorizationInterceptor.REQUEST_ATTRIBUTE_BULK_DATA_EXPORT_OPTIONS, options));
+
+		if (theSearchParamMatcherMatchResult != null) {
+			when(myRuleApplier.getAuthResourceResolver()).thenReturn(myAuthResourceResolver);
+			when(myAuthResourceResolver.resolveCompartmentById(any())).thenReturn(myResource);
+			when(myRuleApplier.getSearchParamMatcher()).thenReturn(mySearchParamMatcher);
+			when(myResource.fhirType()).thenReturn("Group");
```
TipzCM, line 69: I mean, ok.... but probably easier to just provide an actual concrete Group resource, imo.

Either way, this is just a nitpick here.

TipzCM (approved):

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleGroupBulkExportByCompartmentMatcherImplTest.java
jdar8: That would've been my preferred way as well. But I can't import resource types since this module does not have dependencies for the resource models. So I had to live with this work around.

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleGroupBulkExportByCompartmentMatcherImplTest.java
jdar8: I'm not sure how this became a mock either. Removed this class variable and just input an empty set where it was used. Made changes in the Patient version as well

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleGroupBulkExportByCompartmentMatcherImplTest.java
jdar8: Agreed. 
That being said, I just inlined them. Probably makes more sense to anyways since there aren't too many usages.

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBuilderTest.java
jdar8: Sure! Added the test case :)

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
jdar8: That's a good point. Thanks for the suggestion, I've modified the code to a for loop

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/AuthResourceResolver.java
```diff
@@ -0,0 +1,59 @@
...
+import org.hl7.fhir.instance.model.api.IIdType;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+
+/**
+ * Small service class to inject DB access into an interceptor
+ * For example, used in bulk export security to allow querying for resource to match against permission argument filters
+ */
+@Service
```
tadgh: issue: remove service annotation, HAPI-FHIR does not do component scanning

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/AuthResourceResolver.java
```diff
@@ -0,0 +1,59 @@
...
+
+	public AuthResourceResolver(DaoRegistry myDaoRegistry) {
+		this.myDaoRegistry = myDaoRegistry;
+	}
+
+	public IBaseResource resolveCompartmentById(IIdType theResourceId) {
+		return myDaoRegistry
+				.getResourceDao(theResourceId.getResourceType())
+				.read(theResourceId, new SystemRequestDetails());
+	}
```
tadgh, line 48: issue: this method is called "resolve compartment by ID", but as far as I can tell, it simply performs a read using an ID. Should this perhaps be renamed?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
jdar8: Extracted a base abstract class (`BaseRuleBulkExportByCompartmentMatcher`) that handles 
- null inputs
- resource type checking
- export style checking

`RulePatientBulkExportByCompartmentMatcherImpl` and `RuleGroupBulkExportByCompartmentMatcherImpl` now extend this class.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
jdar8: Extracted into base class

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/AuthResourceResolver.java
```diff
@@ -0,0 +1,59 @@
...
+
+	public List<IBaseResource> resolveCompartmentByIds(List<String> theResourceIds, String theResourceType) {
+		TokenOrListParam t = new TokenOrListParam(null, theResourceIds.toArray(String[]::new));
+
+		SearchParameterMap m = new SearchParameterMap();
+		m.add(Constants.PARAM_ID, t);
+		return myDaoRegistry.getResourceDao(theResourceType).searchForResources(m, new SystemRequestDetails());
```
tadgh, line 55: Question: what happens when you have to page here? Or is this not a concern?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/BaseRule.java
```diff
@@ -93,6 +93,31 @@ private boolean applyTesters(
 		return retVal;
 	}
 
+	/**
+	 * Apply testers, and return true if at least 1 tester matches.
+	 * Returns false if all testers do not match.
+	 */
+	boolean applyTestersAtLeastOneMatch(
```
tadgh: suggestion: atLeastOneTesterMatches()

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthResourceResolver.java
```diff
@@ -0,0 +1,41 @@
...
+import org.hl7.fhir.instance.model.api.IIdType;
+
+import java.util.List;
+
+/**
+ * Small service class to inject DB access into an interceptor
+ * For example, used in bulk export security to allow querying for resource to match against permission argument filters
+ */
+public interface IAuthResourceResolver {
+	IBaseResource resolveCompartmentById(IIdType theResourceId);
```
tadgh: question: why does this service need an interface? Are there multiple implementations?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthResourceResolver.java
```diff
@@ -0,0 +1,41 @@
...
+
+	/**
+	 * Resolve a list of resources by ID. All resources should be the same type.
+	 * @param theResourceIds the FHIR id of the resource(s)
+	 * @param theResourceType the type of resource
+	 * @return A list of resources resolved by ID
+	 */
```
tadgh, line 39: issue: I'm confused as to why this class needs to exist at all. Can we not just query the dao?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthRuleBuilderRulePatientMatcherBulkExport.java
```diff
@@ -0,0 +1,34 @@
...
+package ca.uhn.fhir.rest.server.interceptor.auth;
+
+import jakarta.annotation.Nonnull;
+
+/**
+ * @since 8.6.0
+ */
+public interface IAuthRuleBuilderRulePatientMatcherBulkExport {
+	/**
+	 * Allow/deny <b>group-level</b> export rule applies to the Group with the given resource ID, e.g. <code>Group/123</code>
```
tadgh: issue: javadoc does not match method

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBulkExportImpl.java
```diff
@@ -120,11 +120,12 @@ public AuthorizationInterceptor.Verdict applyRule(
 
 		// 1. If each of the requested resource IDs in the parameters are present in the users permissions, Approve
 		// 2. If any requested ID is not present in the users permissions, Deny.
-		if (myWantExportStyle == BulkExportJobParameters.ExportStyle.PATIENT)
+		if (myWantExportStyle == BulkExportJobParameters.ExportStyle.PATIENT) {
```
tadgh, line 123: nice catch.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
jdar8: EDIT - undid these abstraction changes. 

I was talking with @tadgh and we don't really know for sure if these rules should be functionally the same yet, so we should not abstract them yet. 

As an example, from the design discussions for this ticket, I can see a world where Patient rules will need a bunch more functionality in `applyRule()`, particularly around `_typeFilter` matching. 

So unless there is further objection, I will leave them separate for now until things settle some more.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleBuilder.java
jdar8: See https://github.com/hapifhir/hapi-fhir/pull/7310#discussion_r2479866856

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/AuthResourceResolver.java
jdar8: removed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/AuthResourceResolver.java
jdar8: Oops, I've been meaning to rename this. I think I renamed the parameter, but forgot about the method. 

Changed compartment --> resource

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/AuthResourceResolver.java
jdar8: I don't think it's a concern. `searchForResources()` loads all resources into ram according to the javadocs. Internally, looks like it calls `provider.getAllResources()`

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/BaseRule.java
jdar8: renamed

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthResourceResolver.java
jdar8: Yes, we could just query the dao. I added this class for a few reasons:
- Since the dao is slow, we wanted to limit the usage of the dao during any security operations. So we have a new class to limit actions (and probably discourage dao queries by security in the future)
- I added this early on, I wasn't sure how much I would end up needing to query the dao. Turns out (for now) it is just by ID(s), and that's pretty simple. But in the future, will we need to add more functionality?
- Overall, I just wasn't sure about the implications of allowing dao access freely to the auth rules

Do you think we'd be better off without this class by just providing the dao to IRuleApplier?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthRuleBuilderRulePatientMatcherBulkExport.java
jdar8: changed

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthResourceResolver.java
jdar8: No, just a single implementation (for now)

The interface is needed to avoid a circular dependency. 

So we have 
- `AuthResourceResolver` in `hapi-fhir-jpaserver-base`
- `IAuthResourceResolver` in `hapi-fhir-server`
- `AuthorizationInterceptor` and Rules in `hapi-fhir-server`

`AuthResourceResolver` needs to access the dao, so I put it in `jpaserver-base`, but `jpaserver-base` depends on `server`. So I can't directly import the resource resolver or dao from `jpaserver-base` into `AuthorizationInterceptor` where it's needed.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthResourceResolver.java
tadgh: Eh it's fine for now, the discouraging aspect is a good enough reason to retain it.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthRuleBuilderRule.java
```diff
@@ -127,6 +127,22 @@ public interface IAuthRuleBuilderRule {
 	 */
 	IAuthRuleBuilderRuleBulkExport bulkExport();
 
+	/**
+	 * This rule permits the user to initiate a FHIR bulk export
+	 * by providing a filter matcher on Group compartment(s).
+	 *
+	 * @since 8.6.0
+	 */
+	IAuthRuleBuilderRuleGroupMatcherBulkExport bulkExportGroupCompartmentMatcher();
```
jamesagnew: I don't love the names of these new methods - All of the existing methods in this interface are verbs (`update()`, `patch()`, `bulkExport()`, etc) that read fluently, but these new ones break that pattern with a really cryptic noun instead.

Do we need these new interfaces and methods at all? E.g. could we move the new functionality into `IAuthRuleBuilderRuleBulkExport` so that instead of needing to add rule:

```java
builder.allow().bulkExportGroupCompartmentMatcher().groupExportOnGroup("?" + theCompartmentMatcherFilter).withResourceTypes(resourceTypes);
```

could we just allow a syntax like:

```java
```java
builder.allow().bulkExport().groupExportOnFilter("?" + theCompartmentMatcherFilter).withResourceTypes(resourceTypes);
```

This would make the API a lot cleaner, and make the new features more discoverable since we wouldn't have 3 top level methods about the same operation.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/AuthResourceResolver.java
```diff
@@ -0,0 +1,57 @@
...
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.instance.model.api.IIdType;
+
+import java.util.List;
+
+/**
+ * Small service class to inject DB access into an interceptor
+ * For example, used in bulk export security to allow querying for resource to match against permission argument filters
+ */
+public class AuthResourceResolver implements IAuthResourceResolver {
```
jamesagnew, line 37: This makes me nervous. I get the design, but it's breaking the well-established pattern we have that auth interceptor doesn't resolve anything, it just looks at the data it's passed and makes a decision.

Instead of creating a loading infrastructure, could we add a new parameter object to `STORAGE_PRE_INITIATE_BULK_EXPORT` so that `BulkExportJobService` could handle loading the Group and pass it to AuthInterceptor as a part of the interceptor request? That way AuthInterceptor could look at it and make its decisions without needing to know how to fetch anything

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/IAuthRuleBuilderRule.java
jdar8: That’s a good point. I’ve consolidated the API into `IAuthRuleBuilderRuleBulkExport` as suggested. Also decided to just go ahead with @TipzCM's suggestion to refactor the Group/Patient rules to have a `BaseRuleBulkExportByCompartmentMatcher` with shared functionality for the resource types list since it also made the API cleaner in `RuleBuilder`.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/AuthResourceResolver.java
jdar8: As discussed during design, we will stick with this approach since there may be some other use cases for this, such as new features for permissioning with $meta and patch. Also, as discussed, added a query count test (`BulkDataExportAuthorizationQueryCountTest`) to ensure there are 2 select queries during authorization.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
```diff
@@ -0,0 +1,182 @@
...
+		}
+
+		List<IBaseResource> thePatientResources =
+				theRuleApplier.getAuthResourceResolver().resolveResourcesByIds(patientIdOptions, "Patient");
+
+		// Apply the FhirQueryTester (which contains a inMemoryResourceMatcher) to the found Patient compartment
+		// resource,
+		// and return the verdict
+		// All requested Patient IDs must be permitted to return an ALLOW verdict.
+		Map<Boolean, Integer> counts = new HashMap<>();
```
jamesagnew: I'm not sure I follow why these counts are needed. The logic appears to be:
* If `atLeastOneTesterMatches` returns false for any patients, immediately DENY
* If `atLeastOneTesterMatches` returns false for all patients, ALLOW
* If `thePatientResources` is empty, return no verdict

You should be able to implement that without needing to count anything, and it'll be a whole lot more readable

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/BaseRuleBulkExportByCompartmentMatcher.java
```diff
@@ -0,0 +1,108 @@
...
+
+	public void setResourceTypes(Collection<String> theResourceTypes) {
+		myResourceTypes = theResourceTypes;
+	}
+
+	/**
+	 * Remove the resource type and "?" prefix, if present
+	 * since resource type is implied for the rule based on the permission (Patient in this case)
+	 */
+	static String sanitizeQueryFilter(String theFilter) {
```
jamesagnew: Thought: Do any user-supplied strings get passed through this? Could it become an attack vector if someone requested a URL with 2 `?` in it? (which would technically be illegal HTTP but is something someone could do)

E.g., what if I requested a bulk export with a typeFilter like `patient=Patient/BAD,?patient=Patient/GOOD` when I'm only approved to see `patient/GOOD`, could I trick the interceptor?

jdar8: Hmm. It is used when parsing typeFilter, which is user supplied. 

I've switched over to using `Urlutil` and `MatchUrlUtil` to address this and reduce duplication.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RulePatientBulkExportByCompartmentMatcherImpl.java
jdar8: Refactored to use booleans instead of counting. Also added some javadoc for the cases. 
* null/abstain: If the list of patient resources is empty
* null/abstain: If all Patients evaluate to NO match
* DENY: If some Patients evaluate to match, while other Patients evaluate to NO match
* ALLOW: If all Patients evaluate to match

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_6_0/7342-enhance-rulebuilder-to-support-compartments-by-matchers.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: add
+issue: 7342
+title: "Enhanced the bulk export RuleBuilder code to support the identification of allowable Groups/Patients to export by a FHIR query matcher."
```
jamesagnew: QUick note: This file should be moved to the 8.8.0 folder since that will now be the release this change goes into.

jdar8: thanks for catching this. moved it to the current release folder

PR #7311, opened 2025-10-17
Add new methods to IHapiPackageCacheManager to return multiple results

lukedegruchy:
- Add two new methods:  loadPackageAssetsByUrl() and findPackageAssets()
- The intention is to allow duplicates among NPM packages to be handled by the caller, instead of the implicit assumption by loadPackageAssetByUrl() that only one resource will be returned
- loadPackageAssetsByUrl() is similar to loadPackageAssetByUrl() except that it allows the caller to configure paging and will return resource from "current" and "non-current" packages
- Similarly, findPackageAssets() takes a FindPackageAssetRequest for more fine-grained searching and will return multiple resources

Closes https://github.com/hapifhir/hapi-fhir/issues/7314

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java
@@ -586,8 +586,7 @@ public class JpaPackageCache extends BasePackageCacheManager implements IHapiPac
 		// The other APIs are newer and were introduced with newer NPM functionality in mind, including more refined
 		// handling of duplicate canonical URLs across packages
 		final List<NpmPackageVersionResourceEntity> npmPackageVersionResourceEntities =
-				loadPackageInfoByCanonicalUrlCurrentVersionOnly(
-						theFhirVersion, theCanonicalUrl, PageRequest.of(0, 2));
+				loadPackageInfoByCanonicalUrlCurrentVersionOnly(theFhirVersion, theCanonicalUrl, PageRequest.of(0, 2));
 
 		final List<IBaseResource> resources = npmPackageVersionResourceEntities.stream()
 				.map(this::loadPackageEntity)
@@ -681,10 +680,8 @@ public class JpaPackageCache extends BasePackageCacheManager implements IHapiPac
 
 	// We want to load the packages marked as current version true only
 	private List<NpmPackageVersionResourceEntity> loadPackageInfoByCanonicalUrlCurrentVersionOnly(
-		FhirVersionEnum theFhirVersion,
-		String theCanonicalUrl,
-		PageRequest thePageRequest) {
-		return loadPackageInfoByCanonicalUrl(theFhirVersion,  theCanonicalUrl, thePageRequest, null, null, true);
+			FhirVersionEnum theFhirVersion, String theCanonicalUrl, PageRequest thePageRequest) {
+		return loadPackageInfoByCanonicalUrl(theFhirVersion, theCanonicalUrl, thePageRequest, null, null, true);
 	}
 
 	// We want to load the packages whether they're marked as current version or not
@@ -694,7 +691,8 @@ public class JpaPackageCache extends BasePackageCacheManager implements IHapiPac
 			PageRequest thePageRequest,
 			@Nullable String thePackageId,
 			@Nullable String theVersionId) {
-		return loadPackageInfoByCanonicalUrl(theFhirVersion,  theCanonicalUrl, thePageRequest, thePackageId, theVersionId, null);
+		return loadPackageInfoByCanonicalUrl(
+				theFhirVersion, theCanonicalUrl, thePageRequest, thePackageId, theVersionId, null);
 	}
 
 	private List<NpmPackageVersionResourceEntity> loadPackageInfoByCanonicalUrl(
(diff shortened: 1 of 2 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/FindPackageAssetRequest.java
```diff
@@ -21,38 +21,71 @@
...
+import java.util.Optional;
 import java.util.StringJoiner;
 
 /**
  * Request parameters for {@link IHapiPackageCacheManager#findPackageAsset(FindPackageAssetRequest)}
  */
 public class FindPackageAssetRequest {
 
+	private static final int DEFAULT_PAGE_NUMBER = 0;
+	private static final int DEFAULT_PAGE_SIZE = 10;
```
TipzCM: Why so small?

We don't have many packages, but i feel like this number can be at least 50

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java
```diff
@@ -581,53 +581,79 @@ public NpmPackage installPackage(PackageInstallationSpec theInstallationSpec) th
...
+		final List<NpmPackageVersionResourceEntity> npmPackageVersionResourceEntities =
+				loadPackageInfoByCanonicalUrlCurrentVersion(
+						theFhirVersion, theCanonicalUrl, PageRequest.of(0, 2), null, null);
+
+		final List<IBaseResource> resources = npmPackageVersionResourceEntities.stream()
+				.map(this::loadPackageEntity)
+				.toList();
+
+		if (resources.size() > 1) {
+			ourLog.warn(
```
TipzCM, line 596: This is a weird API to me and i don't like that we just log a warning without doing anything to let the caller know that there may be more than 1.

But i know it's a required API and we are not trying to break the signature, so....

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java
```diff
@@ -670,33 +696,84 @@ private List<NpmPackageVersionResourceEntity> loadPackageInfoByCanonicalUrl(
...
+
+		if (slice.isEmpty()) {
+			return List.of();
+		} else {
+			return slice.getContent();
+		}
+	}
+
+	// We want to load the packages whether they're marked as current version or not
+	private List<NpmPackageVersionResourceEntity> loadPackageInfoByCanonicalUrlAnyVersion(
```
TipzCM: This code is so similar to the code above it, all because it's based on "current" or "any" version.

Be nice if we could unify them.

If the queries are complicated enough, we should have a NPMPackageversionsvc that constructs the queries manually using entity manager.

But, in your case, i think it might be easier to instead of duplicating the queries inthe JPA layer, have them take a parameter "currentVersion" and then instead of having `AND e.myPackageVersion.myCurrentVersion = true` on some and not on others, just have `AND (e.myPackageVersion.myCurrentVersion = currentVersion`

i dont' think these are called anywhere but Hapi, so it's probably an easy change. But if not, you can always deprecate the old queries)

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/packages/JpaPackageCacheDuplicateResourcesTest.java
```diff
@@ -118,6 +132,68 @@ void findPackageAssetInfoByUrl_duplicateResources() {
...
+		final List<IBaseResource> duplicatePackageAssets =
+			myPackageCacheManager.findPackageAssets(
+				FindPackageAssetRequest.noVersion(
+					FhirVersionEnum.R4,
+					MEASURE_URL,
+					SIMPLE_ALPHA_PACKAGE));
+
+		assertThat(duplicatePackageAssets).isNotNull().isNotEmpty().hasSize(2);
+
+		// LUKETODO:  assertions
```
TipzCM: should these be done?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/packages/JpaPackageCacheDuplicateResourcesTest.java
```diff
@@ -118,6 +132,68 @@ void findPackageAssetInfoByUrl_duplicateResources() {
...
+			myPackageCacheManager.findPackageAssets(
+				FindPackageAssetRequest.withVersion(
+					FhirVersionEnum.R4,
+					MEASURE_URL,
+					SIMPLE_ALPHA_PACKAGE,
+					VERSION_0_2));
+
+		assertThat(packageAssets).isNotNull().isNotEmpty().hasSize(1);
+
+		// LUKETODO:  assertions
```
TipzCM: here too

lukedegruchy: Test now fully implemented.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/packages/JpaPackageCacheDuplicateResourcesTest.java
lukedegruchy: Test now fully implemented.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/FindPackageAssetRequest.java
lukedegruchy: Bumped up to 50.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java
lukedegruchy: As discussed, we're going to ignore this inconsistency for the time being.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java
lukedegruchy: I've optimized the code paths by introducing HQL that returns records either with currentVersion = true or disregards that field entirely, depending on the Java method parameter.

PR #7313, opened 2025-10-20
[7312] Fix Elasticsearch container OOM failures in CI

fil512:
Added container memory limit of 700MB to prevent out-of-memory issues in CI environments. The JVM heap was already limited to 512MB via ES_JAVA_OPTS, but the container itself had no memory constraint, causing Elasticsearch to request more memory than available and be killed by the OOM killer.


robogary:
Formatting check succeeded!

fil512:
stale

fil512:
stale

PR #7185, opened 2025-08-12
Jd 20250811 add breaking change doc url query parsing

jdar8:

PR #7315, opened 2025-10-21
[7304] fix GET bulk export with typeFilter

YalingPeiS:
### Problem Statement
async GET /$export, _type and _typeFilter parameters not filter the resources based on the criteria.

### Cause
The parsing of the request query string checked the "?" and removed the part before the "?", causing parameter being removed from query string that contains the "?", such as `_type=Patient,Observation,Group&_typeFilter=Patient?gender=female&_outputFormat=application/fhir+ndjson` that would appear in GET `$export`

### Fix
Move the "?" check for resource to the specific method where it is needed.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_6_0/upgrade.md
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_6_0/upgrade.md
@@ -1,7 +1,3 @@
-## Breaking Changes
-
-* The parsing of query URLs has changed such that the '?' character should strictly be treated as the delimiter between path and query components. Any additional '?' characters within the query string must be percent-encoded (e.g. `%3F`). URLs containing unescaped '?' characters in the query portion may cause failures or unexpected behaviour.
-
 ## Changes to package installation
 
 A change has been made to package installation where any client-assigned ID included in the resource being installed will be ignored and replaced with a server-assigned ID.


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/UrlUtil.java
```diff
@@ -353,10 +353,8 @@ public static Map<String, String[]> parseQueryString(String theQueryString) {
 
 	private static void parseQueryString(String theQueryString, HashMap<String, List<String>> map) {
 		String query = defaultString(theQueryString);
-
```
YalingPeiS, line 356: The `if` block being removed here was added in [https://github.com/hapifhir/hapi-fhir/pull/7145/files](https://github.com/hapifhir/hapi-fhir/pull/7145/files) where `parseQueryString(...)` is called in `DaoResourceLinkResolver.extractIdentifierFromUrl(...)`. Moving this `if` block to the calling method to handle the special case so that `parseQuerySting(...)` would parse the query string like the following properly.
```$export?_outputFormat=application/fhir+ndjson&_type=Patient&_typeFilter=Patient?gender=female```

Revert the `?` checking to the previous version.

YalingPeiS:
> Question: With this change/partial revert, do we no longer require additional `?` characters in the URL to be unescaped? I ask because I think the same URL parsing change broke some tests in the past and I assumed this change would be permanent. So I escaped the URLs in the broken tests and [added a breaking change note](https://github.com/hapifhir/hapi-fhir/pull/7185/files#diff-dfb11218d96f6e103ce87f0c8b8f10f15daafec22bb8748b558eab1d08f615eaR1). If this note is no longer needed, we can remove it

Yes. The change here reverted the URL query string parsing to its previous version which parses the string with `?` properly.
I've removed the [breaking change note](https://github.com/hapifhir/hapi-fhir/pull/7185/files#diff-dfb11218d96f6e103ce87f0c8b8f10f15daafec22bb8748b558eab1d08f615eaR1)

Issue #7316, opened 2025-10-21
Enabling an MDM virtualized endpoint  breaks chain search


another-luis:
When working with an MDM virtualized endpoint, MDM tries to expand chain search reference parameters. This leads to misinterpreting the search parameter.

To reproduce this, enable MDM and run a chain search, such as this:

http://localhost:8000/Observation?subject.name=john

This will generate an error saying"Resource john is not known".

PR #7317, opened 2025-10-21
Make sure refparams with chains are not expanded via MDM

another-luis:
Fixes https://github.com/hapifhir/hapi-fhir/issues/7316

### DIFF
--- a/hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/svc/MdmSearchExpansionSvc.java
+++ b/hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/svc/MdmSearchExpansionSvc.java
@@ -147,7 +147,7 @@ public class MdmSearchExpansionSvc {
 		List<IQueryParameterType> toAdd = new ArrayList<>();
 		for (IQueryParameterType iQueryParameterType : orList) {
 			if (iQueryParameterType instanceof ReferenceParam refParam) {
-				if (!refParam.hasChain() && theParamTester.shouldExpand(theParamName, refParam)) {
+				if (theParamTester.shouldExpand(theParamName, refParam)) {
 					ourLog.debug("Found a reference parameter to expand: {}", refParam);
 					// First, attempt to expand as a source resource.
 					IIdType sourceId = newId(refParam.getValue());
--- a/hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/interceptor/MdmReadVirtualizationInterceptor.java
+++ b/hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/interceptor/MdmReadVirtualizationInterceptor.java
@@ -32,6 +32,7 @@ import ca.uhn.fhir.mdm.svc.MdmSearchExpansionSvc;
 import ca.uhn.fhir.rest.api.server.IPreResourceShowDetails;
 import ca.uhn.fhir.rest.api.server.RequestDetails;
 import ca.uhn.fhir.rest.api.server.storage.IResourcePersistentId;
+import ca.uhn.fhir.rest.param.ReferenceParam;
 import ca.uhn.fhir.rest.server.util.ICachedSearchDetails;
 import ca.uhn.fhir.util.FhirTerser;
 import ca.uhn.fhir.util.ResourceReferenceInfo;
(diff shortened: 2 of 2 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/interceptor/MdmReadVirtualizationInterceptorTest.java
```diff
@@ -341,6 +342,38 @@ public void testSearch_RequestDetailsIsReused() {
 
 	}
 
+	@Test
+	public void testSearch_Observation_chainReferencesAreLeftAlone() {
```
michaelabuckley, line 346: question: Does this mean we are returning the normal Patient, not the Golden patient?  I thought this endpoint should always return the golden patient.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/svc/MdmSearchExpansionSvc.java
```diff
@@ -146,9 +146,8 @@ private void expandAnyReferenceParameters(
 		List<IQueryParameterType> toRemove = new ArrayList<>();
 		List<IQueryParameterType> toAdd = new ArrayList<>();
 		for (IQueryParameterType iQueryParameterType : orList) {
-			if (iQueryParameterType instanceof ReferenceParam) {
-				ReferenceParam refParam = (ReferenceParam) iQueryParameterType;
-				if (theParamTester.shouldExpand(theParamName, refParam)) {
+			if (iQueryParameterType instanceof ReferenceParam refParam) {
+				if (!refParam.hasChain() && theParamTester.shouldExpand(theParamName, refParam)) {
```
michaelabuckley: question: we already have a param tester.  Why not modify that param tester at the source (i.e. MdmReadVirtualizationInterceptor.PARAM_TESTER_NO_RES_ID) instead of adding the logic here?

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/interceptor/MdmReadVirtualizationInterceptorTest.java
another-luis: In this test there is not golden record for the patient. I was just trying to show that an SP with a chain will not be expanded (and therefore fail to find the resources).  But in this example there is no golden record, so no substitution happens (which is secondary).

Having said that, I am working on another bug  - the one I discussed with you during office hours. In than one there is an apparent clash between what the documentation says and what the tests do. If PatientA and PatientB point to golden record PatientG, the SP is expanded to include PatientA, PatientB, and PatientG (as expected). The clash is in what happens next. 

According to the test, if PatientA was used in an SP for the search, all references returned should be to PatientA. The documentation, on the other hand, seems to indicate that they should all point to PatientG. But there is a degree of ambiguity in the documentation, because the context of the discussion in the [documentation](https://smilecdr.com/docs/mdm/mdm_virtualized_endpoint.html) is a case in which there is no SP expansion (i.e. search for Observations using a code, which can return  references to PatientA, PatientB, or PatientG). The documentation says that in this the results should all point to PatientG, but it doesn't elaborate on whether this should also be the case when PatientA or PatientB were used as search criteria. 

But all this is in a different bug that is waiting for a separate merge from Gary, who is in the process of refactoring a bunch of the MDM expansion code. Once his changes are in, I will add JamesA to the review to see if I can get some clarification. If, in fact, the golden record is what should be returned, then many (all?) of these tests will have to be updated.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/svc/MdmSearchExpansionSvc.java
another-luis: I didn't want to do that because this requires downcasting (from BaseParam to ReferenceParam):

```
@FunctionalInterface
	public interface IParamTester {

	boolean shouldExpand(String theParamName, BaseParam theParam);
}
```


Since this is code within the MDM expansion svc, I will do it.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/interceptor/MdmReadVirtualizationInterceptorTest.java
michaelabuckley: Ok

michaelabuckley (approved):

tadgh (approved):

PR #7323, opened 2025-10-22
Enable Combo SPs with Uplifred Refchains

jamesagnew:
This PR adds the ability to define a combo search param on an uplifted refchain

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
@@ -2502,13 +2502,26 @@ public class SearchBuilder implements ISearchBuilder<JpaPid> {
 			RuntimeSearchParam theComboParam,
 			List<JpaParamUtil.ComponentAndCorrespondingParam> theComboParamComponents) {
 
-		List<List<IQueryParameterType>> inputs = new ArrayList<>();
-		List<Runnable> runnableRemoveInputsFromParams = new ArrayList<>(theComboParamComponents.size());
+		List<List<IQueryParameterType>> inputs = new ArrayList<>(theComboParamComponents.size());
+		List<Runnable> searchParameterConsumerTasks = new ArrayList<>(theComboParamComponents.size());
 		for (JpaParamUtil.ComponentAndCorrespondingParam nextComponent : theComboParamComponents) {
 			boolean foundMatch = false;
 
+			/*
+			 * The following List<List<IQueryParameterType>> is a list of query parameters where the
+			 * outer list contains AND combinations, and the inner lists contain OR combinations.
+			 * For each component in the Combo SearchParameter, we need to find a list of OR parameters
+			 * (i.e. the inner List) which is appropriate for the given component.
+			 *
+			 * We can only use a combo param when the query parameter is fairly basic
+			 * (no modifiers such as :missing or :below, references are qualified with
+			 * a resource type, etc.) Once we've confirmed that we have a parameter for
+			 * each component, we remove the components from the source SearchParameterMap
+			 * since we're going to consume them and add a predicate to the SQL builder.
+			 */
 			List<List<IQueryParameterType>> sameNameParametersAndList = theParams.get(nextComponent.getParamName());
 			if (sameNameParametersAndList != null) {
+				boolean parameterIsChained = false;
 				for (int andIndex = 0; andIndex < sameNameParametersAndList.size(); andIndex++) {
 					List<IQueryParameterType> sameNameParametersOrList = sameNameParametersAndList.get(andIndex);
 					IQueryParameterType firstValue = sameNameParametersOrList.get(0);
@@ -2525,12 +2538,18 @@ public class SearchBuilder implements ISearchBuilder<JpaPid> {
 					}
 
 					inputs.add(sameNameParametersOrList);
-					runnableRemoveInputsFromParams.add(
-							() -> sameNameParametersAndList.remove(sameNameParametersOrList));
+					searchParameterConsumerTasks.add(() -> sameNameParametersAndList.remove(sameNameParametersOrList));
 					foundMatch = true;
 					break;
 				}
-			} else {
+			} else if (!nextComponent.getParamName().equals(nextComponent.getCombinedParamName())) {
+
+				/*
+				 * If we didn't find any parameters for the parameter name (e.g. "patient") and
+				 * we're looking for a chained parameter (e.g. "patient.identifier"), check if
+				 * there are any matches for the full combined parameter name
+				 * (e.g. "patient.identifier").
+				 */
 				List<List<IQueryParameterType>> combinedNameParametersAndList =
 						theParams.get(nextComponent.getCombinedParamName());
 				if (combinedNameParametersAndList != null) {
@@ -2549,9 +2568,10 @@ public class SearchBuilder implements ISearchBuilder<JpaPid> {
 							}
 
 							inputs.add(combinedNameParametersOrList);
-							runnableRemoveInputsFromParams.add(
+							searchParameterConsumerTasks.add(
 									() -> combinedNameParametersAndList.remove(combinedNameParametersOrList));
 							foundMatch = true;
+							break;
 						}
 					}
 				}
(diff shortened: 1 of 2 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4ComboNonUniqueParamTest.java
```diff
@@ -328,11 +331,11 @@ public void testStringAndToken_MultipleAnd() {
 		assertThat(actual).containsExactlyInAnyOrder(id1.toUnqualifiedVersionless().getValue());
 
 		String sql = myCaptureQueriesListener.getSelectQueries().get(0).getSql(true, false);
-		String expected = "SELECT t0.RES_ID FROM HFJ_IDX_CMB_TOK_NU t0 INNER JOIN HFJ_SPIDX_STRING t1 ON (t0.RES_ID = t1.RES_ID) WHERE ((t0.HASH_COMPLETE = '7545664593829342272') AND ((t1.HASH_NORM_PREFIX = '6206712800146298788') AND (t1.SP_VALUE_NORMALIZED LIKE 'JAY%'))) fetch first '10000' rows only";
+		String expected = "SELECT t0.RES_ID FROM HFJ_IDX_CMB_TOK_NU t0 INNER JOIN HFJ_SPIDX_STRING t1 ON (t0.RES_ID = t1.RES_ID) WHERE ((t0.HASH_COMPLETE = '2215689319713414397') AND ((t1.HASH_NORM_PREFIX = '6206712800146298788') AND (t1.SP_VALUE_NORMALIZED LIKE 'JAY%'))) fetch first '10000' rows only";
```
jamesagnew, line 334: Note to reviewer: These hashes changing seem like a breaking change but they actually aren't. 

I fixed a bug in `ITestDataBuilder` ([here](https://github.com/hapifhir/hapi-fhir/pull/7323/files#diff-6d0343ccfbbb4daec0fd141133c57ce747325ea019540e6ec82b6d5a4022a4ddR68)) which meant that primitive values on resources, such as `Patient.gender` and `Observation.status` were being created incorrectly without their enum factory, and therefore no system URL was set.

So with that fix, this change now represents the hashes that would actually be present if the resource was created normally through the REST API.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -2496,60 +2475,160 @@ private void attemptComboUniqueSpProcessing(
...
+									theComboParam,
+									nextComponent,
+									combinedNameParametersOrList)) {
+								continue;
+							}
+
+							inputs.add(combinedNameParametersOrList);
+							runnableRemoveInputsFromParams.add(
+									() -> combinedNameParametersAndList.remove(combinedNameParametersOrList));
+							foundMatch = true;
```
TipzCM, line 2573: do you want to break after this too?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -2496,60 +2475,160 @@ private void attemptComboUniqueSpProcessing(
...
+					if (!validateParamValuesAreValidForComboParam(
+							theRequest, theParams, theComboParam, nextComponent, sameNameParametersOrList)) {
+						continue;
+					}
+
+					inputs.add(sameNameParametersOrList);
+					runnableRemoveInputsFromParams.add(
+							() -> sameNameParametersAndList.remove(sameNameParametersOrList));
+					foundMatch = true;
+					break;
```
TipzCM, line 2543: we only care about adding this runnableRemoveInputsFromParams for the first occurrence?

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParameterCanonicalizer.java
```diff
@@ -425,11 +426,22 @@ private RuntimeSearchParam canonicalizeSearchParameterR4Plus(IBaseResource theNe
...
-			if (startsWith(definition, "/SearchParameter/")) {
+			if (Strings.CS.startsWith(definition, "/SearchParameter/")) {
 				definition = definition.substring(1);
 			}
 
-			components.add(new RuntimeSearchParam.Component(expression, definition));
+			String comboUpliftChain = null;
+			List<? extends IBaseExtension<?, ?>> componentExtensions = ((IBaseHasExtensions) next).getExtension();
+			for (IBaseExtension<?, ?> nextComponentExtension : componentExtensions) {
+				if (HapiExtensions.EXT_SP_COMBO_UPLIFT_CHAIN.equals(nextComponentExtension.getUrl())
```
TipzCM, line 436: Minor nitpick, but i typically prefer to do ifs with the 'faster' check first (`unique == ComboSearchParamType.NON_UNIQUE`); because if it's not that, then there's no reason to do the url compare.

But i am nitpicking here.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/util/JpaParamUtil.java
```diff
@@ -202,4 +200,106 @@ private static Class<?> getCompositeBindingClass(
...
+		 */
+		ComponentAndCorrespondingParam(
+				@Nonnull RuntimeSearchParam.Component theComponent, @Nonnull RuntimeSearchParam theComponentParameter) {
+			this.myComponent = theComponent;
+			this.myComponentParameter = theComponentParameter;
+
+			int dotIdx = theComponentParameter.getName().indexOf(".");
+			if (dotIdx != -1) {
+				myParamName = theComponentParameter.getName().substring(0, dotIdx);
+				myChain = theComponentParameter.getName().substring(dotIdx + 1);
```
TipzCM, line 251: Dumb question, but there's only ever 1 chain link, right?

`subject.gender=male` and not `composition.patient.identifier` (i know the latter is for bundles; maybe that's not applicable here)?

If it isn't applicable, maybe we should throw or log an error if there are additional '.'s in it (or is that handled already)?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
jamesagnew: Yeah - We only need one match for each component. I'll add more docs to make this more clear.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
jamesagnew: Good call, yes we do.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParameterCanonicalizer.java
jamesagnew: That seems like a false savings at the expense of readability. Look at the internal implementation of String.equals, it already does exactly that:

```java
    public boolean equals(Object anObject) {
        if (this == anObject) {
            return true;
        }
        return (anObject instanceof String aString)
                && (!COMPACT_STRINGS || this.coder == aString.coder)
                && StringLatin1.equals(value, aString.value);
    }
```

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/util/JpaParamUtil.java
jamesagnew: Additional chains are allowed, we only care about the first dot though. I'll add a test to verify that it works.

PR #7326, opened 2025-10-24
increase memory to elastic search container

jdar8:
Increase the allocated memory for the test Elastic Search container to 1GB due to failing test


robogary:
Formatting check succeeded!

PR #7321, opened 2025-10-22
batch2 jobs that are cancelled will not be returned when requesting a…

TipzCM:
closes #7320

PR #7327, opened 2025-10-24
batch2 jobs that are cancelled will not be returned when requesting a…

TipzCM:
…… (#7321)


robogary:
Formatting check succeeded!

PR #7337, opened 2025-10-28
7217 with rel 8 6

### LINKED ISSUE #7217: In a partitioned environment, searching for an existing non-partitionable Resource returns ResourceNotFound.
The issue happens when the requestDetails is a ServletRequestDetails.

TipzCM:
closes #7217

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -39,19 +39,16 @@ jobs:
 
       # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
       # If this step fails, then you should remove it and run the build manually (see below)
-      - name: Autobuild
-        uses: github/codeql-action/autobuild@v2
+      # - name: Autobuild
+      #   uses: github/codeql-action/autobuild@v2
 
       # ℹ️ Command-line programs to run using the OS shell.
       # 📚 https://git.io/JvXDl
 
-      # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
-      #    and modify them (or add more) to build your code if your project
-      #    uses a compiled language
-
-      #- run: |
-      #   make bootstrap
-      #   make release
+      # ✏️ Manual build to avoid Maven wrapper issues with CodeQL autobuild
+      - name: Build with Maven
+        run: |
+          mvn clean install -DskipTests -Dmaven.javadoc.skip=true -Dcheckstyle.skip=true -Dpmd.skip=true -Dspotbugs.skip=true -Denforcer.skip=true -Dlicense.skip=true -Drat.skip=true -Dspotless.check.skip=true
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@v2
--- a/hapi-deployable-pom/pom.xml
+++ b/hapi-deployable-pom/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		 <groupId>ca.uhn.hapi.fhir</groupId>
 		 <artifactId>hapi-fhir</artifactId>
-		 <version>8.5.8-SNAPSHOT</version>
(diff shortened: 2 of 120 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/GenerateRangeChunksStep.java
```diff
@@ -59,7 +65,7 @@ public RunOutcome run(
 			partitionedUrls.forEach(partitionedUrl -> {
 				ChunkRangeJson chunkRangeJson = new ChunkRangeJson(start, end)
 						.setUrl(partitionedUrl.getUrl())
-						.setPartitionId(partitionedUrl.getRequestPartitionId());
```
volodymyr-korzh, line 62: partitionedUrl.getRequestPartitionId() is used in several places as well:

`ca.uhn.fhir.batch2.jobs.expunge.DeleteExpungeJobParametersValidator#validate`
`ca.uhn.fhir.batch2.coordinator.DefaultJobPartitionProvider#getPartitionedUrls`

Should we reuse `getRequestPartitionIdForUrl`  in these places as well?

volodymyr-korzh (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/partition/RequestPartitionHelperSvcTest.java
```diff
@@ -79,6 +83,34 @@ public void testDetermineReadPartitionForSystemRequest_withPartitionIdOnly_retur
...
+		testDetermineReadPartitionRequest_whenResourceIsNonPartitionable_returnsDefaultPartition(srd);
+	}
+
+	public void testDetermineReadPartitionRequest_whenResourceIsNonPartitionable_returnsDefaultPartition(RequestDetails theDetails) {
+		// execute
+		RequestPartitionId result = mySvc.determineReadPartitionForRequestForSearchType(theDetails, "Library");
+
+		// verify
+		assertThat(result.hasDefaultPartitionId(DEFAULT_PARTITION_ID)).isTrue();
+	}
```
tadgh: suggestion: unroll this into the related tests please. This method increases cognitive load, does not reduce it. Method is also named `test` but is not a test. If this "verifying" method was doing something interesting or complex, it may be worth extracting it into a custom assertJ assert.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/jobs/step/GenerateRangeChunksStep.java
TipzCM: This isn't really an easy question to answer.

It's not so much that `getRequestPartitionIdForUrl` is wrong. This is just a POJO after all. It can be set to... whatever.

Problem is, our default partitionId == null. And if we're not in named partition mode, we then create a partition id object with a list of null names (because there are no names).

Unfortunately, when validating the names (which are all null) we expect them not to be null. Or we throw.

----

I've pushed teh code downstream further now, so when we do get null names we just... skip them. Which has the validator then assume it's a default partition and return teh correct value.

I'm... not sure if this is correct. We'll see if the tests fail.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/partition/RequestPartitionHelperSvcTest.java
TipzCM: changed

Issue #7338, opened 2025-10-29
Bug: Validation with unknown profile cause error even when setting FhirInstanceValidator.setErrorForUnknownProfiles(false)


ttnTrifork:
### The Problem
It seems that the fix for https://github.com/hapifhir/hapi-fhir/issues/6607 introduced a new validation error, but does not check if the validator is configured to give error for unknown profiles or not.

The https://github.com/hapifhir/hapi-fhir/issues/6607 fix was ensuring that the [$validate](https://www.hl7.org/fhir/resource-operation-validate.html) operation SHALL return an error if the resource could not be validated against the input `profile` parameter, which is fair. 

But this new validation error is also produced when creating/updating a resource if a RepositoryValidatingInterceptor (with a RequireValidationRule)  or a RequestValidatingInterceptor (that adds the FhirInstanceValidator as a validator module) is registered on the server. 

This means that now the create/update request fails and the resource is not persisted even if having configured `FhirInstanceValidator.setErrorForUnknownProfiles(false)` on the "myInstanceValidator" IInstanceValidatorModule Spring bean. 
Prior to HAPI 8.2 this worked as expected.

See https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/ValidatorWrapper.java#L262-L266
`m.setLevel(ValidationMessage.IssueSeverity.ERROR);`

### The Solution
For other profile errors, ValidatorWrapper consults the `myErrorForUnknownProfiles` boolean property (which comes from the FhirInstanceValidator property with same name).

I am happy to open a PR that adds a check for the property when setting the severity of the new validation message, while still ensuring `$validate` returns an error severity issue if target profile is unknown, regardless of the value of `myErrorForUnknownProfiles`

Issue #7117, opened 2025-07-10
Validator property errorForUnknownProfiles is not respected


daersc:
**Describe the bug**
Validating a resource with unknown profile leads to an error, even if `errorForUnknownProfiles=false`.

Consider a variant of [`FhirInstanceValidatorR4Test#testValidateResourceContainingProfileDeclarationDoesntResolve`](https://github.com/hapifhir/hapi-fhir/blob/8d42e54d9ff789272214a5307795a7f7bbf75a73/hapi-fhir-validation/src/test/java/org/hl7/fhir/r4/validation/FhirInstanceValidatorR4Test.java#L1117) with a single difference in test set up, that `errorForUnknownProfiles` property is set to `false`.

I would expect, that there are no errors at all in this case, leading to some test like the following:
```java
@Test
public void testValidateResourceContainingProfileDeclarationDoesntResolveAndErrorForUnknownProfilesIsFalse() {
	myMockSupport.addValidConcept("http://loinc.org", "12345");

	Observation input = createObservationWithDefaultSubjectPerfomerEffective();

	input.getText().setDiv(new XhtmlNode().setValue("<div>AA</div>")).setStatus(Narrative.NarrativeStatus.GENERATED);
	input.getMeta().addProfile("http://foo/structuredefinition/myprofile");

	input.getCode().addCoding().setSystem("http://loinc.org").setCode("12345");
	input.setStatus(ObservationStatus.FINAL);

	myInstanceVal.setValidationSupport(myValidationSupport);
	myInstanceVal.setErrorForUnknownProfiles(false);
	ValidationResult output = myFhirValidator.validateWithResult(input);
	List<SingleValidationMessage> errors = logResultsAndReturnNonInformationalOnes(output);

	assertThat(errors).isNotEmpty();
	assertThat(errors.stream())
		.noneMatch(r -> r.getSeverity() == ResultSeverityEnum.ERROR);
}
```
But the test fails since there is still a message of `ERROR` severity with message
> Invalid profile. Failed to retrieve profile with url=http://foo/structuredefinition/myprofile

**Expected behavior**
The assertions hold, that none of the validation messages have `ERROR` severity.

**Environment (please complete the following information):**

- HAPI FHIR 8.2.0

daersc:
The `errorForUnknownProfiles` property currently is implemented in

[...]

PR #7339, opened 2025-10-29
fix(7338): Validation with unknown profile no longer cause error when setting FhirInstanceValidator.setErrorForUnknownProfiles(false)

ttnTrifork:
This fixes 
- https://github.com/hapifhir/hapi-fhir/issues/7338
- https://github.com/hapifhir/hapi-fhir/issues/7117

- ValidatorWrapper now adheres to its property 'errorForUnknownProfiles' when validating a resource with an unknown profile 

- Updated BaseHapiFhirResourceDao.validate to ensure `$validate` returns an error severity issue if target profile is unknown, regardless of the value of the ValidatorWrapper `errorForUnknownProfiles` property.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7338-fix-fhir-instance-validator-error-for-unknown-profiles.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7338-fix-fhir-instance-validator-error-for-unknown-profiles.yaml
@@ -1,6 +1,6 @@
 ---
 type: fix
 issue: 7338
-title: "A regression in HAPI FHIR 8.2 meant that setting FhirInstanceValidator.setErrorForUnknownProfiles(false) had no effect.
+title: "A regression in HAPI FHIR 8.2 meant that setting FhirInstanceValidator.setErrorForUnknownProfiles(false) had no effect. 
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

tadgh:
Heya, I appreciate the PR! I do have some concerns about it though. This specifically will change existing behaviour of the system in a non-backwards-compatible way. My understanding is that this PR makes $validate ignore the validator setting. I understand that this is what the spec says, but we prefer not to break existing behaviour unless absolutely necessary. 

Is there a reason that you cannot just set this setting to true, and have the behaviour you intend? 

Forgive me if I'm missing some nuance, validation is not my area of specialization.

tadgh:
> Heya, I appreciate the PR! I do have some concerns about it though. This specifically will change existing behaviour of the system in a non-backwards-compatible way. My understanding is that this PR makes $validate ignore the validator setting. I understand that this is what the spec says, but we prefer not to break existing behaviour unless absolutely necessary.
> 
> Is there a reason that you cannot just set this setting to true, and have the behaviour you intend?
> 
> Forgive me if I'm missing some nuance, validation is not my area of specialization.

Having now read the original issue (sorry I missed it originally), I see the reasoning much more clearly now. Approved pending the addition of a changelog in 8_8_0 changelogs folder.

tadgh (approved):
Approved pending a changelog

ttnTrifork:
Thanks! I updated https://github.com/hapifhir/hapi-fhir/issues/7338 a bit so its easier to read

ttnTrifork:
@tadgh I added a change log. 
Can we merge this now?

PR #7351, opened 2025-11-01
Bump to spring62

tadgh:
- Bump from springframework 6.1 -> 6.2 
- Replace usages of removed Base64Utils with equivalent java.util.Base64
- Correctly insert `@BeforeEach` and `@AfterEach` on each overridden method that was missing it. New versions of Junit have changed method discovery, per:  https://github.com/junit-team/junit-framework/issues/3553


robogary:
Formatting check succeeded!

PR #7364, opened 2025-11-04
Fix validation post processing interceptors

nathandoef:
Previously, if multiple validation post-processing interceptors (`ValidationMessageSuppressingInterceptor`, `ValidationMessagePostProcessingInterceptor`, or `ValidationMessageUnknownCodeSystemPostProcessingInterceptor`) 
were registered that modified a `ValidationResult`, only the first interceptor was executed. This has been fixed.

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/validation/ValidationMessagePostProcessingInterceptorTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/validation/ValidationMessagePostProcessingInterceptorTest.java
@@ -11,6 +11,7 @@ import org.junit.jupiter.api.Test;
 import java.util.List;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/validation/ValidationResult.java
```diff
@@ -62,6 +62,10 @@ public List<SingleValidationMessage> getMessages() {
 		return Collections.unmodifiableList(myMessages);
 	}
 
+	public void setMessages(List<SingleValidationMessage> theMessages) {
+		myMessages = theMessages;
```
jamesagnew, line 68: Recommend chucking a `Validate.notNull(theMessages, "theMessages must not be null");` in here just to be consistent with the constructor (which will NPE if you pass in null, and getMessages() which will also fail if we're null)

nathandoef: Ahh thanks good call. Fix added here: db0f5fcf9457613c65f6019069348fa602be0026

Issue #7284, opened 2025-10-03
Long time for first pre-expand operation for large value set


SevcikMichal:
**Describe the issue**
The pre expansion of larger ValueSets happens only after 10 minutes after startup.
Is it such an expensive operation?
Is there any reason it is implemented like this? In additional context I linked the code that I believe is the reason for this.
Would something break if this schedule was configurable?
I could implement this if it would be fine for it to be configurable.

**Environment (please complete the following information):**

- HAPI FHIR Version 8.4.0
- OS: linux (docker container)
- Browser N/A

**Additional context**
Code that I think is relevant here:
https://github.com/hapifhir/hapi-fhir/blob/eaa02d19dd7d979a9122bba862f4ffed1e6b99bb/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java#L2449C2-L2456C3

SevcikMichal:
Created a proposal for solution in #7370

SevcikMichal:
Hello,

is there any feedback on this?

PR #7370, opened 2025-11-10
Make value set expansion job schedule configurable

SevcikMichal:
Proposal for solution for #7284


robogary:
Formatting check succeeded!

PR #7372, opened 2025-11-10
fixed $translate operation caching

### LINKED ISSUE #7371: Remote Terminology $translate calls not being cached
The $translate operation calls are not cached, i.e. if a request is repeated twice both result in request to the remote terminology server.

mrdnctrk:
closes #7371

## Problem

The ConceptMap `$translate` operation caching was not working properly due to an issue with the cache key comparison. The `TranslateCodeRequest` class was using default object equality for `IBaseCoding` instances, which meant that two semantically identical codings (same system, code, and version) but different object instances were treated as different cache keys. This resulted in unnecessary cache misses.

## Solution
Implemented a `CodingEqualsAdapter` inner class in `TranslateCodeRequest` class. THis provides semantic equality comparison for `IBaseCoding` instances. The adapter compares codings based on their semantic content:
- System
- Code
- Version

The display value is intentionally excluded from the comparison as it is not semantically significant for code translation operations.

### DIFF
--- a/hapi-fhir-validation/src/test/java/ca/uhn/fhir/context/support/TranslateCodeRequestTest.java
+++ b/hapi-fhir-validation/src/test/java/ca/uhn/fhir/context/support/TranslateCodeRequestTest.java
@@ -1,0 +1,189 @@
+package ca.uhn.fhir.context.support;
+
+import org.hl7.fhir.r4.model.Coding;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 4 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-validation/src/test/java/ca/uhn/fhir/context/support/TranslateCodeRequestTest.java
```diff
@@ -0,0 +1,189 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+/**
+ * Tests for {@link ca.uhn.fhir.context.support.IValidationSupport.TranslateCodeRequest}
+ * which is used as a cacheKey for $translate operations.
+ */
+class TranslateCodeRequestTest {
+
+	@Test
+	void testEqualsImpliesHashEqual() {
```
jmarchionatto: nitpick: I think you meant Involves instead of Implies

mrdnctrk: thanks, updated the name.

PR #7388, opened 2025-11-17
Fix Encounter $everything not working in REQUEST_TENANT partitioning

jmarchionatto:


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoEncounter.java
```diff
@@ -63,17 +65,26 @@ public IBundleProvider encounterInstanceEverything(
 		if (theId != null) {
 			paramMap.add("_id", new StringParam(theId.getIdPart()));
 		}
-		IBundleProvider retVal = search(paramMap);
```
jamesagnew, line 66: Suggestion, could we avoid having to copy a bunch of the `search` method here by instead just calling `search(SearchParameterMap, RequestDetails, HttpServletResponse)`?

jmarchionatto: I was able to remove the search registration, but I can't remove parameters handling as those are received as method parameters and parent function would extract them from request.

PR #7391, opened 2025-11-18
7390: call pointcut before updating resource version

another-luis:


robogary:
Formatting check succeeded!

PR #7392, opened 2025-11-18
7386 icd10cmloader is not creating required extensions for concepts

IanMMarshall:
Fixed logic in Icd10CmLoader to append 7th character extensions to the lowest-level child concepts in the hierarchy (rather than to just the next level child concepts).


robogary:
Formatting check succeeded!

PR #7395, opened 2025-11-19
ValueSet code case sensitivity

fil512:
This MR adds a new configuration setting that allows database-stored terminology resources (CodeSystem, ValueSet) to take precedence over built-in default profile validation resources during validation. This addresses scenarios where users need to store custom CodeSystems that override HL7 defaults.

1. Added new `JpaStorageSettings.setAllowDatabaseValidationOverride(boolean)` setting that controls the order of validation support modules in the chain
2. Modified `JpaValidationSupportChain.postConstruct()` to conditionally reorder the JPA validation support before or after the default profile validation support based on the new setting

## Code Review Suggestions

- [ ] Verify that changing the validation support chain order does not have unintended side effects on other validation scenarios beyond the case-sensitivity use case
- [ ] Consider whether the setting name `allowDatabaseValidationOverride` clearly conveys what it does - it specifically changes the order of validation support modules rather than a broad override mechanism
- [ ] Check if there are other places in the codebase that assume a specific order of validation support modules in the chain

### Setup

1. Deploy a HAPI FHIR JPA server with the new code
2. Ensure the server has access to create and manage CodeSystem and ValueSet resources
3. Prepare a test CodeSystem that uses the same URL as a built-in HL7 CodeSystem (e.g., `http://terminology.hl7.org/CodeSystem/insurance-plan-type`)

### Test Cases

(See original bug report for reproduction steps)

- [ ] **Default behavior (setting disabled)**: Create a CodeSystem with the same URL as a built-in HL7 CodeSystem but with different case sensitivity. Verify that ValueSet expansion uses the built-in definition's case sensitivity, not the database version.

- [ ] **Override enabled**: Enable `allowDatabaseValidationOverride=true`, create a case-sensitive CodeSystem with codes that differ only by case (e.g., "drug" and "Drug"). Verify that ValueSet expansion includes both case variants when the database CodeSystem takes precedence.

- [ ] **Validation behavior**: With the override enabled, validate a resource against a ValueSet that references the overridden CodeSystem. Verify that validation uses the database CodeSystem's properties.

- [ ] **Setting change at runtime**: If the setting is changed after server startup, verify behavior is consistent (note: may require server restart or cache invalidation).

- [ ] **FRAGMENT content mode**: Test specifically with CodeSystems using FRAGMENT content mode, as this is the scenario where users need to supplement built-in definitions with additional codes.

### DIFF
--- a/hapi-deployable-pom/pom.xml
+++ b/hapi-deployable-pom/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		 <groupId>ca.uhn.hapi.fhir</groupId>
 		 <artifactId>hapi-fhir</artifactId>
-		 <version>8.7.0-SNAPSHOT</version>
+		 <version>8.7.1-SNAPSHOT</version>
 
 		 <relativePath>../pom.xml</relativePath>
 	 </parent>
--- a/hapi-fhir-android/pom.xml
+++ b/hapi-fhir-android/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.7.0-SNAPSHOT</version>
+		<version>8.7.1-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-base/pom.xml
+++ b/hapi-fhir-base/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.7.0-SNAPSHOT</version>
+		<version>8.7.1-SNAPSHOT</version>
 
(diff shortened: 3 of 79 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7394-valueset-code-case-sensitivity.yaml
```diff
@@ -0,0 +1,7 @@
+---
+type: add
+issue: 7394
+title: "A new setting JpaStorageSettings.setAllowDatabaseValidationOverride(boolean) has been added.
+  When enabled, database-stored CodeSystem and ValueSet resources take precedence over built-in
```
jmarchionatto, line 5: Is the precedence only for validation purposes or any request?
If it is only for validation, it could be indicated here.

fil512: good catch!

PR #7400, opened 2025-11-22
Updated plugins and GitHub actions

madduci:
Updated Plugins and GitHub Actions

Resolving CVE for logback


robogary:
Formatting check succeeded!

tadgh:
Looks good, minus potentially the change to UCUM version, and the spotless failures. 
```
		<ucum_version>1.0.10</ucum_version>
```

Was there a CVE you were resolving with this? I just ask because there may be downstream issues from this update, depending on the contents of the change. I'll have a look at the delta

madduci:
> 10

i reverted back the `ucum` version, i was not aware of the problem in the past

tadgh:
Seems to fail on the parallel build step here. Very strange

madduci:
> Seems to fail on the parallel build step here. Very strange

The issue is that the Step runs the command 

```sh
mvn validate -P CHECKSTYLE --batch-mode
```

but it requires first that the SNAPSHOT Packages are installed (the cache does not contain them yet?)

tadgh:
This apparently breaks our checkstyle run: https://github.com/hapifhir/hapi-fhir/actions/runs/19602202032/job/56135347335?pr=7400 . Can you have a look into the file and see why this might be? looks like its failing to do some resolution. If you can't figure it out let me know and I'll have a look on monday.

madduci:
@tadgh  i've tested it on another PC and i could fully reproduce it.

In order to break the loop, i had to issue the command:

`mvn clean install -DskipTests -f hapi-fhir-checkstyle` 

I couldn't use the dependency, if it not even exists in my repository. The `8.7.0-SNAPSHOT` has never been installed in the cache so far

image

madduci:
@tadgh  any news about it?

PR #7407, opened 2025-11-25
[7401] Smileutil upload-terminology command does not load all concept properties from CSV file

### LINKED ISSUE #7401: Smileutil upload-terminology command does not load all concept properties from CSV file
The `smileutil upload-terminology` command does not support loading more than one `CodeSystem.concept.property` value to the `TRM_CONCEPT_PROPERTY` database table with the same `CodeSystem.concept.property.code` for a given `CodeSystem.concept`.

```
{
  "resourceType" : "CodeSystem",
  "id" : "mycodesystem",
  "url" : "http://example.com/mycodesystem",
  "content" : "complete",
  "concept" : [{
    "code" : "code-a",
    "display" : "Code A",
    "property" : [{
      "code" : "alt_description",
      "valueString" : "code a"
    },{
      "code" : "alt_description",
      "valueString" : "CODE A"
    }]
  }]
}
```

If the above is loaded through a FHIR endpoint via a POST or PUT request, Smile CDR creates two rows in the `TRM_CONCEPT_PROPERTY` table.

Alternately, if this same code system is loaded using `smileutil upload-terminology`, only one row is created.

This is problematic as some ValueSet definitions use `concept.property` values to define their contents.

YalingPeiS:
### Issue
The `smileutil upload-terminology` command does not support loading more than one `CodeSystem.concept.property` value to the `TRM_CONCEPT_PROPERTY` database table with the same `CodeSystem.concept.property.code` for a given `CodeSystem.concept`.

```
{
  "resourceType" : "CodeSystem",
  "id" : "mycodesystem",
  "url" : "http://example.com/mycodesystem",
  "content" : "complete",
  "concept" : [{
    "code" : "code-a",
    "display" : "Code A",
    "property" : [{
      "code" : "alt_description",
      "valueString" : "code a"
    },{
      "code" : "alt_description",
      "valueString" : "CODE A"
    }]
  }]
}
```

If the above is loaded through a FHIR endpoint via a POST or PUT request, Smile CDR creates two rows in the `TRM_CONCEPT_PROPERTY` table. However, only one row is created if loaded using `smileutil upload-terminology`.

This is problematic as some ValueSet definitions use `concept.property` values to define their contents.

### Cause
During concept property processing, the existing `TermConceptProperty` was updated when the code exists in the property list which caused its value being overwritten, resulting in loss of the all the previous property entries with the same code.

### Fix
Create a `TermConceptProperty` object for each property.  

Closes #7401

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/custom/PropertyHandler.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/custom/PropertyHandler.java
@@ -23,15 +23,21 @@ import ca.uhn.fhir.jpa.entity.TermConceptProperty;
 import ca.uhn.fhir.jpa.entity.TermConceptPropertyTypeEnum;
 import ca.uhn.fhir.jpa.term.IZipContentsHandlerCsv;
 import org.apache.commons.csv.CSVRecord;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.util.CollectionUtils;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.apache.commons.lang3.StringUtils.trim;
 
 public class PropertyHandler implements IZipContentsHandlerCsv {
+
+	private static final Logger ourLog = LoggerFactory.getLogger(PropertyHandler.class);
 
 	public static final String CODE = "CODE";
 	public static final String KEY = "KEY";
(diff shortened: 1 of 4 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/custom/PropertyHandler.java
```diff
@@ -50,18 +48,14 @@ public void accept(CSVRecord theRecord) {
...
 
 			List<TermConceptProperty> conceptProperties = myCode2Properties.get(code);
 			if (conceptProperties == null) conceptProperties = new ArrayList<>();
 
-			TermConceptProperty conceptProperty =
-					TermLoaderSvcImpl.getOrCreateConceptProperty(myCode2Properties, code, key);
-			ValidateUtil.isNotNullOrThrowUnprocessableEntity(
-					conceptProperty, "Concept property %s not found in file", conceptProperty);
-
+			TermConceptProperty conceptProperty = new TermConceptProperty();
```
IanMMarshall, line 74: This makes sense, but I have to wonder if this would leave open the possibility of duplicated concept property records in the database, i.e. two or more Concept Property entities with the same "code", "key", "value" and "type" (although currently in this case "type" is always going to be STRING). 

In the ConceptHandler, we have a check for a duplicated "code" values. Would it be possible to include a check here to ensure that the new property has a unique combination of "code", "key", "value", and "type"?

The check would not necessarily have to result in an error. My goal would be mainly to avoid creating unnecessary duplicates in the DB.

We might also want to add a test case for this possibility as well.

YalingPeiS: Updated with duplicate check and added test for it.

IanMMarshall: Nice! Looks good.

PR #7408, opened 2025-11-25
Ehp 2025.10.30 package registry poc

### LINKED ISSUE #7420: update validation support to return multiple resources (as it should) for a provided url
IValidationSupport returns only the first resource found if provided a resourceType and a url.

But multiple resources could exist (across multiple IGs).

all of these should be returned.

TipzCM:
closes #7420


robogary:
Formatting check succeeded!

PR #7415, opened 2025-11-27
Inprove FhirTerser.visit and date searches

jamesagnew:
This PR addresses several issues:

* FhirTerser.visit(IBase, IModelVisitor2) does not consistently visit extensions (it does visit them when visiting composite elements, but does not visit them when visiting primitive elements directly)
* IModelVisitor2 has a redundant `visitExtension` method which is only called for DSTU2 extensions. This method has been deprecated since `visit` covers both.
* The JPA server incorrectly interprets a simple date search (e.g. `Patient?birthdate=2020-01-01`) as a range with the same lower and upper bound, which causes an assertion banning duplicate parameters to fail when performing this search with assertions enabled (this bug does not affect correctness of results when assertions are not enabled)

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
@@ -1311,6 +1311,10 @@ public class FhirTerser {
 		}
 
 		BaseRuntimeElementDefinition<?> def = myContext.getElementDefinition(theElement.getClass());
+
+		// The following method calls itself recursively to descend through and visit each child element
+		// so we call it here with empty lists that get added-to and removed-from as we walk into and out
+		// of the tree
 		visit(theElement, null, def, theVisitor, new ArrayList<>(), new ArrayList<>(), new ArrayList<>());
 	}
 
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
@@ -518,6 +518,11 @@ public class SearchCoordinatorSvcImpl implements ISearchCoordinatorSvc<JpaPid> {
 	}
 
 	private void validateSearch(SearchParameterMap theParams) {
+		/*
+		 * Having duplicate identical params in the search (e.g. Patient?gender=male&gender=male) is not
+		 * technically wrong, but it's inefficient and can slow query execution down. Checking for it also
+		 * adds CPU load itself though, so we only check this in an assert to hopefully catch errors in tests.
+		 */
 		assert checkNoDuplicateParameters(theParams)
 				: "Duplicate parameters found in query: " + theParams.toNormalizedQueryString();
 
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/consent/ConsentInterceptor.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/consent/ConsentInterceptor.java
@@ -461,6 +461,13 @@ public class ConsentInterceptor {
 				if (theElement == outerResource) {
 					return true;
 				}
+
+				// Primitive elements can't contain any embedded resources, so we don't need to
+				// descend into them (and any extensions they might hold)
+				if (theElement instanceof IPrimitiveType<?>) {
+					return false;
+				}
+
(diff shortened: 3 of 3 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
```diff
@@ -1297,20 +1299,19 @@ public Map<Object, Object> newMap() {
...
 					Collections.emptyList(),
 					Collections.emptyList(),
 					Collections.emptyList());
-		} else {
-			theVisitor.acceptElement(
-					theElement, Collections.emptyList(), Collections.emptyList(), Collections.emptyList());
 		}
+
+		BaseRuntimeElementDefinition<?> def = myContext.getElementDefinition(theElement.getClass());
+		visit(theElement, null, def, theVisitor, new ArrayList<>(), new ArrayList<>(), new ArrayList<>());
```
jmarchionatto, line 1318: A comment explaining why visit is called directly could be appropriate.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
```diff
@@ -518,7 +519,7 @@ private Integer getSynchronousMaxResultsToFetch(SearchParameterMap theParams, In
 
 	private void validateSearch(SearchParameterMap theParams) {
 		assert checkNoDuplicateParameters(theParams)
-				: "Duplicate parameters found in query: " + theParams.toNormalizedQueryString(myContext);
+				: "Duplicate parameters found in query: " + theParams.toNormalizedQueryString();
```
jmarchionatto, line 527: If this assert produced an inconsistency already, wouln't be better replaced by a runtime validation?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4Test.java
```diff
@@ -837,6 +837,32 @@ private int createPatientsWithReferences(Map<Integer, Integer> theRefCountToReso
 		return totalNumberOfPatientsCreated;
 	}
 
+	/**
+	 * Ensure that a search for a specific date (which is transformed internally into a
+	 * date range with an identical lower and upper by PatientResourceProvider) doesn't
+	 * cause duplicate parameters.
+	 */
+	@Test
+	public void testSearchDate() {
```
jmarchionatto, line 846: Nice test!

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/consent/ConsentInterceptor.java
```diff
@@ -503,6 +504,9 @@ public boolean acceptElement(
 
 					return shouldCheckChildren;
 				}
+				if (theElement instanceof IPrimitiveType<?>) {
```
jmarchionatto: I would have put this code before the preceding conditional block, but of course I may be misunderstanding the whole thing because this method is unreadable.

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/FhirTerser.java
jamesagnew: Will add

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImpl.java
jamesagnew: Added the following explanation: 

```java
/*
 * Having duplicate identical params in the search (e.g. Patient?gender=male&gender=male) is not
 * technically wrong, but it's inefficient and can slow query execution down. Checking for it also
 * adds CPU load itself though, so we only check this in an assert to hopefully catch errors in tests.
 */
```

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/consent/ConsentInterceptor.java
jamesagnew: It's equivalent whether it's at the top or the bottom since none of the other conditions could match if it's a primitive element. Agree though that it's more clear to put it at the top. I've moved it and added a comment

PR #7425, opened 2025-12-08
fix(validation): return HTTP 400 instead of 500 for malformed JSON/XML

JorisHeadease:
Previously, sending malformed JSON to the validator caused an uncaught `JsonSyntaxException` resulting in HTTP 500. Malformed XML returned HTTP 422 via FATAL validation message.

Now both throw `InvalidRequestException` returning HTTP `400 Bad Request`, which is correct for syntax errors. HTTP `422 Unprocessable Entity` is reserved for syntactically valid input that fails semantic validation.


robogary:
Formatting check succeeded!

JorisHeadease:
Codes added to https://github.com/hapifhir/hapi-fhir/wiki/Bumping-Error-Code

PR #7429, opened 2025-12-10
7390 - Make resource version-before-delete available to subscription interceptor

### LINKED ISSUE #7390: Subscriptions - delivering Delete events with specific criteria fails to match the resource
It works but only when the criteria is ‘Patient?’ and not ‘Patient?identifier=ABC123’

another-luis:
Fixes #7390

### DIFF
--- a/hapi-deployable-pom/pom.xml
+++ b/hapi-deployable-pom/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		 <groupId>ca.uhn.hapi.fhir</groupId>
 		 <artifactId>hapi-fhir</artifactId>
-		 <version>8.7.2-SNAPSHOT</version>
+		 <version>8.7.3-SNAPSHOT</version>
 
 		 <relativePath>../pom.xml</relativePath>
 	 </parent>
--- a/hapi-fhir-android/pom.xml
+++ b/hapi-fhir-android/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.7.2-SNAPSHOT</version>
+		<version>8.7.3-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-base/pom.xml
+++ b/hapi-fhir-base/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.7.2-SNAPSHOT</version>
+		<version>8.7.3-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-bom/pom.xml
+++ b/hapi-fhir-bom/pom.xml
@@ -4,7 +4,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>ca.uhn.hapi.fhir</groupId>
 	<artifactId>hapi-fhir-bom</artifactId>
-	<version>8.7.2-SNAPSHOT</version>
+	<version>8.7.3-SNAPSHOT</version>
 
 	<packaging>pom</packaging>
 	<name>HAPI FHIR BOM</name>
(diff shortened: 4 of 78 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
```diff
@@ -1851,6 +1851,7 @@ public enum Pointcut implements IPointcut {
 	STORAGE_PRECOMMIT_RESOURCE_DELETED(
 			void.class,
 			"org.hl7.fhir.instance.model.api.IBaseResource",
```
tadgh, line 1853: question: You indicated that the resource that was deleted "does not have a body" when it gets to the subscription processor. Does this IBaseResource not contain the body?

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7390-fix-subscription-not-firing-on-delete-resource.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 7390
+title: "Subscriptions were not triggering on delete when subscription criteria uses resource attributes. This has been fixed"
```
tadgh:
```suggestion
title: "Subscriptions were not sending `DELETE` events when the criteria used resource attributes, for example `Patient?gender=male`. This has been fixed."
```

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/submit/interceptor/SubscriptionMatcherInterceptor.java
```diff
@@ -85,8 +86,11 @@ public void resourceCreated(IBaseResource theResource, RequestDetails theRequest
 	}
 
 	@Hook(Pointcut.STORAGE_PRECOMMIT_RESOURCE_DELETED)
-	public void resourceDeleted(IBaseResource theResource, RequestDetails theRequest) {
+	public void resourceDeleted(IBaseResource theResource, RequestDetails theRequest, IIdType previousId) {
 
+		// When matching for deleted resources, we want to match for the version before it was deleted because
+		// the resource version record for the deleted resource doesn't contain a body.
+		theResource.setId(previousId);
```
tadgh, line 93: I don't understand how this fix actually corrects the issue. This is just setting the (I assume versioned) ID, onto the IBaseResource. Can you copy-paste me the contents of `theResource` as seen at this point in time?

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
another-luis: The resource has a body while being handled by the hook, but when the ResourceModifiedMessage is created based on resource, the body is dropped by calling getPayloadLessMessageFromString(..). So it will up to the message handler to reconstitute the original resource by using the version included in the message. If the message points to the deleted version of the resource, it will retrieve that one which - in turn - has no body.

tadgh: gotcha. so it has no actual clinical information in it. It is literally an empty resource?

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/submit/interceptor/SubscriptionMatcherInterceptor.java
another-luis: The ResourceModifiedMessage is a payloadless message, so only the id survives. When picked up by the message handler, the resource is read from the cache/db. If the version retrieved is the deleted version, it will be bodiless, so the resource fields needed by the matching criteria won't be there. We have to retrieve the version right before it was deleted to "inflate" a resource with all its fields populated so that the matching criteria has a fighting chance. That's why I need to set the version to the one before the resource was deleted.

Take a look at theResource **before** I set the id in the next line. It's pointing to version 2 - the deleted version. If we let it continue, that's going to be in the contents of the ResourceModifiedMessage, and the message handler will inflate that (bodiless) version. I had to break the resource into two images because I can't figure out how to zoom out in the debugging window .

Screenshot 2025-12-11 at 5 03 33 PM

Screenshot 2025-12-11 at 5 04 33 PM

tadgh: Ah I gotcha

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
another-luis: yes

PR #7441, opened 2025-12-18
ConceptMap doesn't correctly use sourceVersion

### LINKED ISSUE #7436: ConceptMap doesn't correctly use sourceVersion
Given that a CodeSystem has the same CodeSystem.url stored but they differ in CodeSystem.version, e.g., one being 2024-04-11 and the other 2024-09-11 and the latter added after the former. Note that these are not versions of the same CodeSystem resource. 

When the request validating on FHIR endpoint is enabled, creating a ConceptMap uses source and sourceVersion to identify mappings that should relate to the 2024-04-11 version as specified in sourceVersion, the response incorrectly contains the error 

`The source code 'EC64411004' is not valid in the code system [https://terminology.foobar.com.au/CodeSystem/foobar-pathology-local-order-codes|2024-09-11](https://terminology.foobar.com.au/CodeSystem/foobar-pathology-local-order-codes%7C2024-09-11)}`. 

It is referencing the wrong CodeSystem.version of the code system.

YalingPeiS:
### Problem Statement
Given that a CodeSystem has the same CodeSystem.url stored but they differ in CodeSystem.version, e.g., one being 2024-04-11 and the other 2024-09-11 and the latter added after the former. Note that these are not versions of the same CodeSystem resource.

When the request validating on FHIR endpoint is enabled, creating a ConceptMap uses source and sourceVersion to identify mappings that should relate to the 2024-04-11 version as specified in sourceVersion, the response incorrectly contained errors and referenced the wrong CodeSystem.version of the code system

### Cause
During CodeSystem validation, the source version was not considered for the search. Since the search only uses the URL parameter, it loads the latest version of the CodeSystem stored regardless of the specified source version.

### Fix
For CodeSystem validation, make sure canonical URL containing version as well as the system URL is used in search.

Closes #7436

### DIFF
--- a/hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapterTest.java
+++ b/hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapterTest.java
@@ -230,7 +230,7 @@ public class WorkerContextValidationSupportAdapterTest extends BaseValidationTes
 	}
 
 	@Test
-	public void fetchCodeSystem_withVersion() {
+	public void testFetchCodeSystem_withVersion() {
 		// setup
 		String systemUrl = "http://codesystems.com/system";
 		String version = "v1";


robogary:
Formatting check succeeded!

jamesagnew (approved):

PR #7449, opened 2026-01-06
#7448 Fix _total=accurate by counting resultIterator iterations and prefetching all if an accurate count is requested

### LINKED ISSUE #7448: _total=accurate is not the same as the total amount of results in the result set

**Describe the bug**
The total count returned with a search when we add `_total=accurate` query parameter does not always match the number of search results. We use the total count as part of a bulk data export workflow where we pre-compute all the page requests for the search result set based off the search results id returned by the initial total count query.

**To Reproduce**

1. Run a workflow that is making changes to fhir resources.
2. execute a search with _total=accurate which will overlap with the resources being changed or added.
3. the total count in the search results should always equal the number of resources in the search once you paginate through them all.

**Expected behavior**
The total count from _total=accurate always equals the number of results in the result set.

**Screenshots**

- HAPI FHIR Version 8.2.0
- OS: [e.g. iOS] Linux
- Browser [e.g. chrome, safari] Hapi FHIR HTTP Client

**Additional context**

esecules:
Search results totalCount is set based on the number of items in the results iterator so that it always equals the number of results in the search results set. To make this work, we'll prefetch the entire search if an accurate count is requested along with the search.

If the user requested only a count and nothing else, then we still execute the dedicated COUNT query.

While the total count will always be accurate, Im not sure whether this would be too slow of an approach, particularly if the search would cover a very large number of results.

Fixes #7448

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/tasks/SearchTask.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/tasks/SearchTask.java
@@ -571,7 +571,7 @@ public class SearchTask implements Callable<Void> {
 	@SuppressWarnings({"rawtypes", "unchecked"})
 	private void doSearch() {
 		/*
-		 * If the user has explicitly requested a _count, perform a
+		 * If the user has explicitly only requested a _count, perform a
 		 *
 		 * SELECT COUNT(*) ....
 		 *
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/tasks/SearchTask.java
```diff
@@ -749,7 +751,7 @@ private void doCountOnlyQuery(boolean theParamWantOnlyCount) {
 				.withRequestPartitionId(myRequestPartitionId)
 				.execute(() -> {
 					mySearch.setTotalCount(count.intValue());
-					if (theParamWantOnlyCount) {
+					if (true) {
```
esecules: remove the redundant if.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/tasks/SearchTask.java
```diff
@@ -683,11 +681,12 @@ private void doSearch() {
 			 * matching the search off of the disk and into memory. After
 			 * every X results, we commit to the HFJ_SEARCH table.
 			 */
-			int syncSize = mySyncSize;
```
esecules, line 686: Intellij said this variable was redundant

esecules:
Given the test failures, I think there's a gap in my understanding of how the search results table is populated. When I make a search request to my hapi fhir server using _count=X in the initial request, the searchset still gets populated with all the results that matched at the time and when I issue requests for offsets higher than the initial _count, I dont think i see it going back to fetch from HFJ_RESOURCE except for when it's getting the current version for each of the search results.

esecules:
I believe the format check job just need to be restarted. it failed with a timeout trying to add a comment to this PR. I ran mvn spotless:apply on my branch and nothing changed.

PR #7452, opened 2026-01-06
Avoid unnecessary joins during resource update

jamesagnew:
When a resource update is updating a resource containing one or more references (i.e. references that were already present before the update), we perform a potentially expensive join for the target of the reference even though we really shouldn't need to. We do this for the following reasons (which can be replaced):

* We need to determine the FHIR ID associated with the PID (but we can look that up in the IdHelperService cache)
* We later check whether the resource is a valid target (but this shouldn't be necessary since the resource target was already saved)

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa/configuration.md
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa/configuration.md
@@ -168,7 +168,7 @@ X-Retry-On-Version-Conflict: retry; max-retries=100
 
 Delete with expunge submits a job to delete and expunge the requested resources. This is done in batches. If the DELETE
 ?_expunge=true syntax is used to trigger the delete expunge, then the batch size will be determined by the value
-of [Expunge Batch Size](/apidocs/hapi-fhir-storage/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.html#getExpungeBatchSize())
+of [Expunge Batch Size](/apidocs/hapi-fhir-jpaserver-model/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.html#getExpungeBatchSize())
 property.
 
 # Disabling Non Resource DB History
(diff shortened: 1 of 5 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
```diff
@@ -938,20 +941,33 @@ public ResourceTable updateEntity(
 			existingParams = existingSearchParams.get(entity);
 			if (existingParams == null) {
 				existingParams = ResourceIndexedSearchParams.withLists(entity);
+
 				/*
-				 * If we have lots of resource links, this proactively fetches the targets so
-				 * that we don't look them up one-by-one when comparing the new set to the
-				 * old set later on
+				 * If we have any existing resource links in the existing resource, look them up and
```
michaelabuckley: minor: this is clear, but could be pulled out as a method to be more clear.  Prefetch doesn't change the main flow.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceLink.java
```diff
@@ -483,6 +484,14 @@ public ResourceLink cloneWithoutTargetPid() {
 		return retVal;
 	}
 
+	@Nullable
+	public JpaPid getTargetResourceJpaPid() {
+		if (myTargetResourcePid != null) {
+			return JpaPid.fromId(myTargetResourcePid, myTargetResourcePartitionId);
```
michaelabuckley, line 490: question for someday: can this use your magic annotation stuff and actually use an embedded JpaPid here?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
jamesagnew: Good call. Done.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceLink.java
jamesagnew: Probably - will shelve for later but I will investigate this.

PR #7458, opened 2026-01-08
Add several new pointcuts

jamesagnew:
Several new pointcuts have been added to the storage layer:

* `STORAGE_PRESTORAGE_RESOURCE_PREPATCH` - This is called immediately before a FHIR Patch operation is applied to a resource, and can be used to modify both the patch and the resource that is about to be patched.
* `STORAGE_PREVERIFY_CONDITIONAL_MATCH_CRITERIA` - This is called after a resource has been conditionally created or conditionally modified, but before executing the check that the resource actually matches the given conditional URL.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/repository/IRepositoryRestQueryBuilder.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/repository/IRepositoryRestQueryBuilder.java
@@ -1,3 +1,22 @@
+/*-
+ * #%L
+ * HAPI FHIR - Core Library
+ * %%
+ * Copyright (C) 2014 - 2026 Smile CDR, Inc.
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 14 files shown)


hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.java
```diff
@@ -363,16 +362,6 @@ public class JpaStorageSettings extends StorageSettings {
...
 			IValidationSupport.IssueSeverity.WARNING;
 
-	/**
-	 * This setting allows preventing a conditional update to invalidate the match criteria.
-	 * <p/>
-	 * By default, this is disabled unless explicitly enabled.
-	 *
-	 * @since 6.8.2
-	 */
-	private boolean myPreventInvalidatingConditionalMatchCriteria =
```
jamesagnew, line 373: Note to reviewer: This setting is just being moved from `JpaStorageSettings` up into its superclass `StorageSettings` so that it is visible to `TransactionProcessor`

robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7458-add-several-interceptor-pointcuts.yaml
```diff
@@ -0,0 +1,9 @@
+---
+type: add
+issue: 7456
```
volodymyr-korzh: Issue 7456 is MDM related, and change-log filename has 7458 PR number in it, looks like intention here was to put 7458
```suggestion
issue: 7458
```

jamesagnew: Oops! Thanks!

PR #7460, opened 2026-01-09
Undo SearchConverter deletion

jdar8:
Undo SearchConverter deletion and deprecate it instead to resolve build issues


robogary:
Formatting check succeeded!

PR #7466, opened 2026-01-13
Fix NPE in SearchParamExtractorService when ResourceLink PID not in map

lukedegruchy:
Closes https://github.com/hapifhir/hapi-fhir/issues/7468

## Summary

This merge request fixes a NullPointerException in the search parameter extraction logic that occurs during resource update operations when optimizing resource link matching. The bug manifests when the system attempts to reuse existing ResourceLink entities during resource updates, but encounters a scenario where a ResourceLink's target persistent ID is not present in the ID-to-forced-ID translation map.

1. **NPE Prevention**: Added null check in `SearchParamExtractorService.findMatchingResourceLink()` before calling `Optional.isPresent()` on the result of `PersistentIdToForcedIdMap.get()`, which can return null for keys not in the map rather than an empty Optional.

## Code Review Suggestions

- [ ] **Review the optimization logic context**: The fix is in `findMatchingResourceLink()` which is an optimization path to reuse existing ResourceLink entities during updates. Verify that returning an empty Optional when the map lookup returns null is the correct behavior - this will cause the system to fall back to creating a new ResourceLink rather than reusing an existing one. Consider whether there are scenarios where a missing PID in the map indicates a more serious data inconsistency issue that should be logged or handled differently.

- [ ] **Check PersistentIdToForcedIdMap contract**: The `PersistentIdToForcedIdMap.get()` method delegates directly to `Map.get()`, which returns null for missing keys. Review whether this is the intended contract or if the class should be modified to always return an Optional (empty for missing keys) to prevent this type of bug elsewhere in the codebase.

- [ ] **Verify resource link lifecycle assumptions**: The code assumes that when updating a resource, some existing ResourceLinks may have target PIDs that aren't in the translation map. Confirm this is an expected scenario (e.g., stale links, cross-partition references, or resources deleted between operations) rather than indicating a bug in how the map is populated.

### Setup
- Configure a HAPI FHIR server with JPA persistence
- Enable resource indexing and search parameter extraction
- Create test resources with references (e.g., Group resources with member references to Patient resources)

### Test Cases

- [ ] **Update resource with existing references**: Create a Group resource with Patient references, then update the Group resource (without changing references) and verify no NPE occurs during search parameter extraction.

- [ ] **Update with forced IDs**: Create Patient resources using client-assigned IDs (forced IDs), create a Group referencing these Patients, then update the Group and verify the resource links are correctly maintained.

- [ ] **Update with numeric IDs**: Create Patient resources using server-assigned numeric IDs, create a Group referencing these Patients, then update the Group and verify the resource links are correctly maintained.

- [ ] **Update with deleted reference targets**: Create a Group with Patient references, delete one of the referenced Patients, then update the Group resource and verify the system handles the missing target gracefully without NPE.

- [ ] **Concurrent updates**: Perform concurrent updates to resources with references in a multi-threaded scenario to verify the fix handles race conditions where the PID map may not contain all expected entries.

- [ ] **Cross-partition references** (if partitioning enabled): Create resources in different partitions with cross-partition references, update the referencing resources, and verify no NPE occurs during link resolution.

- [ ] **Bulk operations**: Execute bulk update operations on resources with references and verify the optimization path works correctly without NPEs across the entire batch.

### DIFF
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
@@ -820,8 +820,9 @@ public class SearchParamExtractorService implements ISearchParamExtractorSvc {
 
 		if (thePathAndRef.isCanonical()) {
 			return resourceLinks.stream()
-				.filter(r -> r.getTargetResourceUrl().equals(thePathAndRef.getPath()))
-				.findFirst();
+					.filter(r -> r.getTargetResourceUrl() != null
+							&& r.getTargetResourceUrl().equals(thePathAndRef.getPath()))
+					.findFirst();
 		}
 
 		Set<JpaPid> pids = new HashSet<>();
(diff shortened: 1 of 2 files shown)


robogary:
Formatting check succeeded!

PR #7467, opened 2026-01-14
PatientCompartmentEnforcingInterceptor incorrectly blocks group updates

jamesagnew:
The PatientCompartmentEnforcingInterceptor incorrectly prevented changes to the members of a Group resource, even though the PatientIdPartitionInterceptor does not consider this resource to be patient-specitic. This has been corrected.


robogary:
Formatting check succeeded!

PR #7470, opened 2026-01-14
Fix _has query on partitioned system

jamesagnew:
When using the JPA server with a partitioned configuration, `_has` queries failed if the source and target were in different partitions. This has been corrected.

### DIFF
--- a/hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/jpa/dao/r5/dbpartitionmode/TestDefinitions.java
+++ b/hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/jpa/dao/r5/dbpartitionmode/TestDefinitions.java
@@ -1015,13 +1015,18 @@ abstract class TestDefinitions implements ITestDataBuilder {
 	@Test
 	public void testSearch_ListParam() {
 		// Setup
+		myPartitionSettings.setAllowReferencesAcrossPartitions(PartitionSettings.CrossPartitionReferenceMode.ALLOWED_UNQUALIFIED);
+
 		myPartitionSelectorInterceptor.setNextPartitionId(PARTITION_1);
 		IIdType patId0 = createPatient(withActiveTrue()).toUnqualifiedVersionless();
 		IIdType patId1 = createPatient(withActiveTrue()).toUnqualifiedVersionless();
+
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
```diff
@@ -182,7 +191,7 @@ public DbColumn[] getJoinColumnsForTarget() {
 	}
 
 	public DbColumn[] getJoinColumnsForSource() {
-		return getSearchQueryBuilder().toJoinColumns(getPartitionIdColumn(), myColumnSrcResourceId);
```
michaelabuckley, line 185: problem: This is confusing, and doesn't make sense to me.  How can we join with the target partition, but the src resource id?  Should this fix be higher up?  Or explained?

jamesagnew: I think you're mixing up the line being removed with the line being added.

This method (`getJoinColumnsForSource`) returns the PID+Partition on the source side, so this change is hardcoding it to always use `myColumnSrcPartitionId` and `myColumnSrcResourceId`, which are always going to hold the right columns.

The change above this one (adding `if (myReversed) {` to `getPartitionIdColumn` is the actual fix to the bug: The [getResourceIdColumn()](https://github.com/hapifhir/hapi-fhir/blob/8bc4aa211dbb14be2f3cc5e17e930ade58e6efd2/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java#L211-L217) correctly considered whether this was a forward or a reverse link, but the corresponding `getPartitionIdColumn()` did not.

michaelabuckley: You are correct.  I misunderstood the patch.

PR #7478, opened 2026-01-16
Use new requestdetails factory in Bulk Modification tasks

jamesagnew:
PR !7472 introduced a new mechanism to generate RequestDetails objects. This is just a quick PR to leverage this functionality in the new bulk modification framework.


robogary:
Formatting check succeeded!

PR #7484, opened 2026-01-20
$expunge operation does not delete externally stored resources 

volodymyr-korzh:
- Added support for removing externally stored resources from storage providers during instance-level and type-level `$expunge` operations.
- unit tests
- changelog


robogary:
Formatting check succeeded!

PR #7493, opened 2026-01-21
fix subscription message delivery channel matching

### LINKED ISSUE #7492: subscription matching and resource delivery not delivering resources to correct channel
**Describe the bug**
When using a shared message delivery queue, messages are delivered to the same endpoint and ignores the Subscription configuration settings in `channel.endpoint`. This happens because only the first endpoint is cached in `SubscriptionDeliveringMessageListener` based on the first one encountered [here](https://github.com/hapifhir/hapi-fhir/blob/e8adcda93ec3fe46f27eaf0d2d29bc3c6e4b6565/hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/message/SubscriptionDeliveringMessageListener.java#L141).

**To Reproduce**
1. Create subscription module with a shared delivery queue for all subscriptions
2. Create 2 subscriptions: one that matches on `Patient`s, and another that matches on `Organizations`. The `channel.endpoint`s should be different. Example:
```
{
	"resourceType": "Subscription",
	"status": "requested",
	"criteria": "[Patient]",
	"channel": {
		"type": "message",
		"endpoint": "channel:my-pat-sub",
		"payload": "application/json"
	}
}
```
and 
```
{
	"resourceType": "Subscription",
	"status": "requested",
	"criteria": "[Organization]",
	"channel": {
		"type": "message",
		"endpoint": "channel:my-org-sub",
		"payload": "application/json"
	}
}
```
3. Wait for the Subscriptions to be registered, then submit a Patient and Organization resource

**Expected behaviour**
The Patient is delivered to `channel:my-pat-sub` and the Organization is delivered to `channel:my-org-sub`

**Actual behaviour**
Both resources are delivered to the same channel.

jdar8:
**Problem:** Subscription messages would be delivered to the incorrect endpoint when using a shared delivery queue. This is due to the first endpoint being cached in `SubscriptionDeliveringMessageListener`

**What was done:**
- Fix: instead of caching only the first queue, create a map of queue name and the channel producer
- Tests/changelog

Closes #7492

### DIFF
--- a/hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/message/SubscriptionDeliveringMessageListener.java
+++ b/hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/message/SubscriptionDeliveringMessageListener.java
@@ -180,7 +180,8 @@ public class SubscriptionDeliveringMessageListener extends BaseSubscriptionDeliv
 		try {
 			myChannelProducerMap.values().forEach(producer -> {
 				if (producer instanceof AutoCloseable closeable) {
-					Logs.getSubscriptionTroubleshootingLog().debug("Destroying channel producer {}", producer.getChannelName());
+					Logs.getSubscriptionTroubleshootingLog()
+							.debug("Destroying channel producer {}", producer.getChannelName());
 					IoUtils.closeQuietly(closeable, ourLog);
 				}
 			});


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/message/SubscriptionDeliveringMessageListener.java
```diff
@@ -174,8 +176,15 @@ private String extractQueueNameFromEndpoint(String theEndpointUrl) throws URISyn
 
 	@Override
 	public void close() throws Exception {
-		if (myChannelProducer instanceof AutoCloseable) {
-			IoUtils.closeQuietly((AutoCloseable) myChannelProducer, ourLog);
+		try {
+			myChannelProducerMap.values().forEach(producer -> {
+				if (producer instanceof AutoCloseable closeable) {
+					ourLog.info("Destroying channel producer {}", producer.getChannelName());
```
michaelabuckley: minor: Is this worth `info`, or should it be debug?  This could be very noisy in the logs for large systems.  Existing behaviour doesn't log at all unless close fails in the closeQuietly call.

suggestion: replace the log in this class with the Subscription Troubleshooting log - especially for the debug message when delivering the message.

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/message/SubscriptionDeliveringMessageListener.java
```diff
@@ -138,10 +141,9 @@ public void handleMessage(ResourceDeliveryMessage theMessage) throws MessagingEx
 		ChannelProducerSettings channelSettings = new ChannelProducerSettings();
 		channelSettings.setQualifyChannelName(false);
 
-		if (myChannelProducer == null) {
-			myChannelProducer =
-					myBrokerClient.getOrCreateProducer(queueName, ResourceModifiedJsonMessage.class, channelSettings);
-		}
+		IChannelProducer<ResourceModifiedMessage> producer = myChannelProducerMap.computeIfAbsent(
+				queueName,
+				key -> myBrokerClient.getOrCreateProducer(key, ResourceModifiedJsonMessage.class, channelSettings));
```
tadgh, line 147: Issue: This API smells a bit.. we are doing what the brokerclient should be doing (the caching part). Let's think about this in bug-fix time. We want one centralized caching for channel producer/consumer, and this is evading that concept by caching downstream.

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/message/SubscriptionDeliveringMessageListener.java
jdar8: Applied the suggestion - switched to the subscription troubleshooting log and changed to debug.

Thanks for the review!

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/match/deliver/message/SubscriptionDeliveringMessageListener.java
jdar8: Opened a new ticket to resolve this later. 

If we want to move the caching to IBrokerClient implementations, the I assume the closing of channels will also need to be moved over to those implementations as well. 

Thanks for the review!

Issue #7494, opened 2026-01-21
Race condition in expunge service


jmarchionatto:
**Describe the bug**
The ExpungeEverythingService.deleteAll() method deletes Batch2WorkChunkEntity and Batch2JobInstanceEntity in separate transactions (Propagation.REQUIRES_NEW). This causes intermittent FK constraint violations in tests:

```code
ResourceVersionConflict HAPI-0550: HAPI-2223: JDBC exception executing SQL
[delete from BT2_JOB_INSTANCE bjie1_0 where bjie1_0.ID in (?,?,?,?,?,?)]
[Referential integrity constraint violation: "FK_BT2WC_INSTANCE: PUBLIC.BT2_WORK_CHUNK
FOREIGN KEY(INSTANCE_ID) REFERENCES PUBLIC.BT2_JOB_INSTANCE(ID)"]
```

**To Reproduce**
Reproducing test added

**Environment (please complete the following information):**

- HAPI FHIR Version
v8.6.1

PR #7495, opened 2026-01-21
Fix race condition in expungeEverything with Batch2 entities (#7494)

jmarchionatto:
Fixed a race condition where Batch2WorkChunkEntity and Batch2JobInstanceEntity were deleted in separate transactions, allowing job maintenance to create new chunks between the deletes and cause FK constraint violations (FK_BT2WC_INSTANCE).

The bug occurred because:
1. Transaction 1 deletes all Batch2WorkChunkEntity records
2. Job maintenance can run between transactions
3. Maintenance creates new chunks for completed jobs (workChunksPurged=false)
4. Transaction 2 tries to delete Batch2JobInstanceEntity
5. FK violation: new chunks still reference the instances

The fix deletes both entity types in a single transaction via the new expungeBatch2EntitiesInSingleTransaction() method, eliminating the race window.


robogary:
Formatting check succeeded!

PR #7499, opened 2026-01-24
Fix/preferred

jkiddo:
Before continuing an fixing the other tests that this change introduce - is there a particular reason why the logic in this PR is not the default behaviour - meaning that not following preferred bindings emits as errors - even though the are merely preferred?


robogary:
Formatting check succeeded!

PR #7505, opened 2026-01-27
Issue with creating resources when using PatientId partitioning mode

epeartree:
**The issue:**
This issue happens when the system operates in PatientId partition mode.  
when a resource stored in a patient compartment partition is deleted and re-created, re-creation of the resource fails. 

**Root cause:**
when a resource is deleted, the state of the current/deleted version of the fhir Resource is empty except for its ID that is stored in the userData.  when recreating the resource (PUT /someResource/id), the oldResource and the new resources are submitted to the PatientIdPartitionInterceptor to determine their partition ID based on their referred Patient ID.  since the old version of the resource is empty, the partition id could not be determined and an exception raised. 

**The fix:**
- added logic to extract and return the partition hash from the resource Id stored in the resource userData when the resource is marked as deleted;
- removed the partitionName from the requestPartitionId returned from the interceptor as parition name are not stored in the db and it is impossible to deduce it from the partition hash.

### DIFF
--- a/hapi-deployable-pom/pom.xml
+++ b/hapi-deployable-pom/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		 <groupId>ca.uhn.hapi.fhir</groupId>
 		 <artifactId>hapi-fhir</artifactId>
-		 <version>8.9.0-SNAPSHOT</version>
+		 <version>8.9.1-SNAPSHOT</version>
 
 		 <relativePath>../pom.xml</relativePath>
 	 </parent>
--- a/hapi-fhir-android/pom.xml
+++ b/hapi-fhir-android/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.9.0-SNAPSHOT</version>
+		<version>8.9.1-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
--- a/hapi-fhir-base/pom.xml
+++ b/hapi-fhir-base/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.9.0-SNAPSHOT</version>
+		<version>8.9.1-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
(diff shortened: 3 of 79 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7501-fix-create-issue-in-patientId-partitioning-mode.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 7501
+title: "Previously, re-creating a deleted resource would raise an exception when the system operates in PatientID
```
michaelabuckley:
```suggestion
title: "Previously, re-creating a deleted resource with the same id would raise an exception when the system operates in PatientID 
```

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -126,9 +134,27 @@ public RequestPartitionId identifyForCreate(IBaseResource theResource, RequestDe
 			if (oCompartmentIdentity.isPresent()) {
 				return provideCompartmentMemberInstanceResponse(theRequestDetails, oCompartmentIdentity.get());
 			} else {
-				return getPartitionViaPartiallyProcessedReference(theRequestDetails, theResource)
-						// or give up and fail
-						.orElseGet(() -> throwNonCompartmentMemberInstanceFailureResponse(theResource));
+
+				if (theResource.isDeleted()) {
+					// this conditional statement covers the edge case where a resource was deleted and is being
+					// re-created.
```
michaelabuckley:
```suggestion
					// re-created by PUT with client-assigned id.
```

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -161,6 +161,32 @@ public void testDeleteResource_InPatientCompartment() {
...
+		myCaptureQueriesListener.logUpdateQueriesForCurrentThread();
+
+		// Verify - Assert the observation exists and version is incremented
+		Observation recreated = myObservationDao.read(new IdType("Observation/O"), mySrd);
+		assertThat(recreated).isNotNull();
+		assertThat(recreated.getSubject().getReference()).isEqualTo("Patient/A");
+
+		assertThat(recreated.getIdElement().getVersionIdPartAsLong()).isEqualTo(deletedVersion + 2);
+	}
+
```
michaelabuckley: suggestion: we need a test that verifies we can't change the partition by this delete/create path.

So:
- create two patients
- create observation O pointing to p1
- delete O
- re-create O, but pointing to P2.
- Expect failure.  We can NOT change partition after creation.

Issue #7512, opened 2026-01-30
Configurable version policy for resources installed by package installer


mrdnctrk:
We used to keep the client-assigned fhir ids for IG package resources STORED in the repository. But this made it impossible to have multiple versions of US Core installed at the same time, which breaks multi-version validation. To fix this, there was a change for rules for package loading to match on canonical url+version, and use server assigned ids. This change has broken users who depended on the old behaviour.

This issue is about making this behavior configurable whether to allow installing resources with multiple versions from a package or just allowing a single version as before.

mrdnctrk:
Closed. Fixed by #7516.

centaurusgod:
Hi @mrdnctrk 
This seems to have affected when uploading the **ImplementationGuides** via;
POST `ImplementationGuides/$install`

All the resources of ig are now being saved with server assigned numeric id instead of the client id. 
For eg: The  CodeSystem-v3-ActCode.json present inside **hl7.r4.core** ig has `id=v3-ActCode`, but when installed via `$install`  the id is numeric. 

Any workaround/config to it (jpa-server) so we can still have the client assigned id?

mrdnctrk:
replied on the issue https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/897

PR #7513, opened 2026-01-30
added versionPolicy configuration to package installer

mrdnctrk:
Summary                                                                                                                                                                                                                             
                                                                                                                                                                                                                                      
  Added a new versionPolicy parameter to PackageInstallationSpec that controls how resource versions are handled during package installation in STORE_AND_INSTALL mode.                                                               
                                                                                                                                                                                                                                      
  ### Background                                                                                                                                                                                                                          
                                                                                                                                                                                                                                      
  When installing FHIR packages using STORE_AND_INSTALL mode, users need flexibility in how resources (StructureDefinitions, ValueSets, etc.) are installed. Some use cases require multiple versions of the same profile to   
  coexist (for multi-version validation), while others need a single authoritative version that gets updated when new packages are installed.                                                                                         
                                                                                                                                                                                                                                      
  ### Changes                                                                                                                                                                                                                             
                                                                                                                                                                                                                                      
  New versionPolicy Parameter                                                                                                                                                                                                         
                                                                                                                                                                                                                                      
  Two policies are now available:                                                                                                                                                                                                     
                                                                                                                                                                                                                                      
  - MULTI_VERSION (default): Uses server-assigned IDs for new resources. Existing resources are matched by both canonical URL and version. This allows multiple versions of the same profile to coexist in the repository.            
  - SINGLE_VERSION: Uses client-assigned IDs from the package. Existing resources are matched by canonical URL only (version is ignored). Installing a new version of a profile will overwrite the existing one.                      
                                                                                                                                                                                                                                      
  Note: SearchParameter resources always use client-assigned IDs regardless of the versionPolicy setting, as only one version of a SearchParameter can exist per URL + base combination.                                              
                                                                                                                                                                                                                                      
  Implementation Details                                                                                                                                                                                                              
                                                                                                                                                                                                                                      
  - Added VersionPolicyEnum to PackageInstallationSpec                                                                                                                                                                                
  - Refactored PackageInstallerSvcImpl.createOrUpdateResource() for clarity, as it was too convoluted. I split into multiple methods.                                                                                                                                                           
  - Search queries now include _sort=_pid:desc for deterministic resource selection when multiple versions exist                                                                                                                      
  - meta.source is now set for all package resources regardless of version policy, as this shows from which package and version a resource came from.                                                                                                                                    
                                                                                                                                                                                                                                      
                                                                                                                          
                                                                                                                                                                                                                                      
  Migration Notes                                                                                                                                                                                                                     
                                                                                                                                                                                                                                      
  The current default behavior (MULTI_VERSION) is preserved, so shouldn't be considered a breaking change.      


closes: https://github.com/hapifhir/hapi-fhir/issues/7512

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7512-add-versionPolicy-to-package-installation.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7512-add-versionPolicy-to-package-installation.yaml
@@ -1,0 +1,8 @@
+---
+type: add
+issue: 7512
+title: "Added a new `versionPolicy` parameter to PackageInstallationSpec that 
+controls how resource versions are handled during package installation in STORE_AND_INSTALL mode. 
+The `MULTI_VERSION` policy (default) assigns server-generated IDs and matches existing resources by both canonical URL 
+and version, allowing multiple versions of the installed resources to coexist. The `SINGLE_VERSION` policy retains 
+client-assigned IDs and matches by canonical URL only, overwriting existing resources when a new version is installed."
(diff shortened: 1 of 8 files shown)


robogary:
Formatting check succeeded!

PR #7514, opened 2026-01-30
potential fix for flaky ResourceProviderR4Test

mrdnctrk:
The following test is failing intermittently, I couldn't reproduce the failure locally, but it is possible that it is related to parser options is not reset to default properly for stripping versions after another test. So if a test suite is changing this config, another test that relies on the default behaviour could fail. This PR attempts to fix this restoring strip version policy to the default values after each test. 
 
ResourceProviderR4Test.testPreserveVersionsOnAuditEvent:3626 Patient/14796 ==> expected: <true> but was: <false>


robogary:
Formatting check succeeded!

PR #7525, opened 2026-02-04
Add retry logic for FK violations during Batch2 entity expunge

jmarchionatto:
The expungeBatch2EntitiesInSingleTransaction method can encounter ResourceVersionConflictException when background job maintenance inserts new work chunks between the DELETE of chunks and DELETE of job instances. H2 MVCC allows concurrent INSERTs despite SELECT FOR UPDATE locks on the parent rows, and the built-in retry mechanism in HapiTransactionService has maxRetries=0 for this code path.

Wrap the transaction execution in a try/catch that retries up to 10 times on ResourceVersionConflictException before rethrowing.

### DIFF
--- a/.github/actions/assemble-test-reports/action.yml
+++ b/.github/actions/assemble-test-reports/action.yml
@@ -57,7 +57,9 @@ runs:
       uses: actions/upload-artifact@v4
       with:
         name: junit-test-results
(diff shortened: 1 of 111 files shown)


robogary:
Formatting check succeeded!

Issue #3150, opened 2021-11-09
False searchInclude and searchRevInclude information in the Capability Statement


npsuraweera:
After upgrading to HAPI FHIR version 5.4.0 it started showing wrong information related to supported include and revinclude parameters in the CapabilityStatement.
![image](https://user-images.githubusercontent.com/63303677/140889144-fded2bf3-aced-473d-8963-85783d6833c7.png)
Truly we havent implemented the support for include or revinclude parameters for This particular resource. But HAPI FHIR adds all attributes as include and  revinclude parameters.
In Old versions this section was added only when we use the @include annotation on the methods

PR #7539, opened 2026-02-06
Issues #3042 and #3150 fix: remove implicit addition of _include and _revinclude in CapabilityStatement

### LINKED ISSUE #3042: Capability Statement Reverse Includes Issue
If a resource provider (eg. Coverage) has no reverse include parameters on its @Search endpoint, the HAPI capability statement generates reverse includes for all possibilities rather than generating an empty reverse includes.

For example:

1) CoverageResourceProvider.java:
@Search 
public IBundleProvider search(@OptionalParam(name="patient") TokenParam patient) { **// No reverse includes**
..

2) MedicationDispenseResourceProvider.java
@Search
public IBundleProvider search(@OptionalParam(name="patient") ReferenceParam patientRef) { **// Not reverse searchable**
...

Generated Statement:
=====
"type": "Coverage",
                    "profile": "http://hl7.org/fhir/StructureDefinition/Coverage",
..
"searchRevInclude": [
                        "HealthcareService:location",    
                        "MedicationDispense:patient",   **<-- Is this wrong? the resource does not define any reverse includes**
                        "OrganizationAffiliation:location",
                        "PractitionerRole:location",
                        "Procedure:patient"
                    ],

-----
Note that the capability statement does work properly when when your resource **does** define a reverse includes:

ConditionResourceProvider.java

@Search
public IBundleProvider search(
    @IncludeParam(reverse = true, allow = {"Provenance:target"}) Set<Include> myParam)) { **// Reverse searchable**
...

"type": "Condition",
                    "profile": "http://hl7.org/fhir/StructureDefinition/Condition",
..
"searchRevInclude": [
                        "Provenance:target",   **<-- Correct!**
                    ],

tanmaydarmorha:
Fixes #3042 and #3150 — plain servers that do not declare `@IncludeParam` on their `@Search` methods were still advertising `searchInclude` / `searchRevInclude` entries in the generated `CapabilityStatement`. Per the [FHIR spec](https://hl7.org/fhir/R4/capabilitystatement-definitions.html#CapabilityStatement.rest.resource.searchInclude), the list should be empty when the server does not support these.

### Background

The original revision of this PR simply removed the auto-inference logic in `ServerCapabilityStatementProvider`. @jamesagnew correctly pointed out that this would regress the JPA server, whose tinder-generated resource providers declare bare `@IncludeParam` / `@IncludeParam(reverse = true)` (no `allow` values) and rely on the auto-inference to populate their `CapabilityStatement`.

The root cause is that the current code conflates two distinct states, both of which yield an empty `Set<String>` from `BaseMethodBinding.getIncludes()`:

1. **No `@IncludeParam` on any `@Search` method** — server does not accept `_include` at runtime.
2. **`@IncludeParam` with no `allow`** — server accepts any `_include` at runtime (JPA's model).

### Approach

Split the two states apart so the generated `CapabilityStatement` mirrors what the dispatcher actually accepts at runtime:

| Resource state | `searchInclude` output |
|---|---|
| No `@IncludeParam` on any `@Search` method | Emit nothing (fixes #3042 / #3150) |
| `@IncludeParam` with no `allow` | Auto-infer from reference-type search params (unchanged — preserves JPA) |
| `@IncludeParam(allow = {...})` | Emit the explicit list (unchanged) |

Same three-way logic for `searchRevInclude`.

### Changes

- `BaseMethodBinding` — new `hasIncludeParameter(boolean theReverse)` that reports the *presence* of `@IncludeParam`, orthogonal to the existing `getIncludes()` / `getRevIncludes()` which report `allow` values.
- `ServerCapabilityStatementProvider` — per-resource tracking of whether any `@Search` method declared `@IncludeParam` (include and reverse directions tracked independently). The emission block is gated by this tracking and restores the auto-inference branch for the bare-annotation case.
- `ServerCapabilityStatementProviderR4Test` — explicit coverage for all three scenarios on each axis:
  - Includes: `testIncludes_NotSpecifiedReturnsEmptyList`, `testIncludes_Inferred`, `testIncludes_Explicit`
  - RevIncludes: `testRevIncludes_NotSpecifiedReturnsEmptyList`, `testRevIncludes_Inferred`, `testRevIncludes_Explicit`, `testRevIncludes_DisabledReturnsEmpty`
  - Gate independence: `testIncludes_PresentDoesNotEmitRevIncludes`, `testRevIncludes_PresentDoesNotEmitIncludes`

### Verification

- `ServerCapabilityStatementProviderR4Test` — 42/42 pass.
- Cross-version capability tests — DSTU2 (19/19), DSTU3 (26 pass + 2 skip), R4B (23 pass + 1 skip), R5 (23 pass + 1 skip), HL7.org DSTU2 (14/14).
- JPA integration test — `ServerCapabilityStatementProviderJpaR4Test` (14/14) confirms JPA's bare-`@IncludeParam` providers continue to emit auto-inferred includes/revincludes as before. `ServerDstu3Test` and `ServerR5Test` equivalents also pass.

### Why this over `@IncludeParam(includeInCapabilityStatement = false)`

- No new annotation attribute for plain-server implementers to learn.
- JPA works with zero changes to its providers or tinder template.
- `CapabilityStatement` generation becomes structurally consistent with runtime `_include` routing — both driven by the same `IncludeParameter` presence check.


jamesagnew (changes requested):
Hi @tanmaydarmorha . Thanks for the contribution.

I'm worried that this change will break the JPA server, which depends on the automatic inferring of include/revincludes. I.e. which this change, the JPA server will no longer list any include/revincludes in its own CapabilityStatement.

WOuld it make sense to add a new option attribute to the `@IncludeParam` annotation such as `includeInCapabilityStatement` which would default to true, but that you could explicitly set to false if you didn't want this behaviour?

tanmaydarmorha:
> Hi @tanmaydarmorha . Thanks for the contribution.
> 
> I'm worried that this change will break the JPA server, which depends on the automatic inferring of include/revincludes. I.e. which this change, the JPA server will no longer list any include/revincludes in its own CapabilityStatement.
> 
> WOuld it make sense to add a new option attribute to the `@IncludeParam` annotation such as `includeInCapabilityStatement` which would default to true, but that you could explicitly set to false if you didn't want this behaviour?

Hi @jamesagnew — thanks for the review. You're right that JPA relies on this. Removing it outright would ship empty include lists for every JPA resource — that's a real regression.

Before adding a new annotation flag, I'd like to float an alternative. Today the capability generation conflates two states: "no `@IncludeParam` parameter at all" and "`@IncludeParam` with empty `allow`" both yield an empty set. If we split them, the generated statement can mirror the dispatcher exactly:

- No `@IncludeParam` on any search method → emit nothing (what #3042 / #3150 asked for).
- `@IncludeParam(allow = {...})` → emit that list (unchanged).
- `@IncludeParam` without `allow` → auto-infer (unchanged — what JPA uses).

Mechanically: add a `hasIncludeParameter(boolean reverse)` on `BaseMethodBinding` and use it to gate the auto-infer branch.

Why I prefer this over the annotation flag: JPA keeps working untouched, plain-server implementers don't need to know about a new attribute, and the CapabilityStatement ends up consistent with runtime routing — both driven by the same `IncludeParameter` presence check.

Does this direction sound reasonable before I rework the PR?

jamesagnew:
@tanmaydarmorha Yeah, that's a good suggestion. If you wanted to rework this to work in that way, I'd certainly be happy to get that merged.

tanmaydarmorha:
> @tanmaydarmorha Yeah, that's a good suggestion. If you wanted to rework this to work in that way, I'd certainly be happy to get that merged.

I've updated the PR, please take a look whenever you get time

tanmaydarmorha:
@jamesagnew I’ve updated the PR following our agreed approach. Could you please take another look and merge if everything looks good to go?

jamesagnew (approved):
This looks great. Will kick off the CI and get it merged assuming everything passes. Thanks for the contribution!

tadgh:
Formatting check succeeded!

PR #7546, opened 2026-02-10
Fix mdm resource matching during delete cleanup bypasses interceptors

volodymyr-korzh:
Added:
- Fix for MdmStorageInterceptor - read , instead of readByPid to make sure Interceptor are invoked
- IT test for MdmStorageInterceptor to make sure `STORAGE_PRESHOW_RESOURCES` interceptor is properly invoked for all related resources when deleting a match patient
- changelog

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7541-fix-mdm-resource-matching-during-delete-cleanup.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7541-fix-mdm-resource-matching-during-delete-cleanup.yaml
@@ -1,5 +1,6 @@
 ---
 type: fix
 issue: 7541
-title: "Previously, MDM delete cleanup operations would prevent interceptors from being invoked when reading 
-the golden resource and possible match resources. This has been fixed."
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/7541-fix-mdm-resource-matching-during-delete-cleanup.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+issue: 7541
+title: "Previously, MDM delete cleanup operations would prevent interceptors from being invoked when reading
```
tadgh: suggestion: could you be more precise here? What's an MDM Delete Cleanup Operation?

volodymyr-korzh: Agree, this is not clear enough. 
Added details:

> Previously, when deleting a source resource with MDM links, interceptors would not be invoked 
> when reading the golden resource and possible match resources during link cleanup, preventing custom 
> interceptor logic from executing for these operations. This has been fixed.

PR #7548, opened 2026-02-10
Add duplicate resource blocker to public test server

jamesagnew:
This PR adds a new duplicate resource blocking interceptor to the public test server (and makes this interceptor available if anyone else wants to use it).

It also modifies the InterceptorService to not log noisy stack traces if interceptor hook methods throw a HAPI FHIR BaseServerResponseException (since this often happens as a part of routine processing)

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TerminologyValidationSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TerminologyValidationSvcImpl.java
@@ -1,0 +1,19 @@
+/*-
+ * #%L
+ * HAPI FHIR JPA Server
+ * %%
+ * Copyright (C) 2014 - 2026 Smile CDR, Inc.
+ * %%
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * #L%
+ */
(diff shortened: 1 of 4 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/executor/BaseInterceptorService.java
```diff
@@ -605,11 +606,14 @@ public Object invoke(HookParams theParams) {
 				return invokeMethod(args);
 			} catch (InvocationTargetException e) {
 				Throwable targetException = e.getTargetException();
-				ourLog.error(
-						"Exception thrown by interceptor for pointcut {}: {}",
-						getPointcut(),
-						targetException.toString(),
-						targetException);
+
+				ourLog.atWarn()
```
tadgh: Thought: I suppose the change to a warning is intentional here. I assume the reasoning is "Well its an interceptor, so its not something that the system admin can fix, and thus its a warning"?  My opinion is that error is more suitable, but my opinion is not strongly held.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7548-add-dupe-checking-interceptor.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: add
+issue: 7548
+title: "A new interceptor called MemoryBasedDuplicateBlockingInterceptor has been added. This interceptor
+  does a basic check upon resource creation to detect other duplicate resources recently created, and
+  blocks them."
```
tadgh: suggestion: define what counts as a duplicate

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/executor/BaseInterceptorService.java
jamesagnew: Yeah, I guess you're right - I'll switch it back to error

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7548-add-dupe-checking-interceptor.yaml
jamesagnew: Done.

PR #7554, opened 2026-02-12
Fix: Replace probabilistic batch2 expunge retry with deterministic se…

jmarchionatto:
…maphore hold

The $expunge-everything operation races with JobMaintenanceServiceImpl on the batch2 tables (BT2_JOB_INSTANCE, BT2_WORK_CHUNK). Maintenance can insert or update work chunks while expunge deletes them, causing FK violations and deadlocks. The previous fix (SELECT FOR UPDATE + retry up to 10 times) was probabilistic and still failed under load.

Hold the existing maintenance semaphore for the duration of batch2 entity deletion so no maintenance pass can start or be in-flight while rows are being removed. This eliminates the race entirely.

- Add IJobMaintenanceService.holdMaintenanceForExpunge() returning a Closeable that acquires the maintenance semaphore (with 5-min timeout) and releases it on close
- Implement MaintenanceHold inner class with idempotent close via AtomicBoolean
- Simplify ExpungeEverythingService: remove pessimistic locking, retry loop, and custom three-step SELECT/DELETE/DELETE; reuse existing expungeEverythingByTypeWithoutPurging() for both entity types under the hold
- Make IJobMaintenanceService a required dependency (jpaserver-base always has storage-batch2 on the classpath)

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeEverythingService.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeEverythingService.java
@@ -20,6 +20,7 @@
 package ca.uhn.fhir.jpa.dao.expunge;
 
 import ca.uhn.fhir.batch2.api.IJobMaintenanceService;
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

Issue #7555, opened 2026-02-12
HAPI-FHIR POM is not compatible with Maven 4.


another-luis:
For example, hapi-fhir-jpaserver-base/pom.xml has two declaration for ca.uhn.hapi.fhir:hapi-tinder-plugin

>The project ca.uhn.hapi.fhir:hapi-fhir-jpaserver-base:8.7.11-SNAPSHOT (.../hapi-fhir-jpaserver-base/pom.xml) has 1 error
>[ERROR]     'build.plugins.plugin.(groupId:artifactId)' must be unique but found duplicate declaration of plugin ca.uhn.hapi.fhir:hapi-tinder-plugin @ line 524, column 4

PR #7558, opened 2026-02-12
7555: consolidate both hapi-tinder-plugin plugins into a single one

another-luis:
Fixes: #7555


robogary:
Formatting check succeeded!

PR #7551, opened 2026-02-11
Fix: Validate resource existence in $everything operations

epeartree:
### Summary
  Fixes Patient and Encounter `$everything` operations to return 404 Not Found when invoked on non-existent resources, per FHIR HTTP
   specification.

  ### Changes
  - **JpaResourceDaoPatient**: Added existence validation in `patientInstanceEverything()`
  - **JpaResourceDaoEncounter**: Added existence validation in `encounterInstanceEverything()`
  - **Tests**: Added comprehensive test coverage for both operations

  ### Technical Details
  Both operations now call `read(theId, theRequest)` to validate resource existence before querying related resources. This:
  - Throws `ResourceNotFoundException` (404) for non-existent resources
  - Throws `ResourceGoneException` (410) for deleted resources
  - Properly handles partitioning and authorization

  ### Testing
  - ✅ `JpaPatientEverythingTest#testEverythingOperation_shouldReturn404ForNonExistentPatient`
  - ✅ `ResourceProviderR4EverythingTest#testEncounterEverythingOperation_shouldReturn404ForNonExistentEncounter`

Closes #7550

PR #7560, opened 2026-02-13
delay check for instance everything to fix broken test

jdar8:
### **Context:**
A [bug fix yesterday](https://github.com/hapifhir/hapi-fhir/pull/7551) related to instance `$everything` on non-existent resources added a check (calls `read()` in `JpaResourceDaoPatient`) to ensure the resource exists before continuing with `$everything`. However, the `read()` seems to be ignoring the cross partition setting.

It seems that the expected behvaiour for $everything with mdm and partitioning is that if we have

- `PartitionA/Patient/123`
- `PartitionB/Patient/456`
- `Patient/123` and `Patient/456` are linked to the same golden resource 
- Cross partition is enabled/allow unqualified

Then a query `PartitionA/Patient/456/$everything?_mdm=true` should perform cross partition MDM expansion for all patients matching `Patient/456`, but then only return related resources on the requested `Partition/A` --> so this returns  only `PartitionA/Patient/123`

So, when we check to see if the resource in the initial requests exists, `PartitionA/Patient/456` returns resource not found, but we do have cross-partition enabled and we want MDM expansion for the `$everything` operation.  So I think this check kind of ignores the cross partition MDM expansion part that was supposed to happen. 

### **Suggested Fix/What was Done:**
- Move the check to see if the instance resource exists down to the `SearchBuilder#createChunkedQueryForEverythingSearch()`, where we already query for the DB when converting the ID provided in the request to a PID. This uses the `myRequestPartitionId` used in `SearchBuilder` instead of what is provided by the user in the `RequestDetails`.
    - Another option would be to make the `read()` in `JpaResourceDaoPatient` determine the partition from settings and request details (mdm expansion/cross partitioning). This would fail earlier, but I'm not sure if this is better as it would add extra logic there. Also because it seems like the partition can be [modified by interceptors after registering](https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java#L2205) the search 
- Undo `read()` and additional query counts

### DIFF
--- a/hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/config/TestElasticsearchContainerHelper.java
+++ b/hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/config/TestElasticsearchContainerHelper.java
@@ -34,7 +34,7 @@ public class TestElasticsearchContainerHelper {
 	private static final Logger ourLog = LoggerFactory.getLogger(TestElasticsearchContainerHelper.class);
 
 
-	public static final String ELASTICSEARCH_VERSION = "7.17.3";
+	public static final String ELASTICSEARCH_VERSION = "7.17.25";
 	public static final String ELASTICSEARCH_IMAGE = "docker.elastic.co/elasticsearch/elasticsearch:" + ELASTICSEARCH_VERSION;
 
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

PR #7565, opened 2026-02-17
Fix: Handle versioned StructureDefinition URLs and ensure fallback to non-versioned URLs. 

### LINKED ISSUE #7564: DefaultProfileValidationSupportBundleStrategy cannot resolve versioned core profile canonicals
### Problem

The HL7 IG Publisher has started emitting **versioned canonical URLs** in `meta.profile`, e.g.: `http://hl7.org/fhir/StructureDefinition/Medication|4.0.1`

When validating such resources with HAPI FHIR (R4/R4B), this leads to lookup failures:
`Unable to resolve the profile reference http://hl7.org/fhir/StructureDefinition/Medication|4.0.1`

The reason is that `DefaultProfileValidationSupportBundleStrategy.fetchStructureDefinition()` currently only resolves **unversioned** canonical URLs, while the built-in StructureDefinitions are stored without the `|<version>` suffix.

---

### Expected Behavior

Versioned canonicals pointing to **HL7 base resources** should still resolve correctly, since the core resources shipped with HAPI are not keyed by version.

This is already the behavior implemented for CodeSystem/ValueSet resolution via `fetchCodeSystemOrValueSet()`.

patrick-werner:
# Fix for Versioned Profile URL Resolution

## Problem
#7564

### Implementation Details
The fix extracts the version from URLs containing a pipe character (`|`), then:
1. Looks up the resource using the non-versioned URL
2. For HL7 base URLs (`http://hl7.org/*` and `http://terminology.hl7.org/*`), the version is ignored (fallback behavior)
3. For custom/non-HL7 URLs, the version must match exactly

This matches the behavior already implemented for CodeSystem and ValueSet lookups.

### Files Modified
- `hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/DefaultProfileValidationSupportBundleStrategy.java`
  - Added version extraction logic to `fetchStructureDefinition()` method
  - Implements fallback to non-versioned URLs for HL7 base resources

### Test Added
- `hapi-fhir-validation/src/test/java/org/hl7/fhir/dstu3/hapi/validation/DefaultProfileValidationSupportR4Test.java`
  - Added `testLoadStructureDefinitionWithVersion()` test to verify versioned URL handling

## Behavior
- **HL7 base URLs**: Version is stripped, any version will resolve to the base resource
  - `http://hl7.org/fhir/StructureDefinition/Medication|4.0.1` → resolves to `Medication` SD
  - `http://hl7.org/fhir/StructureDefinition/Medication|999` → also resolves to `Medication` SD
- **Custom URLs**: Version must match exactly (or resource is not found)
  - `http://example.com/fhir/StructureDefinition/MyProfile|1.0.0` → only matches if version is 1.0.0

This approach is consistent with the JPA starter workaround but implemented directly in the core validation support.

fixes #7564

### DIFF
--- a/hapi-fhir-validation/src/test/java/org/hl7/fhir/r4/validation/DefaultProfileValidationSupportR4Test.java
+++ b/hapi-fhir-validation/src/test/java/org/hl7/fhir/r4/validation/DefaultProfileValidationSupportR4Test.java
@@ -1,4 +1,4 @@
-package org.hl7.fhir.dstu3.hapi.validation;
+package org.hl7.fhir.r4.validation;
 
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.context.support.DefaultProfileValidationSupport;
--- a/hapi-fhir-validation/src/test/java/org/hl7/fhir/r5/validation/DefaultProfileValidationSupportR5Test.java
+++ b/hapi-fhir-validation/src/test/java/org/hl7/fhir/r5/validation/DefaultProfileValidationSupportR5Test.java
@@ -1,4 +1,4 @@
-package org.hl7.fhir.dstu3.hapi.validation;
+package org.hl7.fhir.r5.validation;
 
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.context.support.DefaultProfileValidationSupport;


robogary:
Formatting check succeeded!

PR #7566, opened 2026-02-17
Add support for post-hoc case insensitive enum checking with LenientErrorHandler enabled

dnaworks:
This PR adds support for a post-hoc object enum creation and population when JSON mapping to an enum has been received but is the wrong case aka Active vs active. This PR will, if LenientErrorHandler is enabled and myErrorOnInvalidValue is set to false, allow for minor case errors in enums to successfully generate the HAPI objects.

Tests enabled for ensuring that StrictErrorHandler still throws on bad cased enums, and that loosened validation applies ONLY to case and not other


robogary:
Formatting check succeeded!

PR #7568, opened 2026-02-17
Fix: Parse query string parameters for POST _search requests

jmarchionatto:
Previously, query string parameters were only parsed for GET requests in RestfulServer. This meant POST /Resource/_search?param=value requests without a form-encoded body had their query string params silently ignored. Extend the condition to also parse query strings for POST requests while keeping PUT/DELETE/PATCH unchanged.


robogary:
Formatting check succeeded!

PR #7570, opened 2026-02-18
Exclude commons-logging from htmlunit to avoid spring-jcl conflict wa…

jmarchionatto:
Exclude commons-logging from htmlunit to avoid spring-jcl conflict warning

htmlunit 4.6.0 transitively pulls in commons-logging 1.3.4, which
conflicts with spring-jcl and causes a warning on stdout at startup.
Exclude it, matching the existing pattern for naming-java and json-lib.


robogary:
Formatting check succeeded!

PR #7576, opened 2026-02-19
Fix Batch2JobHelper to treat FAILED/CANCELLED jobs as terminal

jmarchionatto:
awaitNoJobsRunning() used `!= COMPLETED` which made it wait forever for FAILED/CANCELLED jobs since they never transition to COMPLETED. hasRunningJobs() used isIncomplete() which also considers FAILED as "still running". Both now use !isEnded() which correctly identifies COMPLETED, FAILED, and CANCELLED as terminal states.


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/Batch2JobHelper.java
```diff
@@ -258,8 +258,12 @@ public void awaitAllJobsOfJobDefinitionIdToComplete(String theJobDefinitionId) {
 			theJobDefinitionId,
 			BATCH_SIZE,
 			0);
-		// then await completion status
-		awaitJobCompletions(instances);
+		// Only wait for jobs that haven't reached a terminal state yet.
```
epeartree, line 267: Nice catch!

PR #7577, opened 2026-02-19
[7571] FHIR endpoint is not handling Consent resourceType on GET request

YalingPeiS:
### Problem Statement

When the `Populate Response Coding Displays` property was enabled on the FHIR endpoint, requests
to retrieve a Consent resource would fail with error "HAPI-0905: Unknown type: BOOLEAN" if the Consent contained a CodeSystem concept with a BOOLEAN-typed property

### Cause

When `Populate Response Coding Displays` is enabled in the FHIR endpoint, CDR attempts to generate displays for the encoded values based on the loaded code systems. However, the existing service did not handle Boolean type when process CodeSystem concept properties.

### Fix

Handled Boolean-typed term concept properties


robogary:
Formatting check succeeded!

PR #7581, opened 2026-02-21
SMILE-7421: Fix terminology pre-seed conflict (HAPI-0848)

fil512:
## Summary

Fixes SMILE-7421: When a CodeSystem is pre-seeded via package install (`STORE_AND_INSTALL`) and then the same CodeSystem URL is uploaded via `$upload-external-code-system`, the upload would fail with "Resource Already Exists" (`Msg.code(848)`).

**Root cause:** When the upload CodeSystem has an explicit ID (assigned by the loader), `TermVersionAdapterSvcR4.createOrUpdateCodeSystem()` performs a direct update by ID, creating a **new** FHIR resource instead of reusing the pre-seeded one. The existing `TermCodeSystemVersion` from the pre-seed still points to the old resource, and `checkForCodeSystemVersionDuplicate()` sees this resource ID mismatch and throws.

**Fix:** In `storeNewCodeSystemVersion()`, create a copy of the CodeSystem resource with the ID stripped (except for LOINC) and content set to `NOTPRESENT` before calling `createOrUpdateCodeSystem()`. This forces a conditional update by URL, ensuring the existing pre-seeded resource is reused. The `NOTPRESENT` content prevents the `storeNewCodeSystemVersionIfNeeded()` interceptor from creating a competing `TermCodeSystemVersion`. This approach is consistent with how existing SNOMED and LOINC loaders already work.

## Changes

- **`TermCodeSystemStorageSvcImpl.java`**: Create a `.copy()` of the CodeSystem before `createOrUpdateCodeSystem()` — strip ID (except LOINC), set content=NOTPRESENT, clear concepts. The caller's in-memory object is left unmodified.
- **`TermCodeSystemStorageSvcTest.java`**: Add `testPreSeedCodeSystemThenUploadFullTerminology()` — pre-seeds a CodeSystem via `dao.create()`, then uploads full terminology via `storeNewCodeSystemVersion()` with a different explicit ID. Verifies the upload succeeds, concepts are stored, and persisted content mode is NOTPRESENT.

## Review Tips

- The key change is in `TermCodeSystemStorageSvcImpl.storeNewCodeSystemVersion()` (resource overload, lines ~376-397)
- Compare the two storage entry points: package install path (`dao.create()` → `storeNewCodeSystemVersionIfNeeded()`) vs upload path (`storeNewCodeSystemVersion()` → `createOrUpdateCodeSystem()`)
- The LOINC exclusion (`!url.contains(LOINC_LOW)`) preserves existing LOINC behavior which requires an explicit ID

## QA Scenarios

- [ ] Pre-seed US Core 6.1.0 with `STORE_AND_INSTALL`, then upload SNOMED via `$upload-external-code-system` — should succeed without Msg.code(848)
- [ ] Upload a CodeSystem with no pre-seed — should work as before
- [ ] Upload LOINC terminology — should continue to use explicit ID path (LOINC excluded from ID stripping)
- [ ] Verify pre-seeded CodeSystem resource is reused (not duplicated) after upload

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
@@ -3,6 +3,7 @@ package ca.uhn.fhir.jpa.term;
 import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.jpa.entity.TermCodeSystem;
 import ca.uhn.fhir.jpa.entity.TermCodeSystemVersion;
+import ca.uhn.fhir.jpa.entity.TermConcept;
 import ca.uhn.fhir.jpa.model.dao.JpaPid;
 import ca.uhn.fhir.jpa.model.entity.ResourceTable;
 import ca.uhn.fhir.jpa.test.BaseJpaR4Test;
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

PR #7582, opened 2026-02-23
SMILE-7421: Fix NOTPRESENT CodeSystem duplicate resource conflict on pre-seed

fil512:
## Summary

This fix resolves a conflict that occurs during CodeSystem pre-seeding when the same NOTPRESENT CodeSystem URL is created via `dao.create()` more than once — a real-world scenario with `PackageInstallerSvc` where a second pre-seed run or `$upload-external-code-system` creates a new FHIR resource for a URL that already has a `TermCodeSystemVersion` entry.

1. In `TermCodeSystemStorageSvcImpl.storeNewCodeSystemVersionIfNeeded()`, before the duplicate-check guard rail runs in the NOTPRESENT early-return path, the fix checks whether the existing `TermCodeSystemVersion` has zero concepts. If it does (it is a placeholder), the version's resource pointer is updated to the incoming resource so the downstream `checkForCodeSystemVersionDuplicate()` call sees matching resources and does not throw HAPI-0848.
2. Four integration tests cover the bug scenarios and guard rails: unversioned NOTPRESENT placeholder re-point (was failing), versioned NOTPRESENT placeholder re-point (was failing), different-version NOTPRESENT creates separate entries (regression guard), and COMPLETE CodeSystem with concepts still rejects a duplicate (guard rail preserved).
3. The production change is minimal — 7 lines inserted into the NOTPRESENT early-return branch, with no changes to the duplicate check logic itself or any other code path.

## Code Review Suggestions

- [ ] `myConceptDao.countByCodeSystemVersion(codeSystemVersion.getPid()) == 0` — this issues a COUNT query inside a code path that is already inside a transaction; confirm the TermConceptDao method does not cause an N+1 or flush issue when called from saveAllDeferred() context.
- [ ] `codeSystemVersion.setResource(theResourceEntity)` followed by `myCodeSystemVersionDao.saveAndFlush()` — the flush is intentional so the downstream `getOrCreateDistinctTermCodeSystem()` call sees the updated row.
- [ ] The fix only runs when `codeSystemVersion != null` in the NOTPRESENT block. Confirm the null-else branch (new version creation) is still correct.
- [ ] The guard `conceptCount == 0` is the sole criteria for "safe to re-point." Confirm there is no edge case where a 0-concept TermCodeSystemVersion should NOT be re-pointed.

## QA Test Suggestions

- [ ] POST a NOTPRESENT CodeSystem with `url: http://snomed.info/sct`. Then POST a second NOTPRESENT CodeSystem with the same URL. The second POST should return HTTP 201.
- [ ] Repeat with explicit `version: "5.0.0"` on both resources. Both creates should succeed.
- [ ] POST a COMPLETE CodeSystem with concepts, then POST a NOTPRESENT CodeSystem with the same URL. Verify the second POST is rejected with HAPI-0848.
- [ ] Simulate a PackageInstallerSvc double-seed scenario and verify no error on the second install.


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
```diff
@@ -88,6 +90,191 @@ public void testStoreNewCodeSystemVersionForExistingCodeSystemVersionId() {
 
 	}
 
+	// Created by claude-opus-4-6
+	@Test
+	void storeNewCodeSystemVersionIfNeeded_notPresentPlaceholderWithDifferentResource_shouldUpdatePlaceholder() {
```
epeartree, line 95: i dont' think smile stuff should go in hapi comments.  please verify

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
```diff
@@ -88,6 +90,191 @@ public void testStoreNewCodeSystemVersionForExistingCodeSystemVersionId() {
...
+		// Reproduces SMILE-7421: When a NOTPRESENT CodeSystem already exists (e.g. from package
+		// pre-seeding), creating a second FHIR resource for the same CodeSystem URL should succeed
+		// because the existing TermCodeSystemVersion is a 0-concept placeholder.
+		//
```
epeartree, line 99: the explanation is very much welcomed but method names should be omitted as they may chance with time and the comment will become impossible to understand.  consider removing method names

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
```diff
@@ -88,6 +90,191 @@ public void testStoreNewCodeSystemVersionForExistingCodeSystemVersionId() {
...
+		runInTransaction(() -> {
+			TermCodeSystem tcs = myTermCodeSystemDao.findByCodeSystemUri("http://snomed.info/sct");
+			assertThat(tcs).isNotNull();
+			assertThat(tcs.getCurrentVersion()).isNotNull();
+			long conceptCount = myTermConceptDao.countByCodeSystemVersion(tcs.getCurrentVersion().getPid());
+			assertThat(conceptCount).isEqualTo(0);
+		});
+
+		// Step 2: Create a second NOTPRESENT CodeSystem with the same URL.
+		// This simulates the scenario where a new FHIR resource is created for the same
```
epeartree, line 126: Remove explicit method names in the comments, since those method names can change in the future. The comment is still useful, but should refer to operations in general terms rather than specific method names

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
```diff
@@ -88,6 +90,191 @@ public void testStoreNewCodeSystemVersionForExistingCodeSystemVersionId() {
...
+		// subsequent pre-seed run when conditional search fails to find the existing resource,
+		// or $upload-external-code-system's createOrUpdateCodeSystem() creating a new resource).
+		CodeSystem cs2 = new CodeSystem();
+		cs2.setUrl("http://snomed.info/sct");
+		cs2.setContent(CodeSystem.CodeSystemContentMode.NOTPRESENT);
+		cs2.setStatus(Enumerations.PublicationStatus.ACTIVE);
+
+		// This should succeed — the existing TermCodeSystemVersion is a 0-concept placeholder
+		// from the first pre-seed. It should be updated to point to the new resource.
+		// Previously threw UnprocessableEntityException (Msg.code(848)) — fixed by SMILE-7421.
```
epeartree, line 136: reference to smile

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
```diff
@@ -88,6 +90,191 @@ public void testStoreNewCodeSystemVersionForExistingCodeSystemVersionId() {
...
+			assertThat(conceptCount).isEqualTo(0);
+		});
+
+		// Step 2: Create second versioned NOTPRESENT CodeSystem with same URL and version
+		CodeSystem cs2 = new CodeSystem();
+		cs2.setUrl("http://snomed.info/sct");
+		cs2.setVersion("5.0.0");
+		cs2.setContent(CodeSystem.CodeSystemContentMode.NOTPRESENT);
+		cs2.setStatus(Enumerations.PublicationStatus.ACTIVE);
+
```
epeartree, line 182: referencing smile

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
```diff
@@ -88,6 +90,191 @@ public void testStoreNewCodeSystemVersionForExistingCodeSystemVersionId() {
...
+			assertThat(myTermCodeSystemVersionDao.findByCodeSystemPidVersionIsNull(tcs.getPid())).isNotNull();
+			assertThat(myTermCodeSystemVersionDao.count()).isEqualTo(2);
+		});
+	}
+
+	// Created by claude-opus-4-6
+	@Test
+	void storeNewCodeSystemVersionIfNeeded_completeCodeSystemThenNotPresentDifferentResource_shouldRejectDuplicate() {
+		// Guard rail: When a COMPLETE CodeSystem with real concepts already exists,
+		// creating a second NOTPRESENT CodeSystem with a different FHIR resource for the
```
epeartree, line 238: referencing smile

epeartree (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TermCodeSystemStorageSvcTest.java
fil512: let's keep it for traceability

fil512:
Thanks for the feedback! Removed explicit method names from the test comments — they now refer to operations in general terms. Pushed in the latest commit.

PR #7584, opened 2026-02-23
SMILE-8708: Handle non-numeric ETag in If-Match header

fil512:
## Summary

This fix resolves a server crash when a client sends an `If-Match` header containing a non-numeric version string (e.g., `W/"abc"` or `W/"A23"`) on update or delete requests. Previously, both code paths called `Long.parseLong()` directly without first validating the version string, causing a `NumberFormatException` that propagated as a 500 Internal Server Error. The fix introduces an `isVersionIdPartValidLong()` guard before the parse in both affected paths, so a non-numeric ETag version now correctly short-circuits to a 409 Conflict response — consistent with how a numeric-but-wrong version is already handled.

### Changes

1. `BaseStorageResourceDao.doUpdateForUpdateOrPatch()` — guarded the `Long.parseLong()` call with `isVersionIdPartValidLong()` so non-numeric versions are treated as a version mismatch (update and patch path).
2. `BaseHapiFhirResourceDao` delete path — same guard applied to the delete version-check block.
3. `ResourceProviderR4Test` — integration tests added for the non-numeric ETag update case, plus adjacent regression tests for wrong-numeric and correct ETags on update.
4. `FhirResourceDaoR4DeleteTest` — unit-level test added for the non-numeric ETag delete case.
5. Changelog entry added under `8_10_0`.

## Code Review Suggestions

- [ ] **Logic inversion in `doUpdateForUpdateOrPatch`**: The original code threw on `versionPart != entity.getVersion()`. The new code computes `versionMatches = isValidLong && parsedVersion == entity.getVersion()` and throws when `!versionMatches`. Verify that a non-numeric version correctly evaluates `isVersionIdPartValidLong()` to `false`, making `versionMatches = false`, and therefore always throws — this is the intended behavior.
- [ ] **Patch path coverage**: `doUpdateForUpdateOrPatch` is shared by both `update` and `patch` operations. Verify the patch path also exercises this guard correctly — no separate patch test is included.
- [ ] **`isVersionIdPartValidLong()` contract**: This method is called on `IIdType` — confirm its behavior for `null`, empty string, and leading/trailing whitespace edge cases so the guard is truly exhaustive.

## QA Test Suggestions

- [ ] **Non-numeric ETag on PUT (update)**: `PUT /Patient/{id}` with `If-Match: W/"abc"` — confirm response is `409 Conflict`, not `500`.
- [ ] **Non-numeric ETag on DELETE**: `DELETE /Patient/{id}` with `If-Match: W/"xyz"` — confirm response is `409 Conflict`, not `500`.
- [ ] **Wrong numeric ETag (regression)**: `PUT /Patient/{id}` with `If-Match: W/"999"` — confirm still returns `409 Conflict`.
- [ ] **Correct ETag on PUT (regression)**: `PUT /Patient/{id}` with `If-Match: W/"1"` — confirm returns `200 OK`.
- [ ] **PATCH with non-numeric ETag**: `PATCH /Patient/{id}` with `If-Match: W/"A23"` — confirm `409 Conflict` rather than `500`.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-8708-non-numeric-etag.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-8708-non-numeric-etag.yaml
@@ -1,0 +1,6 @@
+---
+type: fix
+jira: SMILE-8708
+title: "Previously, supplying a non-numeric version in an If-Match header (e.g. If-Match: W/\"abc\") on an
+  update or delete request would result in a 500 Internal Server Error due to a failed integer parse.
+  Non-numeric ETag values are now handled gracefully and correctly return a 409 Conflict response."


robogary:
Formatting check succeeded!

PR #7588, opened 2026-02-24
SMILE-9739: Fix incorrect default REST-hook endpoint URL validation regex

fil512:
## Summary

Fix for the default REST-hook subscription endpoint URL validation regex which incorrectly allowed URLs with `htt://` scheme. The regex alternation `http?|https?` was parsed as `htt` + optional `p` OR `http` + optional `s`, so `htt://` matched. The corrected regex `https?://[...]` properly requires `http://` or `https://`.

- `BaseSubscriptionSettings.java` — `DEFAULT_RESTHOOK_ENDPOINTURL_VALIDATION_REGEX` replaced with `https?://[-%()_.!~*';/?:@&=+$,A-Za-z0-9]+`
- `RestHookChannelValidatorTest.java` — Two new test cases for `htt://` and `httpx://` URLs

Closes https://gitlab.com/simpatico.ai/cdr/-/issues/7150

## Test plan

- [x] Reproduction test confirms `htt://example.com:8080/fhir` is now rejected
- [x] All 284 subscription module tests pass
- [x] All 13 downstream `SubscriptionValidatingInterceptorTest` tests pass
- [x] Spotless clean, checkstyle clean (no new violations)

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7150-fix-resthook-url-regex.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7150-fix-resthook-url-regex.yaml
@@ -1,6 +1,9 @@
 ---
 type: fix
+issue: 7150
 jira: SMILE-9739
 title: "Previously, the default REST-hook endpoint URL validation regex allowed malformed URLs
   such as those beginning with htt:// because the p in http was treated as optional. The
-  regex has been corrected to properly require http:// or https:// as the URL scheme."
+  regex has been corrected to properly require http:// or https:// as the URL scheme.
+  Additionally, fragment identifiers (#) and IPv6 bracket notation ([ ]) are now accepted
+  in the default regex."
--- a/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/config/BaseSubscriptionSettings.java
+++ b/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/config/BaseSubscriptionSettings.java
@@ -33,7 +33,7 @@ public abstract class BaseSubscriptionSettings {
 	public static final String DEFAULT_EMAIL_FROM_ADDRESS = "noreply@unknown.com";
 	public static final String DEFAULT_WEBSOCKET_CONTEXT_PATH = "/websocket";
 	public static final String DEFAULT_RESTHOOK_ENDPOINTURL_VALIDATION_REGEX =
-			"https?://[-%()_.!~*';/?:@&=+$,A-Za-z0-9]+";
+			"https?://[-%()_.!~*';/?:@&=+$,#A-Za-z0-9\\[\\]]+";
 	public static final long DEFAULT_SUBMISSION_INTERVAL_IN_MS = 5000;
 
 	private final Set<Subscription.SubscriptionChannelType> mySupportedSubscriptionTypes = new HashSet<>();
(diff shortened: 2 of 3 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7150-fix-resthook-url-regex.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
```
jmarchionatto, line 2: question: shouldn't GL Issue be used in file name?

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/config/BaseSubscriptionSettings.java
```diff
@@ -33,7 +33,7 @@ public abstract class BaseSubscriptionSettings {
 	public static final String DEFAULT_EMAIL_FROM_ADDRESS = "noreply@unknown.com";
 	public static final String DEFAULT_WEBSOCKET_CONTEXT_PATH = "/websocket";
 	public static final String DEFAULT_RESTHOOK_ENDPOINTURL_VALIDATION_REGEX =
-			"((((http?|https?)://))([-%()_.!~*';/?:@&=+$,A-Za-z0-9])+)";
+			"https?://[-%()_.!~*';/?:@&=+$,A-Za-z0-9]+";
```
jmarchionatto: ### Regex rejects URLs with `#` fragments and `[` `]` for IPv6

The character class `[-%()_.!~*';/?:@&=+$,A-Za-z0-9]` is missing:
- `#` — fragment identifiers (e.g., `https://example.com/path#section`)
- `[` and `]` — IPv6 host addresses (e.g., `https://[::1]:8080/callback`)
- `:` — port numbers (e.g., `https://example.com:8443/hook`)

The **`:` omission is a real problem** — any REST-hook URL with an explicit port will be rejected by the default regex. URLs like `https://internal-service:8080/webhook` are a common pattern in enterprise deployments.

**Note:** This pre-existed in the old regex. However, since the regex is being fixed anyway, this is the right time to address it. At minimum, add `:` to the character class. Consider also adding `#` and `[]`.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7150-fix-resthook-url-regex.yaml
fil512: Fixed. Renamed the changelog file to `7150-fix-resthook-url-regex.yaml` using the GitLab issue number as prefix, and added both `issue: 7150` and `jira: SMILE-9739` fields to match the convention used by other dual-tracked changelog entries (e.g., `7550-fix-patient-encounter-everything-nonexistent-resource.yaml`).

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/config/BaseSubscriptionSettings.java
fil512: Fixed. Added `#` (fragment identifiers) and `[]` (IPv6 bracket notation) to the regex character class. Note that `:` was already present in the character class (between `?` and `@`), so port numbers like `https://internal-service:8080/webhook` were already supported -- added an explicit test case to confirm this. New test cases cover fragment URLs, IPv6 addresses, and port-based URLs. All 29 tests pass.

PR #7592, opened 2026-02-25
SMILE-11707: Fix bundle validation profile match error for bundle-internal references

fil512:
## Summary

This PR fixes a false-positive validation error that occurred when validating a FHIR Bundle containing cross-referencing resources against a custom Implementation Guide that constrains reference target profiles. The underlying `InstanceValidator` in `org.hl7.fhir.core` hardcodes `CHECK_VALID` for bundle-internal and contained references, bypassing the `policyForReference` method on the policy advisor entirely — so even when the server was configured with `IGNORE` or another non-validating policy, profile-match checks were silently forced on bundle-internal references.

### Changes

1. **`FhirDefaultPolicyAdvisor.isSuppressMessageId`** now suppresses `Reference_REF_CantMatchChoice` when `getReferencePolicy()` is not `CHECK_VALID`, working around the hardcoded behaviour in the core validator.
2. **`ValidatorPolicyAdvisor.isSuppressMessageId`** applies the same suppression logic keyed on `ValidationSettings.getLocalReferenceValidationDefaultPolicy()`, consistent with how `policyForReference` already uses that setting.
3. **5 new integration tests** in `FhirResourceDaoR4ValidateTest` cover the primary failure scenario and four adjacent cases (individual resource validation, no custom IG, base-profile-only constraints, and explicit CHECK_VALID policy negative test).

## Code Review Tips

- The fix is a targeted suppression via `isSuppressMessageId` rather than a core validator patch — the core library (`org.hl7.fhir.core`) is an external dependency we do not own.
- The suppression only fires when the configured reference policy would NOT have resulted in `CHECK_VALID`. Users who explicitly set `CHECK_VALID` continue to see profile match errors.
- The `policyForContained` method still returns `CHECK_VALID` but any resulting `Reference_REF_CantMatchChoice` messages are also covered by the `isSuppressMessageId` hook.

## QA Scenarios

- [ ] Validate a batch Bundle with cross-referencing resources + custom IG (policy=IGNORE) → no profile match error
- [ ] Same bundle with policy=CHECK_VALID → profile match error IS reported
- [ ] Validate resources individually (not in bundle) → no error
- [ ] Bundle without custom IG loaded → no error
- [ ] Bundle where referenced resource satisfies the custom profile → no error

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11707-bundle-validation-profile-match.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11707-bundle-validation-profile-match.yaml
@@ -1,0 +1,9 @@
+---
+type: fix
+jira: SMILE-11707
+title: "Previously, validating a Bundle containing cross-referencing resources could incorrectly fail with
+  \"Unable to find a profile match\" errors when a custom Implementation Guide constrained reference target
+  profiles and the configured reference validation policy was not CHECK_VALID. This was caused by the
+  underlying validator hardcoding CHECK_VALID for bundle-internal references, bypassing the configured
+  policy. This has been fixed so that profile match errors for bundle-internal references are suppressed
+  when the configured reference policy would not have validated those references."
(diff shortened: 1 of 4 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11707-bundle-validation-profile-match.yaml
```diff
@@ -0,0 +1,9 @@
+---
+type: fix
+jira: SMILE-11707
```
epeartree, line 3: i'm under the impression that changelog require the 'issue' property

fil512:
Closing in favor of a v2 PR with improved test structure (nested class with shared helpers, extra specialty test case), corrected changelog (jira property), and code review refinements (I18nConstants constant). The production fix is identical.

PR #7509, opened 2026-01-29
  Transform relative URLs in search parameter documentation to absolu…

jkiddo:
…te HL7 FHIR URLs

  Search parameter descriptions from the FHIR spec contain relative markdown
  links like [CarePlan](careplan.html) which don't resolve correctly when
  displayed in capability statements outside the HL7 website context.

  This change transforms these relative links to absolute URLs pointing to
  the correct FHIR version-specific HL7 specification, e.g.:
  [CarePlan](careplan.html) -> [CarePlan](https://hl7.org/fhir/R4/careplan.html)

  The transformation is FHIR version-aware, using the appropriate base URL
  for each version (DSTU2, STU3, R4, R4B, R5).

This relates to https://github.com/hapifhir/hapi-fhir/issues/7508

PR #7593, opened 2026-02-26
Fix/adjust generated links

jamesagnew:
Adds credit for #7509


robogary:
Formatting check succeeded!

PR #7594, opened 2026-02-26
Enable clients to mark more resources, such as Measure, as non-partit…

lukedegruchy:
…ionable.


robogary:
Formatting check succeeded!

PR #7595, opened 2026-02-26
SMILE-11707: Fix bundle validation profile match error for bundle-internal references

fil512:
## Summary

This PR fixes a regression introduced in HAPI-FHIR 8.8.0 (CDR 2026.02) where bundle validation incorrectly produces "Unable to find a profile match" errors for bundle-internal references even when the server's local reference validation policy is configured as IGNORE. The root cause is that `InstanceValidator` in `org.hl7.fhir.core` hardcodes `CHECK_VALID` for INTERNAL (bundle-internal) and contained references, entirely bypassing the `policyForReference` hook exposed by `IValidationPolicyAdvisor`. The fix works around this by implementing `isSuppressMessageId` in both `ValidatorPolicyAdvisor` (JPA server layer) and `FhirDefaultPolicyAdvisor` (validation library layer) to suppress the `REFERENCE_REF_CANTMATCHCHOICE` message when the configured policy would not have performed profile matching in the first place.

1. `ValidatorPolicyAdvisor.isSuppressMessageId` now suppresses `REFERENCE_REF_CANTMATCHCHOICE` when `localReferenceValidationDefaultPolicy` is not `CHECK_VALID`, aligning bundle-internal reference behavior with the server's configured reference policy.
2. `FhirDefaultPolicyAdvisor.isSuppressMessageId` applies the same suppression logic keyed on `getReferencePolicy()`, ensuring standalone (non-JPA) validator usage is also corrected.
3. Six integration-style tests are added to `FhirResourceDaoR4ValidateTest` covering: custom IG profile with constrained reference targets, with/without specialty, external reference, no custom profiles, base-profile constraint, and explicit `CHECK_VALID` opt-in.

**Version Analysis:** Confirmed regression — CDR 2025.05.R01 (HAPI 8.2.0) passes bundle validation without error; CDR 2026.02.R01 (HAPI >= 8.8.0) fails with "Unable to find a profile match" on the same payload.

---

## Code Review Suggestions

- [ ] **Suppression scope**: The suppression fires for any `REFERENCE_REF_CANTMATCHCHOICE` message whenever the policy is not `CHECK_VALID`. If a future `org.hl7.fhir.core` release emits this message ID for an external reference that genuinely failed a profile constraint, the suppression would silently hide it.
- [ ] **Two parallel fix sites**: Confirm that both `FhirDefaultPolicyAdvisor` and `ValidatorPolicyAdvisor` need the fix, and that there is no code path that hits both.
- [ ] **`CHECK_VALID` test confirms errors are produced**: The guard rail test verifies suppression is correctly gated.
- [ ] **`I18nConstants.REFERENCE_REF_CANTMATCHCHOICE` stability**: Confirm this constant won't be renamed in a near-term `org.hl7.fhir.core` upgrade.

---

## QA Test Suggestions

- [ ] Validate a batch Bundle with cross-referencing resources + custom IG (policy=IGNORE) → no profile match error
- [ ] Same bundle with policy=CHECK_VALID → profile match error IS reported
- [ ] Validate resources individually (not in bundle) → no error
- [ ] Bundle without custom IG loaded → no error
- [ ] Bundle where referenced resource satisfies the custom profile → no error

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11707-bundle-validation-profile-match-v2.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11707-bundle-validation-profile-match-v2.yaml
@@ -1,5 +1,6 @@
 ---
 type: fix
+issue: 7602
 jira: SMILE-11707
 title: "Previously, validating a FHIR Bundle containing cross-referencing resources against a custom
   Implementation Guide that constrains reference target profiles would produce false-positive


robogary:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11707-bundle-validation-profile-match-v2.yaml
```diff
@@ -0,0 +1,11 @@
+---
+type: fix
+jira: SMILE-11707
```
epeartree, line 4: this PR is missing an 'issue'

fil512: Created GitHub issue #7602 and added the `issue: 7602` field to the changelog. Fixed in cf47f89.

PR #7596, opened 2026-02-27
SMILE-11603: Fix batch2 gated job advancement race condition

fil512:
## Summary

This PR fixes a race condition in the batch2 gated job framework where work chunks produced by slow workers — finishing after a gated step has already advanced — could be permanently stuck in `GATE_WAITING` status and never processed.

This is iteration 3. Iteration 2 attempted to fix this with a per-chunk DB read in `getOnCreateStatus` (looking up `currentGatedStepId` at creation time). That approach was removed because it added a DB read per chunk creation and still had a narrow race window. Iteration 3 instead uses two mechanisms that together eliminate the race without the extra DB read:

1. **Advancement-time flip in `advanceJobStepAndUpdateChunkStatus`** (`JpaJobPersistenceImpl`): When the job advances to a new step, the existing code that flips `GATE_WAITING` chunks to `READY` is now also called as a private helper (`flipGateWaitingChunks`) at advancement time. Previously the flip only happened in the maintenance path; now it also happens atomically at the moment of advancement, covering any chunks already in the DB for that step.

2. **Maintenance safety net via `enqueueGateWaitingChunksForCurrentStep`** (`IJobPersistence` + `JpaJobPersistenceImpl`): A new interface method flips any remaining `GATE_WAITING` (and legacy `QUEUED`) chunks for the current gated step to `READY` (or `REDUCTION_READY`). Called on every maintenance run from `JobInstanceProcessor`, this catches late-arriving chunks created by slow workers after advancement — chunks that start as `GATE_WAITING` because the creation path always uses `GATE_WAITING` for gated jobs and no longer reads the DB.

**Key files changed:**
- `IJobPersistence.java` — new `enqueueGateWaitingChunksForCurrentStep` interface method (default impl returns 0 for backward compatibility)
- `JpaJobPersistenceImpl.java` — extracted `flipGateWaitingChunks` private helper; called at advancement time and exposed via the new interface method; `getOnCreateStatus` reverted to `static` (no per-chunk DB read)
- `JobInstanceProcessor.java` — calls `enqueueGateWaitingChunksForCurrentStep` on every maintenance pass for running gated jobs
- `JpaJobPersistenceImplTest.java` — five new integration tests covering the fix
- `Batch2JobMaintenanceDatabaseIT.java` — updated test to reflect that a `QUEUED` chunk now gets flipped to `READY` by maintenance

## Code Review Suggestions

- [ ] `flipGateWaitingChunks` is called from `advanceJobStepAndUpdateChunkStatus` as a direct private call (not via `this.enqueueGateWaitingChunksForCurrentStep`) — this is intentional to avoid Spring AOP self-invocation bypassing `@Transactional(REQUIRES_NEW)`. Confirm the comment explains this clearly enough.
- [ ] `enqueueGateWaitingChunksForCurrentStep` is a default method on the interface returning 0 — any alternative persistence implementations will silently be a no-op. Consider whether that is acceptable or whether a warning log should be emitted.
- [ ] The `@Transactional(propagation = Propagation.REQUIRES_NEW)` on `enqueueGateWaitingChunksForCurrentStep` in `IJobPersistence` matches the annotation on `advanceJobStepAndUpdateChunkStatus` — verify consistency is intentional.
- [ ] `JobInstanceProcessor` calls the new method inside the `hasGatedStep() && isRunning()` guard — confirm no gated-but-not-running scenario could leave chunks stranded.
- [ ] The legacy `QUEUED` TODO comment (remove after 7.6.0) is now in `flipGateWaitingChunks` — make sure it's still visible and tracked.

### Setup
- Deploy a build with this fix against a real HAPI JPA server.
- Use a gated batch2 job definition with at least two steps (e.g., a bulk export or reindex).

### Scenarios
- [ ] Run a gated batch2 job end-to-end — all chunks should complete and no chunks should remain stuck in `GATE_WAITING` after the job finishes.
- [ ] Simulate a slow worker by pausing a worker thread after step 1 starts but before it produces its output chunks, then let step advancement happen, then resume the worker — verify the late-produced chunks are eventually flipped to `READY` by the next maintenance run.
- [ ] Confirm a non-gated batch job is unaffected — chunks are created as `READY` immediately.
- [ ] Confirm a reduction-step gated job works correctly — late `GATE_WAITING` chunks for the reduction step should flip to `REDUCTION_READY` (not `READY`) by maintenance.
- [ ] After applying the fix to an environment with pre-existing stuck `GATE_WAITING` chunks, verify the maintenance run rescues them without manual intervention.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7596-fix-batch2-gated-job-advancement-race-condition.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7596-fix-batch2-gated-job-advancement-race-condition.yaml
@@ -4,8 +4,7 @@ issue: 7596
 jira: SMILE-11603
 title: "Fixed a race condition in the batch2 framework where work chunks produced by slow workers
   after a gated job step had already advanced could be permanently stuck in GATE_WAITING status.
-  Work chunks for gated jobs now check the job instance's current gated step at creation time
-  and start in READY status when they target the step already being processed, preventing orphaned
-  chunks in the common case. The maintenance run retains a safety net that flips any remaining
-  GATE_WAITING chunks for the current step to READY on each pass, covering the narrow race window
-  between step advancement and chunk creation."
+  Step advancement now atomically flips existing GATE_WAITING chunks for the new current step to
+  READY. Additionally, the maintenance run flips any remaining GATE_WAITING chunks for the current
+  step to READY on each pass, catching late-arriving chunks produced by slow workers after
+  advancement."
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
@@ -141,26 +141,17 @@ public class JpaJobPersistenceImpl implements IJobPersistence {
 	/**
 	 * Gets the initial onCreate state for the given workchunk.
 	 * For non-gated jobs, chunks always start in READY.
-	 * For gated jobs, chunks targeting the current gated step start in READY (they are already
-	 * eligible for processing). Chunks targeting a future step start in GATE_WAITING and will be
-	 * transitioned to READY during maintenance pass when all chunks in the previous step are COMPLETED.
-	 * This handles the common late-arrival race condition where a slow worker produces a chunk for
-	 * the current step after advancement has already occurred.
+	 * For gated jobs, chunks always start in GATE_WAITING and will be transitioned to READY
+	 * when the job advances to their target step. Late-arriving chunks (produced by slow workers
+	 * after advancement) are caught by {@link #enqueueGateWaitingChunksForCurrentStep} on the
+	 * next maintenance pass.
 	 */
-	private WorkChunkStatusEnum getOnCreateStatus(WorkChunkCreateEvent theBatchWorkChunk) {
-		if (!theBatchWorkChunk.isGatedExecution) {
+	private static WorkChunkStatusEnum getOnCreateStatus(WorkChunkCreateEvent theBatchWorkChunk) {
+		if (theBatchWorkChunk.isGatedExecution) {
+			return WorkChunkStatusEnum.GATE_WAITING;
+		} else {
 			return WorkChunkStatusEnum.READY;
 		}
-
-		// Check if this chunk targets the step that the job has already advanced to.
-		// If so, the chunk is immediately eligible for processing — no need to wait.
-		Batch2JobInstanceEntity instance =
-				myEntityManager.find(Batch2JobInstanceEntity.class, theBatchWorkChunk.instanceId);
-		if (instance != null && theBatchWorkChunk.targetStepId.equals(instance.getCurrentGatedStepId())) {
-			return WorkChunkStatusEnum.READY;
-		}
-
-		return WorkChunkStatusEnum.GATE_WAITING;
 	}
 
 	@Override
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_batch/batch2_states.md
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_batch/batch2_states.md
@@ -59,8 +59,8 @@ stateDiagram-v2
     state FAILED
     state COMPLETED
    direction LR
-   [*]             --> READY            : on create - non-gated jobs, or gated jobs targeting the current step
-   [*]             --> GATE_WAITING     : on create - gated jobs targeting a future step
+   [*]             --> READY            : on create - non-gated jobs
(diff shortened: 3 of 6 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -130,6 +130,27 @@ public String onWorkChunkCreate(WorkChunkCreateEvent theBatchWorkChunk) {
...
 		entity.setStatus(getOnCreateStatus(theBatchWorkChunk));
 
+		// If the chunk would be GATE_WAITING, check if the job has already advanced to this step.
+		// Late-arriving chunks (produced by slow workers after advanceJobStepAndUpdateChunkStatus has
+		// already flipped existing chunks to READY) would otherwise be stuck in GATE_WAITING forever.
+		// This should technically never happen, but we have seen it at customer sites so we add this guard to cover
+		// this case just in case the system somehow gets into this invalid state.
+		if (entity.getStatus() == WorkChunkStatusEnum.GATE_WAITING) {
+			Optional<Batch2JobInstanceEntity> instanceOpt =
+					myJobInstanceRepository.findById(theBatchWorkChunk.instanceId);
```
michaelabuckley: performance: this adds a DB round trip to every chunk write.  
Instead, what do you think about Ellie's suggested fix: have the maintenance run flip chunks from gate_watiting -> ready for completed steps.

fil512: Where can I find Ellie's suggested fix? I want to ensure I understand the proposal clearly.

fil512: Do we hunt for these blocked jobs with every maintenance pass?  I.e. at end of maintenance pass: identify any job that would be blocked because it is in this inconsistent state, and then clean it up?

fil512: This is what needs to be done:

The current behaviour has a single method advanceJobStepAndUpdateChunkStatus - [https://github.com/hapifhir/hapi-fhir/blob/6526b89e8e849cc0ef964524e675f5feced7283[…]src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java](https://github.com/hapifhir/hapi-fhir/blob/6526b89e8e849cc0ef964524e675f5feced7283b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java#L654-L654)

This advances the job step (in this case from loadIds->loadResources), and flips all chunks in loadIds from GATE_WAITING->READY.

What we need to do is to separate those.  Have one database transaction advance the job step.  This only happens once per step.
Then in a separate transaction, on every maintenance run, flip all chunks for complete steps from GATE_WAITING->READY.

In normal operation, this would be a no-op after the first run per step. BUT it would also catch these failure cases.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7596-fix-batch2-gated-job-advancement-race-condition.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: fix
+issue: 7596
+jira: SMILE-11603
+title: "Fixed a race condition in the batch2 framework where work chunks produced by slow workers
```
fil512, line 5: Claude: Don't merge this MR until I've done a final review of the most recent changes

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
fil512: Fixed in latest push.

Implemented the approach described by @fil512:

1. **Removed the extra DB round trip** from `onWorkChunkCreate` -- chunks are no longer checked against the job instance on every write.

2. **Separated step advancement from chunk flipping** -- `advanceJobStepAndUpdateChunkStatus` still advances the step and flips existing GATE_WAITING chunks in the same call (for immediate processing of chunks that are already there).

3. **Added `enqueueGateWaitingChunksForCurrentStep`** -- a new method that flips GATE_WAITING/QUEUED chunks for the current gated step to READY. This is called:
   - By `advanceJobStepAndUpdateChunkStatus` for the initial flip when the step advances
   - By `JobInstanceProcessor.process()` on **every maintenance run** for the current gated step, catching any late-arriving chunks

In normal operation the maintenance call is a no-op (all chunks were already flipped during advancement). But if a slow worker produces a chunk after advancement, it will be created as GATE_WAITING and the next maintenance run will flip it to READY.

fil512:
replaced by https://github.com/hapifhir/hapi-fhir/pull/7636

PR #7598, opened 2026-02-27
Fix search in Patient Id when query references are mixed between Patient and non-Patient

michaelabuckley:
Search by subject fails to find resources on the default partition when the subject list is mixed between Patient and non-Patient references.

Replicated in attached changes.

This likely never happens in real query loads, but we should fix it, or throw an error as unsupported.


robogary:

Issue #7319, opened 2025-10-22
Calling newJsonParser().encodeToString() on a FhirContext object passing in a Parameters object causes an exception


cw85865:
**Describe the bug**
Calling `newJsonParser().encodeToString()` on a `FhirContext` object passing in an `org.hl7.fhir.r4.model.Parameters` object causes an exception. This does not happen on every `Parameters` object, but only on specific ones that include a `RequestGroup` and a `Bundle` that contains the same `RequestGroup` along with a `ServiceRequest`. What seems to be happening is that at line 1534 of `FhirTerser`, the call to `getContainedResourceList()` is returning the default, immutable empty list. Following the code through into `getContainedResourceList()`, it initializes `containedResources` to `Collections.emptyList()`, but in our case neither of the if-else if statements are executed (i.e. it doesn't detect the `Parameters` object as being an instance of `IResource` nor `IDomainResource`), therefore it returns the immutable empty list. Back at line 1534, it then attempts to call `add()` on the immutable empty list, which causes an exception to be thrown.

There will be an attached zip file that contains a binary file to load in the test. I know this may give you pause, but it was the easiest way to show how to reproduce the issue without having to spend hours and hours trying to manually construct the `Parameters` object in the test code since the object is an input to API handler code. Also, the data is from a development environment and does not contain any real PHI.

**To Reproduce**
Use the following Groovy test along with the attached binary `Parameters` object file to reproduce (ensure that you extract the zip archive and replace the path to the file with the appropriate path to the .bin file):
```groovy
import ca.uhn.fhir.context.FhirContext
import org.hl7.fhir.r4.model.Bundle
import org.hl7.fhir.r4.model.Parameters
import org.hl7.fhir.r4.model.Reference
import org.hl7.fhir.r4.model.RequestGroup
import spock.lang.Specification

class WorkflowServiceSpec extends Specification {

  def "Reproduce Parameters JSON encoding exception"() {
    given:
    def VERSIONED_PATHS = List.of("Provenance.target")
    def ctx = FhirContext.forR4()
    ctx.setParserOptions(ctx.getParserOptions().setDontStripVersionsFromReferencesAtPaths(VERSIONED_PATHS))

    ObjectInputStream ois = null;
    Parameters params = null;
    try {
      def inStream = new FileInputStream("path/to/parameters.bin")
      ois = new ObjectInputStream(inStream)
      params = ois.readObject() as Parameters
    } catch (Exception ignored) {
[...]

PR #7600, opened 2026-02-28
SMILE-11238: Fix encoding of Parameters with inline Reference resources

fil512:
## Summary
- Fixes `UnsupportedOperationException` thrown when encoding a `Parameters` resource that contains a `Reference` parameter with an inline resource set via `Reference.setResource()`
- Root cause is in `FhirTerser.containResourcesForEncoding()`, which unconditionally attempted to add contained resources for all resource types, even those like `Parameters`, `Bundle`, and `Binary` that extend `Resource` directly (not `DomainResource`) and have no `contained` field
- A `canContain` guard is added so containment is only attempted for `IResource`/`IDomainResource` instances; non-containable types now emit a single warning log instead of throwing

## Root Cause
`FhirTerser.containResourcesForEncoding()` iterated over all inline references and called `getContainedResourceList(theResource).add(resource)`. When the top-level resource is a `Parameters` (which extends `Resource`, not `DomainResource`), `getContainedResourceList()` returns `Collections.emptyList()` — an immutable list. Attempting to add to it throws `UnsupportedOperationException`.

## Fix
A `boolean canContain` check (`theResource instanceof IResource || theResource instanceof IDomainResource`) guards the containment loop in `FhirTerser`. When containment is not supported, the code emits one `WARN`-level log message and skips the loop. The inline resource data is not emitted, which is correct per the FHIR spec's restriction of `contained` to `DomainResource`.

## Review Tips
- The `canContain` check uses `IResource || IDomainResource` covering both DSTU2-era and R4/R5 class hierarchies
- The warning log fires once per encode call (guarded by `break`)
- Existing containment logic (the `canContain == true` branch) is structurally unchanged

## QA Scenarios
- [ ] Encode a `Parameters` resource with a `Reference` parameter containing an inline resource — confirm no exception
- [ ] Verify the encoded output contains `valueReference` but no `contained` array
- [ ] Encode a `Patient` with inline `Organization` reference — confirm containment works as before
- [ ] Check for `WARN` log message about skipped containment

Resolves https://github.com/hapifhir/hapi-fhir/issues/7319

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11238-parameters-encode-fix.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11238-parameters-encode-fix.yaml
@@ -1,0 +1,10 @@
+---
+type: fix
+issue: 7319
+jira: SMILE-11238
+title: "Previously, encoding a Parameters resource containing inline Reference resources with
+  an embedded resource (e.g. a Reference with a resource object set rather than a URL) would
+  throw an UnsupportedOperationException. This was caused by FhirTerser.getContainedResourceList()
+  returning an immutable empty list for non-DomainResource types such as Parameters. The encoder
+  now correctly skips containment processing for resource types that do not support contained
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

PR #7601, opened 2026-02-28
SMILE-11050: Fix NPE in search param extraction for Timing with partial boundsPeriod

fil512:
## Summary

Fixes a NullPointerException in `BaseSearchParamExtractor.addDate_Timing()` that occurs when persisting a resource whose `occurrenceTiming` has a `boundsPeriod` with only a start date. The root cause is that FHIR model accessor methods (e.g. `getEndElement()`) auto-create an empty `DateTimeType` with a null internal value; the extractor was guarding against a null element reference but not against a null element _value_, so passing that empty wrapper to `DateStringWrapper(Date, String)` caused the NPE.

1. Added `&& start.getValue() != null` and `&& end.getValue() != null` guards in `addDate_Timing()` for the `boundsPeriod` start/end extraction, consistent with the existing null-value check already applied to `Timing.event` dates in the same method.
2. Added five new regression tests in `FhirResourceDaoR4SearchNoFtTest` covering the primary NPE scenario (auto-created end element), symmetric auto-created start element, clean start-only period, event-dates-only Timing, and empty `boundsPeriod`.
3. Added changelog entry `SMILE-11050-timing-date-npe.yaml` for the 8.10.0 release.

## Code Review Suggestions

- [ ] **Consistency of the fix pattern**: The event-date loop at the top of `addDate_Timing()` already uses the same `theEvent.getValue() != null` guard — the fix is symmetric.
- [ ] **No other `DateStringWrapper` call sites affected**: Verify all other callers guard against null values.
- [ ] **Range behaviour with start-only boundsPeriod**: With only a start indexed, dates.first() == dates.last() (point-in-time semantics).

## QA Test Suggestions

- [ ] POST a `ServiceRequest` with `occurrenceTiming.repeat.boundsPeriod` having only `start` — confirm `201 Created` (no NPE)
- [ ] POST with both `start` and `end` populated — regression check
- [ ] POST with `occurrenceTiming.event[]` only (no bounds) — regression check
- [ ] POST with empty `boundsPeriod` (no start, no end) — confirm no error, no search match
- [ ] Trigger `$reindex` on existing affected resources

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11050-timing-date-npe.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-11050-timing-date-npe.yaml
@@ -1,0 +1,8 @@
+---
+type: fix
+jira: SMILE-11050
+title: "Previously, persisting a resource with a Timing search parameter whose boundsPeriod contained only
+  a start date (no end date) could result in a NullPointerException during search parameter extraction
+  if the end element had been auto-created as an empty DateTimeType by earlier processing (e.g. validation,
+  serialization, or interceptors). The boundsPeriod date extraction now correctly checks for null values
+  before constructing date wrappers, matching the existing null-value check used for Timing event dates."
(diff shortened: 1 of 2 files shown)


robogary:
Formatting check succeeded!

PR #7603, opened 2026-03-02
SMILE-10730: Fix NPE when Subscription uses :missing modifier on token search params

fil512:
## Summary

This PR fixes a `NullPointerException` in `InMemoryResourceMatcher.supportedQualifier()` that caused an HTTP 500 error whenever a FHIR Subscription was created with a criteria string using the `:missing` modifier on a token search parameter (e.g., `MessageHeader?event:missing=true`).

The root cause is that `:missing` is handled by `BaseParam` (which sets `myMissing = true`) rather than being a member of the `TokenParamModifier` enum. This means `TokenParam.getModifier()` returns `null` for `:missing` searches, and the `switch` statement on this null value causes the NPE.

### Changes

1. **Added a null guard in `InMemoryResourceMatcher.supportedQualifier()`**: When `tokenParam.getModifier()` is `null`, the method now returns `false` (unsupported), correctly directing subscriptions to use the `DATABASE` matching strategy instead of crashing.
2. **Tests at 3 levels**:
   - Unit: `InMemoryResourceMatcherR5Test` — verifies `canBeEvaluatedInMemory` returns unsupported for `:missing` on token params
   - Module: `SubscriptionStrategyEvaluatorTest` — verifies strategy evaluator routes to `DATABASE`
   - Integration: `SubscriptionsR4Test` — end-to-end subscription creation via FHIR client matching the exact customer scenario

## Code Review Suggestions

- [ ] **Null guard scope**: Only the `TOKEN` branch received the null check. Other param types may have similar latent issues if their modifier enums are incomplete, though `_source:missing` already works via a different code path (`matchSourceWithModifiers`).
- [ ] **Future consideration**: `:missing` could be added to `TokenParamModifier` with explicit in-memory support, but this is a larger change — the current null guard correctly forces DATABASE fallback for the narrow case where it matters.

## QA Test Suggestions

- [ ] POST a Subscription with `criteria: "MessageHeader?event:missing=true"` — confirm HTTP 201 (not 500)
- [ ] Verify matching strategy is `DATABASE` (not `IN_MEMORY`) for the `:missing` subscription
- [ ] POST a Subscription with `criteria: "Observation?code:not=FOO"` — confirm still works (IN_MEMORY)
- [ ] POST a Subscription with `criteria: "Observation?_source:missing=true"` — confirm still works (IN_MEMORY)
- [ ] Create `:missing` subscription, then POST a MessageHeader without an event field — confirm subscription fires

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
```diff
@@ -693,6 +693,10 @@ private boolean supportedQualifier(RuntimeSearchParam theParamDef, IQueryParamet
 		switch (theParamDef.getParamType()) {
 			case TOKEN:
 				TokenParam tokenParam = (TokenParam) theParam;
+				// :missing is handled by BaseParam, not TokenParamModifier, so getModifier() returns null
+				if (tokenParam.getModifier() == null) {
+					return false;
+				}
```
tadgh, line 699: nitpick: this implementation smells a bit. I imagine upstream, we only invoke `supportedQualifier()` when there.... you know, is a qualifier. But then the implementation here manually checks `getModifier() == null`. Having looked through the code now, this will work, but it still feels odd.

hapi-fhir-jpaserver-subscription/src/test/java/ca/uhn/fhir/jpa/subscription/module/matcher/SubscriptionStrategyEvaluatorTest.java
```diff
@@ -16,6 +16,12 @@ public class SubscriptionStrategyEvaluatorTest extends BaseSubscriptionDstu3Test
 	@Autowired
 	SubscriptionStrategyEvaluator mySubscriptionStrategyEvaluator;
 
+	// SMILE-10730: :missing on token param should fall back to DATABASE (not NPE)
+	@Test
+	void testTokenParamMissingModifier() {
+		assertDatabase("MessageHeader?event:missing=true");
+	}
```
tadgh, line 23: nice!

tadgh (approved):

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/matcher/InMemoryResourceMatcher.java
fil512: it's odd because this modifier is treated differently from the others. the real fix would be to normalize modifiers in our model

PR #7607, opened 2026-03-02
SMILE-9706: Fix auto-versioning error for missing reference targets

fil512:
## Summary

This PR fixes a regression where creating a resource with `autoVersionReferenceAtPaths` configured would return HTTP 404 (HAPI-0523) instead of HTTP 400 (HAPI-1094) when a referenced resource did not exist. The root cause was that `performAutoVersioning()` in `BaseStorageDao` eagerly threw a `ResourceNotFoundException` when the target of an auto-version reference was not found, short-circuiting before `DaoResourceLinkResolver` had a chance to produce its richer, path-aware error.

1. **One-line production fix** in `BaseStorageDao.performAutoVersioning()`: replace `throw new ResourceNotFoundException(...)` with `continue`, allowing the loop to skip unresolvable references and pass them downstream to `DaoResourceLinkResolver`, which already handles missing references correctly with HAPI-1094.
2. **Regression test** (`testAutoVersionReferenceToNonExistentResource_throwsInvalidRequestException`) verifies that the bug scenario now yields HTTP 400 with a message containing `"not found, specified in path"`.
3. **Control test** (`testAutoVersionReferenceToNonExistentResource_withoutAutoVersionPaths_throwsInvalidRequestException`) confirms that the non-auto-versioned code path, which already worked, continues to produce the same HTTP 400 + path-aware message.
4. **Happy path test** (`testAutoVersionReferenceToExistingResource_succeeds`) guards the normal flow: when the target exists, the reference is versioned and persisted successfully.

## Code Review Suggestions

- [ ] **`continue` vs `throw` semantics**: The `continue` causes `performAutoVersioning()` to silently skip the reference for that iteration. Confirm that the reference being left unversioned (still a bare `Patient/X` reference string) is the correct pre-condition for `DaoResourceLinkResolver` to fire and produce the HAPI-1094 error. If `DaoResourceLinkResolver` only runs when referential integrity enforcement is enabled, there may be a configuration-dependent gap where neither error is thrown and the dangling reference is silently persisted.
- [ ] **Import removal**: `ResourceNotFoundException` is removed from the import list in `BaseStorageDao`. Verify no other usages exist in that class that could now cause a compile error (the diff shows only the one `throw` site was using it).
- [ ] **`autoCreatePlaceholderReferenceTargets` interaction**: The existing case 2 logic (`version = 1L`) still runs when `isAutoCreatePlaceholderReferenceTargets()` is true. The new `continue` only covers case 3. Confirm that when placeholder creation is on, the downstream resolver still creates the placeholder correctly — this path is not covered by the new tests.
- [ ] **Test placement**: The three new tests are added inside the `@Nested` class (implied by indentation and surrounding context). Verify they are placed in the correct nested class to pick up the `observationAutoVersionExtension` setup from `@BeforeEach`, otherwise the first test will use a null extension.

### Setup
- [ ] Deploy a HAPI-FHIR JPA server (or use `hapi-fhir-jpaserver-starter`) with `autoVersionReferenceAtPaths` configured to include `Observation.subject`.
- [ ] Ensure `autoCreatePlaceholderReferenceTargets` is **disabled** (the default).

### Test Cases
- [ ] **Regression case (the bug)**: POST an `Observation` with `subject` referencing a non-existent `Patient/X`. Confirm the response is HTTP 400 (not 404) and the body contains a message like `"Resource Patient/X not found, specified in path: Observation.subject"`.
- [ ] **Happy path**: POST a `Patient`, then POST an `Observation` with `subject` referencing that patient's unversioned ID. Confirm the response is HTTP 201 and the stored `Observation.subject` reference has been upgraded to include the version number (e.g., `Patient/1/_history/1`).
- [ ] **Non-auto-versioned path (control)**: Disable `autoVersionReferenceAtPaths` and repeat the missing-reference POST. Confirm HTTP 400 with the same path-aware message — verifying the control case still works.
- [ ] **Placeholder targets enabled**: Enable `autoCreatePlaceholderReferenceTargets` and POST an `Observation` referencing a non-existent `Patient/Y`. Confirm the server creates the placeholder patient and the Observation is persisted successfully (regression guard on case 2).
- [ ] **Bundle transaction**: Submit a Bundle containing an `Observation` with a forward reference to a `Patient` entry in the same bundle. Confirm that auto-versioning still resolves correctly within a transaction context (these share the same `performAutoVersioning` call path).

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-9706-autoversion-missing-ref-error.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-9706-autoversion-missing-ref-error.yaml
@@ -1,0 +1,8 @@
+---
+type: fix
+jira: SMILE-9706
+title: "Previously, when autoVersionReferenceAtPaths was configured and a resource was stored
+  referencing a non-existent target, the server would return HTTP 404 with a cryptic
+  HAPI-0523 error instead of the expected HTTP 400 with a descriptive referential integrity
+  error (HAPI-1094). The auto-versioning logic now skips unresolvable references and allows
+  the downstream referential integrity check to produce the correct error response."
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4VersionedReferenceTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4VersionedReferenceTest.java
@@ -9,6 +9,7 @@ import ca.uhn.fhir.rest.api.server.IBundleProvider;
 import ca.uhn.fhir.rest.api.server.SystemRequestDetails;
 import ca.uhn.fhir.rest.param.TokenParam;
 import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
+import ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException;
 import ca.uhn.fhir.rest.server.servlet.ServletRequestDetails;
 import ca.uhn.fhir.util.BundleBuilder;
 import ca.uhn.fhir.util.ClasspathUtil;
@@ -55,6 +56,7 @@ import java.util.stream.Collectors;
 
 import static ca.uhn.fhir.util.HapiExtensions.EXTENSION_AUTO_VERSION_REFERENCES_AT_PATH;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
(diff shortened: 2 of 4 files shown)


robogary:
Formatting check succeeded!

fil512:
Pre-merge review note (from a second-pass quality review):

The shift from `throw` to `continue` in the autoversion path is the right call — it lets `DaoResourceLinkResolver` surface the richer HAPI-1094 error with path context. One gap worth closing before merge:

**Untested configuration: `enforceReferentialIntegrityOnWrite=false`**

Under the default (enforce=true) the `continue` is safe: the later link-resolver stage catches the missing target and throws. But if referential integrity enforcement is disabled, the `continue` path would now silently persist an unversioned reference that previously would have been caught upstream. That's a behavior change for that configuration.

Suggested follow-up: add a one-line test asserting behavior when `enforceReferentialIntegrityOnWrite=false` — either confirming the write proceeds (if that's the intended semantics) or that some other guard fires. Either outcome documented in a test closes the loop.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-9706-autoversion-missing-ref-error.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: fix
+jira: SMILE-9706
+title: "Previously, when autoVersionReferenceAtPaths was configured and a resource was stored
```
tadgh, line 4:
```suggestion
title: "Previously, when `autoVersionReferenceAtPaths` was configured and a resource was stored
```

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-9706-autoversion-missing-ref-error.yaml
```diff
@@ -0,0 +1,8 @@
...
+  referencing a non-existent target, the server would return HTTP 404 with a cryptic
+  HAPI-0523 error instead of the expected HTTP 400 with a descriptive referential integrity
```
tadgh, line 6:
```suggestion
  `HAPI-0523` error instead of the expected HTTP 400 with a descriptive referential integrity
```

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-9706-autoversion-missing-ref-error.yaml
```diff
@@ -0,0 +1,8 @@
...
+  error (HAPI-1094). The auto-versioning logic now skips unresolvable references and allows
```
tadgh, line 7:
```suggestion
  error ,`HAPI-1094`. The auto-versioning logic now skips unresolvable references and allows
```

tadgh (approved):

PR #7608, opened 2026-03-02
Update org.hl7.fhir.core to 8.6.2

dotasek:


robogary:
Formatting check succeeded!

hapi-fhir-structures-r5/src/main/java/org/hl7/fhir/r5/hapi/ctx/HapiWorkerContext.java
```diff
@@ -270,6 +271,21 @@ public String getVersion() {
 		return myCtx.getVersion().getVersion().getFhirVersionString();
 	}
 
+	@Override
+	public int getDefinitionsVersion() {
+		throw new UnsupportedOperationException(Msg.code(2861));
```
tadgh, line 279: Is this accurate for error code? feels like we should be higher up..

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapter.java
```diff
@@ -656,6 +658,19 @@ public String getVersion() {
 		return myFhirContext.getVersion().getVersion().getFhirVersionString();
 	}
 
+	@Override
+	public int getDefinitionsVersion() {
+		return 0;
```
tadgh: question: this impl returns 0, the other throws NotSupported. What should we be doing?

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/validator/HostApplicationServicesTest.java
```diff
@@ -154,7 +155,7 @@ private static class RecordingHostServices extends FhirInstanceValidator.NullEva
 
 		@Override
 		public org.hl7.fhir.r5.model.Base resolveReference(
-			FHIRPathEngine engine, Object appContext, String url, org.hl7.fhir.r5.model.Base refContext)
+			FHIRPathEngine engine, Object appContext, String url, Identifier identifier, org.hl7.fhir.r5.model.Base refContext)
```
tadgh, line 158: Question: do we not need to take this new parameter into account in the logic of any of the implementations?

hapi-fhir-structures-r5/src/main/java/org/hl7/fhir/r5/hapi/ctx/HapiWorkerContext.java
dotasek: https://github.com/hapifhir/hapi-fhir/wiki/Bumping-Error-Code is currently on 2872 (building on the errors I reserved for this PR)

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapter.java
dotasek: After some investigation, `getDefinitionsVersion` is implemented in the core libs, but never actually used.

IntelliJ's search doesn't find any uses, and this is corroborated by @grahamegrieve: 
> actually, I ended up using store/retrieveAnalysis direclty, and never using definitions version otherwise

The `store/retrieveAnalysis` methods he's using will generate what they need on `null` returns, but at the cost of lower performance on FHIRPath and validation. Implementing these would require some care, as in org.hl7.fhir.core, all the implementations are threadsafe, and HAPI's implementations appear less straightforward.

@tadgh would you advise on spending time implementing these, or leaving them as is?

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/validator/HostApplicationServicesTest.java
dotasek: From my conversations with @grahamegrieve:

> David Otasek: `resolveReference(FHIRPathEngine engine, ... Identifier identifier, ...)` doesn't take into account the Identifier.
> Grahame Grieve: that's fine
> David Otasek: It's fine that HAPI ignores the identifier? I'm assuming that there will be a validation or a FHIRPath case that eventually runs into this, unless it was only implemented in org.hl7.fhir.core to manage a core libs specific case.
> Grahame Grieve: there will be a test case, but the HAPI provider doesn't have to pass our test cases?

Following up on this, I looked at all the implementations of this method in org.hl7.fhir.core, and found that it was unused in almost all of them:

| Code Reference | Usage notes |
|----------------|-------------|
| org.hl7.fhir.r5.comparison.ComparisonRenderer#resolveReference | not used |
| org.hl7.fhir.r5.liquid.LiquidEngine#resolveReference | re-used in a method call, but not directly |
| org.hl7.fhir.r5.test.FHIRPathTests.FHIRPathTestEvaluationServices#resolveReference | test code, NotImplementedException |
| org.hl7.fhir.r5.test.SnapShotGenerationTests.SnapShotGenerationTestsContext#resolveReference | test code, not used, returns null |
| org.hl7.fhir.r5.testfactory.TestDataHostServices#resolveReference | test code, not used, returns null |
| org.hl7.fhir.r5.utils.LiquidJsonTest#resolveReference | test code, not used, returns null |
| org.hl7.fhir.r5.utils.sql.Runner#resolveReference | Error("Not implemented yet: resolveReference") |
| org.hl7.fhir.r5.utils.structuremap.FHIRPathHostServices#resolveReference | not used |
| org.hl7.fhir.conversion.tests.SnapShotGenerationXTests.SnapShotGenerationTestsContext#resolveReference | test code, not used, returns null |
| org.hl7.fhir.validation.instance.InstanceValidator.ValidatorHostServices#resolveReference | used, but see note |
| org.hl7.fhir.validation.tests.ValidationTests#resolveReference | test code, not used |
 
Note: in all these places `Identifier identifier` is only really used here:
https://github.com/hapifhir/org.hl7.fhir.core/blob/648e039b7a31dfa579a43e1d850e5eb1adc88fea/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/InstanceValidator.java#L439

In hopes of finding specific test cases that actually utilize this, I put a breakpoint there and ran all the tests in org.hl7.fhir.validation. It was never hit. The three failed tests are due to unrelated updates to IGs.

Coverage agrees on this:

Screenshot 2026-03-10 at 11 23 09 AM

Without a clear test case, I think implementation on HAPI's side isn't advisable.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapter.java
tadgh: I would leave them if we do not use them. At most, a comment about why it is the way it is.

hapi-fhir-validation/src/test/java/org/hl7/fhir/common/hapi/validation/validator/HostApplicationServicesTest.java
tadgh: Totally fine. I would recommend leaving a comment on the javadoc explaining this.

dotasek: Resolved in https://github.com/hapifhir/hapi-fhir/pull/7608/commits/b2caa952aa9d501fff5fad909128de5e293a1238

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/validator/WorkerContextValidationSupportAdapter.java
dotasek: Resolved in https://github.com/hapifhir/hapi-fhir/pull/7608/commits/b2caa952aa9d501fff5fad909128de5e293a1238

dotasek:
Just a note: This PR is already closed, so cannot be edited, but the title is a typo. This was an update to 6.8.2 of org.hl7.fhir.core

PR #7612, opened 2026-03-02
GL-8464: Add null guard for myValidationSettings in ValidatorPolicyAdvisor.isSuppressMessageId

fil512:
## Summary

This PR fixes a NullPointerException regression in `ValidatorPolicyAdvisor` introduced by PR #7595 (SMILE-11707). That PR added message suppression logic in `isSuppressMessageId()` and reference policy logic in `policyForReference()` that both access `@Autowired` fields (`myValidationSettings` and `myFhirContext`) without null guards. When `ValidatorPolicyAdvisor` is instantiated directly with `new ValidatorPolicyAdvisor()` — a legitimate usage pattern outside of a Spring application context — those fields are never injected, causing an NPE at runtime. The fix adds null checks to both methods so they degrade gracefully (returning `false` / `IGNORE` respectively) when Spring has not wired the dependencies.

1. Added `myValidationSettings != null` guard to `isSuppressMessageId()` so the method returns `false` instead of throwing NPE when called outside Spring.
2. Wrapped the body of `policyForReference()` in a `myFhirContext != null && myValidationSettings != null` guard, preserving the existing `IGNORE` fallback return.
3. Added `ValidatorPolicyAdvisorTest` with three unit tests covering: no-Spring NPE regression, IGNORE policy suppression of `REFERENCE_REF_CANTMATCHCHOICE`, and CHECK_VALID policy non-suppression.

## Code Review Suggestions

- [ ] The null guard in `isSuppressMessageId()` silently returns `false` when `myValidationSettings` is null. Confirm this is the desired behavior rather than logging a warning.
- [ ] The `policyForReference()` null guard wraps both field accesses together. If only one is null, the method falls through to `IGNORE`. Confirm both fields are always either both-null or both-injected in practice.
- [ ] The test uses `ReflectionTestUtils.setField()` to inject `ValidationSettings` — confirm this is acceptable vs constructor injection.

## QA Test Suggestions

- [ ] Build and run: `mvn -pl hapi-fhir-storage test -Dtest=ValidatorPolicyAdvisorTest`
- [ ] Validate a FHIR resource using `new ValidatorPolicyAdvisor()` (no Spring) — confirm no NPE
- [ ] In Spring-wired JPA server with IGNORE policy, validate bundle with internal references — confirm `REFERENCE_REF_CANTMATCHCHOICE` messages are suppressed
- [ ] In Spring-wired JPA server with CHECK_VALID policy, validate bundle with invalid internal reference — confirm error messages appear

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/GL-8464-null-guard-isSuppressMessageId.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/GL-8464-null-guard-isSuppressMessageId.yaml
@@ -1,0 +1,7 @@
+---
+type: fix
+jira: GL-8464
+title: "Previously, instantiating ValidatorPolicyAdvisor directly (e.g. new ValidatorPolicyAdvisor())
+  without a Spring context would result in a NullPointerException when isSuppressMessageId was
+  called, because the injected myValidationSettings field was null. A null guard has been added
+  so that the class can be safely used outside of a Spring-managed context."


robogary:
Formatting check succeeded!

Issue #7482, opened 2026-01-19
Synchronous search does not work correctly with _include and _revinclude:iterate


mlitcher:
**Describe the bug**
When a synchronous search is invoked with a query that contains both an `_include` and `_revinclude:iterate`, the `_revinclude:iterate` is applied before the `_include`, which does not honor the stated order:

1. non-iterated _revincludes
2. non-iterated _includes
3. iterated _revincludes
4. iterated _includes

Documentation: https://hapifhir.io/hapi-fhir/docs/server_jpa/search.html#-include-and-revinclude-order

**To Reproduce**

1. Using the public HAPI FHIR instance, execute the following FHIR query:
https://hapi.fhir.org/baseR4/Appointment?_id=48096739&_include=Appointment:practitioner&_revinclude:iterate=PractitionerRole:practitioner&_count=100
2. Note the inclusion of PractitionerRole resources.
3. Append `&_offset=0` to the FHIR query to force a synchronous search:
https://hapi.fhir.org/baseR4/Appointment?_id=48096739&_include=Appointment:practitioner&_revinclude:iterate=PractitionerRole:practitioner&_count=100&_offset=0
4. Note that the PractitionerRole resources are absent.

**Expected behavior**
The expectation is that both the following queries would return the exact same result:

1. https://hapi.fhir.org/baseR4/Appointment?_id=48096739&_include=Appointment:practitioner&_revinclude:iterate=PractitionerRole:practitioner&_count=100
2. https://hapi.fhir.org/baseR4/Appointment?_id=48096739&_include=Appointment:practitioner&_revinclude:iterate=PractitionerRole:practitioner&_count=100&_offset=0

**Environment (please complete the following information):**
- HAPI FHIR Version: 8.5.3 (also tested locally in 6.10.5)
- OS: macOS
- Browser: chrome

**Additional context**
[...]

PR #7613, opened 2026-03-02
SMILE-9135: Fix _include/_revinclude transitive results without _iterate

### LINKED ISSUE #7626: RevInclude including results from included results when iterate is not specified
**Describe the bug**
Under the FHIR spec for includes and revincludes it states:
`_include` and `_revinclude` are, by default, invoked only on the initial results of the search set, not on any 'included' resources. To repeatedly perform the `_include` and `_revinclude` instructions on included resources, use the `:iterate` modifier.
[FHIR Spec](https://build.fhir.org/search.html#_revinclude)
Currently it seems like if we did a request such as (where replaces is array of references to other ServiceRequests): `/ServiceRequest?_id=123&_include=ServiceRequest:replaces&_revinclude=ServiceRequest:replaces`
We end up getting `ServiceRequests` in the result bundle, that were referenced by `ServiceRequests` that were returned as part of the revIncludes on the initial `ServiceRequest` we are reading.

**To Reproduce**

1. Create 3 service requests, with the 3rd one referencing the Ids of the first two in the replaces array
```
1326
{
  "resourceType": "ServiceRequest",
  "status": "draft",
  "intent": "order",
  "occurrenceDateTime": "2023-01-15T14:30:00Z",
  "code": {
    "coding": [
      {
        "system": "http://snomed.info/sct",
        "code": "384958002",
        "display": "Administration of medication"
      }
    ]
  }
}

1327
{
  "resourceType": "ServiceRequest",
  "status": "draft",
  "intent": "order",
  "occurrenceDateTime": "2023-01-15T14:30:00Z",
  "code": {
    "coding": [
      {
        "system": "http://snomed.info/sct",
        "code": "384958002",
        "display": "Administration of medication"
      }
    ]
  }
}

1328
{
  "resourceType": "ServiceRequest",
  "status": "draft",
  "intent": "order",
  "occurrenceDateTime": "2023-01-15T14:30:00Z",
  "replaces": [
    {
        "reference": "ServiceRequest/1326"
    },
    {
        "reference": "ServiceRequest/1327"
    }
  ],
  "code": {
    "coding": [
      {
        "system": "http://snomed.info/sct",
        "code": "384958002",
        "display": "Administration of medication"
      }
    ]
  }
}
```
2. `http://localhost:8000/ServiceRequest?_id=1326&_include=ServiceRequest:replaces&_revinclude=ServiceRequest:replaces`
3. Returns all 3 resources

**Expected behavior**
Returns only ServiceRequest with id=1326 and id=1328, and not id=1327

jdar8:
## Summary

This fixes a FHIR spec violation where `_include` was incorrectly applied to resources found via `_revinclude`, producing transitive include results that are only permitted when `_iterate` is specified. The root cause was that both `PersistedJpaBundleProvider` and `SynchronousSearchSvcImpl` mutated the working PID set after loading `_revinclude` results, then passed the expanded set to the subsequent `_include` call. A secondary issue discovered during review was that `SynchronousSearchSvcImpl` was missing `_iterate` support entirely — it only ran `_revinclude` then `_include` with no phase separation.

1. `PersistedJpaBundleProvider`: capture original search result PIDs before `_revinclude` expansion, pass `originalPids` (not the expanded set) to the non-iterate `_include` call.
2. `SynchronousSearchSvcImpl`: replace the flat 2-phase include/revinclude sequence with the correct 4-phase algorithm — non-iterate revinclude, non-iterate include, iterate revinclude, iterate include — matching `PersistedJpaBundleProvider`.
3. `ResourceProviderRevIncludeTest`: 8 new tests cover the non-iterate and iterate combinations on both the async (`myClient`) and synchronous DAO paths, including a dedicated test that exercises `SynchronousSearchSvcImpl` directly via `SearchParameterMap.newSynchronous()`.

---

## Code Review Suggestions

- [ ] **Phase ordering in `SynchronousSearchSvcImpl`** — The new 4-phase sequence runs non-iterate revinclude (Phase 1) before non-iterate include (Phase 2). `PersistedJpaBundleProvider` uses the same ordering. The FHIR spec does not mandate ordering between non-iterate `_include` and `_revinclude` when used together, but both sides of the fix must agree on which PID set is passed to each phase. Confirm the ordering is intentional and consistent between the two implementations.
- [ ] **`everythingMode` guard moved** — In the original `SynchronousSearchSvcImpl`, the `everythingMode == null` guard wrapped the entire include+revinclude block. In the new code, the guard is applied to Phase 2 (non-iterate includes) and Phase 4 (iterate includes) separately, but not to Phase 1 (non-iterate revinclude) or Phase 3 (iterate revinclude). Verify this is the correct scope — the guard was previously inherited from `PersistedJpaBundleProvider` behaviour and may need to cover revinclude phases too.
- [ ] **`maxIncludes` budget shared across all 4 phases** — The single `maxIncludes` counter is decremented across phases in order. If Phase 1 exhausts the budget, Phases 2-4 are silently skipped (the `maxIncludes > 0` guard short-circuits). This matches the existing behaviour in `PersistedJpaBundleProvider`. Confirm this page-level cap is still the intended ceiling for the synchronous path.
- [ ] **`allIncludedPidsList` used for total count correction** — The bundle size calculation subtracts `allIncludedPidsList.size()` from `resources.size()`. The old code subtracted `includedPidsList.size()`, which in the original also accumulated both include and revinclude results. The semantics are the same, but worth verifying that the list is accumulated identically across all 4 phases (no phase conditionally skips adding to it when it should not).
- [ ] **`originalPids` as `HashSet` vs the incoming `pids` list type** — `originalPids` is a `HashSet<JpaPid>` snapshot. `loadIncludes` accepts a `Collection<JpaPid>`; confirm `HashSet` iteration order does not affect correctness (it should not, since inclusion is a set operation, but worth a second look given the PID type).

---

### Setup

- [ ] Stand up a HAPI JPA server (R4). Create three `ServiceRequest` resources: SR/A (no `replaces`), SR/B (no `replaces`), SR/C (`replaces` both SR/A and SR/B).

### Non-iterate path (the bug fix)

- [ ] `GET /ServiceRequest?_id=<A>&_include=ServiceRequest:replaces&_revinclude=ServiceRequest:replaces` — response must contain SR/A and SR/C only; SR/B must not appear.
- [ ] `GET /ServiceRequest?_id=<A>&_revinclude=ServiceRequest:replaces` — response must contain SR/A and SR/C only; SR/B must not appear.
- [ ] `GET /ServiceRequest?_id=<A>&_include=ServiceRequest:replaces` — response must contain SR/A only; SR/B and SR/C must not appear.

### Iterate path (regression guard — must still work)

- [ ] `GET /ServiceRequest?_id=<A>&_include:iterate=ServiceRequest:replaces&_revinclude:iterate=ServiceRequest:replaces` — response must contain all three: SR/A, SR/B, and SR/C.
- [ ] Repeat the iterate test using a client configured to go through the synchronous code path (small page size, or a DAO-layer call with `SearchParameterMap.newSynchronous()`) to exercise `SynchronousSearchSvcImpl` Phase 3 and Phase 4 directly.

### Edge cases

- [ ] Run the same `_include`/`_revinclude` non-iterate query against a resource type that has no matching revinclude results — confirm the response contains only the direct search result with no extras and no server error.
- [ ] Run a search that would hit the `maxIncludes` page cap across multiple phases — confirm the total count in the returned bundle is correct and does not include included resources in the match count.

Closes #7626

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-9135-revinclude-no-iterate.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/SMILE-9135-revinclude-no-iterate.yaml
@@ -2,8 +2,8 @@
 type: fix
 jira: SMILE-9135
 issue: 7626
-title: "Previously, when a search used both _revinclude and _include without _iterate, the
-  _include parameter was incorrectly applied to revincluded resources in addition to the
+title: "Previously, when a search used both `_revinclude` and `_include` without `:iterate`, the
+  `_include` parameter was incorrectly applied to revincluded resources in addition to the
   initial search results. This caused extra resources to be returned that the FHIR specification
-  only permits when _iterate is specified. The _include expansion is now correctly scoped to
+  only permits when `:iterate` is specified. The `_include` expansion is now correctly scoped to
   the original search result set only, matching the expected non-iterative behaviour."
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/PersistedJpaBundleProvider.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/PersistedJpaBundleProvider.java
@@ -476,11 +476,11 @@ public class PersistedJpaBundleProvider implements IBundleProvider {
 		if (mySearchEntity.getSearchType() == SearchTypeEnum.SEARCH) {
 			Integer maxIncludes = myStorageSettings.getMaximumIncludesToLoadPerPage();
 
-			// Save original search result PIDs — non-iterate _include must apply only to initial results, not to
-			// revinclude results
+			// Save original search result PIDs — non-iterate `_include` must apply only to initial results, not to
+			// `_revinclude` results
 			Set<JpaPid> originalPids = new HashSet<>(thePids);
 
-			// Load non-iterate _revincludes
+			// Load non-iterate `_revinclude`
 			Set<JpaPid> nonIterateRevIncludedPids = theSearchBuilder.loadIncludes(
 					myContext,
 					myEntityManager,
@@ -497,8 +497,8 @@ public class PersistedJpaBundleProvider implements IBundleProvider {
 			thePids.addAll(nonIterateRevIncludedPids);
 			includedPidList.addAll(nonIterateRevIncludedPids);
 
-			// Load non-iterate _includes (use originalPids so _include only applies to the
-			// initial search results, not to revincluded resources — per FHIR spec, without _iterate)
+			// Load non-iterate `_include` (use originalPids so `_include` only applies to the
+			// initial search results, not to revincluded resources — per FHIR spec, without `:iterate`)
 			Set<JpaPid> nonIterateIncludedPids = theSearchBuilder.loadIncludes(
 					myContext,
 					myEntityManager,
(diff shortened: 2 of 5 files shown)


robogary:
Formatting check succeeded!

mlitcher:
This PR should resolve #7482.

PR #7618, opened 2026-03-04
7614 install only

### LINKED ISSUE #7614: add INSTALL_ONLY option to NPM package
we want a means to install the contents of a package without storing the package locally

TipzCM:
closes #7614

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
@@ -438,8 +438,7 @@ public class PackageInstallerSvcImpl implements IPackageInstallerSvc {
 		if (theInstallationSpec.isDryRun()) {
 			constructDryRunReport(theResource, existingResource, map, theOutcome);
 		} else {
-			isInstalled =
-					createOrUpdateResource(dao, theResource, existingResource, theInstallationSpec);
+			isInstalled = createOrUpdateResource(dao, theResource, existingResource, theInstallationSpec);
 		}
 		if (isInstalled) {
 			theOutcome.incrementResourcesInstalled(myFhirContext.getResourceType(theResource));


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImplCreateTest.java
```diff
@@ -249,6 +261,73 @@ void testSearchParameterUsesClientId_regardlessOfPolicy(
 		assertThat(actual.getMeta().getSource()).isEqualTo(PACKAGE_ID_1 + "|" + PACKAGE_VERSION);
 	}
 
+	@ParameterizedTest
+	@ValueSource(booleans = { true, false })
+	public void install_installOnly_persistsResourcesButNotThePkg(boolean theEmbedContent, @TempDir Path theTempDir) throws IOException {
```
volodymyr-korzh, line 266: Question: Can we run installOnly with dryRun= true? Looks like a valid use case. There is no test for this combination right now, I would suggest adding a test for this case.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -457,17 +466,49 @@ protected boolean createOrUpdateResource(
...
+			}
 		} else {
-			return updateExistingResource(theDao, theResource, theExistingResource);
+			FhirTerser terser = myFhirContext.newTerser();
+			String resourceType = theResource.fhirType();
+			RuntimeResourceDefinition rtDefinition = myFhirContext.getResourceDefinition(resourceType);
+			List<RuntimeSearchParam> sps = rtDefinition.getSearchParams();
+			StringBuilder uniqueIdentifierSpBuilder = new StringBuilder();
+			for (Map.Entry<String, List<List<IQueryParameterType>>> spSet : theSpMap.entrySet()) {
+				String key = spSet.getKey();
```
volodymyr-korzh: Question: are these else block changes were done to fix dry run behaviour? (we've been saving resources instead of executing dry run?)

1) I don't see any new tests or test assertions that cover changes in this branch. Please add or improve current tests for dry run.
All these test are passing regardless of changes in this else branch 
```
  - dryRunInstall_singleDependencyDifferentVersionsAreConsideredNew_doesntInstall
  - dryRunInstall_flagsDuplicates_andDoesntInstall
  - dryRunInstall_withDependencies_wontInstallDependencies
```

2) This logic should be consolidated with `constructDryRunReport` method, right now it is a copy paste. (valueOp.isEmpty() is correctly processed in `constructDryRunReport` method, but here we use valueOp.get() without isEmpty check)

3) `.orElseThrow();` - we can throw meaningful Exception message here instead, even if the case of exception is unlikely to happen, this will allow to remove TODO comment.

4) We can simply do it like this instead of reverse !thePackageInstallationSpec.isDryRun() logic
```
if (thePackageInstallationSpec.isDryRun()) {
     return executeDryRun(...)
}
....
```

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImplCreateTest.java
TipzCM: added

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
TipzCM: i just deleted it. it's the result of the drift between this branch and my (previously unmerged) dry-run branch.

There is no need for a 'dryrun' branch in this location, after all, since the condition leading here necessitates it not being a dry run.

PR #7487, opened 2026-01-20
Add Patient ID mode resource policies

jamesagnew:
The Patient ID Partition Interceptor has been enhanced to allow configurable policies per-resource-type, allowing specific resources to have multiple or missing patient compartment information.

PR #7619, opened 2026-03-04
merge operation patient id mode support

mrdnctrk:
- Add cross-partition merge support, specifically when using (via PatientIdPartitionInterceptor). When source and target patients reside in different partitions , the $merge operation now copies compartment resources from
   the source partition to the target partition, updates all references, deletes the originals, and creates a Provenance resource to enable undo. Cross-partition merge currently supports synchronous execution only.                
  - Fix a bug where $merge did not set active=false on the source resource when the active field was not previously set (#7669).
  - Reject standalone $replace-references for cross-partition requests with NotImplementedOperationException, since cross-partition behavior is not well defined for all partition modes — it is only supported through $merge via    
  CrossPartitionReplaceReferencesSvc for patient-id mode.                                                                                                                                                                             
  - Remove explicit partition ID from PreviousResourceVersionRestorer and UndoReplaceReferencesRequest to simplify partition handling.                                                                                                
  - Fix Provenance search in patient-id partitioning mode after #7487 rebase.

### DIFF
--- a/hapi-fhir-storage/src/test/java/ca/uhn/fhir/merge/MergeOperationInputParameterNamesTest.java
+++ b/hapi-fhir-storage/src/test/java/ca/uhn/fhir/merge/MergeOperationInputParameterNamesTest.java
@@ -1,0 +1,40 @@
+// Created by claude-opus-4-6
+package ca.uhn.fhir.merge;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * These tests assert against hardcoded strings to guard against accidental renames
+ * of FHIR operation parameter names. The parameter names are part of the wire protocol
+ * (e.g., "source-patient", "target-resource") — changing them is a breaking change for
+ * clients. If a constant in ProviderConstants is renamed, these tests will fail and
+ * force a deliberate decision rather than a silent drift.
+ */
+class MergeOperationInputParameterNamesTest {
+
+	@Test
+	void testPatientParameterNames() {
+		PatientMergeOperationInputParameterNames names = new PatientMergeOperationInputParameterNames();
+
+		assertThat(names.getSourceResourceParameterName()).isEqualTo("source-patient");
+		assertThat(names.getTargetResourceParameterName()).isEqualTo("target-patient");
+		assertThat(names.getSourceIdentifiersParameterName()).isEqualTo("source-patient-identifier");
+		assertThat(names.getTargetIdentifiersParameterName()).isEqualTo("target-patient-identifier");
+		assertThat(names.getResultResourceParameterName()).isEqualTo("result-patient");
+		assertThat(names.getDeleteSourceParameterName()).isEqualTo("delete-source");
+	}
+
+	@Test
+	void testGenericParameterNames() {
+		GenericMergeOperationInputParameterNames names = new GenericMergeOperationInputParameterNames();
+
+		assertThat(names.getSourceResourceParameterName()).isEqualTo("source-resource");
+		assertThat(names.getTargetResourceParameterName()).isEqualTo("target-resource");
+		assertThat(names.getSourceIdentifiersParameterName()).isEqualTo("source-resource-identifier");
+		assertThat(names.getTargetIdentifiersParameterName()).isEqualTo("target-resource-identifier");
+		assertThat(names.getResultResourceParameterName()).isEqualTo("result-resource");
+		assertThat(names.getDeleteSourceParameterName()).isEqualTo("delete-source");
+	}
+}


robogary:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/StorageResponseCodeEnum.java
```diff
@@ -26,40 +26,53 @@
...
+	StorageResponseCodeEnum(String theDisplay, boolean theNoChange) {
 		myDisplay = theDisplay;
+		myNoChange = theNoChange;
+	}
+
+	/**
+	 * Returns {@code true} if this outcome code represents a no-change operation,
+	 * meaning no data was actually modified in storage.
+	 */
+	public boolean isNoChange() {
```
jamesagnew, line 74: Cool! This is a neat quality-of-life enhancement

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/PreviousResourceVersionRestorer.java
```diff
@@ -89,49 +90,63 @@ private void restoreToPreviousVersions(
...
 			}
-			Long referenceVersion = referenceId.getVersionIdPartAsLong();
-
-			// Restore previous version (version - 1)
-			long previousVersion = referenceVersion - 1;
-			if (previousVersion < 1) {
-				throw new IllegalArgumentException(Msg.code(2731)
-						+ "Resource cannot be restored to a previous as the provided version is 1: " + referenceStr);
-			}
+			long referenceVersion = referenceId.getVersionIdPartAsLong();
```
jamesagnew, line 94: Nitpick: There were several comments in this method that you have removed as a part of this change that I thought made it much easier to understand what is going on.. IMO you should keep 'em

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/ReplaceReferencesPatchBundleSvc.java
```diff
@@ -77,14 +77,29 @@ public Bundle patchReferencingResources(
...
+	 * participate in an already-open outer transaction. Used by sync merge to keep everything atomic.
+	 */
+	public Bundle patchReferencingResourcesInNestedTransaction(
+			ReplaceReferencesRequest theReplaceReferencesRequest,
+			List<IdDt> theResourceIds,
+			RequestDetails theRequestDetails) {
+		Bundle patchBundle = buildPatchBundle(theReplaceReferencesRequest, theResourceIds, theRequestDetails);
+		IFhirSystemDao<Bundle, Meta> systemDao = myDaoRegistry.getSystemDao();
+		Bundle result = systemDao.transactionNested(theRequestDetails, patchBundle);
+		result.setId(UUID.randomUUID().toString());
```
jamesagnew, line 91: Question: What's with assigning the ID to the result? I a probably missing something but I don't see it getting used anywhere. I actually thought BaseTransactionProcessor already did this but I looked and it doesn't. Perhaps we should just modify it to handle adding an ID to the response bundle if we have a reason to need one

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/ReplaceReferencesPatchBundleSvc.java
```diff
@@ -77,14 +77,29 @@ public Bundle patchReferencingResources(
...
+
 	private Bundle buildPatchBundle(
 			ReplaceReferencesRequest theReplaceReferencesRequest,
 			List<IdDt> theResourceIds,
 			RequestDetails theRequestDetails) {
 		BundleBuilder bundleBuilder = new BundleBuilder(myFhirContext);
-		theResourceIds.forEach(referencingResourceId -> {
-			IFhirResourceDao<?> dao = myDaoRegistry.getResourceDao(referencingResourceId.getResourceType());
-			IBaseResource resource = dao.read(referencingResourceId, theRequestDetails);
+		theResourceIds.forEach(theReferencingResourceId -> {
```
jamesagnew: Thought: `theReferencingResourceId` isn't a method parameter so IMO it should be `referencingResourceId`

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/ReplaceReferencesProvenanceSvc.java
```diff
@@ -234,13 +232,17 @@ protected List<Provenance> getProvenancesOfTargetsFilteredByActivity(
...
 
-		theTargetIds.forEach(tId -> map.add("target", new ReferenceParam(tId)));
+		theTargetIds.forEach(tId -> map.add("target", new ReferenceParam(tId.toUnqualifiedVersionless())));
 
 		// Add sort by recorded field, in case there are multiple Provenance resources for the same source and target,
 		// we want the most recent one.
 		map.setSort(new SortSpec("recorded", SortOrderEnum.DESC));
 
-		IBundleProvider searchBundle = myProvenanceDao.search(map, theRequestDetails);
+		// Use all-partitions request because in patient-id partitioning mode with NON_UNIQUE_COMPARTMENT_IN_DEFAULT
```
jamesagnew: This doesn't feel right to me. I would have expected the PatientIdPartitionInterceptor to automatically supply a more specific partition set here (presumably 2 partitions: the compartment partition for the target, and the default partition). But I assume it's just supplying the copartment one since the map has a subject. I think this is a bug in the policy logic to be honest... That's probably beyond the scope of this ticket, but can you add the following comment in, just to remind us to clean this up:

// TODO: PatientIdPartiitonInterceptor should be returning the default partition as a part of the response when Provenance has policy `NON_UNIQUE_COMPARTMENT_IN_DEFAULT`. It doesn't currently. We shouldn't hardcode allPartitions once we fix that

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/merge/CrossPartitionMergePatientIdModeR4Test.java
```diff
@@ -0,0 +1,870 @@
...
+		registerInterceptor(myPartitionInterceptor);
+
+		myPartitionSettings.setPartitioningEnabled(true);
+		myPartitionSettings.setUnnamedPartitionMode(true);
+		myPartitionSettings.setAllowReferencesAcrossPartitions(
+			PartitionSettings.CrossPartitionReferenceMode.ALLOWED_UNQUALIFIED);
+
+		// Disable search cache: merge mutates data mid-test and subsequent searches
+		// would return stale cached results
+		myStorageSettings.setReuseCachedSearchResultsForMillis(null);
```
jamesagnew: This feels a little dangerous given that most real deployments won't have the setting set like this, making this test less valid than it should be..

What fails if you remove this? If it's just searches in this test itself, can they be replaced with synchronous searches? If it's searches as a part of the actual merge operation, it probably make sense to make them synchronous

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/ReplaceReferencesPatchBundleSvc.java
mrdnctrk: The reason is that in the async execution of replace-references operation, we are creating a Task resource, and we end up with several output bundles, as the async version processes multiple transaction bundles. At the reduction step of the async job, we collect all these result bundles and add them as contained resources in the Task resource's output here:  [link](https://github.com/hapifhir/hapi-fhir/blob/edb924f73ba5b8ce9620cd4e94fde4c4c133e938/hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/replacereferences/ReplaceReferenceUpdateTaskReducerStep.java#L157), where we use this id as a reference to the contained resource. 

Although in this method I added in this PR,  we don't actually need it since it is not used in the async flow currently, I've added for consistency. 

Let me know if you want me to set this in the BaseTransactionProcessor instead.

hapi-fhir-base/src/main/java/ca/uhn/fhir/model/api/StorageResponseCodeEnum.java
mrdnctrk: thanks

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/ReplaceReferencesPatchBundleSvc.java
mrdnctrk: makes sense, fixed it

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/PreviousResourceVersionRestorer.java
mrdnctrk: yes, restored.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/replacereferences/ReplaceReferencesProvenanceSvc.java
mrdnctrk: yes, added the comment, i will fix this next and remove the TODO in a separate PR.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/merge/CrossPartitionMergePatientIdModeR4Test.java
mrdnctrk: yes, this isn't necessary. I removed it here, and from other merge tests as well

PR #7633, opened 2026-03-10
Fix: Preserve meta.source during NOP updates

### LINKED ISSUE #7632: Bug: meta.source gets cleared during NOP updates
## Summary
When updating a FHIR resource with identical content but a different `meta.source` value, the source property gets cleared or overwritten with stale data. This affects resource provenance tracking.

## Environment
- **HAPI FHIR Version**: Latest (main branch)

## Steps to Reproduce
1. Create a resource with `meta.source` set (e.g., `"initial-request-123"`)
2. Update the same resource with identical content but different `meta.source` (e.g., `"updated-request-456"`)
3. Observe that `meta.source` is either cleared or shows the old value instead of the new one

## Expected Behavior
The `meta.source` should reflect the value from the current request (`"updated-request-456"`), even when the resource content hasn't changed and the update is detected as a NOP (No Operation).

## Actual Behavior
The `meta.source` gets cleared or shows stale data from previous requests (`"initial-request-123"` or `null`).

## Impact
- **Resource Provenance**: Affects tracking of resource origins and request correlation
- **Audit Trails**: Makes it difficult to trace which system/request last touched a resource
- **Integration**: Impacts systems relying on `meta.source` for workflow routing or compliance

## Related
- Original issue: GitLab 5271
- Affects both direct `meta.source` setting and header-based source capture

jorgepinilla-smile:
## Summary
Fixes the issue where `meta.source` gets cleared during NOP (No Operation) updates.

## Issue
Closes #7632

## Changes
- Added test case that reproduces the issue where `meta.source` is not preserved during NOP updates
- Test demonstrates that when updating a resource with identical content but different `meta.source` value, the source should reflect the current request's value

## Test Case
The test `testUpdateWithNoChanges_PreservesMetaSource()` in `FhirResourceDaoR4UpdateTest`:
1. Creates a practitioner with `meta.source = "initial-request-123"`
2. Updates with same content but `meta.source = "updated-request-456"`  
3. Verifies that the final `meta.source` reflects the updated value

## Status
- [x] Test case added (currently failing - reproduces the bug)
- [x] Fix implementation

## Impact
- Fixes resource provenance tracking during NOP updates
- Ensures audit trails correctly reflect the latest request source
- Maintains compliance for systems relying on `meta.source`

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
@@ -1040,10 +1040,13 @@ public abstract class BaseHapiFhirDao<T extends IBaseResource> extends BaseStora
 			if (theResource != null) {
 				myJpaStorageResourceParser.updateResourceMetadata(entity, theResource);
 
-				if (entity.getCurrentVersionEntity() != null) {
-					String sourceId = entity.getCurrentVersionEntity().getSourceUri();
-					String requestId = entity.getCurrentVersionEntity().getRequestId();
-					MetaUtil.populateResourceSource(myFhirContext, sourceId, requestId, theResource);
+				ResourceHistoryTable entityCurrentVersion = entity.getCurrentVersionEntity();
+				if (entityCurrentVersion != null) {
+					MetaUtil.populateResourceSource(
+							myFhirContext,
+							entityCurrentVersion.getSourceUri(),
+							entityCurrentVersion.getRequestId(),
+							theResource);
 				}
 			}
 			entity.setUnchangedInCurrentOperation(true);


jorgepinilla-smile:
Primary Root Cause:
  Incomplete Metadata Population in NOP Detection Path (This is directly related to the issue):
  When HAPI FHIR detects a No Operation (NOP) update, the system takes a shortcut path that calls.
  updateResourceMetadata() but fails to populate meta.source field.

Secondary Root Cause
  Insufficient Change Detection Logic (This is a side effect of the current logic):
  The original NOP detection only considered resource content changes, ignoring metadata changes like
  meta.source. This caused legitimate updates (where only source tracking changed) to be incorrectly
  classified as NOPs.

Impact Analysis

  Affected Scenarios
  - NOP updates with X-Request-ID header changes
  - Meta.source updates without content changes
  - Provenance tracking in high-frequency update systems
  - Audit trail integrity

  FHIR Compliance Issue
  According to FHIR R4 spec, any update should increment version, including metadata-only changes. Original
  NOP behavior violated this by suppressing version increments for valid updates.

robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
```diff
@@ -1039,6 +1059,11 @@ public ResourceTable updateEntity(
 					entity.getIdDt().toUnqualified().getValue());
 			if (theResource != null) {
 				myJpaStorageResourceParser.updateResourceMetadata(entity, theResource);
+				String sourceId =
+						((ResourceTable) theEntity).getCurrentVersionEntity().getSourceUri();
```
jmarchionatto: Use casted `entity` from L938

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4UpdateTest.java
```diff
@@ -1343,4 +1344,37 @@ public void invoke(IPointcut thePointcut, HookParams theArgs) {
...
+		updatePractitioner.setActive(true);
+		updatePractitioner.getMeta().setSource("updated-request-456");
+
+		IIdType updatedId = myPractitionerDao.update(updatePractitioner, mySrd).getId();
+
+		// Read the updated resource
+		Practitioner updated = myPractitionerDao.read(updatedId.toUnqualifiedVersionless(), mySrd);
+
+		// ASSERTION: meta.source should reflect the updated value from the request
+		// Currently this will fail because populateResourceMetadata() overwrites
```
jmarchionatto: suggestion: last two comment lines  are unnecessary

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4UpdateTest.java
```diff
@@ -1343,4 +1344,37 @@ public void invoke(IPointcut thePointcut, HookParams theArgs) {
...
+
+		// Now perform UPDATE with same content but different meta.source
+		// This should be detected as NOP but source should still be updated
+		Practitioner updatePractitioner = new Practitioner();
+		updatePractitioner.setId(id);
+		updatePractitioner.addName().setFamily("TestPractitioner");
+		updatePractitioner.setActive(true);
+		updatePractitioner.getMeta().setSource("updated-request-456");
+
+		IIdType updatedId = myPractitionerDao.update(updatePractitioner, mySrd).getId();
```
jmarchionatto: The test verifies that meta.source is updated, but doesn't assert that the update was actually treated as a NOP (version should remain at 1). Without this, the test would pass even if the fix accidentally causes a full version bump for every source change — creating extra history entries for what should be NOPs.

jorgepinilla-smile: You're right, I should update the ticket description cause I noticed that according to fhir spec we should be incrementing the version whenever there's a meta data update. So it shouldn't be a NOP update but just a regular update.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4UpdateTest.java
jorgepinilla-smile: removed unnecessary comment lines

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
jorgepinilla-smile: Good catch! I've updated the changes accordingly.

jorgepinilla-smile:
> Missing a changelog

Good catch! Added change log

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7559-fix-meta-source-nop-updates.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: fix
+jira: SMILE-7559
```
epeartree, line 3: Previously, the value provided in header `X-Request-ID` was disappearing from the meta.source when performing a NOP resource update.  This issue is Fixed.

jorgepinilla-smile: I've updated the change log with the suggested message.

Issue #7630, opened 2026-03-10
Bug: Integer overflow in BaseInterceptorService.BaseInvoker.compareTo


ttnTrifork:
**Describe the bug**
Integer overflow in `BaseInterceptorService.BaseInvoker.compareTo`

**To Reproduce**
Register 3 interceptors

1. with `@Interceptor(order = Int.MIN_VALUE)`
2. with `@Interceptor(order = -1)`
3. with `@Interceptor`

**Expected behavior**
The interceptors are invoked in turn, from lowest to highest order value

**Actual behavior**
Due to the implementation of compareTo
```
		@Override
		public int compareTo(IInvoker theInvoker) {
			return myOrder - theInvoker.getOrder();
		}
```

An integer overflow happens when the 3. interceptor (with order 0) is compared to the 1. interceptor (with order Int.MIN_VALUE)

0 - (-2,147,483,648) = 2,147,483,648  → exceeds Integer max value  → Overflow to negative → -2,147,483,648

This causes 2. interceptor to be invoked last

**The fix**
`return Integer.compare(myOrder, theInvoker.getOrder());`

I will open a PR with tests

PR #7641, opened 2026-03-11
fix(7630): Integer overflow during comparison of `@Interceptor` and `@Hook` order values

ttnTrifork:
Comparison of `@Interceptor` and `@Hook` order values can no longer cause integer overflow and lead to incorrect order when the difference between them overflows int range (e.g., Integer.MIN_VALUE - 100)

This fixes https://github.com/hapifhir/hapi-fhir/issues/7630


robogary:
Formatting check succeeded!

PR #7642, opened 2026-03-11
Add isPartitioningEnabled() to IDefaultPartitionSettings.

lukedegruchy:
- Add isPartitioningEnabled() to IDefaultPartitionSettings.
- Decouple PackageInstallerSvcImpl from PartitionSettings
- Promote isPartitioningEnabled() from the concrete PartitionSettings class to the IDefaultPartitionSettings interface, reducing unnecessary coupling to JPA-specific types. Refactor PackageInstallerSvcImpl to depend on the interface instead of the concrete class, since it only uses methods available on IDefaultPartitionSettings.

Closes https://github.com/hapifhir/hapi-fhir/issues/7643

### DIFF
--- a/hapi-fhir-base/src/test/java/ca/uhn/fhir/interceptor/model/RequestPartitionIdTest.java
+++ b/hapi-fhir-base/src/test/java/ca/uhn/fhir/interceptor/model/RequestPartitionIdTest.java
@@ -187,6 +187,11 @@ public class RequestPartitionIdTest {
 			public boolean isPartitioningEnabled() {
 				return false;
 			}
+
+			@Override
+			public boolean isIncludePartitionInSearchHashes() {
+				return false;
+			}
 		};
 		return new ContainsTestCase[]{
 			new ContainsTestCase("all vs all", allPartitions(), allPartitions(), EQUAL),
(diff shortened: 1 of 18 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-base/src/test/java/ca/uhn/fhir/interceptor/model/RequestPartitionIdTest.java
```diff
@@ -182,7 +182,12 @@ void testContains(ContainsTestCase theTestCase) {
 	}
 
 	static ContainsTestCase[] getContainsTestCases() {
-		IDefaultPartitionSettings nullDefaultPartition = new IDefaultPartitionSettings() {};
+		IDefaultPartitionSettings nullDefaultPartition = new IDefaultPartitionSettings() {
+			@Override
+			public boolean isPartitioningEnabled() {
```
codeforgreen, line 187: maybe the interface can have this default behaviour?

lukedegruchy: I made the explicit decision to force implementors to implement this contract.

My opinion is that by defaulting to false, there's a risk of some implementation in the future forgetting to make its own decision of how to handle partition, and we could end up falsely claiming there's no partitioning.

PR #7645, opened 2026-03-11
7644 make sure reasources are inserted before sps when processing bundle

### LINKED ISSUE #7644: Token search param row is being inserted before its parent resource row for some search params
When a bundle is being inserted, conditions may lead to inserting resource search param rows before the resource is inserted. When this happens, a foreign key violation error is raised by the database.

another-luis:
When a transaction bundle is submitted that causes auto-placeholder creation (e.g. a Composition with a relatesTo reference to another Composition that doesn't exist), Hibernate queues an INSERT for the placeholder's HFJRESOURCE row and INSERTs for its search param rows (e.g. HFJSPIDX_TOKEN).

With JDBC batch mode enabled, Hibernate's TopologicalSortInsertActionSorter sorts these INSERTs so that parent rows are always inserted before their dependents. It does this by following @ManyToOne object references — if a search param entity holds a reference to its parent ResourceTable, the sorter knows to insert the resource first.

The bug is that setResource() on ResourceIndexedSearchParamToken, String, and Uri was not assigning the myResource field, leaving it null on newly created entities. The sorter therefore saw no dependency between the token row and the resource row, and could emit them in either order.

On the first bundle submission this went unnoticed because all entities in the session were new INSERTs and the sorter produced a valid order. On the second submission (after deleting the placeholder), the Observation already existed and was being updated, which changed the shape of the Hibernate session enough for the sorter to emit the placeholder's token INSERTs before the placeholder's resource INSERT — violating the FK constraint.

The fix assigns myResource in setResource() for the three affected entity types, giving the sorter the information it needs to always order the resource INSERT first.

Fixes #7644

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoSearchParamSynchronizer.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoSearchParamSynchronizer.java
@@ -131,16 +131,11 @@ public class DaoSearchParamSynchronizer {
 			next.setResourceId(theEntity.getId().getId());
 			next.setPartitionId(theEntity.getPartitionId());
 			next.calculateHashes();
-			// Populate @ManyToOne entity references so that Hibernate's InsertActionSorter can detect
-			// FK dependencies and order INSERT statements correctly when JDBC batch mode is active.
-			// Without these references, search param and link rows may be inserted before their parent
-			// HFJ_RESOURCE row, causing FK constraint violations.
+			// set the resource table for some search parameters because otherwise there's a risk the search
+			// parameter could be inserted into the DB before the resource it depends on.
 			if (next instanceof BaseResourceIndexedSearchParam searchParam) {
 				searchParam.setResource(theEntity);
 			}
-			// For resource links, populate the target @ManyToOne reference using a Hibernate proxy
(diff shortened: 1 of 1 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/index/DaoSearchParamSynchronizer.java
```diff
@@ -129,6 +131,20 @@ private <T extends BaseResourceIndex> void synchronize(
 			next.setResourceId(theEntity.getId().getId());
 			next.setPartitionId(theEntity.getPartitionId());
 			next.calculateHashes();
+			// Populate @ManyToOne entity references so that Hibernate's InsertActionSorter can detect
+			// FK dependencies and order INSERT statements correctly when JDBC batch mode is active.
+			// Without these references, search param and link rows may be inserted before their parent
+			// HFJ_RESOURCE row, causing FK constraint violations.
```
TipzCM: This comment is very confusing to me.

It's basically trying to say that it needs to set the `resource` value for some search parameters because otherwise there's a risk the search parameter could be inserted into the DB before the resource it's depending on is.

But, like... in a more complicated way?

It also references the link (for a similar order-of-operations thing), but it's also commented on separately below (is this intentional?)

PR #7657, opened 2026-03-16
SMILE-11858: Fix no-op PUT version bump with null myResourceTypeId

tadgh:
Co-authored by Claude Opus 4.6

## Summary

This fix prevents spurious version increments on `HFJ_RESOURCE` when a no-op PUT is performed on a resource that was created before the V8_4_0 database migration introduced the `RES_TYPE_ID` column. For such pre-upgrade resources the column is `NULL`, and HAPI's DAO eagerly populates it during `populateResourceIntoEntity()`. Because `ResourceTable.myResourceTypeId` lacked `@OptimisticLock(excluded = true)`, Hibernate's dirty-checking treated the `NULL`→value transition as a version-triggering change — even when no-op detection had already decided not to create a new `ResourceHistoryTable` entry. The result was `RES_VER` incrementing without a matching history row, breaking version-history integrity. The fix is a one-line annotation addition that aligns `myResourceTypeId` with the 33+ other housekeeping fields in `ResourceTable` that already carry this annotation.

1. Added `@OptimisticLock(excluded = true)` to `ResourceTable.myResourceTypeId` so that lazy population of the column after a schema migration does not trigger Hibernate's optimistic-lock version increment.
2. Added a primary reproduction test (`testNoOpPutWithNullResourceTypeId_VersionShouldNotBeIncremented`) that simulates a pre-upgrade resource by nulling `RES_TYPE_ID` via native SQL, performs a no-op PUT, and asserts that both the DAO return value and the raw database row remain at version 1 with no orphan history entry.
3. Added three adjacent tests covering: no-op PUT where `RES_TYPE_ID` is already populated (never-broken path), content-change PUT with a null `RES_TYPE_ID` (version must still increment correctly), and create-then-no-op PUT (field already populated on create, so version stays at 1).

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7657-fix-noop-put-version-bump.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7657-fix-noop-put-version-bump.yaml
@@ -1,0 +1,10 @@
+---
+type: fix
+jira: SMILE-11858
+title: "Previously, performing a no-op PUT (update with unchanged content) on a resource that existed
+  before the V8_4_0 database migration (which introduced the RES_TYPE_ID column) would incorrectly
+  increment the resource version number without creating a corresponding history entry. This was caused
+  by Hibernate treating the lazy population of the RES_TYPE_ID field (from null to a resolved value)
+  as a dirty change, triggering an unintended version bump at transaction commit. The ResourceTable
+  entity field has been annotated to exclude it from optimistic locking, so no-op PUTs on pre-upgrade
+  resources no longer cause spurious version increments."


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4UpdateTest.java
```diff
@@ -1343,4 +1343,157 @@ public void invoke(IPointcut thePointcut, HookParams theArgs) {
...
+		IIdType id = myPatientDao.create(patient, mySrd).getId().toUnqualified();
+		assertThat(id.getVersionIdPart()).isEqualTo("1");
+
+		Long pid = id.getIdPartAsLong();
+
+		// Simulate a pre-upgrade state: null out RES_TYPE_ID using native SQL
+		// to avoid Hibernate dirty checking on the null-out itself
+		runInTransaction(() -> {
+			myEntityManager.createNativeQuery(
+				"UPDATE HFJ_RESOURCE SET RES_TYPE_ID = NULL WHERE RES_ID = " + pid
```
elavy-harris, line 1368: Shouldn't be checking `RES_TYPE` and `FHIR_ID`, not `RES_ID`?

tadgh: This is just for fetching the affected row for update, so I believe any lookup works.

There is only one entity in the DB at this point during the test. I could even remove the RES_ID clause entirely and the test is effectively the same. Is there some nuance you think would be improved if I were to replace this WHERE clause with `WHERE FHIR_ID=1 AND RES_TYPE="Patient"`?

 Happy to do so if you think it will make a material difference, but this is just DB hacking to get the DB into a state that "looks" like the failed upgrade situation.

elavy-harris: > This is just for fetching the affected row for update, so I believe any lookup works.
> 
> There is only one entity in the DB at this point during the test. I could even remove the RES_ID clause entirely and the test is effectively the same. Is there some nuance you think would be improved if I were to replace this WHERE clause with `WHERE FHIR_ID=1 AND RES_TYPE="Patient"`?
> 
> Happy to do so if you think it will make a material difference, but this is just DB hacking to get the DB into a state that "looks" like the failed upgrade situation.

It should be fine as a test.

elavy-harris:
Re-asking because I think my question got lost before:

Will this also address reindex? It currently causes the same behavior as a no-op PUT.

tadgh:
> Re-asking because I think my question got lost before:
> 
> Will this also address reindex? It currently causes the same behavior as a no-op PUT.

Feel free to build the jars and check. I don't see why it wouldn't though(reindex effectively does a no-op put when it finds no delta, by dirtying the hibernate session).

PR #7665, opened 2026-03-17
`PointcutLatch` causes cascading test failures

### LINKED ISSUE #7664: PointcutLatch causes cascading test failures
For tests using `PointcutLatch`.
When a test fails with a `PointcutLatchException`, the subsequent tests can also fail depending on the test setup which makes it hard to investigate the failure.

**To Reproduce**

Setup: ourDeliveryLatch is a static field shared across tests.                                                                                                                                             
                                                                                                                                                                                                           
  ---                                                                                                                                                                                                        
  Test A — sets up a latch but fails before awaiting:                                                                                                                                                        
```
  ourDeliveryLatch.setExpectedCount(1);  // session started
  createAndStorePatient();// test fails here (assertion, timeout, etc.)                                                                                                                       
  ourDeliveryLatch.awaitExpected(); // NEVER REACHED — session still active   
```                                                                                                                          
                                                                                                                                                                                                             
  @AfterEach — the old version just called clear() which threw AssertionError if unexpected invocations exist
 OR leaves stale session if clear() isn't called at all

  ---
  Test B — runs next, tries to start fresh:
```
  ourDeliveryLatch.setExpectedCount(1);
  // throws: "setExpectedCount() called before previous awaitExpected() completed"
  // ↑ This is Test A's stale session, not Test B's fault
```

  Result: Test B fails with a confusing error pointing at Test A's stack trace. If many tests share the latch, the failures keep cascading down the suite.

  ---
  The two root causes:

  1. Stale session — myPointcutLatchSession is never nulled out because awaitExpected() was skipped. The next setExpectedCount() sees it and throws.
  2. @AfterEach clear() itself throwing — if clear() checks exceptions and throws, it masks the original failure with a cleanup error, making the real cause even harder to find.

**Expected behaviour**
`PointcutLatch` should only cause failures with the problematic test and should isolate the issue.

- HAPI FHIR Version 8.8

codeforgreen:
Fixes #7664

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7664-fix-pointcut-latch-cascading-failure.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7664-fix-pointcut-latch-cascading-failure.yaml
@@ -1,0 +1,6 @@
+---
+type: fix
+issue: 7664
+title: "Fixed PointcutLatch to avoid cascade test failures by no longer throwing immediately
+when invoke() is called outside of a session. Unexpected invocations are now collected
+silently and surfaced as an AssertionError on the test thread when clear() is called."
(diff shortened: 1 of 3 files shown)


robogary:
Formatting check succeeded!

PR #7666, opened 2026-03-18
Updated - Fix search in Patient Id when query references are mixed between Patient and non-Patient

epeartree:


robogary:
Formatting check succeeded!

Issue #7331, opened 2025-10-27
ApacheHttp5Request contains the relative URL path which breaks UrlTenantSelectionInterceptor


bcrosbyva:
**Describe the bug**
When using `UrlTenantSelectionInterceptor` on a client that is also using `ApacheHttp5RestfulClientFactory` for connections, the validation at [UrlTenantSelectionInterceptor#L82](https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-client/src/main/java/ca/uhn/fhir/rest/client/interceptor/UrlTenantSelectionInterceptor.java#L82) gets tripped:

> java.lang.IllegalArgumentException: Request URI /hapi/v1/fhir/Appointment/1 does not start with server base http://example.com/hapi/v1/fhir
> 	at org.apache.commons.lang3.Validate.isTrue(Validate.java:572)
> 	at ca.uhn.fhir.rest.client.interceptor.UrlTenantSelectionInterceptor.request(UrlTenantSelectionInterceptor.java:82)

This is happening because ApacheHttp5Request.getUri() is delegating to HttpUriRequest.getRequestUri() which returns the URI path, instead of HttpUriRequest.getUri() which returns the full URI.

**To Reproduce**
```java
FhirContext ctx = FhirContext.forR4Cached();
ApacheHttp5RestfulClientFactory clientFactory = new ApacheHttp5RestfulClientFactory(ctx);
ctx.setRestfulClientFactory(clientFactory);
ctx.getRestfulClientFactory().setServerValidationMode(ServerValidationModeEnum.NEVER);
final IGenericClient client = ctx.getRestfulClientFactory()
                .newGenericClient("http://example.com/hapi/v1/fhir");

UrlTenantSelectionInterceptor tenantSelection = new UrlTenantSelectionInterceptor();
client.registerInterceptor(tenantSelection);
tenantSelection.setTenantId("123");
Appointment appt = client.read().resource(Appointment.class).withId("1").execute();
```

**Environment (please complete the following information):**

- HAPI FHIR Version: 8.4.0
- OS: iOS
- Browser: N/A

PR #7675, opened 2026-03-19
7331: Fix ApacheHttp5Request.getUri() returning path instead of full URI

kathbigra:
Fixes: #7331
  getRequestUri() returns only the request path, causing UrlTenantSelectionInterceptor
  to throw IllegalArgumentException when used with ApacheHttp5RestfulClientFactory.
  Fixed by using getUri() which returns the full URI.


robogary:
Formatting check succeeded!

PR #7679, opened 2026-03-19
7347 invalidate valueset cache when underlying codesystem is updated

### LINKED ISSUE #7678: Updating a CodeSystem with a dependent ValueSet could cause validation to incorrectly reject associated codes

another-luis:
Closes #7678

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoCodeSystem.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoCodeSystem.java
@@ -252,7 +252,7 @@ public class JpaResourceDaoCodeSystem<T extends IBaseResource> extends BaseHapiF
 				myTerminologyCodeSystemStorageSvc.storeNewCodeSystemVersionIfNeeded(
 						cs, (ResourceTable) theEntity, theRequest);
 
-				String codeSystemUrl = cs.getUrl();
+				String codeSystemUrl = cs != null ? cs.getUrl() : null;
 				if (isNotBlank(codeSystemUrl)) {
 					int invalidated = myTerminologySvc.invalidatePreCalculatedExpansionOfValueSetsContainingCodeSystem(
 							codeSystemUrl);


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/JpaResourceDaoCodeSystem.java
```diff
@@ -251,6 +251,18 @@ public ResourceTable updateEntity(
 
 				myTerminologyCodeSystemStorageSvc.storeNewCodeSystemVersionIfNeeded(
 						cs, (ResourceTable) theEntity, theRequest);
+
+				String codeSystemUrl = cs.getUrl();
+				if (isNotBlank(codeSystemUrl)) {
+					int invalidated = myTerminologySvc.invalidatePreCalculatedExpansionOfValueSetsContainingCodeSystem(
```
codeforgreen, line 257: Did you have a chance to look at what happens in `JpaResourceDaoValueSet`? When a `ValueSet` is updated, it will delete all expansion together with existing `TermValueSet`. There is a Quartz job which picks up `NOT_EXPANDED` `TermValueSet` and expands in the background via `TermReadSvcImpl#preExpandDeferredValueSetsToTerminologyTables`. So I assume something similar can be done here. I could be wrong but I thought to ask.

another-luis: Isn't that similar to what I am doing here? The method invalidatePreCalculatedExpansionOfValueSetsContainingCodeSystem (in TermReadSvcImpl) does exactly that - it looks for TermValueSets that depend on the updated CodingSystem, that also have a status of EXPANDED, or EXPANSION_IN_PROGRESS. If it finds any, the method sets their status to NOT_EXPANDED so that the Quartz job expands them asynchronously.

Did I miss something?

codeforgreen: The difference would be, there `TermValueSet` is being deleted an re-created while here they are being updated with only the expansion status. I wonder if there is any other change which needs to be reflected from the `CodeSystem`. Probably not in which case it's all good.

another-luis: It makes sense to delete the TermValueSets when the ValueSet is updated to cover the possibility that entries in the ValueSet have been removed. Otherwise they could stay in the DB forever. Updating the CodeSystem is a bit different because all the entries in the ValueSet continue to be valid, they just need to be re-expanded to use the updated values in the udated CodeSystem.

Let me know if you agree so I can close the conversation. If not, let's meet to discuss more efficiently.

PR #7695, opened 2026-03-23
Rework XXE protection

jamesagnew:
This PR does not make any functional changes. XXE vulnerabilities were mitigated in the HAPI FHIR codebase roughly 10 years ago.

This PR moves the XXE protection in `XmlUtil` so that it is less likely to produce a false positive by vulnerability scanners.


robogary:
Formatting check succeeded!

Issue #1061, opened 2018-08-23
Results not returned when using _sort


seanmcilvenna:
**Describe the bug**
If I do a search and _sort by "name,title", some entries do not get returned (entries that either don't have a name or title).
In my case, this turns a list of 77 potential results into 44 simply because of sorting.
Is this expected behavior, and if so, is there a way to change it?

**To Reproduce**
1. Load several StructureDefinition resources into an instance of the FHIR server, where at least one of the StructureDefinition resources does *not* have a title.
2. Perform a GET /baseDstu3/StructureDefinition?_sort=name,title

**Expected behavior**
The results returned do not include the StructureDefinition that does not have a title, even though it should; and the one(s) missing the title should be last in the results.

**Environment (please complete the following information):**
 - HAPI FHIR Version: 3.5.0-SNAPSHOT
 - OS: Windows Server 2016
 - Browser: Chrome/Postman

akanduru:
I see similar problem. I am using version 3.6.0 for R4 resources. 
I have four Questionnaires with three of them missing date field. A query like `/baseR4/Questionnaire` returns four items and `/baseR4/Questionnaire/_search?_sort=date` returns one.

I don't see the problem on 3.1.0 which I use for STU3 Questionnaires. It looks like the bug is introduced after 3.1.0.

nicozorza:
Any updates on this subject? I have the same problem.

michaelabuckley:
This was fixed in 2022.  It does not reproduce in the current release.  Please file a new ticket if you encounter this.

See https://github.com/hapifhir/hapi-fhir/pull/4099

Issue #818, opened 2018-01-09
Sort parameter may reduce number of results


jbalbien:
In some instances, the use of the _sort parameter reduces the number of return results. For example:

https://fhirtest.uhn.ca/baseDstu3/ProcedureRequest -- **1181 total results**

https://fhirtest.uhn.ca/baseDstu3/ProcedureRequest?_sort=_lastUpdated -- **1181 total results**

https://fhirtest.uhn.ca/baseDstu3/ProcedureRequest?_sort=authored -- **1085 total results**

https://fhirtest.uhn.ca/baseDstu3/ProcedureRequest?_sort=priority -- **1126 total results**

gansoper:
Hello, found the same issue for Slots. 

baseDstu3/Slot?_count=100&&schedule=215&_sort=start returns all the slots which have "free" status. 
But if I remove _sort=start - it starts to return all slots (expected behavior with sort)

nicozorza:
Any updates on this subject? I have the same problem.

davidballester:
We found that, when specifying `_sort=-_lastUpdated`, having pagination enabled with 10 results by default, each page would get just 4 resources out of a total of 14, and we'd just get 8 results, so 4 were missing.

However, doing the same query and specifying `_count=10`, which, again, is the default page size of our database backed pagination, will behave properly: 10 results on the first page, and the remaining 4 results in the next page.

michaelabuckley:
This was fixed in 2022.  It does not reproduce in the current release.  Please file a new ticket if you encounter this.

See https://github.com/hapifhir/hapi-fhir/pull/4099

PR #7697, opened 2026-03-23
Add sorting tests

jamesagnew:
Just testing whether these tests fail on CI.

Trying to determine whether #818 and #1061 are still an issue


robogary:
Formatting check succeeded!

PR #7700, opened 2026-03-24
Fix PatientIdPartitionInterceptor to include default partition for NON_UNIQUE_COMPARTMENT_IN_DEFAULT (#7699)

### LINKED ISSUE #7699: PatientIdPartitionInterceptor does not include default partition for NON_UNIQUE_COMPARTMENT_IN_DEFAULT on read
When PatientIdPartitionInterceptor is active with the NON_UNIQUE_COMPARTMENT_IN_DEFAULT policy, resources that reference multiple patients (e.g., a merge Provenance targeting Patient/A and Patient/B) are stored in the default partition. However, when searching for those resources by patient compartment param (e.g., target=Patient/A), the interceptor only returns the patient-specific partition — missing the default partition where the resource actually lives.

The interceptor fetches the policy but does not use it to include the default partition on read. The fix is to consult the policy getPartitionIdForNonUniqueCompartment() and merge the default partition into the search partition set.

Workaround was to use SystemRequestDetails.forAllPartitions() to bypass the interceptor entirely, which is overly broad.

mrdnctrk:
Fixes #7699

## Summary
- There was a bug where PatientIdPartitionInterceptor.identifyForRead did not include the default partition when searching for resources with NON_UNIQUE_COMPARTMENT_IN_DEFAULT policy. Resources stored in the default partition (due to referencing multiple patients, e.g., merge Provenance) were missed during searches.
- The fix merges the non-unique compartment partition (default partition) into the search partition set when the policy allows it.
- Removed the forAllPartitions() workarounds in ReplaceReferencesProvenanceSvc and ReplaceReferencesTestHelper that were bypassing the interceptor.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7699-patientid-interceptor-nonunique-default-partition.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7699-patientid-interceptor-nonunique-default-partition.yaml
@@ -1,0 +1,6 @@
+---
+type: fix
+issue: 7699
+title: "There was a bug where PatientIdPartitionInterceptor did not include the default partition when
+  searching for resources with NON_UNIQUE_COMPARTMENT_IN_DEFAULT policy, causing resources stored in the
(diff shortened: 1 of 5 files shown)


robogary:
Formatting check succeeded!

PR #7706, opened 2026-03-25
[7702] Add support for ValueSet expansion filters using the EQUAL ope…

### LINKED ISSUE #7702: Fix error when expand ValueSet with compose.include.filter that has "=" operator
GET /ValueSet/SemanticDrugVS/$expand` returned a 400 error when Hibernate Search was disabled and the ValueSet contains compose.include.filter that has "=" operator

> HAPI-0894: Don't know how to handle op=EQUAL on property TTY - Note that Hibernate Search is disabled on this server so not all ValueSet expansion functionality is available.

YalingPeiS:
GET /ValueSet/SemanticDrugVS/$expand` returned a 400 error when Hibernate Search was disabled and the ValueSet contains compose.include.filter that has "=" operator

> HAPI-0894: Don't know how to handle op=EQUAL on property TTY - Note that Hibernate Search is disabled on this server so not all ValueSet expansion functionality is available.

Added support to ValueSet expansion that contains `compose.include` filters using the EQUAL operator on concept properties when Hibernate Search is disabled.

Closes #7702

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
@@ -1952,6 +1952,8 @@ public class TermReadSvcImpl implements ITermReadSvc, IHasScheduledJobs {
 
 		if (!equalPropertyFilters.isEmpty()) {
 			Collection<TermConcept> concepts = myConceptDao.fetchConceptsAndPropertiesByVersionPid(theVersion.getPid());
+			// Pre-load designations into the session cache to avoid N+1 lazy loading in the loop below
+			myConceptDao.fetchConceptsAndDesignationsByVersionPid(theVersion.getPid());
 			for (TermConcept next : concepts) {
 				if (conceptMatchesAllPropertyFilters(next, equalPropertyFilters)) {
 					addCodeIfNotAlreadyAdded(
@@ -2017,8 +2019,8 @@ public class TermReadSvcImpl implements ITermReadSvc, IHasScheduledJobs {
 			TermConcept theConcept, List<ValueSet.ConceptSetFilterComponent> theFilters) {
 		for (ValueSet.ConceptSetFilterComponent filter : theFilters) {
 			boolean matchesFilter = theConcept.getProperties().stream()
-					.anyMatch(p -> filter.getProperty().equals(p.getKey())
-							&& filter.getValue().equals(p.getValue()));
+					.anyMatch(p -> Objects.equals(filter.getProperty(), p.getKey())
+							&& Objects.equals(filter.getValue(), p.getValue()));
 			if (!matchesFilter) {
 				return false;
 			}
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4ValueSetHSearchDisabledTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4ValueSetHSearchDisabledTest.java
@@ -43,7 +43,6 @@ public class ResourceProviderR4ValueSetHSearchDisabledTest extends BaseJpaTest {
 
 	private static final org.slf4j.Logger ourLog = org.slf4j.LoggerFactory.getLogger(ResourceProviderR4ValueSetHSearchDisabledTest.class);
 
-
 	@Autowired
 	private FhirContext myFhirCtx;
 	@Autowired
@@ -151,31 +150,12 @@ public class ResourceProviderR4ValueSetHSearchDisabledTest extends BaseJpaTest {
 
 	@Test
 	void testExpandByPropertyEqualFilter_whenHibernateSearchDisabled() {
-		// Set up - CodeSystem with 4 concepts, each with a TTY property
-		CodeSystem codeSystem = new CodeSystem();
-		codeSystem.setId("v3-rxNorm");
-		codeSystem.setUrl("http://www.nlm.nih.gov/research/umls/rxnorm");
-		codeSystem.setContent(CodeSystem.CodeSystemContentMode.COMPLETE);
-		codeSystem.addConcept().setCode("748856").setDisplay("Yaz 28 Day Pack")
-			.addProperty().setCode("TTY").setValue(new StringType("BPCK"));
-		codeSystem.addConcept().setCode("748832").setDisplay("Levonorgestrel Pack")
-			.addProperty().setCode("TTY").setValue(new StringType("GPCK"));
-		codeSystem.addConcept().setCode("95267").setDisplay("glucose 500 MG/ML Oral Solution")
-			.addProperty().setCode("TTY").setValue(new StringType("SCD"));
-		codeSystem.addConcept().setCode("91668").setDisplay("tetracycline Ophthalmic Ointment")
-			.addProperty().setCode("TTY").setValue(new StringType("SBD"));
+		// Set up - CodeSystem with 3 concepts, each with a TTY property
+		CodeSystem codeSystem = mockCodeSystem("cs-id-1");
 		persistCodeSystem(codeSystem);
 
-		// ValueSet filtering on TTY=SBD and TTY=SCD
-		ValueSet valueSet = new ValueSet();
-		valueSet.setId("NonPackSemanticDrugVS");
-		valueSet.setUrl("http://hl7.org/fhir/us/davinci-drug-formulary/ValueSet/NonPackSemanticDrugVS");
-		valueSet.getCompose().addInclude()
-			.setSystem("http://www.nlm.nih.gov/research/umls/rxnorm")
-			.addFilter().setProperty("TTY").setOp(ValueSet.FilterOperator.EQUAL).setValue("SBD");
-		valueSet.getCompose().addInclude()
-			.setSystem("http://www.nlm.nih.gov/research/umls/rxnorm")
-			.addFilter().setProperty("TTY").setOp(ValueSet.FilterOperator.EQUAL).setValue("SCD");
+		// ValueSet filtering on TTY=SBD
+		ValueSet valueSet = mockValueSet("vs-id-1", "cs-id-1");
 		persistValueSet(valueSet);
 
 		// Execute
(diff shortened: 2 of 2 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
```diff
@@ -1992,6 +2013,19 @@ private void expandWithoutHibernateSearch(
 		}
 	}
 
+	private boolean conceptMatchesAllPropertyFilters(
+			TermConcept theConcept, List<ValueSet.ConceptSetFilterComponent> theFilters) {
+		for (ValueSet.ConceptSetFilterComponent filter : theFilters) {
+			boolean matchesFilter = theConcept.getProperties().stream()
+					.anyMatch(p -> filter.getProperty().equals(p.getKey())
+							&& filter.getValue().equals(p.getValue()));
```
volodymyr-korzh: Suggestion: use `Objects.equals(v1, v2)` here to avoid potential nullPointer. 
While `filter.getProperty()` is impossible to be null in runtime,` filter.getValue()` can be null and can throw nullPointer here. using `Objects.equals(v1, v2)` in both cases is defensive and will avoid this.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4ValueSetHSearchDisabledTest.java
```diff
@@ -146,5 +149,49 @@ public void testExpandById() throws Exception {
 
 	}
```
volodymyr-korzh, line 150: Reproduction test for N+1 issue (one designation query per concept (N+1))
I would suggest including this check (adding designation) in existing test, to help ensuring (N+1) issue is fixed
```
	@Test
	void testExpandByPropertyEqualFilter_doesNotCauseNPlus1DesignationQueries() {
		// Set up - CodeSystem with 3 concepts, each with a property and a designation
		CodeSystem codeSystem = new CodeSystem();
		codeSystem.setId("cs-n-plus-1");
		codeSystem.setUrl("http://example.org/cs-n-plus-1");
		codeSystem.setContent(CodeSystem.CodeSystemContentMode.COMPLETE);
		for (int i = 0; i < 3; i++) {
			CodeSystem.ConceptDefinitionComponent concept = codeSystem.addConcept()
				.setCode("code-" + i)
				.setDisplay("Display " + i);
			concept.addProperty().setCode("TTY").setValue(new StringType("SBD"));
			concept.addDesignation().setLanguage("en").setValue("Designation " + i);
		}
		persistCodeSystem(codeSystem);

		// ValueSet filtering on TTY=SBD
		ValueSet valueSet = new ValueSet();
		valueSet.setId("NPlusOneTestVS");
		valueSet.setUrl("http://example.org/ValueSet/NPlusOneTestVS");
		valueSet.getCompose().addInclude()
			.setSystem("http://example.org/cs-n-plus-1")
			.addFilter().setProperty("TTY").setOp(ValueSet.FilterOperator.EQUAL).setValue("SBD");
		persistValueSet(valueSet);

		// Execute with query capture
		Parameters responseParam = myServer.getFhirClient()
			.operation()
			.onInstance(myExtensionalVsId)
			.named("expand")
			.withNoParameters(Parameters.class)
			.execute();
		ValueSet expanded = (ValueSet) responseParam.getParameter().get(0).getResource();

		// Verify all 3 concepts are returned
		List<String> codes = expanded.getExpansion().getContains().stream()
			.map(ValueSet.ValueSetExpansionContainsComponent::getCode)
			.toList();
		assertThat(codes).containsExactlyInAnyOrder("code-0", "code-1", "code-2");

		// Check for N+1: count queries that hit the designation table.
		// With proper eager fetching, designations should be loaded in the same query as concepts.
		// With N+1, there would be one extra query per matching concept.
		long designationQueries = myCaptureQueriesListener.getSelectQueries().stream()
			.map(q -> q.getSql(true, false))
			.filter(sql -> sql.toLowerCase().contains("trm_concept_desig"))
			.count();
		ourLog.info("Designation queries: {}", designationQueries);
		myCaptureQueriesListener.logSelectQueries();

		// Should be at most 1 query for designations (eagerly fetched), not N
		assertThat(designationQueries)
			.as("Expected at most 1 designation query (eager fetch), but got N+1 lazy loads")
			.isLessThanOrEqualTo(1);
	}
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ITermConceptDao.java
```diff
@@ -40,6 +40,10 @@ public interface ITermConceptDao extends JpaRepository<TermConcept, TermConcept.
 			+ "WHERE t.myCodeSystemVersionPid = :pid")
 	List<TermConcept> fetchConceptsAndDesignationsByVersionPid(@Param("pid") Long theCodeSystemVersionPid);
 
+	@Query("SELECT DISTINCT t FROM TermConcept t LEFT JOIN FETCH t.myProperties "
+			+ "WHERE t.myCodeSystemVersionPid = :pid")
+	List<TermConcept> fetchConceptsAndPropertiesByVersionPid(@Param("pid") Long theCodeSystemVersionPid);
```
volodymyr-korzh, line 45: problem: this query can cause N+1 query issue. t.myDesignations are lazy fetched, on line 1967 of  TermReadSvcImpl when we do `next.getDesignations()`

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
```diff
@@ -1948,6 +1950,25 @@ private void expandWithoutHibernateSearch(
 			}
 		}
 
+		if (!equalPropertyFilters.isEmpty()) {
+			Collection<TermConcept> concepts = myConceptDao.fetchConceptsAndPropertiesByVersionPid(theVersion.getPid());
+			for (TermConcept next : concepts) {
```
volodymyr-korzh, line 1957: Potential fix for N+1 query problem (you can try to find a better solution if possible):
```suggestion
			Collection<TermConcept> concepts = myConceptDao.fetchConceptsAndPropertiesByVersionPid(theVersion.getPid());
myConceptDao.fetchConceptsAndDesignationsByVersionPid(theVersion.getPid());
			for (TermConcept next : concepts) {
```
Hibernate does not allow to load both Properties and Designations by one query, we can mitigate it by using 2 queries total instead of N+1. The second query loads designations into Hibernate's session cache, so next.getDesignations() won't trigger lazy loads.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ITermConceptDao.java
YalingPeiS: This query uses `JOIN FETCH` and `fetchConceptsAndDesignationsByVersionPid` uses `JOIN FETCH` too. Would this override the lazy fetch mapping since the associated entities are initialized as part of the initial query result?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
YalingPeiS: Would `JOIN FETCH` resolve the N+1 query issue?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ITermConceptDao.java
volodymyr-korzh: Yes, `JOIN FETCH` override the lazy fetch mapping, but in this query it does it only for `t.myProperties`.
Later added code executes  `next.getDesignations()` and Designations are lazy fetched. 

Ideally we need something like: 
`	@Query("SELECT DISTINCT t FROM TermConcept t LEFT JOIN FETCH t.myProperties LEFT JOIN FETCH t.myDesignations WHERE t.myCodeSystemVersionPid = :pid")`
But this is not supported by Hibernate.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
volodymyr-korzh: Yes, but we cant do this for 2 properties simultaneously in one query, I've tried and got error: 
```
Caused by: java.lang.IllegalArgumentException: org.hibernate.loader.MultipleBagFetchException: cannot simultaneously fetch multiple bags: [ca.uhn.fhir.jpa.entity.TermConcept.myDesignations,                              
ca.uhn.fhir.jpa.entity.TermConcept.myProperties]      
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
```diff
@@ -1948,6 +1950,25 @@ private void expandWithoutHibernateSearch(
...
+							theValueSetCodeAccumulator,
+							theAddedCodes,
+							theAdd,
+							theSystem,
+							theInclude.getVersion(),
+							next.getCode(),
+							next.getDisplay(),
+							next.getId(),
+							next.getParentPidsAsString(),
+							next.getDesignations());
```
volodymyr-korzh, line 1969: Just FYI: Here is where N+1 query happens to fetch Designations for each concept.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ITermConceptDao.java
volodymyr-korzh: If we execute `fetchConceptsAndPropertiesByVersionPid` and then `fetchConceptsAndDesignationsByVersionPid`, we will generate 2 queries (which is not ideal, but still better than N+1). Second query merges Designations to the session cache. (technically, yes, it overrides lazy fetch)

YalingPeiS: Updated. Thanks Volo!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
YalingPeiS: Good catch. I missed this line.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderR4ValueSetHSearchDisabledTest.java
YalingPeiS: Updated. Thanks for adding the test case.

PR #7720, opened 2026-03-30
[7718] Inefficient SQL generation for Coverage beneficiary search, multiple joins to HFJ_RES_LINK despite 1:1 cardinality

### LINKED ISSUE #7718: Inefficient SQL generation for Coverage beneficiary search, multiple joins to HFJ_RES_LINK despite 1:1 cardinality

YalingPeiS:
When performing a Coverage search using beneficiary attributes, the SQL generated by the persistence layer joins the HFJ_RES_LINK table multiple times; once for each search criterion applied to the beneficiary.

### Cause ###
The main QueryStack constructor initialized join-reuse with only PredicateBuilderTypeEnum.DATE.
Since REFERENCE was excluded, createOrReusePredicateBuilder() always created a new                        
ResourceLinkPredicateBuilder — and thus a new HFJ_RES_LINK join — for each chained parameter               
(patient.family, patient.given, patient.gender, etc.), even though they all traverse the same            
Coverage→beneficiary relationship.

### Fix ###
* Enable the existing join-cache mechanism to reuse the same ResourceLinkPredicateBuilder (and its   
single HFJ_RES_LINK join) for all chained parameters on the same reference path
* Added unit test

Closes #7718


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -179,7 +179,7 @@ public QueryStack(
 				theSqlBuilder,
 				theSearchParamRegistry,
 				thePartitionSettings,
-				EnumSet.of(PredicateBuilderTypeEnum.DATE));
+				EnumSet.of(PredicateBuilderTypeEnum.DATE, PredicateBuilderTypeEnum.REFERENCE));
```
another-luis, line 182: I thought the change was pretty reasonable, but when I ran the code-review command using Claude, it pointed out a few things. Most of them are minor, but the one below stood out as a potential issue. I looked some more at the code and after a bit of back-and-forth with Claude, it looks like it may have found an edge case that is not being covered:  _(multi-valued (0..*) reference parameters where a resource can have multiple references of different types — such as  DiagnosticReport.performer)._

Below is Claude's final statement, which you can use to generate a test that covers the edge case to make sure Claude's finding is, in fact, and issue. I think it is, based on what I read and my limited understanding of this code.

-----------------------


Potential regression for qualified chained params on multi-valued references

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java

PredicateBuilderCacheKey is keyed on (sourceJoinColumn, type, paramName) — qualifiers are not included. For a multi-valued (0..) reference parameter that can point to multiple resource types, a query using two different type qualifiers on the same param name is affected. Example: DiagnosticReport.performer is 0.. and accepts Practitioner | PractitionerRole | Organization | CareTeam:

GET /DiagnosticReport?performer:Practitioner.name=Smith&performer:Organization.name=Lab
Previously generated two separate HFJ_RES_LINK JOINs, one per AND clause. A DiagnosticReport with performer[0]=Practitioner/P1 (name=Smith) and performer[1]=Organization/O1 (name=Lab) would match because t1 and t2 could bind to different rows.

With this fix, both AND clauses share the same HFJ_RES_LINK JOIN. The combined WHERE condition becomes:

HFJ_RES_LINK.TGT_RES_ID IN (<Practitioner pids>)
AND HFJ_RES_LINK.TGT_RES_ID IN (<Organization pids>)
Since Practitioner and Organization resources have disjoint PIDs, the same HFJ_RES_LINK row cannot satisfy both conditions — the query always returns 0 results.

Note: for single-valued (0..1 or 1..1) references the AND was already impossible (a resource can't simultaneously reference both a Practitioner and an Organization via the same single-valued field), so the behavior is unchanged there.

A test covering multi-valued mixed-type chained queries is missing.

YalingPeiS: Updated and added the test for multi-valued mixed type chained search.

another-luis (approved):
Looks good!

PR #7724, opened 2026-03-31
org.hl7.fhir.core 6.9.4

dotasek:
Updates org.hl7.fhir.core to 6.9.4

Notes for reviewers:
## IWorkerContext.VersionResolutionRules
There are several method signature changes that include a new parameter 'IWorkerContext.VersionResolutionRules'. This is explained by @grahamegrieve here:
 > The validator and IG-publisher support the recently defined extension http://hl7.org/fhir/StructureDefinition/version-resolution-method, as described here: https://build.fhir.org/ig/FHIR/ig-guidance/pinning.html#specifying-not-to-use-package-dependency-based-resolution (that is, |* - |@ and |? are not yet supported)

Note that this new extension is only mentioned in the build for FHIR R6. 

HAPI will not yet support this new extension in this change.

## New methods in IPackageCacheManager

There are two new methods in IPackageCacheManager:
* org.hl7.fhir.utilities.npm.IPackageCacheManager#getLatestVersion(java.lang.String, boolean)
* org.hl7.fhir.utilities.npm.IPackageCacheManager#getLatestVersion(java.lang.String, java.lang.String)

These are only called in org.hl7.fhir.core in the following classes:

* org.hl7.fhir.validation.IgLoader - Used only for ValidationEngine and ValidationService, implementations specific to the HL7 Validator CLI
* org.hl7.fhir.validation.instance.type.ImplementationGuideValidator - Used by InstanceValidator for validating IGs. This is only ever done using an object that is explicitly a generated FilesystemPackageCacheManager instance (org.hl7.fhir.core internal class). No way for a HAPI implementation to get there.
* org.hl7.fhir.utilities.npm.FilesystemPackageCacheManager - An implementation specific to the HL7 Validator CLI. We use our own implementations in its place.

In the moment, implementing these methods in HAPI is unnecessary.


robogary:
Formatting check succeeded!

dotasek:
Note for future developers: 

For the implementation of the new IPackageCacheManager methods, this method contains logic that queries the package servers appropriately: https://github.com/hapifhir/org.hl7.fhir.core/blob/f9dedeb9e093eb99deddf7b05369567d84119911/org.hl7.fhir.utilities/src/main/java/org/hl7/fhir/utilities/npm/BasePackageCacheManager.java#L75

I believe if some simple changes were made to that class, that logic could be re-used in HAPIs PackageLoaderSvc to greatly simplify a proper implementation with shared logic.

hapi-fhir-structures-r5/src/main/java/org/hl7/fhir/r5/hapi/ctx/HapiWorkerContext.java
```diff
@@ -72,16 +70,26 @@ public HapiWorkerContext(FhirContext theCtx, IValidationSupport theValidationSup
...
-	public CodeSystem fetchCodeSystem(String theSystem) {
+	public CodeSystem fetchCodeSystem(String theSystem, IWorkerContext.VersionResolutionRules rules) {
 		if (myValidationSupport == null) {
 			return null;
 		} else {
 			return (CodeSystem) myValidationSupport.fetchCodeSystem(theSystem);
 		}
 	}
 
+	public CodeSystem fetchCodeSystem(
```
jdar8, line 81:
```suggestion

	@Override
	public CodeSystem fetchCodeSystem(
```

hapi-fhir-structures-r5/src/main/java/org/hl7/fhir/r5/hapi/ctx/HapiWorkerContext.java
```diff
@@ -359,9 +368,9 @@ public <T extends Resource> T fetchResource(Class<T> class_, String uri, FhirPub
 	}
```
jdar8: nitpick: this looks like an unused method  that doesn't override anything now

nsabale7:
Hi @tadgh when can we expect the next release with this change?

PR #7725, opened 2026-04-01
Add missing test scope for jena-shex test dependencies in hapi-fhir-structures-r4

Rene2000k:
jena-shex is used as a test dependency, but the `<scope>test</scope>` is missing.

Advantage of adding the test scope:  jena-shex and all it's transitive dependencies are not packaged with hapi-fhir-resources-r4 anymore.


robogary:
Formatting check succeeded!

PR #7731, opened 2026-04-02
GL-8591: Fix cross-version package dependency resolution

fil512:
## Important Note

I have confirmed that U.S. Core 8.0.1 loads without errors with this fix.

## Summary

When an R4 server installs an Implementation Guide with `fetchDependencies=true`, transitive dependencies that declare a FHIR version incompatible with the server (for example, `hl7.fhir.uv.extensions` which declares FHIR 5.0.0) would immediately fail with HAPI-1288 even though a version-specific counterpart (e.g., `hl7.fhir.uv.extensions.r4`) exists on the package registry. This PR adds automatic cross-version package substitution to `PackageInstallerSvcImpl` so the installer recovers gracefully by trying the versioned variant before raising a compatibility error.

1. Added `substituteVersionSpecificPackageIfNeeded()` — called after each transitive dependency is loaded; if the package's FHIR version is incompatible with the server, the method appends a version suffix (`.r4`, `.r5`, `.r3`) to the package ID and attempts to load the variant from the cache/registry.
2. Extracted `areFhirVersionsCompatible()` as a shared private static method, removing duplicate version-comparison logic from `assertFhirVersionsAreCompatible()`.
3. Added `getVersionSpecificSuffix()` mapping R4/R4B → `.r4`, R5 → `.r5`, DSTU3 → `.r3`; unknown/unsupported versions return `null` and skip substitution.
4. Added a `CrossVersionDependencyTest` nested test class with four scenarios: successful substitution, no variant available (expected HAPI-1288 failure), excluded dependency (verify `loadPackage` is never called), and R4/R4B compatibility regression guard.
5. Added `hapi-fhir-docs/server_jpa/packages.md` documenting package upload, transitive dependency fetching, cross-version substitution behavior, and validation usage.

## Code Review Suggestions

- [ ] **Fallback behavior after failed variant lookup** — `substituteVersionSpecificPackageIfNeeded` catches `IOException` and returns the original (incompatible) package with only a warning log. The caller (`fetchAndInstallDependencies`) will subsequently call `assertFhirVersionsAreCompatible`, which will throw HAPI-1288. Confirm this is the intended user experience: a warning is logged, then the error fires a few lines later, rather than failing immediately with a clearer message in the substitution method itself.
- [ ] **`NpmPackage.fhirVersion()` contract** — The substitution logic calls `theDependency.fhirVersion()` to detect incompatibility. Verify that `fhirVersion()` returns a value that `FhirVersionEnum.forVersionString()` can parse for the known cross-version packages (e.g., `hl7.fhir.uv.extensions` which may return `"5.0.0"` rather than `"R5"`). The test mocks use `FhirVersionEnum.R5.getFhirVersionString()` — confirm the real package metadata uses the same format.
- [ ] **R4B suffix mapping** — R4B is mapped to `.r4` (same as R4). Check whether there are real packages with an `.r4b` suffix that should be preferred for R4B servers, or whether `.r4` is intentionally correct for R4B.
- [ ] **`areFhirVersionsCompatible` null-return semantics** — When either version string cannot be resolved to a `FhirVersionEnum`, the method returns `false`. This means `assertFhirVersionsAreCompatible` will throw HAPI-1288 on an unrecognized version string, whereas the previous code threw an `IllegalArgumentException` via `Validate.notNull`. Confirm this change in error type is acceptable to callers.
- [ ] **`files.properties` entry** — Confirm the new `packages.md` doc page is registered correctly in `files.properties` and that the cross-links to `instance_validator.html` resolve in the built docs site.

### Setup
- [ ] Stand up a HAPI FHIR JPA server configured for **R4** (FhirContext R4).
- [ ] Ensure the server has outbound access to `packages.fhir.org`, or pre-load the relevant packages into the local package cache.

### Test Cases
- [ ] **Happy path — cross-version substitution succeeds**: Install `hl7.fhir.us.core` (latest R4 version) with `fetchDependencies=true`. Confirm the installer logs a warning about `hl7.fhir.uv.extensions` FHIR version mismatch and then logs a success message for `hl7.fhir.uv.extensions.r4`. Confirm the install completes without HAPI-1288.
- [ ] **No variant available**: Install a test package whose dependency declares FHIR 5.0.0 but has no `.r4` variant on the registry. Confirm the installer logs the "Could not find version-specific variant" warning and then fails with HAPI-1288.
- [ ] **Excluded cross-version dependency**: Install the same IG with `addDependencyExclude("hl7\\.fhir\\.uv\\.extensions")`. Confirm the install completes, the excluded package is never fetched, and no HAPI-1288 error is raised.
- [ ] **R4B server**: Configure the server for R4B and install an IG with a cross-version dependency. Confirm the installer attempts the `.r4` variant (not `.r4b`) and succeeds.
- [ ] **R5 server**: Configure the server for R5 and install an IG that depends on a cross-version package with a `.r5` variant. Confirm the `.r5` variant is substituted and installed successfully.
- [ ] **Regression — R4/R4B compatibility unchanged**: Install an R4B package on an R4 server (or vice versa) without cross-version packages in the dependency tree. Confirm there is no HAPI-1288 error and no substitution warning logged.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
@@ -62,7 +62,6 @@ import ca.uhn.hapi.converters.canonical.VersionCanonicalizer;
 import com.google.common.annotations.VisibleForTesting;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.PostConstruct;
-import org.apache.commons.lang3.Validate;
 import org.hl7.fhir.instance.model.api.IBase;
 import org.hl7.fhir.instance.model.api.IBaseResource;
 import org.hl7.fhir.instance.model.api.IIdType;
@@ -403,19 +402,12 @@ public class PackageInstallerSvcImpl implements IPackageInstallerSvc {
 		String dependencyFhirVersion = theDependency.fhirVersion();
 		String serverFhirVersion = myFhirContext.getVersion().getVersion().getFhirVersionString();
 
-		FhirVersionEnum dependencyVersionEnum = FhirVersionEnum.forVersionString(dependencyFhirVersion);
-		FhirVersionEnum serverVersionEnum = FhirVersionEnum.forVersionString(serverFhirVersion);
-
-		if (dependencyVersionEnum == null || serverVersionEnum == null) {
+		if (areFhirVersionsCompatible(dependencyFhirVersion, serverFhirVersion)) {
 			return theDependency;
 		}
 
-		boolean versionsAreCompatible = dependencyVersionEnum.equals(serverVersionEnum);
-		if (!versionsAreCompatible && dependencyFhirVersion.startsWith("R4") && serverFhirVersion.startsWith("R4")) {
-			versionsAreCompatible = true;
-		}
-
-		if (versionsAreCompatible) {
+		FhirVersionEnum serverVersionEnum = FhirVersionEnum.forVersionString(serverFhirVersion);
+		if (serverVersionEnum == null) {
 			return theDependency;
 		}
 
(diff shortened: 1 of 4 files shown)


robogary:
Formatting check succeeded!

volodymyr-korzh (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -363,6 +362,11 @@ private void fetchAndInstallDependencies(
 
 						// resolve in local cache or on packages.fhir.org
 						NpmPackage dependency = myPackageCacheManager.loadPackage(id, ver);
+
+						// If the dependency's FHIR version is incompatible with the server,
+						// attempt to load a version-specific variant (e.g., {id}.r4 for R4 servers)
```
volodymyr-korzh, line 367: FYI: Change-log is missing for this fix

fil512: The changelog already exists at `hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8591-cross-version-package-resolution.yaml` — it was added in a separate commit.

PR #7733, opened 2026-04-05
GL-8588: Handle unqualified references in cross-partition search

fil512:
## Summary

This fix resolves a crash in `ResourceLinkPredicateBuilder.calculatePotentialCrossPartitionsForPredicateTarget` when a FHIR search uses an unqualified reference (a bare resource ID without a resource type prefix, e.g. `?subject=123` instead of `?subject=Patient/123`) in a system with cross-partition references enabled. Previously, the method unconditionally required a fully-qualified `[resourceType]/[id]` reference and threw HAPI-2870 for anything less. The fix introduces a three-branch strategy — no-ID falls back to all partitions, unqualified-ID falls back to the search parameter's declared target types, and qualified-ID continues with the exact-partition lookup — eliminating the hard error without sacrificing partition-narrowing when information is available.

1. **Unqualified reference handling**: When only a bare ID is supplied (no resource type prefix), the fix iterates the search parameter's declared `targets` list, calls `determineReadPartitionForRequestForSearchType` for each supported target type, and merges the resulting partition IDs. This mirrors the approach already used in the chained-reference branch of the same method.
2. **Graceful degradation (query scoping only, not authorization)**: If the search parameter declares no target types, or if partition resolution fails, the method falls back to `RequestPartitionId.allPartitions()` rather than throwing. This variable is a **query-narrowing hint** used to scope the SQL JOIN against `HFJ_RES_LINK` — it is **not** an authorization decision and does not widen the requester's visibility. The user's partition authorization is still enforced by the enclosing request's partition context (set upstream by `IRequestPartitionHelperSvc` / partition interceptors). The sibling chained-reference branch already uses this same `allPartitions()` fallback for the no-targets case; this PR makes the unqualified-direct branch consistent. The practical cost is a broader partition scan, not broader data access.
3. **Three regression tests**: Tests cover the unqualified-reference cross-partition path (the bug), the qualified-reference cross-partition path (regression guard), and the non-partitioned path (confirms the fix has no effect outside cross-partition mode).

## Code Review Suggestions

- [ ] **Exception swallowing in the fallback loop**: Caught exceptions during `determineReadPartitionForRequestForSearchType` are logged at `debug` level and silently skipped. If every target type fails resolution, `predicateTargetPartitionId` may still be `null` at the end of the loop (the `allPartitions()` fallback only triggers when `theParam.getTargets().isEmpty()`). Reviewers should confirm whether a `null` result here is safe downstream, or whether the no-targets fallback should also be reached when all per-type resolutions throw.
- [ ] **Log level in the catch block**: The committed code logs at `ourLog.debug(...)` with the exception included. The diff comment says `ourLog.trace(...)` without the exception — confirm the intended verbosity level is consistent with how similar catch blocks are logged elsewhere in this class.
- [ ] **`determineReadPartitionForRequestForSearchType` semantics**: This method is used in the chained-reference branch but is now also called for the direct-reference unqualified case. Reviewers familiar with partition helper semantics should verify that `ForSearchType` (as opposed to `ForRead`) is the correct call here — specifically, that it doesn't introduce different partition visibility rules that could yield false positives or missed results.
- [ ] **Multiple target types producing a merged partition set**: When a search parameter targets several resource types (e.g. `subject` targets both `Patient` and `Group`), all their partitions are merged. In an environment where different resource types are pinned to different named partitions, this could cause a wider-than-necessary partition scan. This is intentional for correctness but worth calling out for performance-sensitive deployments.

### Setup

- Configure a Smile CDR instance with partitioning enabled and cross-partition references allowed (`AllowReferencesAcrossPartitions = ALLOWED_UNQUALIFIED` or equivalent).
- Create at least two named partitions (e.g. `partition-a` and `partition-b`).
- Ingest an HL7v2 ADT message (or create resources directly via FHIR API) so that a `Patient` lands in `partition-a` and a linked `Observation` (or `DiagnosticReport`) lands in `partition-b`.

### Test Cases

- [ ] **Unqualified reference search succeeds**: Perform a FHIR search using `?subject=<bare-patient-id>` (no `Patient/` prefix) from a request scoped to `partition-b`. Confirm the search returns the expected Observation and does not throw a 400/500 with HAPI-2870.
- [ ] **Qualified reference search still works**: Repeat the same search with `?subject=Patient/<id>`. Confirm the result is identical to the unqualified case and no regression is introduced.
- [ ] **HL7v2 ingest path (original bug scenario)**: Re-ingest an HL7v2 message that generates a reference lookup with a bare ID (as in the GL-8588 bug report). Confirm the message is processed without error and the resulting FHIR resource is stored correctly.
- [ ] **Non-partitioned system unaffected**: In a Smile CDR instance with partitioning disabled, perform a search with an unqualified reference. Confirm there is no change in behaviour and no error.
- [ ] **Search parameter with no declared targets**: If there is a custom search parameter with an empty target list and cross-partition mode is enabled, confirm that a search using an unqualified reference returns results from all partitions rather than throwing.


robogary:
Formatting check succeeded!

fil512:
Pre-merge review notes (from a second-pass quality review):

Two items to look at before merge:

**1. Broad `catch (Exception e)` at debug-log level**

The catch swallows any failure from `determineReadPartitionForRequestForSearchType` across *all* target types. If every target throws (e.g. a misconfigured partition interceptor), `predicateTargetPartitionId` stays `null` — and the existing `allPartitions()` safety net only fires on the *empty-targets* branch, not on this failure path. Worth either tightening the catch to expected exception types, or adding an explicit fallback to `allPartitions()` when no target resolved successfully.

**2. Weakened assertion in `PartitioningAllowedUnqualifiedR4Test.testSearch_UnqualifiedTarget`**

This test previously asserted `assertThatThrownBy(...).hasMessageContaining("must be in the format")` and was changed to `assertThat(results).isEmpty()`. The change is legitimate — the new behavior *is* "return empty" instead of "throw" — but reviewers should consciously accept that the contract for unqualified refs on unpartitioned deployments has shifted from "loud failure" to "silent empty." A brief comment on the assertion (or in the PR description) confirming this was intentional would help future archeologists.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
```diff
@@ -359,15 +358,40 @@ private RequestPartitionId calculatePotentialCrossPartitionsForPredicateTarget(
...
-				if (isBlank(resourceType) || isBlank(resourceId)) {
-					throw new InvalidRequestException(Msg.code(2870) + "Parameter \""
-							+ UrlUtil.sanitizeUrlPart(paramName) + "\" must be in the format [resourceType]/[id]");
-				}
 
-				IIdType id = myFhirContext.getVersion().newIdType(resourceType, resourceId);
-				RequestPartitionId nextPartitionId =
-						myRequestPartitionHelperSvc.determineReadPartitionForRequestForRead(theRequest, id);
-				predicateTargetPartitionId = nextPartitionId.mergeIds(predicateTargetPartitionId);
+				if (isBlank(resourceId)) {
```
codeforgreen, line 362: **question:** There is a difference in behaviour. Now we silently skip instead of throwing an exception. Can this condition be true? We're the ones computing the parameters of this method including the parameter list, so I presume. this is a no-op.

hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/jpa/dao/r5/CrossPartitionReferencesTest.java
codeforgreen, line 1: **thought:** The tests are in cross partition mode but the actual resources are not in different partitions. Let's say we refactor the indexing logic and eventually the condition that checks if cross partition is enabled is not the one that determines the fork in logic. Then the tests would need to be updated as well. I guess this is ok because supposably all tests will be reworked with that refactoring exercise.

codeforgreen (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
```diff
@@ -359,15 +358,40 @@ private RequestPartitionId calculatePotentialCrossPartitionsForPredicateTarget(
...
+					/*
+					 * If no ID part is present at all, we can't narrow the partition
+					 * by a specific resource. Just skip this parameter value — the
+					 * downstream search will handle it (and likely match nothing).
+					 */
+					continue;
+				} else if (isBlank(resourceType)) {
```
codeforgreen, line 369: **suggestion:** I would consider extracting this entire method onto a separate class entirely and splitting this method as it's becoming too long. 
**nitpick:** I also wonder if there is any part of this logic that can be found elsewhere, although I assume not given how the code is.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
fil512: Good question. In practice this branch is unlikely to be hit because callers normally pass parsed ReferenceParam instances that have either a resource type, an ID, or both. However, ReferenceParam does not enforce non-blank idPart at the type level (it accepts arbitrary user-supplied search values), so a defensive guard here keeps us robust to malformed input — e.g. an empty value coming through chained processing — without throwing. The previous code path didn't reach this point with a blank id either, so behaviour for well-formed callers is unchanged; we just avoid an NPE/IAE on the malformed-input edge. Happy to convert it to an assertion if you'd prefer to make the invariant explicit.

hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/jpa/dao/r5/CrossPartitionReferencesTest.java
fil512: Agreed — the test fixture exercises the cross-partition code path via configuration (cross-partition reference mode enabled) rather than by physically placing resources in distinct partitions, because the bug being reproduced is in the partition-determination logic itself, not in cross-partition storage. You're right that if a future refactor decoupled the cross-partition flag from the fork in logic, these tests would need to be revisited as part of that exercise. I'll leave it as-is for this targeted fix.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
fil512: Fair point on length. I'd prefer to keep the extraction/split out of scope for this bugfix MR to avoid expanding the diff and risk surface — the change here is intentionally minimal (the unqualified-reference branch only). Happy to file a follow-up for a focused refactor of calculatePotentialCrossPartitionsForPredicateTarget if you'd like. On the nitpick: I looked for an existing partition-resolution helper that could be reused and didn't find one that handles both the qualified and target-types fallback cases, so I agree there's nothing obvious to delegate to today.

PR #7750, opened 2026-04-09
Support Implied Profiles in RepositoryValidatingRuleBuilder

### LINKED ISSUE #6028: Support Implied Profiles in RepositoryValidatingRuleBuilder
The Core libraries now support the concept of "implied profiles", which indicates that a resource should undergo validation against this "implied profile" whether it is present in the incoming resource's profile section or not. This request is to add the capability to our RepositoryValidatingRuleBuilder to provide the ability to make use of this new feature in the core libraries. 

A potential API could look like: 
```
 	ruleBuilder
				.forResourcesOfType("Patient")
                                .impliedProfile("http://www.hl7.org/fhir/us/core/StructureDefinition-us-core-patient.html")
				.requireAtLeastProfile("http://www.hl7.org/fhir/us/core/StructureDefinition-us-core-patient.html");
```

Which would cause any incoming Patients to be validated against US Core, whether or not they indicate that they support the profile in their meta.profile section.

jorgepinilla-smile:
## Description

This PR implements support for implied profile validation in the RepositoryValidatingRuleBuilder, addressing issue #6028.

### Core Implementation
- **New Rule Class**: Added `RuleImpliedProfileValidation` with two validation modes:
  - `IF_NOT_EXPLICIT`: Apply validation only when no profile is declared
  - `ALWAYS`: Apply validation regardless of declared profiles

### API Extensions
- **impliedProfileIfNotExplicit(String theProfile)**: Validates against specified profile when `Resource.meta.profile` is empty
- **impliedProfileAlways(String theProfile)**: Validates against specified profile regardless of existing profile declarations

### Example Usage
```java
ruleBuilder
   .forResourcesOfType("Patient")
   .impliedProfileIfNotExplicit("http://hl7.org/fhir/us/core/StructureDefinition/us-core-patient")
   .impliedProfileAlways("http://hl7.org/fhir/us/core/StructureDefinition/someOtherProfile");
```

## Documentation

- Updated `repository_validating_interceptor.md` with new "Implied Profile Validation" section
- Added code examples in `RepositoryValidatingInterceptorExamples.java`
- Documented use cases and behavior differences between modes

## Validation Logic

The implementation:
1. Creates a copy of the resource to avoid modifying the original
2. For `ALWAYS` mode: validates against the implied profile as well as the profiles specified in the resource meta 
3. For `IF_NOT_EXPLICIT` mode: adds implied profile only if no profiles are declared
4. Uses existing FHIR validator infrastructure for consistent validation behavior
5. Properly handles validation results and OperationOutcome generation

## Backwards Compatibility

This is a purely additive change with no breaking modifications to existing APIs or behavior.

Fixes #6028


robogary:
Formatting check succeeded!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
```diff
@@ -0,0 +1,144 @@
...
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+import org.hl7.fhir.common.hapi.validation.validator.FhirInstanceValidator;
+import org.hl7.fhir.instance.model.api.IBaseResource;
+import org.hl7.fhir.instance.model.api.IPrimitiveType;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+class RuleImpliedProfileValidation extends BaseTypedRule {
```
codeforgreen: Missing javadoc

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RepositoryValidatingRuleBuilder.java
```diff
@@ -145,6 +145,51 @@ public FinalizedTypedRule requireAtLeastOneProfileOf(String... theProfileUrls) {
...
 		}
 
+		/**
+		 * Apply validation against the specified profile when no profile is explicitly declared in the incoming resource.
+		 * This method should cause validation against the specific profile to be applied to the resource type,
+		 * when no profile is explicitly declared in the incoming resource.
+		 *
+		 * @param theProfile The profile canonical URL to validate against
+		 */
+		public FinalizedTypedRule impliedProfileIfNotExplicit(String theProfile) {
```
codeforgreen, line 155: It seems we are allowed to add both rules and they should be exclusive. I think we should throw an exception if there is already a rule defined from this set.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/6529-feature-implied-profiles-validation.yaml
```diff
@@ -0,0 +1,5 @@
+---
+type: add
+issue: 6529
+jira: SMILE-9966
+title: "Added implied validation profile rules in order to enforce specific implementation guides so that resources are validated when no profiles are specified under meta."
```
codeforgreen: link to documentation would be helpful and more specific details as to what API was added.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
```diff
@@ -0,0 +1,144 @@
...
+			return RuleEvaluation.forSuccess(this);
+		} else if (!declaredProfiles.contains(myImpliedProfileUrl)) {
+			resourceCopy.getMeta().addProfile(myImpliedProfileUrl);
+		}
+
+		FhirValidator validator = getFhirContext().newValidator();
+		validator.setInterceptorBroadcaster(
+				CompositeInterceptorBroadcaster.newCompositeBroadcaster(myInterceptorBroadcaster, theRequestDetails));
+		validator.registerValidatorModule(myValidator);
+		ValidationResult outcome = validator.validateWithResult(resourceCopy);
```
codeforgreen: This is an expensive operation. I believe we perform the actual resource validation elsewhere. `RepositoryValidatingRuleBuilder` is purely configuration for validation. The only exception to that is .`requireValidationToDeclaredProfiles` which actually performs validation. Looking at the names of these new additions they do suggest configuration only. I need to look into this more to see if it makes sense.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
codeforgreen: nitpick: I know currently we have both flavors. I think Rule (noun) should be at the end of the name rather than the beginning.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
codeforgreen: It looks like `RepositoryValidatingInterceptor` is the one we are targeting with this feature and for this case, we call evaluate on all rules. 

Concerns with this approach:
- there seem to be validation rules which can co-exist and those which should not. This is not being checked for currently.
- how can we avoid calling validate multiple times? e.g. `requireValidationToDeclaredProfiles` is also part of the rules, or we have multiple rules of the same type which invoke validate.
- does it make sense to separate configuration from the actual validation? now that we have 3 rules which can call validate, I believe the current design warrants a change.
#6028 references a field called `supportedProfiles` which could potentially be used to store the actual profiles which are applied. Reading through the documentation it is one way to we could go which could help us achieve the separation configuration/validation. There could be other solutions as well.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/6529-feature-implied-profiles-validation.yaml
tadgh:
```suggestion
title: "Added  new rules for implicit profile validation in `RepositoryValidatingRuleBuilder`. The methods `impliedProfileIfNotExplicit()` and `impliedProfileAlways()` can now be used to force validation against a given profile regardless of the profile declaration in the resource. More information can be found in [the documentation](//insert-link-to-docs-here)
```

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/validation/repository_validating_interceptor.md
```diff
@@ -120,6 +120,35 @@ Rules can declare that a specific profile is not allowed.
...
+Use the following rule to apply validation against a specific profile regardless of what profiles are declared in the incoming resource:
+
+```java
+{{snippet:classpath:/ca/uhn/hapi/fhir/docs/RepositoryValidatingInterceptorExamples.java|impliedProfileAlways}}
+```
+
+This rule will:
+- Always apply validation against the specified profile
+- Apply validation even if the resource declares conformance to different profiles
+- Effectively enforce a mandatory Implementation Guide across all resources of the specified type
```
tadgh: issue: I believe this is not a mandatory IG, but a mandatory Profile (as far as verbiage goes)

tadgh: question: If profile A is declared on the Patient, and Profile B is declared in the `implied profile always` rule, do _both_ profiles get validated? They should.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
```diff
@@ -0,0 +1,144 @@
+/*-
```
tadgh: thought: I question the need for a full rule here. It appears to me as though this is effectively a special case of RequireValidationRule. Is it not possible to re-use its functionality for this behaviour? I hesitate to put this in the tree as it is a 50% copy-paste of RequireValidationRule, with _less_ configurability (e.g. RequireValidationRule allows you to set myTagOnSeverity, which is an important customization for customers) 


Is it at all possible to just use the RequireValidationRule with some pre-processing here?

tadgh (changes requested):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
jorgepinilla-smile: of course, I can look into refactoring this to make it a care for the require validation rule.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/6529-feature-implied-profiles-validation.yaml
jorgepinilla-smile: Great suggestion! I've updated the change log's message.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/validation/repository_validating_interceptor.md
jorgepinilla-smile: Updated to wording to use profile instead of implementation guide.

jorgepinilla-smile: I see what you mean. I've updated the documentation as well as made changes in the logic and unit tests to reflect this. Validation will be applied to both the implied profile as well as the declared profiles in the "Always" scenario.

codeforgreen (commented):
Sorry I seem to have forgotten to click submit on my review comments. I appologize.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
codeforgreen: Looks like this is gone which is good, and we're reusing the existing rule which calls validate, that is a way to solve it without making many changes.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
jorgepinilla-smile: I've removed this file.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/6529-feature-implied-profiles-validation.yaml
jorgepinilla-smile: Link to documentation has been added.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
jorgepinilla-smile: correct, I've refactored the code to be a special case for require validation rule

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RuleImpliedProfileValidation.java
jorgepinilla-smile: File no longer exists

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RepositoryValidatingRuleBuilder.java
jorgepinilla-smile: I believe the idea is to provide flexibility to allow user to specify a profile in case the meta profile is empty and allow user to have a profile specified for a particular resource to always be validated against.

For example, we could set up the always implied profile to be validated on top of the profiles that the resource meta has. If the meta happens to not have any resources, check to see if there are any implied when profile meta is empty and validate against that as well.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RequireValidationRule.java
```diff
@@ -46,26 +47,55 @@
...
 	private final FhirInstanceValidator myValidator;
 	private final IInterceptorBroadcaster myInterceptorBroadcaster;
+	private final String myImpliedProfileUrl;
+	private final ImpliedProfileMode myImpliedProfileMode;
 	private ResultSeverityEnum myRejectOnSeverity = ResultSeverityEnum.ERROR;
 	private List<TagOnSeverity> myTagOnSeverity = Collections.emptyList();
 
+	public enum ImpliedProfileMode {
+		IF_NOT_EXPLICIT,
+		ALWAYS
```
tadgh, line 72: note: add docs to these enum values

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RequireValidationRule.java
```diff
@@ -74,11 +104,18 @@ void setBestPracticeWarningLevel(BestPracticeWarningLevel theBestPracticeWarning
 	@Override
 	public RuleEvaluation evaluate(RequestDetails theRequestDetails, @Nonnull IBaseResource theResource) {
 
+		IBaseResource resourceToValidate = theResource;
```
tadgh, line 122: question: why this assignment? does it do anything? its still just an object reference, you aren't gaining anything by reassigning.

tadgh: Ah, I see, you're silently replacing it with a copy down below so you don't mutate the original `theResource`. Fair enough

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RequireValidationRule.java
```diff
@@ -104,6 +141,33 @@ public RuleEvaluation evaluate(RequestDetails theRequestDetails, @Nonnull IBaseR
 		return RuleEvaluation.forSuccess(this);
 	}
 
+	private IBaseResource handleImpliedProfileValidation(IBaseResource theResource) {
```
tadgh, line 161: issue: leave an overarching comment on this method that indicates that to perform implied profile validation, we clone the resource first and apply the profiles to it, and explain that the profiles are not actually added to the persisted resource.

tadgh (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RequireValidationRule.java
jorgepinilla-smile: Added docs to enum values

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/validation/RequireValidationRule.java
jorgepinilla-smile: Good suggesting, I've added a short comment that indicates such behaviour.

PR #7752, opened 2026-04-10
Jr 20260407 recurse dependencies

JasonRoberts-smile:
This MR builds on https://github.com/hapifhir/hapi-fhir/pull/7736 to add support for dependency processing to the async package loader job. 

Key features:

- new `InitializeDependenciesStep` determines whether the package has any dependencies to process, and if so, launches a child job for each of them
- new `ConsolidateDependenciesStep` waits for all the child jobs launched in the previous step to complete, then merges the `PackageInstallOutcomeJson` results from each child job into the parent outcome object
- `ProcessPackageStep` and `FinalizeInstallationStep` have been collapsed into a single step, since the latter turned out to be providing little value


robogary:
Formatting check succeeded!

volodymyr-korzh (approved):

PR #7765, opened 2026-04-13
Handling long running tasks

### LINKED ISSUE #7670: fix issue with batch2 to handle slow arriving chunks
Slow arriving chunks sometimes show up during a maintenance run and this results in jobs stalling indefinitely.

TipzCM:
closes #7670


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -129,6 +129,7 @@ public String onWorkChunkCreate(WorkChunkCreateEvent theBatchWorkChunk) {
 		entity.setCreateTime(new Date());
 		entity.setStartTime(new Date());
 		entity.setStatus(getOnCreateStatus(theBatchWorkChunk));
+		entity.setLastHeartbeat(new Date());
```
michaelabuckley: suggestion: if heartbeat is to mark processing work, then this should probably be back-dated, or some beginning-of-time sentinel.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/entity/Batch2WorkChunkEntity.java
```diff
@@ -149,6 +149,16 @@ public class Batch2WorkChunkEntity implements Serializable {
 	@Column(name = "POLL_ATTEMPTS", nullable = true)
 	private Integer myPollAttempts;
 
+	/**
+	 * Records the last time that this chunk registered work being done.
+	 * By default, this is creation time.
+	 * But upon processing, this value will be updated by workers to
+	 * let other workers know the chunk is still being processed.
+	 */
+	@Column(name = "LAST_HEARTBEAT", nullable = true)
```
michaelabuckley, line 158: question: is this ever null, since you initialize it on creation?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/broker/api/ChannelConsumerSettings.java
```diff
@@ -23,6 +23,11 @@ public class ChannelConsumerSettings extends BaseChannelSettings {
 	public static final Integer DEFAULT_CHANNEL_CONSUMERS = 2;
 
 	private Integer myConcurrentConsumers = DEFAULT_CHANNEL_CONSUMERS;
+	/**
+	 * This is the timeout setting of the broker.
+	 * It is the time after which message redelivery happens.
+	 */
+	private long myAckTimeoutMs;
```
michaelabuckley: minor: suggest using Duration and you can drop the Ms units.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/broker/api/ChannelConsumerSettings.java
```diff
@@ -39,4 +44,12 @@ public ChannelConsumerSettings setConcurrentConsumers(int theConcurrentConsumers
 		myConcurrentConsumers = theConcurrentConsumers;
 		return this;
 	}
+
+	public long getAckTimeoutMs() {
+		return myAckTimeoutMs;
+	}
+
+	public void setAckTimeoutMs(long theAckTimeoutMs) {
```
michaelabuckley: question: I can't figure out how this is set.  All I see are copy constructors.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/channel/subscription/SubscriptionChannelFactory.java
```diff
@@ -86,6 +86,7 @@ protected ChannelConsumerSettings newConsumerConfigForDeliveryChannel(ChannelCon
 		config.setConcurrentConsumers(getDeliveryChannelConcurrentConsumers());
 		if (theOptions != null) {
 			config.setRetryConfiguration(theOptions.getRetryConfigurationParameters());
+			config.setAckTimeoutMs(theOptions.getAckTimeoutMs());
 		}
 		return config;
```
michaelabuckley, line 86: minor: should this be a copy constructor, or factory method on ChannelConsumerSettings?
question: is this missing setQualifyChannelName on purpose?  This seems copy-pasted in several methods here.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/broker/jms/SpringMessagingReceiverAdapter.java
```diff
@@ -44,13 +44,17 @@ public class SpringMessagingReceiverAdapter<T> implements IChannelConsumer<T> {
 	private MessageHandler myMessageHandler;
 	private boolean myClosed;
 
+	protected final long myRedeliveryDelayMs;
```
michaelabuckley: minor: suggest Duration
question: why is this called "redeliveryDelay" but "ackTimeout" elsewhere?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/DelayChunkException.java
```diff
@@ -0,0 +1,7 @@
+package ca.uhn.fhir.batch2.api;
+
+public class DelayChunkException extends RuntimeException {
```
michaelabuckley, line 7: problem: I thought this was dead?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobDataSink.java
```diff
@@ -67,4 +67,14 @@ default void accept(OT theData) {
...
+
+	/**
+	 * A keep-alive method.
+	 * Useful for long-running processes.
+	 * Calling accept() will automatically call this method.
+	 *
+	 * But if no new workchunks are being created during the long-running
+	 * process, this method is exposed for use at the job-creators discretion.
+	 */
+	void heartbeat();
```
michaelabuckley: question: I thought this was dead?  Isn't the heartbeat an automatic timer?  Or must processors call this if not generating chunks?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
```diff
@@ -142,4 +171,35 @@ private void handleFastTracking(BaseDataSink<PT, IT, OT> theDataSink) {
...
+			String workchunkId = (String) context.getMergedJobDataMap().get(CHUNK_ID);
+
+			WorkChunkStatusEnum status = myWorkChunkPersistence.getWorkChunkStatus(workchunkId);
+			ourLog.info("WorkChunk {} has status {}", workchunkId, status.name());
+
+			if (status == WorkChunkStatusEnum.IN_PROGRESS) {
+				myWorkChunkPersistence.onWorkChunkHeartbeat(workchunkId);
+			} else {
+				/*
+				 * If it's no longer in progress, it's either going to be
```
michaelabuckley: question: is this conditional needed? I wouldn't expect any logic in the heartbeat system.  We shouldn't be starting the job if the chunk is not being processed.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/channel/BatchJobSender.java
```diff
@@ -22,23 +22,37 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.function.Function;
+
 public class BatchJobSender {
 	private static final Logger ourLog = LoggerFactory.getLogger(BatchJobSender.class);
 	private final IChannelProducer<JobWorkNotification> myWorkChannelProducer;
 
+	private Function<JobWorkNotification, JobWorkNotificationJsonMessage> myMessageCreationFn;
```
michaelabuckley: suggestion: add a comment explaining this is a field to allow partition injection in tests.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/channel/BatchJobSender.java
```diff
@@ -22,23 +22,37 @@
...
 	}
+
+	@VisibleForTesting
+	public void setMessageCreationFn(
+			Function<JobWorkNotification, JobWorkNotificationJsonMessage> theMessageCreationFn) {
+		myMessageCreationFn = theMessageCreationFn;
+		if (myMessageCreationFn == null) {
+			myMessageCreationFn = JobWorkNotificationJsonMessage::new;
+		}
+	}
```
michaelabuckley, line 73: suggestion: simpler to Validate.notNull.  Silently fixing nulls is bad.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/config/BaseBatch2Config.java
```diff
@@ -152,14 +155,27 @@ public WorkChannelMessageListener workChannelMessageListener(
...
-				myInterceptorService);
+				myInterceptorService,
+				theSchedulerSvc);
+	}
+
+	/**
+	 * Spring ensures SmartInitializingSingleton is created/run after
+	 * all non-lazy beans in this class (and inheriters) are instantiated.
+	 * This allows us to set the property, even if this class is overriden.
+	 */
```
michaelabuckley, line 171: suggestion: explain that we pull this value from Kafak, and that can't happen until after we have a connection to the broker.

spelling: overriden

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/FinalStepDataSink.java
```diff
@@ -49,6 +51,11 @@ public void accept(WorkChunkData<VoidModel> theData) {
 		throw new JobExecutionFailedException(Msg.code(2045) + msg);
 	}
 
+	@Override
+	public void heartbeat() {
+		// TODO - nothing - but we could... should we?
```
michaelabuckley: question: what is going on here?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
```diff
@@ -71,18 +88,30 @@ public class JobStepExecutor<PT extends IModelJson, IT extends IModelJson, OT ex
...
+	@WithSpan(JOB_STEP_EXECUTION_SPAN_NAME)
+	public void processstep() {
 		BatchJobOpenTelemetryUtils.addAttributesToCurrentSpan(
 				myInstance.getJobDefinitionId(),
 				myInstance.getJobDefinitionVersion(),
 				myInstance.getInstanceId(),
 				myCursor.getCurrentStepId(),
 				myWorkChunk == null ? null : myWorkChunk.getId());
 
+		scheduleHeartbeat();
```
michaelabuckley: problem: where is this heartbeat canceled?  I'd feel safer with a try/finally block, or a try-with-resources pattern.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutorFactory.java
```diff
@@ -65,6 +86,8 @@ JobStepExecutor<PT, IT, OT> newJobStepExecutor(
 				myJobStepExecutorSvc,
 				myJobMaintenanceService,
 				myJobDefinitionRegistry,
-				myInterceptorService);
+				myInterceptorService,
+				myIHapiScheduler,
+				Math.min(myAckTimeoutMS, DEFAULT_ACK_TIMEOUT));
```
michaelabuckley: question: why Min?  If myAckTimeoutMS is -1, then what is going on?

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7670-handling-slow-batch-worker.yaml
```diff
@@ -0,0 +1,11 @@
+---
+type: add
+issue: 7670
+jira: SMILE-11603
+title: "Added infrastructure to Batch2 to allow
+      slow running workers to have more time to process
+      larger jobs.
+      The time utilized will still depend on broker level
```
michaelabuckley: suggtion: mention the heartbeat mechanism.  Hapi is our technical audience.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -179,7 +180,9 @@ public Optional<WorkChunk> onWorkChunkDequeue(String theChunkId) {
 			return Optional.empty();
 		} else {
 			Optional<Batch2WorkChunkEntity> chunk = myWorkChunkRepository.findById(theChunkId);
-			return chunk.map(this::toChunk);
+			WorkChunk chunkToReturn = chunk.map(this::toChunk).orElseThrow();
+			chunkToReturn.setPreviousStatus(chunkLock.getStatus());
+			return Optional.of(chunkToReturn);
 		}
```
michaelabuckley: problem: this changes semantics.  Why add the orElseThrow() but keep the optional?  The old semantics returned an empty if not present.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
```diff
@@ -71,18 +88,30 @@ public class JobStepExecutor<PT extends IModelJson, IT extends IModelJson, OT ex
...
+		ScheduledJobDefinition definition = new ScheduledJobDefinition();
+		definition.setJobClass(HeartbeatJob.class);
+		definition.setId(String.format("BATCH2-HEARTBEAT-%s-%s", myInstanceId, myWorkChunk.getId()));
+		definition.addJobData(CHUNK_ID, myWorkChunk.getId());
+		// we don't want a time that's <100ms (or worse, negative)
+		myIHapiScheduler.scheduleClusteredJob(Math.max(myAckTimeoutMs / 3, 500), definition);
+	}
 
+	@WithSpan(JOB_STEP_EXECUTION_SPAN_NAME)
+	public void processstep() {
```
michaelabuckley: problem: camel case

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
TipzCM: will use nullable as discussed (and remove this line)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/entity/Batch2WorkChunkEntity.java
TipzCM: i've removed the initialization

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
```diff
@@ -142,4 +171,35 @@ private void handleFastTracking(BaseDataSink<PT, IT, OT> theDataSink) {
...
+				 * In either case, this listener is no longer required to keep track
+				 * of it (and we'll stop the job/heartbeat)
+				 */
+				try {
+					context.getScheduler().unscheduleJob(context.getTrigger().getKey());
+				} catch (SchedulerException ex) {
+					ourLog.info("Failed to unschedule job {}", getClass().getName(), ex);
+					throw new RuntimeException(Msg.code(2912), ex);
+				}
+			}
```
michaelabuckley: question: I don't get why we don't do this in the same place we schedule the job.  I expect this job to be _much_ dumber.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -540,6 +548,14 @@ public Stream<WorkChunk> fetchAllWorkChunksForStepStream(String theInstanceId, S
 				.map(this::toChunk);
 	}
 
+	@Override
+	public WorkChunkStatusEnum getWorkChunkStatus(String theWorkChunkId) {
+		Batch2WorkChunkEntity chunkOp =
```
michaelabuckley: I am skeptical we need this.

michaelabuckley (changes requested):

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/ReductionStepDataSink.java
```diff
@@ -127,6 +127,13 @@ public void accept(WorkChunkData<OT> theData) {
 		}
 	}
 
+	@Override
+	public void heartbeat() {
+		throw new UnsupportedOperationException(
+				Msg.code(2914)
+						+ " Reduction steps are not associated with any one workchunk, but with a multitude of cunks being 'reduced' to a single one.");
```
michaelabuckley: spelling: cunks

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/JobWorkNotification.java
```diff
@@ -20,10 +20,11 @@
 package ca.uhn.fhir.batch2.model;
 
 import ca.uhn.fhir.model.api.IModelJson;
+import ca.uhn.fhir.rest.server.messaging.IHasPayloadMessageKey;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import jakarta.annotation.Nonnull;
 
-public class JobWorkNotification implements IModelJson {
+public class JobWorkNotification implements IHasPayloadMessageKey, IModelJson {
```
michaelabuckley: problem?: implements IHasPayloadMessageKey but not getPayloadMessageKey.  Why does getPayloadMessageKey have a default method?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/broker/jms/SpringMessagingReceiverAdapter.java
TipzCM: I tried to make the name correct for the actual broker.

It's called "acktimeout" lazily everywhere else.

But realistically it's different per broker:
* Kafka -> max.poll.interval (since kafka listeners poll the topic periodically, and this is the setting used for determining rebalance, not acknowledgement timeouts)
* pulsar -> acktimeout (since pulsar only cares about the delay between "delivery" and "acknowledgement" and has no such concept of polling)
* spring messaging receiver doesn't use either of these but just delays between redelivery attempts (spring doesn't even have this issue since it doesn't have partitions and doesn't care about acknowledgements or polling timeouts; but it's part of our stack all the same and i need to provde *something* so i'm using the retry delay, since that's the 'most obvious choice' imo)

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobDataSink.java
TipzCM: Originally i kept it because i thought it wouldn't "do any harm" to leave calling heartbeat on datasink.accept

but i'll remove it, because there may be race conditions or other things i'm not considering that could be problematic in the future. best to just leave it to the timer

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/channel/BatchJobSender.java
TipzCM: added

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/config/BaseBatch2Config.java
TipzCM: updated comment

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutorFactory.java
TipzCM: you're right; should be max (to ensure we have a > 0 value)

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
TipzCM: updated; simplified job and put the cancelling logic into a finally block

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
TipzCM: Put into a finally block to deschedule.

It's only used in this class, so i'm hoping a finally block is sufficient here.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/DelayChunkException.java
TipzCM: No, it's still used in WorkChannelMessageListener for flow control.

We discussed using RetryChunkLaterException (As it already existed), but that is used (and caught) elsewhere.

We're just using DelayChunkException in order to throw this back to the queue after we've waited a small amount of time

hapi-fhir-storage/src/main/java/ca/uhn/fhir/broker/api/ChannelConsumerSettings.java
TipzCM: Updated all to use Duration

hapi-fhir-storage/src/main/java/ca/uhn/fhir/broker/api/ChannelConsumerSettings.java
TipzCM: Kafka properties and Pulsar properties work different.

**Kafka**

Kafka's properties are injected under the property key `module.clustermgr.config.kafka.consumer.properties.text`.

_(aside: Technically a "file" could also be specified, but i'm just going to follow the string route for simplicity (but it's basically the same)_

This is a set of whitespace delimited kafka properties (similar, but not exactly, what kafka takes, since kafka uses some kinda yaml format, with hierarchy starting at "kafka" then consumers, and producers and security... etc..... here it's just the "consumer" set of things).

1. `KafkaBrokerConfig` creates `KafkaProperties` Bean, which is where this giant white-space delimited text blob exists.

2. `KafkaConsumerConfig` (which is imported by `KafkaBrokerConfig`) injects the `KafkaProperties` bean and uses them to instantiate a bunch of `ConsumerFactory` beans (one per message type we consume)

3. All ConsumerFactory bean methods will create a NullDroppingKafkaProperties object (which is just a Map) and calls `addCommonListenerProperties`, which takes the giant string of properties from KafkaProperties, parses them into a series of key-value pairs, and sets them on NullDroppingKafkaProperties map.

------

aside: this is partially why i prefer separate responsibilities for ModuleConfigs/Bean creation and Property Value injection. Here our `@Configuration` bean, which defines module properties and constructs beans, is *also* our "properties injection" bean.

So we have to construct multiple third objects (NullDroppingKafkaProperties) so we can read those properties in from our configuration and pass sets of them to all our various ConsumerFactory objects, instead of just passing in a single configuration bean directly because passing in an `@Configuration` object *as* a properties object is potentially exposing something to consumers we shouldn't expose and is very confusing (at least to me) to trace out the path; and that's assuming it's not going to result in circular dependencies

-----

4. It then creates the `DefaultKafkaConsumerFactory` with this map (ie, the injected kafka properties) which is our actual ConsumerFactory bean.

5. `KafkaConsumerFactoryRegistry` is lazily instantiated in `KafkaConsumerConfig` (same place). All these factories are injected and registered in `KafkaConsumerFactoryRegistry`

6. `IMessageBrokerSvc` is created in `ClusterMgrCtxConfig` using the `messagebroker.type` config to determine what it wants (Kafka/Pulsar/whatever)

7. In the case of Kafka we get a `KafkaMessageBrokerSvcImpl`. This object injects our `KafkaConsumerFactoryRegistry` via property injection.

8. CdrSubscriptionChannelConfig (which overrides SubscriptionChannelConfig) to instantiate the IBrokerClient (concrete class - `CdrBrokerClient`).

_aside: SubscriptionChannelConfig creates a LinkedBlockingChannelFactory, which doesn't use any of this stuff, because it lives solely in hapi_

9. CdrBrokerClient injects (via field injection, not constructor injection) the `IMessageBrokerSvc` that we created in step 7 above.

10. `BaseBatch2Config` instantiates the IChannelConsumer for JobWorkNotification. But this bean is overwritten by `CdrJpaBatch2Config`, which does this work instead (so i won't refer to what BaseBatch2Config does because it's not important here)

11. CdrJpaBatch2Config injects the `IBrokerClient` (and our WorkChannelMessageListener), creates a CdrChannelConsumerSettings object and sets current consumers to it (another properties `@value` injected into a Configuration object, so we have to create yet another ne wobject). Then calls `getOrCreateConsumer` on the `IBrokerClient` to create (and return) the IChannelConsumer

12. IBrokerClient (which for us is the CdrBrokerClient constructed in step 8) will use these passed in settings (or if they are not an instance of CdrChannelConsumerSettings, use a copy constructor and uses that) as well as the IMessageListener (WorkChannelMessageListener) to call `IMessageBrokerSvc.createChannelConsumer`

13. IMessageBrokerSvc is actually the KafkaMessageBrokerSvcImpl created in step 7, so it calls down into `BaseMessageBrokerSvc` (this is actually the same for `PulsarMessageBrokerSvcImpl`) where it delegates to `doCreateChannelConsumer`

14. `doCreateChannelConsumer` is implemented differently for our 3 brokers (Kafka, Pulsar, ActiveMq) and (for Kafka) will use the KafkaConsumerFactoryRegistry (constructed and populated in step 5) to get the ConsumerFactory, which it then uses (along with injected kafka-level broker properties) to create a new `CdrKafkaListenerConfig` object, passing that (and the listener) to `doCreateKafkaConsumer` - which creates the IChannelConsumer, and then calls `start` on it.

15. `IChannelConsumer is (for Kafka) KafkaChannelConsumer (implements IChannelConsumer and now has a copy of the ConsumerFactory that it got via the CdrKafkaListenerConfigs) internally creates its own `ICdrKafkaListenerContainer` using the passed in `CdrKafkaListenerConfig`.

16. CdrKafkaListenerContainer  (concrete class of ICdrKafkaListenerContainer) contains both the IMessageListener and the CdrKafkaListenerConfigs. So when `start` is called, it:

17. creates a CdrKafkaListener (either a CdrKafkaAcknowledgingListener or CdrKafkaListener)...

18. constructs ContainerProperties (this is a SpringKafka object, not ours) and adds the IMessageListener to it (the api takes an 'object', even though CdrKafkaListener implements MessageListener - also a springkafka object) and passes this to  `createListenerContainer`

19. createListenerContainer creates a `ConcurrentMessageListenerContainer` using both the passed in ContainerProperties as well as the ConsumerFactory from the CdrKakaListenerConfig.

This is where things kinda split for our purposes.

ConcurrentMessageListenerContainer is a spring kafka object, and it (internally) uses the ConsumerFactory directly to obtain the ConfigurationProperties by calling `getConfigurationProperties` (which returns the map of configs we passed in in step 3), which presumably uses `max.poll.interval.ms` somewhere inside to determine timeouts.

For *this* work, the only thing that's relevant is that `KafkaChannelConsumer` (our IChannelConsumer created in CdrJpaBatch2Config in step 11) *also* contains the CdrListenerConfig, which contains the ConsumerFactory, which is used in BaseBatch2Config to create the SmartInitializingSingleton and from which we can pull the ackTimeout and provide to the IMessageListener (which is also created in this bean, but extended and actually created in CdrJpaBatch2Config in step 12)

----

As you can see, it's all very straightforward.

Pulsar would be similar, but we don't actually care because PulsarBrokerConfig creates PulsarClient without any listener threads (so there's only 1, and no partitions, so no rebalance issues)

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/channel/subscription/SubscriptionChannelFactory.java
TipzCM: Created copy constructor.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
TipzCM: this has been removed

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/FinalStepDataSink.java
TipzCM: this has been removed

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7670-handling-slow-batch-worker.yaml
TipzCM: updated

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
TipzCM: updated

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
TipzCM: removed

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/JobWorkNotification.java
TipzCM: removed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
TipzCM: changed

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/ReductionStepDataSink.java
TipzCM: This has been removed

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/channel/BatchJobSender.java
TipzCM: Actually this is deliberate.

This method is visiblefortesting only.

So setting 'null' is to allow it to reset to the default value.

But i'e changed the code so that the msg function is only used if set, and setting it to null is still allowed (to remove it), but we won't set it back to a default and instead set it in the `sendWorkChannelMessage`

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
TipzCM: this has been removed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -179,7 +179,15 @@ public Optional<WorkChunk> onWorkChunkDequeue(String theChunkId) {
...
-			return chunk.map(this::toChunk);
+			if (chunk.isEmpty()) {
+				return Optional.empty();
+			}
+			Optional<WorkChunk> chunkToReturn = chunk.map(c -> {
+				WorkChunk ret = toChunk(c);
+				ret.setPreviousStatus(chunkLock.getStatus());
+				return ret;
+			});
+			return chunkToReturn;
```
michaelabuckley:
```suggestion
			return chunk.map(c -> {
				WorkChunk ret = toChunk(c);
				ret.setPreviousStatus(chunkLock.getStatus());
				return ret;
			});
```

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/channel/subscription/SubscriptionChannelFactory.java
```diff
@@ -101,12 +102,13 @@ protected ChannelProducerSettings newProducerConfigForMatchingChannel(ChannelPro
...
 	protected ChannelConsumerSettings newConsumerConfigForMatchingChannel(ChannelConsumerSettings theOptions) {
-		ChannelConsumerSettings config = new ChannelConsumerSettings();
-		config.setConcurrentConsumers(getMatchingChannelConcurrentConsumers());
+		ChannelConsumerSettings config;
 		if (theOptions != null) {
-			config.setQualifyChannelName(theOptions.isQualifyChannelName());
-			config.setRetryConfiguration(theOptions.getRetryConfigurationParameters());
+			config = new ChannelConsumerSettings(theOptions);
+		} else {
+			config = new ChannelConsumerSettings();
```
michaelabuckley, line 116:
```suggestion
config = ChannelConsumerSettings.nullSafeCopy(theOptions);
```

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/DelayChunkException.java
michaelabuckley, line 7: trivial: a line of javadoc would be kind

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
```diff
@@ -142,4 +171,28 @@ private void handleFastTracking(BaseDataSink<PT, IT, OT> theDataSink) {
...
 	}
+
+	private ScheduledJobDefinition scheduleHeartbeat() {
+		String jobId = String.format("%s-%s-%s", SCHEDULED_JOB_ID_PREFIX, myInstanceId, myWorkChunk.getId());
+		ScheduledJobDefinition definition = new ScheduledJobDefinition();
+		definition.setJobClass(HeartbeatJob.class);
+		definition.setId(jobId);
+		definition.addJobData(CHUNK_ID, myWorkChunk.getId());
+		// we don't want a time that's <100ms
+		myIHapiScheduler.scheduleClusteredJob(Math.max(myAckTimeout.toMillis() / 3, 500), definition);
```
michaelabuckley: suggestion: this could probably be a local job.  That would be lighter.  clustered jobs leave garbage behind in the db.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
```diff
@@ -71,20 +87,33 @@ public class JobStepExecutor<PT extends IModelJson, IT extends IModelJson, OT ex
...
 				myInstance.getInstanceId(),
 				myCursor.getCurrentStepId(),
 				myWorkChunk == null ? null : myWorkChunk.getId());
 
-		JobStepExecutorOutput<PT, IT, OT> stepExecutorOutput =
-				myJobExecutorSvc.doExecution(myCursor, myInstance, myWorkChunk);
+		JobStepExecutorOutput<PT, IT, OT> stepExecutorOutput;
+		ScheduledJobDefinition scheduledJobDef = null;
+		try {
+			scheduledJobDef = scheduleHeartbeat();
```
michaelabuckley: suggestion: wrap this up with a AutoCloseable handle that unregisters on close.  Then you can use a try-with-resources.  That will also let you do finer testing on this, outside the complicated context of job step execution.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/channel/BatchJobSender.java
michaelabuckley: I was wrong.  You were right.  Add a comment to explain that null means reset to default behaviour

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/JobWorkNotification.java
michaelabuckley: I meant - the null default method in IHasPayloadMessageKey looks dangerous.  Please remove the default body.

michaelabuckley (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
TipzCM: updated

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/DelayChunkException.java
TipzCM: added

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
TipzCM: changed

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/subscription/channel/subscription/SubscriptionChannelFactory.java
TipzCM: updated

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobStepExecutor.java
TipzCM: updated

PR #7770, opened 2026-04-14
7769 consent reject bundle write missing hfj res ver

### LINKED ISSUE #7769: Consent REJECT during bundle lead to DB inconsistencies — HFJ_RESOURCE row created without matching HFJ_RES_VER
The toMethodOutcome() fires STORAGE_PREACCESS_RESOURCES at the end of every write operation, including bundle entries. When a ConsentInterceptor returns REJECT from canSeeResource, this hook nulls the resource on the DaoMethodOutcome. resolveReferencesThenSaveAndIndexResources() skips any outcome where getResource() == null, so createHistoryEntry() is never called. The result is an HFJ_RESOURCE row with no corresponding HFJ_RES_VER row — silent DB corruption — while the server returns 200 OK.

another-luis:
Fixes #7769

## Problem

When a `ConsentInterceptor` returned `REJECT` from `canSeeResource` during a FHIR batch or transaction bundle write (PUT, POST, or PATCH), the `STORAGE_PREACCESS_RESOURCES` hook nulled the resource on the `DaoMethodOutcome`. The second pass (`resolveReferencesThenSaveAndIndexResources`) skipped any outcome with a null resource, so `HFJ_RESOURCE` rows were persisted without matching `HFJ_RES_VER` rows — leaving the database in an inconsistent state. The server returned 200 OK with no indication of the problem.

For PATCH entries the problem was compounded: the patch document (`Binary` or `Parameters`) was a different resource type than the target resource, so a naive restore using the bundle entry resource would have caused a server error.

## Fix

- `DaoMethodOutcome` gains a `myWrittenResource` field, populated in `BaseStorageDao.toMethodOutcome()` alongside `setResource()`, **before** `STORAGE_PREACCESS_RESOURCES` fires. This always holds the actual persisted resource (e.g. `Patient`), regardless of what the bundle entry body contained.
- `BaseTransactionProcessor.handleTransactionCreateOrUpdateOutcome()` restores `outcome.setResource(outcome.getWrittenResource())` after all response-building has completed. Consent filtering of the response is preserved; only the internal write-completion path sees the restored resource.

## Tests

Three regression integration tests added to `ConsentInterceptorResourceProviderR4IT`:
- `testRejectCanSeeResource_batchBundleWrite_dbIntegrityPreservedAndApiFiltered`
- `testRejectCanSeeResource_transactionBundleWrite_dbIntegrityPreservedAndApiFiltered`
- `testRejectCanSeeResource_transactionBundlePatch_dbIntegrityPreservedAndApiFiltered`

Each test verifies both DB integrity (HFJ_RES_VER row exists) and that the resource is hidden from the FHIR API (read → 404, search → empty bundle).


robogary:
Formatting check succeeded!

another-luis:
Closing in favour of a different architectural approach: instead of silently completing the write, consent REJECT during a bundle PUT/POST/PATCH will now bubble a 403/412, causing the transaction to roll back with no DB changes.

PR #7775, opened 2026-04-14
Fix stale `$lookup` result after `CodeSystem` concept updates

### LINKED ISSUE #7754: `$lookup` operation returns stale results after updating `CodeSystem` concepts
**Describe the bug**

When a `CodeSystem` is updated (e.g., via PUT /CodeSystem/{id} with different concepts), subsequent `$lookup` calls can return stale results — either failing to find newly added concepts or still finding concepts from the replaced version.  
                                         
**To Reproduce**                                                                                                                                                                                                                      
                       
1. Create a `CodeSystem` with `content=complete` and concept `codeA`                                                                                                                                                                               
2. Call `$lookup` for `codeA` — succeeds
3. Update the `CodeSystem` with different concepts (`codeB`, `codeC`), removing `codeA` via PUT operation                                                                             
4. Call `$lookup` for `codeB` — fails (expected: found)                                                          
5. Call `$lookup` for `codeA` — succeeds (expected: not found)                                                                                                                                   

Failing test added as part of #7746, where the bug was discovered: `TerminologySvcImplR4Test#lookupCode_codeSystemVersionReplaced_reflectsNewConcepts`

This issue was likely introduced by #6522

**Expected behaviour**
The removed concepts `codeA` and `codeB` should not be found, so the result of the `$lookup` operation should be not found.

- HAPI FHIR Version 8.8

codeforgreen:
Fixes #7754 

**Root cause**

`TermReadSvcImpl` cached the current `CodeSystem` version PID in a `userData` map on the `TermCodeSystem` JPA entity object held in memory. When a `content=complete` `CodeSystem` was updated, `storeNewCodeSystemVersion` created a new `TermCodeSystemVersion` with a new PID, but the in-memory `TermCodeSystem` object still carried the stale version PID in `userData`. Subsequent `$lookup` calls resolved concepts against the old, replaced version and returned stale or missing results.  

**Changes**

Core fix — Caffeine cache replacing `UserData`:
- `myCodeSystemVersionCache`: `Cache<String, Optional<TermCodeSystemVersionDetails>>` keyed by `CodeSystem` URL → version PID + version string
- `myValueSetCache`: `Cache<String, Optional<TermValueSet>> `keyed by `ValueSet` URL → `TermValueSet` entity
- Cache initialized in `start()` from configurable `JpaStorageSettings` (timeout, max size)

Selective invalidation wired to mutations:
- `invalidateCodeSystemCaches()` — called from `applyDeltaCodeSystemsAdd`, `applyDeltaCodeSystemsRemove`, `deleteCodeSystemVersion`, `deleteCodeSystem`
- `invalidateValueSetCaches()` — called from `storeTermValueSet`, `deleteValueSetAndChildren`, `afterValueSetExpansionStatusChange`

Write-through on `CodeSystem` update:
- `updateCodeSystemVersionCache`(url, TermCodeSystemVersion) called at the end of `storeNewCodeSystemVersion` — immediately populates the cache with the new version, avoiding a cold-cache DB hit on the first subsequent lookup

Other fixes:
- `validateCode`: null-checked `getCurrentCodeSystemVersion()` result before accessing `myCodeSystemVersionId`
- `lookupCode`: reuses `myTxTemplate` instead of allocating a new `TransactionTemplate` on every call


**What was achieved**

- `$lookup` after `CodeSystem` update — previously returned stale results due to wrong version PID in cache; now correct because the new version is written to cache     
immediately via write-through after storeNewCodeSystemVersion
- `$lookup`/`$validate-code` after delta add/remove — previously returned stale results because the chain cache was not cleared; now correct because                     
invalidateCodeSystemCaches() clears the version cache and the chain cache is also cleared                                                                            
- `$lookup`/`$validate-code` repeated calls — `CodeSystem` version resolution (URL → version PID) is now served from cache, eliminating one DB query per call
- Concurrent requests for the same `CodeSystem` — Caffeine's atomic per-key loading ensures only one DB query is issued when multiple threads request the same version 
simultaneously, instead of each thread querying independently                                                                                                        
- `$validate-code` with `ValueSet` URL — `TermValueSet` entity lookup is now served from cache, eliminating one DB query per call
- `$validate-code` after `ValueSet` update/delete — previously could use stale pre-expansion status; now correct because storeTermValueSet and deleteValueSetAndChildren 
invalidate the ValueSet cache                                                                                                                                        
                                                                                                                                                                   
What remains uncached: individual concept lookups (`findByCodeSystemAndCode`) still hit the DB on every `$lookup` and `$validate-code` call — the concept-level cache can be a future improvement.


robogary:
Formatting check succeeded!

codeforgreen:
Maybe `MemoryCacheService` should integrated as well. Need to check.

TipzCM (approved):

PR #7781, opened 2026-04-16
Jr 20260414 dry run

JasonRoberts-smile:
This is step 4 of the async package loading epic. It hooks up dry run processing. This MR only changes the dependency initialization step, because the logic for dry run mode for the other steps is already encapsulated in the shared services they call.


robogary:
Formatting check succeeded!

volodymyr-korzh (approved):

PR #7788, opened 2026-04-17
Jr 20260415 error handling

JasonRoberts-smile:
This is the fifth and final stage of the async package installation epic.

Notable changes include:
- ensure that the server supports package installation before launching the root job
- if a package cannot be retrieved or cannot be installed, fail the job
- if a dependency job cannot be launched, or terminates in failure, log the outcome, skip the dependency, and proceed to install the rest of the package
- hook up the cross-version package dependency resolution code that was implemented on the synchronous path in https://github.com/hapifhir/hapi-fhir/pull/7731

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
@@ -403,9 +403,7 @@ public class PackageInstallerSvcImpl implements IPackageInstallerSvc {
 					// If the dependency's FHIR version is incompatible with the server,
 					// attempt to load a version-specific variant (e.g., {id}.r4 for R4 servers)
 					dependency = substituteVersionSpecificPackageIfNeeded(
-							dependency,
-							nextPackage.name(),
-							nextPackage.version());
+							dependency, nextPackage.name(), nextPackage.version());
 
 					// recursive call to install dependencies of a package before
 					// installing the package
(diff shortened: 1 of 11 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -255,6 +242,23 @@ public PackageInstallOutcomeJson install(PackageInstallationSpec theInstallation
...
+	private void validatePackageAlreadyExists(PackageInstallationSpec theInstallationSpec) {
+		boolean exists = myTxService
+				.withSystemRequest()
+				.withRequestPartitionId(myPartitionSettings.getDefaultRequestPartitionId())
+				.execute(() -> {
+					Optional<NpmPackageVersionEntity> existing = myPackageVersionDao.findByPackageIdAndVersion(
+							theInstallationSpec.getName(), theInstallationSpec.getVersion());
+					return existing.isPresent();
+				});
+		if (exists) {
```
TipzCM, line 254: Q: i'm not sure how this is validating anything?

if the package exists, it logs, but allows the code to continue.

Given patterns in hapi, i would (as a caller) expect a method called "validate" to throw an exception. but it doesn't do that even.

TipzCM (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
JasonRoberts-smile: Yeah, that's fair. I'll admit that I didn't really understand the purpose of this check when I encountered it, and that does make choosing a good name difficult. I'll try again.

PR #7797, opened 2026-04-20
Auto-populate subject for auto-created placeholder targets in PID mode

jamesagnew:
When operating in Patient ID Partition Mode with Auto-create placeholder reference targets enabled (i.e. when `PatientIdPartitionInterceptor` is in use with `StroageSettings#isAutoCreatePlaceholderReferenceTargets()`), if a resource in the patient compartment is being stored, and it results in the creation of a placeholder resource which is also in the Patient compartment, the placeholder will be assigned the same patient/subject value as the source resource in order to place it automatically in the same compartment.


robogary:
Formatting check succeeded!

epeartree (approved):
Nice work James.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -304,6 +309,82 @@ public RequestPartitionId identifyForCreate(IBaseResource theResource, RequestDe
...
+
+									IBaseReference patientReference = (IBaseReference) myFhirContext
+											.getElementDefinition("Reference")
+											.newInstance();
+									patientReference.setReference(
+											sourceCompartmentOpt.get().getValueAsString());
+
+									child.getMutator().addValue(target, patientReference);
+									return;
+								}
```
tadgh, line 379: question: so this not only places it into the partition, but also mutates it such that the placeholder now actually **refers** to the Subject at hand?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -304,6 +309,82 @@ public RequestPartitionId identifyForCreate(IBaseResource theResource, RequestDe
...
+				.filter(t -> t.getResourceType().equals(PATIENT_STR))
+				.filter(t -> isNotBlank(t.getIdPart()))
+				.findFirst();
+
+		// Given all the possible SPs that define the patient compartment for the newly
+		// created placeholder resource, find one whose path places a reference directly
+		// off the root of the resource, and use it to set the compartment reference.
+		List<RuntimeSearchParam> targetPatientCompartmentSPs =
+				targetDefinition.getSearchParamsForCompartmentName(PATIENT_STR);
+		if (sourceCompartmentOpt.isPresent() && !targetPatientCompartmentSPs.isEmpty()) {
```
tadgh, line 356: question: cyclomatic complexity of this code is quite high, I'm having trouble parsing it. Is my understanding correct here?

```
1. Fetch all patient compartment SPs for target (e.g. Coverage). -> returns a list of RuntimeSearchParam objects. 
2. For each SP, for each Path in the SP,  find whether:  
    a. it has a period in it, and
    b. If  the path prefix  `resourceType`  matches the target type (e.g. 'Coverage.subject'), then it
    c. Extracts the child (e.g. `subject`, in our `Coverage.subject` example)
    d. if that `subject` is non-null (not sure how it could ever be null, i imagine those fhirpaths must resolve for the SP to even exist) and  
    e. If that target path definition has name "Reference"
    f. THEN, we slam the patient ref into it. 
```



Couple thoughts: 
1. Why not just skip any that have >1 period. I imagine all of those are non-candidates, as they wouldn't be at the root of the resource. 
2. The comparison at line 363 feels odd to me. We just fetched SPs specifically targeting coverage (L#354) . Which SP would not have a prefix of `Coverage`? Is this to cover weird Base SPs, that can span multiple resource types?

Either way, non-blocking

hapi-fhir-storage/src/main/java/ca/uhn/fhir/storage/interceptor/AutoCreatePlaceholderReferenceTargetRequest.java
```diff
@@ -30,15 +30,27 @@
...
+	private final IBaseResource mySourceResource;
 
+	private final IBaseResource myTargetResourceToCreate;
 	/**
 	 * Constructor
 	 */
-	public AutoCreatePlaceholderReferenceTargetRequest(IBaseResource theTargetResourceToCreate) {
+	public AutoCreatePlaceholderReferenceTargetRequest(
+			IBaseResource theSourceResource, IBaseResource theTargetResourceToCreate) {
+		mySourceResource = theSourceResource;
```
tadgh, line 41: thought: this is a nice addition to the API for the request object.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/storage/interceptor/AutoCreatePlaceholderReferenceTargetRequest.java
```diff
@@ -30,15 +30,27 @@
...
 
-	private final IBaseResource myTargetResourceToCreate;
+	private final IBaseResource mySourceResource;
 
+	private final IBaseResource myTargetResourceToCreate;
 	/**
 	 * Constructor
 	 */
-	public AutoCreatePlaceholderReferenceTargetRequest(IBaseResource theTargetResourceToCreate) {
+	public AutoCreatePlaceholderReferenceTargetRequest(
```
tadgh, line 39: Suggestion: Create a javadoc for this class, with an example, explaining what the "source" and "target" of this example are. e.g. `If we are creating an Observation, and it's subject field does not exist in the repo, then for the purposes of this class,  Source == Observation, Target==Patient`. I often find the source/target nomenclature open to interpretation

tadgh (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
jamesagnew: It actually only does the second thing. Then the standard PatientIdPartitionInterceptor path handles the first.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
jamesagnew: 1- I thought about adding that skip, but it wouldn't actually speed things up. The test for a child with the given name will return null if there are additional periods in the string, that that's a super fast hashmap lookup
2- This is for SPs that match multiple types.. E.g. the some SPs apply to multiple resource types and have paths for each of them.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/storage/interceptor/AutoCreatePlaceholderReferenceTargetRequest.java
jamesagnew: Done

tadgh (approved):

Issue #7796, opened 2026-04-20
TermReadSvcImpl doesn't honour CodeSystem.content=not-present, short-circuits algorithmic validators (e.g. UCUM)


fil512:
### Summary

When a `CodeSystem` with `content = not-present` is stored in the JPA terminology tables (e.g. the `http://unitsofmeasure.org` stub from `hl7.terminology.r4`), `TermReadSvcImpl` claims to support it and returns a non-null "code not found" error from `validateCode(...)`. This breaks `ValidationSupportChain`'s iteration and prevents downstream validators like `CommonCodeSystemsTerminologyService` (which can validate UCUM algorithmically via `UcumEssenceService`) from ever being consulted.

### Observed behaviour

Given a typical JPA validation chain:

```
DefaultProfileValidationSupport
JpaPersistedResourceValidationSupport
TermReadSvcImpl
InMemoryTerminologyServerValidationSupport
CommonCodeSystemsTerminologyService    <-- handles UCUM algorithmically
```

…after loading `hl7.terminology.r4` (which ships a `v3-ucum` CodeSystem with `content = not-present`, canonical URL `http://unitsofmeasure.org`):

```
POST /CodeSystem/$validate-code
{"resourceType":"Parameters","parameter":[
  {"name":"url","valueUri":"http://unitsofmeasure.org"},
  {"name":"code","valueCode":"mg/dL"}]}
```

returns `result=false` with message `Unable to validate code http://unitsofmeasure.org#mg/dL - Code is not found in CodeSystem: http://unitsofmeasure.org` — when the algorithmic validator in the chain would have returned success.

### Root cause

`TermReadSvcImpl` doesn't inspect `CodeSystem.content` at all (no `content`, `NOTPRESENT`, or `CodeSystemContentMode` references in the file).

- `isCodeSystemSupported` (`TermReadSvcImpl.java:297`) returns `true` purely on a `TRM_CODESYSTEM` row existing.
- `validateCode` (`TermReadSvcImpl.java:2955`) calls `findCode()`; since `not-present` means concepts aren't enumerated, it always returns empty, and at line 3002-3003 returns `createCodeNotFoundErrorForValidationResult(...)` — a non-null error.
- `ValidationSupportChain.validateCode` (`ValidationSupportChain.java:844`) breaks on any non-null result, so `CommonCodeSystemsTerminologyService` never gets called.

### Expected behaviour

[...]

PR #7798, opened 2026-04-21
GH-7796: TermReadSvcImpl falls through on NOTPRESENT CodeSystem code miss

fil512:
## Summary

This fix corrects a regression in `TermReadSvcImpl` where validating or looking up a code against a CodeSystem with `content=not-present` returned a hard "code not found" error instead of falling through to downstream validators in the `ValidationSupportChain`. The practical impact was that loading an Implementation Guide that pulls in `hl7.terminology.r4` (which ships a not-present UCUM stub) silently broke UCUM code validation on JPA servers — the stub intercepted the UCUM URL before algorithmic validators like `CommonCodeSystemsTerminologyService` could answer.

1. Three call sites in `TermReadSvcImpl` (`validateCode`, `lookupCode`, and `validateCodeIsInPreExpandedValueSet`) now return `null` instead of a non-null "not found" result when the CodeSystem is `content=not-present`, allowing the `ValidationSupportChain` to continue to the next handler.
2. A new package-private helper `isCodeSystemContentNotPresent` centralizes the not-present check; it is null-safe against a missing CodeSystem and against `VersionCanonicalizer` returning null.
3. An existing integration test (`TerminologySvcImplR4Test`) was updated to assert `null` rather than `false` on a code miss against the not-present fixture CodeSystem used throughout that test class.
4. New unit tests in `TermReadSvcImplTest` cover all four branches of the fix (NOTPRESENT/COMPLETE × validateCode/lookupCode) plus a null-canonicalizer guard test.
5. The validation support modules documentation was updated to explain the not-present fall-through behavior.

## Code Review Suggestions

- [ ] **`validateCodeIsInPreExpandedValueSet` call site** — The not-present check is injected just before the final `createCodeNotFoundErrorForValidationResult` call at line ~3010. Confirm this is the correct guard point: a code miss that reaches this method has already failed pre-expanded ValueSet lookup, so returning null here correctly hands off to the chain. Verify there is no path where returning null here causes the chain to silently swallow an actual validation failure (e.g., a ValueSet that wraps a complete CodeSystem alongside a not-present one).
- [ ] **`isCodeSystemContentNotPresent` fetches from root support** — The helper calls `theValidationSupportContext.getRootValidationSupport().fetchCodeSystem(...)`. If the root support is the JPA module itself (not the full chain), a not-present CodeSystem stored only in a downstream module would not be found and the helper would return `false`, falling back to "code not found" rather than null. Confirm the root support at validation time always has visibility into loaded IGs.
- [ ] **Existing test flip** — `TerminologySvcImplR4Test.testValidateCodeAgainstBuiltInSystem` previously asserted `assertFalse(validation.isOk())` on a miss; it now asserts `assertNull(validation)`. Confirm that the fixture CodeSystem created by `BaseTermR4Test#createCodeSystem()` is indeed `content=not-present` and that changing this assertion does not mask a real regression for COMPLETE CodeSystems in that test class.
- [ ] **`lookupCode` null vs. not-found semantics** — `IValidationSupport.lookupCode` returning `null` means "I don't handle this" while returning a `LookupCodeResult` with `found=false` means "I handle this and it's not found." Verify that returning `null` from the not-present path is consistent with how callers of `lookupCode` distinguish "not supported" from "supported but absent."

### Setup
- Stand up a HAPI JPA R4 server with `CommonCodeSystemsTerminologyService` (UCUM support) included in the `ValidationSupportChain`.
- Load the `hl7.terminology.r4` NPM package via `$import` or the package loader. This installs a `CodeSystem` resource for `http://unitsofmeasure.org` with `content=not-present`.

### Test Cases
- [ ] **UCUM code validates successfully after IG load** — POST a `Parameters` resource to `CodeSystem/$validate-code` with `system=http://unitsofmeasure.org` and `code=mg/dL`. Expect HTTP 200 with `result=true`. Before the fix this returned a "code not found" error.
- [ ] **UCUM code validated through a ValueSet** — Load or create a ValueSet that includes `http://unitsofmeasure.org` (e.g., the standard UCUM ValueSet from `hl7.terminology.r4`). Call `ValueSet/$validate-code` with `code=mg/dL`. Expect success.
- [ ] **Invalid UCUM code still fails** — Call `CodeSystem/$validate-code` with `system=http://unitsofmeasure.org` and `code=INVALID_UNIT`. Expect a validation error from the UCUM algorithmic validator (not a JPA "code not found" error — the error message origin matters).
- [ ] **COMPLETE CodeSystem miss still returns error** — Without loading `hl7.terminology.r4`, upload a small custom CodeSystem with `content=complete` and validate a code not in it. Expect HTTP 200 with `result=false` and an appropriate error message (not null/fall-through).
- [ ] **`CodeSystem/$lookup` for UCUM code** — Call `CodeSystem/$lookup` with `system=http://unitsofmeasure.org` and `code=mg/dL` on a server with `hl7.terminology.r4` loaded. Expect a successful lookup result from the algorithmic validator.
- [ ] **Server without UCUM algorithmic support** — On a server that does NOT include `CommonCodeSystemsTerminologyService`, validate a UCUM code after loading `hl7.terminology.r4`. Expect a "code not found" error (the chain exhausted all validators), not a silent success.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
@@ -3019,8 +3019,38 @@ public class TermReadSvcImpl implements ITermReadSvc, IHasScheduledJobs {
 			}
 		}
 
+		if (isCodeSystemContentNotPresent(theValidationSupportContext, theCodeSystemUrl)) {
+			// See GH-7796: a CodeSystem with content=not-present asserts only that the system
+			// exists — it does not enumerate codes. Returning a "code not found" error here
+			// short-circuits the ValidationSupportChain and prevents downstream algorithmic
+			// validators (e.g. UCUM, SNOMED via a remote terminology server) from being
+			// consulted. Returning null lets the chain continue to the next validator.
+			return null;
+		}
+
 		return createCodeNotFoundErrorForValidationResult(
 				theCodeSystemUrl, theCode, null, createMessageAppendForCodeNotFoundInCodeSystem(theCodeSystemUrl));
+	}
+
+	/**
+	 * Returns true if the backing {@link CodeSystem} resolved from the validation support
+	 * chain for {@code theCodeSystemUrl} declares {@code content = not-present}. A
+	 * not-present CodeSystem is a bare assertion that the system exists; it does not
+	 * enumerate its codes, so terminology-backed validators must not claim a code is
+	 * invalid just because it is not indexed.
+	 */
+	private boolean isCodeSystemContentNotPresent(
+			@Nonnull ValidationSupportContext theValidationSupportContext, String theCodeSystemUrl) {
+		if (isBlank(theCodeSystemUrl) || myVersionCanonicalizer == null) {
+			return false;
+		}
+		IBaseResource codeSystem =
+				theValidationSupportContext.getRootValidationSupport().fetchCodeSystem(theCodeSystemUrl);
+		if (codeSystem == null) {
+			return false;
+		}
+		CodeSystem canonical = myVersionCanonicalizer.codeSystemToCanonical(codeSystem);
+		return canonical != null && canonical.getContent() == CodeSystem.CodeSystemContentMode.NOTPRESENT;
 	}
 
 	IValidationSupport.CodeValidationResult validateCodeInValueSet(
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
@@ -2917,7 +2917,7 @@ public class TermReadSvcImpl implements ITermReadSvc, IHasScheduledJobs {
 				return result;
 
 			} else {
-				if (isCodeSystemContentNotPresent(theValidationSupportContext, theSystem)) {
+				if (isNotBlank(theSystem) && isCodeSystemContentNotPresent(theValidationSupportContext, theSystem)) {
 					// Fall through to the next validator in the chain (e.g. UCUM, remote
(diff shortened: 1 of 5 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
```diff
@@ -2890,6 +2896,11 @@ public IValidationSupport.LookupCodeResult lookupCode(
 				return result;
 
 			} else {
+				if (isCodeSystemContentNotPresent(theValidationSupportContext, theSystem)) {
```
epeartree: in method 'validateCodeIsInPreExpandedValueSet', invocation of isCodeSystemContentNotPresent is prefix with checking that 'theSystem' isNotBlank.  should we not do it here as well?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
```diff
@@ -2999,10 +3010,40 @@ public IValidationSupport.CodeValidationResult validateCode(
 			}
 		}
 
+		if (isCodeSystemContentNotPresent(theValidationSupportContext, theCodeSystemUrl)) {
```
epeartree: same here

epeartree (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
fil512: Fixed in latest push.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
fil512: Fixed in latest push.

PR #7812, opened 2026-04-24
7708 - DB - Change HFJ_RESOURCE.IDX_RES_DATE Index to include RES_ID and PARTITION_ID

### LINKED ISSUE #7708: DB - Change HFJ_RESOURCE.IDX_RES_DATE Index to include RES_ID and PARTITION_ID
The proposal in this Issue is to change an existing DB Index on the **HFJ_RESOURCE** Table called **IDX_RES_DATE** which is [defined liked this](https://github.com/hapifhir/hapi-fhir/blob/master/hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java#L105) currently:
```
@Index(name = "IDX_RES_DATE", columnList = BaseHasResource.RES_UPDATED)
```

The new version would look like this:
```
@Index(name = "IDX_RES_DATE_V2", columnList = "RES_UPDATED, RES_ID, PARTITION_ID")
```

This would accommodate a search that use a query string parameter that includes the `lastUpdated` value similar to this example:
`&_lastUpdated=ge2026-03-09T03:00:00Z`

Currently, the Postgres DB Planner uses the `idx_res_type_del_updated` Index when attempting to find search results but it causes it to scan all of the data due to the fact that this Index does not have the `lastUpdated` column in the first position and it therefore uses an "index condition":

Image

One assumption here is that even though this change is based on the behaviour with a Postgres Database, the change will still have a beneficial effect on other database platforms as well.

KevinDougan:
NOTE: This must wait for PR #7894 to merge and be pulled in to this Branch first!

Closes #7708


robogary:
Formatting check succeeded!

PR #7816, opened 2026-04-26
fix: TermReadSvcImpl returns null for content=not-present CodeSystems (#7796)

### LINKED ISSUE #7796: TermReadSvcImpl doesn't honour CodeSystem.content=not-present, short-circuits algorithmic validators (e.g. UCUM)

devimano2011:
Fixes #7796

## Summary

When a `CodeSystem` with `content=not-present` is stored in the JPA terminology tables (e.g. the UCUM stub from `hl7.terminology.r4`), `TermReadSvcImpl.validateCode()` was returning a hard 'code not found' error instead of `null`. This broke `ValidationSupportChain` iteration, preventing downstream validators like `CommonCodeSystemsTerminologyService` (which validates UCUM algorithmically) from ever being reached.

## Root Cause

`TermReadSvcImpl` did not inspect `CodeSystem.content` at all. It claimed authority over any CodeSystem with a `TRM_CODESYSTEM` row, even `not-present` stubs where no concepts are stored.

## Fix

Added an `isNotPresent()` helper and a null-return guard in `validateCode()`: when `findCode()` returns empty AND the backing CodeSystem has `content=not-present`, return `null` instead of a 'code not found' error, allowing the chain to continue.

## Testing

Manually verified: UCUM code `mg/dL` now validates correctly when the `hl7.terminology.r4` stub is present.


devimano2011:
Hi maintainers — thanks for looking at this.

During conflict resolution I noticed that upstream master already introduced `isCodeSystemNotPresentAndHasNoLocalContent()` which addresses the same root cause. Happy to close this PR if the fix is already covered upstream, or I can pivot this PR to add tests and a changelog entry for the existing fix if that would be useful. Let me know how you'd like to proceed!

PR #7821, opened 2026-04-28
Expose getPolicyForResourceType and add additional tests for ResourceCompartmentStoragePolicy

mrdnctrk:
## Summary
- Exposes `getPolicyForResourceType(ReadPartitionIdRequestDetails)` as `public` so cross-module wiring tests can verify the active policy.
- Wraps `DataFormatException` from `FhirContext.getResourceDefinition` in a `ConfigurationException` (Msg.code 2866) so unknown FHIR resource types passed to `setResourceTypePolicies` produce a clear configuration error, as original author intended, rather than a low-level format exception.
- Adds `ResourceCompartmentStoragePolicyTest` covering valid policy names and the error paths for unknown policy names, malformed `ALWAYS_USE_PARTITION_ID/<n>`, and empty input.
- Adds `testSetResourceTypePolicies_UnknownResourceType_ThrowsConfigurationException` to lock in the new exception wrapping.

### DIFF
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
@@ -41,6 +41,7 @@ import ca.uhn.fhir.jpa.searchparam.extractor.ISearchParamExtractor;
 import ca.uhn.fhir.jpa.util.ResourceCompartmentUtil;
 import ca.uhn.fhir.model.api.IQueryParameterType;
 import ca.uhn.fhir.model.primitive.IdDt;
+import ca.uhn.fhir.parser.DataFormatException;
 import ca.uhn.fhir.rest.api.Constants;
 import ca.uhn.fhir.rest.api.RequestTypeEnum;
 import ca.uhn.fhir.rest.api.server.RequestDetails;
@@ -195,12 +196,16 @@ public class PatientIdPartitionInterceptor {
 								resourceType));
 			}
 
-			RuntimeResourceDefinition resourceDef = myFhirContext.getResourceDefinition(resourceType);
-			if (resourceDef == null) {
-				throw new ConfigurationException(Msg.code(2866)
-						+ String.format(
-								"Resource type '%s' is not a valid resource type, can not apply resource type policy",
-								resourceType));
+			RuntimeResourceDefinition resourceDef;
+			try {
+				resourceDef = myFhirContext.getResourceDefinition(resourceType);
+			} catch (DataFormatException e) {
+				throw new ConfigurationException(
+						Msg.code(2866)
+								+ String.format(
+										"Resource type '%s' is not a valid resource type, can not apply resource type policy",
+										resourceType),
+						e);
 			}
 
 			ResourceCompartmentStoragePolicy strictnessMode = entry.getValue();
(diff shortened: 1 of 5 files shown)


robogary:
Formatting check succeeded!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -195,12 +196,16 @@ public void setResourceTypePolicies(
...
 
-			RuntimeResourceDefinition resourceDef = myFhirContext.getResourceDefinition(resourceType);
-			if (resourceDef == null) {
-				throw new ConfigurationException(Msg.code(2866)
-						+ String.format(
-								"Resource type '%s' is not a valid resource type, can not apply resource type policy",
-								resourceType));
+			RuntimeResourceDefinition resourceDef;
+			try {
+				resourceDef = myFhirContext.getResourceDefinition(resourceType);
```
JasonRoberts-smile: question: the previous code had a null check here. That has been removed. I can't determine definitively whether it is possible for `getResourceDefinition()` to return null. It seems like it might be possible in the case of a misconfiguration. Should we retain the null check in addition to the exception handling, for maximum safety?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -567,7 +572,7 @@ private IBaseResource getPreviousVersion(IBaseResource theResource) {
 	 * @see #setResourceTypePolicies(Map) for a description of the default policies
 	 */
 	@Nonnull
-	protected ResourceCompartmentStoragePolicy getPolicyForResourceType(ReadPartitionIdRequestDetails theReadDetails) {
+	public ResourceCompartmentStoragePolicy getPolicyForResourceType(ReadPartitionIdRequestDetails theReadDetails) {
```
JasonRoberts-smile, line 572: It seems that you have only exposed this publicly to facilitate testing in CDR. Would the `@VisibleForTesting` annotation be appropriate here?

hapi-fhir-storage/src/test/java/ca/uhn/fhir/jpa/interceptor/ResourceCompartmentStoragePolicyTest.java
```diff
@@ -0,0 +1,52 @@
...
+			"MANDATORY_SINGLE_COMPARTMENT",
+			"OPTIONAL_SINGLE_COMPARTMENT",
+			"NON_UNIQUE_COMPARTMENT_IN_DEFAULT",
+			"ALWAYS_USE_PARTITION_ID/7"
+	})
+	void testParse_validPolicyName_returnsPolicyWithMatchingName(String thePolicyName) {
+		assertThat(ResourceCompartmentStoragePolicy.parse(thePolicyName).getName())
+				.isEqualTo(thePolicyName);
+	}
+
```
JasonRoberts-smile, line 27: suggestion: for the `ALWAYS_USE_PARTITION` case, add a test to ensure that a valid partition id is present in the policy

JasonRoberts-smile (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
mrdnctrk: yes, added.

hapi-fhir-storage/src/test/java/ca/uhn/fhir/jpa/interceptor/ResourceCompartmentStoragePolicyTest.java
mrdnctrk: sure, added.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
mrdnctrk: yes, introduced a new method for isKnownResourceType as suggested, which takes into account the null case as well. null is actually a possibility, at least in theory, `if (IBaseResource.class.isAssignableFrom(clazz)) {` were to evaluate to false, within `myFhirContext.getResourceDefinition`. Thanks for the suggestion.

PR #7825, opened 2026-04-28
GL-7974: Add partition support to MDM metric service

fil512:
## Summary

This change makes the MDM metrics endpoint (`$mdm-metrics`) partition-aware in multi-tenant Smile CDR deployments. Previously, the three DAO searches in `BaseMdmMetricSvc` always used a plain `new SystemRequestDetails()`, which scopes results to the default partition — causing cross-partition deployments to receive an undercount for golden resources, source resources, and excluded resources. After this fix, the partition context is resolved by CDR before the HAPI service is called and propagated through `GenerateMdmMetricsParameters`.

1. **HAPI — `GenerateMdmMetricsParameters`**: Added a `RequestPartitionId` field so callers can declare the desired partition scope before the metric service runs.
2. **HAPI — `BaseMdmMetricSvc`**: Replaced the three `new SystemRequestDetails()` calls with `SystemRequestDetails.forRequestPartitionId(theParameters.getRequestPartitionId())` so each DAO search respects the requested scope.
3. **CDR — `CdrMdmControllerSvc`**: Added `resolveMetricsPartitionId()` logic: when `search_all_partition=true` it sets `RequestPartitionId.allPartitions()`; when the caller supplies explicit `_partitionIds` it builds a scoped id; otherwise it passes `null` (preserving the original single-partition default for non-partitioned deployments).
4. **CDR — `MdmController`**: Added an optional `_partitionIds` query parameter to the REST endpoint; the resolved `RequestPartitionId` is now passed through to the service.
5. **CDR IT — `MdmMultitenantR4WithCrossPartitionIT`**: New test seeds one patient in `TENANT_A` and one in `DEFAULT`, then asserts that `generateMdmMetrics()` aggregates both source records when `search_all_partition=true`.

This is a **cross-repo change**. The HAPI-side modifications live in the sibling worktree and are not yet in a separate HAPI PR:
- `hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/api/params/GenerateMdmMetricsParameters.java`
- `hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/api/BaseMdmMetricSvc.java`

A separate HAPI PR can be opened if reviewers prefer; otherwise these changes can ship together.

---

## Code Review Suggestions

(See companion Smile CDR MR https://gitlab.com/simpatico.ai/cdr/-/merge_requests/7138 for the full cross-repo description and review notes.)

### DIFF
--- a/hapi-deployable-pom/pom.xml
+++ b/hapi-deployable-pom/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		 <groupId>ca.uhn.hapi.fhir</groupId>
 		 <artifactId>hapi-fhir</artifactId>
-		 <version>8.11.0-SNAPSHOT</version>
+		 <version>8.11.1-SNAPSHOT</version>
 
 		 <relativePath>../pom.xml</relativePath>
 	 </parent>
--- a/hapi-fhir-android/pom.xml
+++ b/hapi-fhir-android/pom.xml
@@ -5,7 +5,7 @@
 	<parent>
 		<groupId>ca.uhn.hapi.fhir</groupId>
 		<artifactId>hapi-deployable-pom</artifactId>
-		<version>8.11.0-SNAPSHOT</version>
+		<version>8.11.1-SNAPSHOT</version>
 
 		<relativePath>../hapi-deployable-pom/pom.xml</relativePath>
 	</parent>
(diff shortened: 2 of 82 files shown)


robogary:
Formatting check succeeded!

TipzCM (approved):

PR #7832, opened 2026-04-29
7811 display mismatch policy not applied to remote terminology

jmarchionatto:
## Summary

`display_mismatch_policy` was being ignored when code validation went through Remote Terminology — a code-with-wrong-display response always came back as `ERROR`, regardless of the configured severity. In-memory and JPA-backed validation already honored the policy; remote did not.

## Fix 

- Lifted `getIssueSeverityForCodeDisplayMismatch` / setter from `InMemoryTerminologyServerValidationSupport` up to `BaseValidationSupport`, so `RemoteTerminologyServiceValidationSupport` inherits the same configurable severity.
- `RemoteTerminologyServiceValidationSupport.createCodeValidationResult` now applies the policy on `result=false` responses with per-issue granularity. The **response-level gate** (request `display` ≠ response `display`) must fire; when it does, an issue qualifies for severity override if the issue itself, or its `diagnostics` primitive element, carries an extension whose URL is either:
  - `operationoutcome-issue-displayMismatched`, or
  - `operationoutcome-message-id` with a primitive value containing "display" (case-insensitive) — e.g. the i18n key `Display_Name_for__should_be_one_of__instead_of` that OntoServer emits.
- Unrelated issues (e.g. `invalid-code`) keep their server severity. Result-level severity is the max across adjusted issues. `result.code` is surfaced when ≥1 issue qualified.
- Synthesized-from-`message` fallback for servers returning no structured `issues`: when displays differ AND the message text contains "display", the synthesized issue gets the configured severity.
- New public constants `DISPLAY_MISMATCH_EXTENSION_URL` and `MESSAGE_ID_EXTENSION_URL`.
- R4 and DSTU3 conversion paths share a single `buildCodeValidationIssues` method via an internal `IssueData` record; per-version code is reduced to a thin extractor that maps the version-specific model to the shared record.


robogary:
Formatting check succeeded!

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/BaseValidationSupport.java
```diff
@@ -2,10 +2,12 @@
 
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.context.support.IValidationSupport;
+import jakarta.annotation.Nonnull;
 import org.apache.commons.lang3.Validate;
 
 public abstract class BaseValidationSupport implements IValidationSupport {
 	protected final FhirContext myCtx;
+	private IssueSeverity myIssueSeverityForCodeDisplayMismatch = IssueSeverity.WARNING;
```
jamesagnew: Thought: It might make sense to create a new subclass of `BaseValidationSupport` called something like `BaseTerminologyServerValidationSupport` or something like that to add this field to. Otherwise, this will be a config item on other ivalidationsupport instance types which don't actually respect this setting.

jamesagnew (approved):

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/BaseValidationSupport.java
jmarchionatto: Good catch! Added

PR #7837, opened 2026-04-29
Package install fails with HAPI-0848 when the version (search) index is stale

### LINKED ISSUE #7836: Package install fails with `HAPI-0848` when `CodeSystem.version` (search) index is stale
**Describe the bug**                                                                                                                                                                        
Installing an IG package that contains a CodeSystem resource fails with HAPI-1286 → HAPI-0848 (duplicate CodeSystem url/version) even though the URL+version pair should match an existing resource that the installer would normally update. The failure occurs whenever the FHIR search index for CodeSystem.version is out of sync with the actual resource state.

**To Reproduce**

1. Start with a server where CodeSystem resource at http://example.org/cs version 1 is already installed (e.g. seeded via a previous IG install).
2. Install an IG package that ships a SearchParameter touching CodeSystem.version (this is common — many IGs add their own SPs). The search-param registry refreshes.
3. Either (a) mark_resources_for_reindexing_upon_search_parameter_change is disabled (often turned off in large deployments to avoid expensive reindex storms), or (b) the background reindex job is still queued and hasn't completed yet.
4. Install a second IG package — typically a re-install, an upgrade, or a dependency — that contains the same CodeSystem at http://example.org/cs version 1.

Install fails with: 
```
HAPI-1286: Error installing IG <pkg>#<ver>:
HAPI-0848: Can not create multiple CodeSystem resources with CodeSystem.url
"http://example.org/cs" and CodeSystem.version "1", already have one withresource ID: CodeSystem/<existing-id>

**Related tests**
**PackageInstallerSvcDuplicateCodeSystemR4Test.install_multiVersionPackageWithStaleVersionSearchIndex_succeeds** (currently failing — locks in the desired behaviour).

```
**Expected behaviour**
The installer detects the existing resource and updates it.

- HAPI FHIR Version 8.10

**Additional context**
Related to #7820

codeforgreen:
Fixes #7836


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -600,6 +606,41 @@ && isContentNotPresentCodeSystem(existingResource)
 		}
 	}
 
+	// Generated by Claude Opus 4.6
+	private Optional<IBaseResource> findExistingResource(
+			IFhirResourceDao<?> theDao, IBaseResource theResource, IBundleProvider theSearchResult) {
+		if (!theSearchResult.isEmpty()) {
+			return theSearchResult.getAllResources().stream().findFirst();
+		}
+		if ("CodeSystem".equals(theResource.fhirType())) {
```
JasonRoberts-smile, line 636: Suggestion for follow-up: if CodeSystem is failing to update due to out-of-date SP indices, it is likely that other resource types are also failing to update, perhaps in less obvious ways. It might be worth putting a ticket in the backlog to investigate and fix additional problems of this category

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -600,6 +606,41 @@ && isContentNotPresentCodeSystem(existingResource)
...
+	private Optional<IBaseResource> findExistingCodeSystemViaTermLayer(
+			IFhirResourceDao<?> theDao, IBaseResource theResource) {
+		String url = extractSimpleValueIfPresent(theResource, "url");
+		if (url.isEmpty()) {
+			return Optional.empty();
+		}
+		String version = extractSimpleValueIfPresent(theResource, "version");
+		return myTxService
+				.withSystemRequest()
+				.withRequestPartitionId(myPartitionSettings.getDefaultRequestPartitionId())
```
JasonRoberts-smile: Elsewhere, we guard calls to `getDefaultRequestPartitionId()` by ensuring that partitioning is enabled first. Is it safe to make this call here if partitioning is disabled?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -600,6 +606,41 @@ && isContentNotPresentCodeSystem(existingResource)
...
+				.execute(() -> myTermCodeSystemStorageSvc.findExistingCodeSystemResourcePid(
+						url, version.isEmpty() ? null : version))
```
JasonRoberts-smile: nitpick (non-blocking): it bugs me a bit that we went out of our way to get a non-null `version` above only to convert it back to `null` here.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcCodeSystemVersionSearchParamR4Test.java
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * <p>The installer locates an existing resource via a {@code (url, version)} FHIR search
+ * before deciding between create and update. When that index is stale (post-SP-change with
+ * reindex suppressed) or misconfigured (active SP pointing at the wrong path), the search
+ * misses, the installer attempts a fresh create, and the term layer rejects it with
+ * {@code HAPI-0848} — even though the same {@code (url, version)} pair already exists.
+ *
+ * <p>Both tests assert the desired contract: the installer should detect the existing
+ * resource via the term layer's mapping and route the install to update it. Both currently
+ * fail; they document the desired fix.
```
JasonRoberts-smile: This comment should be updated, since the fix has been implemented and the test no longer fails

JasonRoberts-smile (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
codeforgreen: Sure. I created https://github.com/hapifhir/hapi-fhir/issues/7896.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
codeforgreen: Good point. I think another round of refinement for the utility methods is needed. Need some variants where you can provide the default value if absent (e.g. null, empty).

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
codeforgreen: I reused the `createRequestDetails` in this class for consistency which performs the check, but it's probably not needed.

PR #7844, opened 2026-05-01
Widen Awaitility window in SubscriptionTopicPartitionR5IT to deflake CI

jmarchionatto:
The negative-assertion test testTopicSubscriptionDoesNotFireAcrossPartitions held a 300ms/500ms during/atMost window — only 200ms of slack — which timed out intermittently on slow CI runners. Widen to 1s/5s; the assertion still fails correctly if the partition guard regresses, only timing-induced false failures are removed.


tadgh:
Formatting check succeeded!

volodymyr-korzh (approved):

PR #7846, opened 2026-05-01
GL-8676: Suppress write-response body when willSeeResource REJECTs

fil512:
# Fix consent `willSeeResource` REJECT not suppressing write response body

## Summary

When a consent service returns REJECT from `willSeeResource` on a write operation (POST/PUT/PATCH), HAPI was correctly nulling the response resource inside `ConsentInterceptor`, but the wire response still contained the full resource body with the original 201/200 status code. The root cause was a two-line ordering bug in `BaseOutcomeReturningMethodBinding.returnResponse()`: the outcome resource was snapshotted from `theMethodOutcome.getResource()` **before** `fireResourceViewCallbacks()` was called, so any PRESHOW mutation (including the consent interceptor's resource suppression) was silently discarded. A secondary fix adds a `myRequestWriteRejectedNoOutcomeKey` userData marker so that `ConsentInterceptor.interceptOutgoingResponse` can set the wire status to `204 No Content` for the write-REJECT-no-OperationOutcome case, matching the existing read-side contract.

Significant changes:

1. **`BaseOutcomeReturningMethodBinding`** — reorder `fireResourceViewCallbacks()` to precede the `outcome = theMethodOutcome.getResource()` snapshot, so PRESHOW mutations reach the wire response for all write methods.
2. **`ConsentInterceptor.interceptPreShow`** — when a REJECT fires on a write operation and no replacement OperationOutcome is supplied, record `myRequestWriteRejectedNoOutcomeKey = true` on the request's userData map.
3. **`ConsentInterceptor.interceptOutgoingResponse`** — on the early-return path (response resource is null), check for the write-rejected marker and set `responseCode = 204 No Content`.
4. **`ConsentInterceptorResourceProviderR4IT`** — four new integration tests covering POST-REJECT, POST-REJECT-with-OperationOutcome, PUT-REJECT, and the PROCEED control case; one pre-existing bundle test updated to source the resource ID from the `Location` header instead of the now-correctly-empty response body.
5. **Docs** (`consent_interceptor.md`) — new section documenting the `willSeeResource → REJECT` behaviour contract for write operations.

Related GitLab issue: https://gitlab.com/simpatico.ai/cdr/-/issues/8676

The customer's smoke test (`cdr.ConsentServiceTest.testConsentService` in `smile-cdr-smoke-test`) was asserting that a POST of a V-tagged Patient yields an empty response body; that assertion was failing because the full resource was leaking through.

## Code Review Suggestions

- [ ] **`BaseOutcomeReturningMethodBinding` reorder** (lines 218-227): the two-line swap (`fireResourceViewCallbacks()` before `getResource()`) is the entire functional fix for the body-leak. Verify that no other call site reads `theMethodOutcome.getResource()` between construction and this binding, and that `fireResourceViewCallbacks()` is idempotent (safe to call once here and nowhere else in the write path).
- [ ] **userData marker scope** (`myRequestWriteRejectedNoOutcomeKey`): the key is instance-scoped (includes `myInstanceIndex`) so multiple `ConsentInterceptor` instances on the same server won't collide. Confirm this is correct if CDR registers more than one consent interceptor per request cycle.
- [ ] **`isWriteOperation` helper**: the marker is gated on `isWriteOperation(theRequestDetails)` — check that PATCH is included in that predicate, since the issue-context lists PATCH as an affected method but it is easy to overlook in a POST/PUT-focused helper.
- [ ] **OperationOutcome short-circuit independence**: the new 204-status path fires only when `getResponseResource() == null` AND the marker is set. If the consent service supplies a replacement OperationOutcome, `getResponseResource()` will be non-null, so the 204 path is skipped and the OO body is returned with its own status — confirm the OO test (`testCreate_WillSeeResourceReject_WithOperationOutcome`) validates this independently.
- [ ] **Pre-existing test fix** (`testGetBundle_WhenWillSeeReturnsRejectForABundle...`): the test now extracts the resource ID from the `Location` header. Confirm the header is always present for a successful create on the JPA provider — if `Location` is absent in some server configurations, the test would fail with a null-pointer rather than a meaningful assertion failure.

### Setup
- [ ] Configure a HAPI JPA server (or Smile CDR `federated_endpoint` module) with a `ConsentInterceptor` whose `IConsentService` returns `proceed()` from `canSeeResource` on write operations and `reject()` from `willSeeResource` for resources matching a test tag (e.g. security label `V`).

### Test cases
- [ ] **POST suppress (primary scenario)**: POST a Patient with the V tag. Assert response status is `204 No Content` and the response body is empty. Assert the resource was actually persisted (GET by ID from a non-consented client returns the resource).
- [ ] **PUT suppress**: PUT an update to a V-tagged Patient. Assert `204 No Content` and empty body. Assert the update was persisted.
- [ ] **PATCH suppress**: PATCH a V-tagged Patient. Assert `204 No Content` and empty body. Assert the patch was applied.
- [ ] **POST PROCEED (control)**: POST an untagged Patient (willSeeResource returns PROCEED). Assert the original 201/200 status and non-empty resource body are returned — confirm the reorder does not affect the happy path.
- [ ] **POST with OperationOutcome replacement**: Configure willSeeResource to REJECT and supply a replacement OperationOutcome. Assert the response body is the OperationOutcome (not empty) and the status is not 204 — the OO short-circuit must remain independent of the new 204 path.
- [ ] **Smoke test regression** (`cdr.ConsentServiceTest.testConsentService`): run the `smile-cdr-smoke-test` suite against a CDR build that includes this HAPI snapshot. The `V tag should yield empty response body` assertion must now pass.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8676-consent-willsee-reject-write-response.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8676-consent-willsee-reject-write-response.yaml
@@ -1,0 +1,7 @@
+---
+type: fix
+title: "ConsentInterceptor now correctly suppresses the response body when consentWillSeeResource
+  returns REJECT on a write operation (POST/PUT/PATCH). Previously, the interceptor returned
+  the original status code (201 or 200) with the full resource body. It now returns 204 No
+  Content with an empty body, consistent with the behavior of an explicit REJECT on read
+  operations."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/security/consent_interceptor.md
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/security/consent_interceptor.md
@@ -32,6 +32,28 @@ This is slower than the normal path, and will prevent the reuse of the results f
 The `willSeeResource()` operation supports reusing cached search results, but removed resources may be 'visible' as holes in returned bundles.
 Disabling `canSeeResource()` by returning `false` from `processCanSeeResource()` will enable the search cache.
 
+## Write Operation Responses
+
+`willSeeResource()` is invoked for every resource that is about to be returned to the client, including the resource
+body that is echoed back from a successful write operation (`POST`/`PUT`/`PATCH`) when the client has requested the
+`Prefer: return=representation` behaviour (the FHIR default). This lets the consent service mask, redact, or fully
+suppress the resource that the server echoes back after a write, using the same logic that protects read and search
+responses.
+
+The behaviour for a write response follows the same rules as a read response:
+
+* `PROCEED` / `AUTHORIZED`: The (possibly modified) resource is returned in the response body and the original
+  success status (`200 OK` or `201 Created`) is preserved.
+* `REJECT` *with* a replacement `OperationOutcome` on the `ConsentOutcome`: The `OperationOutcome` is returned in
+  the response body in place of the written resource. The original success status is preserved.
+* `REJECT` *without* a replacement resource: The response body is suppressed and the status code is changed to
+  `204 No Content`. The write itself is **not** rolled back &mdash; the resource is still persisted on the server,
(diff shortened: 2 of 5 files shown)


tadgh:
Formatting check succeeded!

PR #7850, opened 2026-05-04
GL-8604: Fix chained search params throwing HAPI-1322 in patient compartment mode

epeartree:
## Summary

Fixes a HTTP 405 HAPI-1322 error thrown when searching in MegaScale PATIENT_ID partition mode using a chained search param alongside a direct compartment reference on the same parameter (e.g. `Encounter?subject=Patient/abc&subject.gender=female`). The root cause was that `PatientIdPartitionInterceptor.getResourceIdsForSearchParam` iterated all values for a given search param name — including chained `ReferenceParam` instances — and Guard 1 raised HAPI-1322 when it encountered the `.gender` qualifier before ever reaching the direct reference that carries the patient ID. Chained params refine results but play no role in partition selection; the FHIR search engine applies them after the partition is resolved.

1. **Core fix (HAPI-FHIR):** Added an early `continue` in `getResourceIdsForSearchParam` for any `ReferenceParam` whose `getChain()` is non-null, so chained params are silently skipped during partition-ID extraction. The now-unreachable Guard 2 (HAPI-1323) block is removed alongside.
2. **HAPI-FHIR unit tests:** 5-row `@ParameterizedTest` covering `subject=Patient/abc`, mixed `subject=Patient/abc&subject.gender=female/male`, and `patient=...` variants (including chain-first ordering), plus a standalone chain-only test documenting the all-partitions fall-through in `PatientIdPartitionInterceptorTest`.
3. **CDR IT test:** `testSearchEncounterBySubjectIdentifierAndGender` in `MegaScalePatientIdPartitionModeIT` validates the end-to-end flow including pre-resolve identifier conversion (`subject.identifier=http://patient|ABC` → `subject=Patient/<pid>`) followed by a `subject.gender=female` qualifier.

## Code Review Suggestions

- [ ] **`getResourceIdsForSearchParam` early `continue` placement** — confirm the `continue` fires before Guard 1 (the `getChain() != null` check must precede the `instanceof ReferenceParam` block that reads the patient ID literal). A misplaced guard could silently skip direct references too.
- [ ] **Guard 2 removal scope** — verify no other call site depended on the HAPI-1323 error path (e.g. a caller that catches it to provide a user-facing message). A quick `git grep HAPI-1323` across the CDR repo is worth checking.
- [ ] **Pre-resolve ordering assumption** — the CDR IT test relies on `subject.identifier` being converted to a direct reference *before* `PatientIdPartitionInterceptor` runs partition selection. Confirm the interceptor ordering in `BaseMegaScalePatientIdPartitionMode` makes this guarantee explicit, not accidental.
- [ ] **Chain-only query behaviour** — `Observation?patient.gender=female` with no direct patient reference is included in the parameterized test and should return `isAllPartitions() == false`. Verify the expected partition IDs in the test match the hash of the implicit compartment lookup, not a hardcoded value.

### Setup
- [ ] Deploy Smile CDR with MegaScale PATIENT_ID mode enabled and at least two shard databases configured.
- [ ] Configure a pre-resolve patient identifier system (e.g. `http://patient`) in the MegaScale interceptor settings.
- [ ] Load: one female Patient (`PAT-001`, identifier `http://patient|P001`), one male Patient (`PAT-002`, identifier `http://patient|P002`), one Encounter per patient.

### Test cases
- [ ] **Direct ref + chained qualifier (the reported bug):** `GET /fhir/Encounter?subject=Patient/PAT-001&subject.gender=female` — must return HTTP 200 with the correct Encounter (not HTTP 405 HAPI-1322).
- [ ] **Pre-resolve identifier + chained qualifier:** `GET /fhir/Encounter?subject.identifier=http://patient|P001&subject.gender=female` — must resolve to PAT-001's partition and return the matching Encounter.
- [ ] **Pre-resolve identifier + non-matching chained qualifier:** `GET /fhir/Encounter?subject.identifier=http://patient|P001&subject.gender=male` — must route to the correct partition and return an empty bundle (not an error).
- [ ] **Chain-only query (no direct reference):** `GET /fhir/Observation?patient.gender=female` — must not throw HAPI-1322; should either return results or an appropriate empty bundle.
- [ ] **Unaffected query regression:** `GET /fhir/Encounter?subject=Patient/PAT-001` — must continue to work identically to pre-fix behaviour.
- [ ] **Invalid resource type policy config still fails fast:** Configure a `setResourceTypePolicies` map containing `"NotAResource"` — verify startup throws `ConfigurationException` (not a `NullPointerException` or silent pass-through).


tadgh:

PR #7852, opened 2026-05-04
GH-7847: Fix $everything _type filter leaking unfiltered resource types

### LINKED ISSUE #7847: Request GET /Patient/{id}/$everything with _type parameter returns all linked resources
Both FHIR web and gateway search GET /Patient/{id}/$everything?_type={resource_type}, return all resources linked to the {resource_type} resources. 
For FHIR web search, it can be reproduced with the following steps 
* start smile cdr
* create a patient using PUT http://localhost:8000/Patient/P-1
* Create an encounter http://localhost:8000/Encounter/Enc-1
* create an observation that references the newly created patient and encounter
* GET http://localhost:8000/Patient/P-1/$everything?_type=Observation 

Expected: 1 observation

Actual: 1 patient, 1 observation and 1 encounter

YalingPeiS:
## Summary

This PR fixes `GET /Patient/{id}/$everything?_type=<type>` returning resources of types not requested via the `_type` parameter. When a client requested only Observations, the response also included Encounters and other resources transitively linked through Observation references.

**Key behavior change:** When `_type` is specified, the `_type` filter now applies to _all_ returned resources including the Patient anchor resource itself. **If `Patient` is not listed in `_type`, the Patient anchor is omitted from the response.** To include the Patient resource when using `_type`, add `Patient` to the type list explicitly (e.g., `_type=Observation,Patient`).

The root cause was in `IncludesIterator.fetchNext()` inside `SearchBuilder.java`: it constructed `Include("Observation:*", true)` entries and invoked `loadIncludes()` through a simple overload that did not pass `desiredResourceTypes`. As a result, forward-include expansion followed all outbound references from matching resources without any type filter. The existing `loadIncludesMatchAll()` SQL layer already supported an `AND target_resource_type IN (...)` predicate — it simply was not being invoked in this code path.

1. `IncludesIterator.fetchNext()` now constructs a `SearchBuilderLoadIncludesParameters` object and populates `desiredResourceTypes` from the parsed `_type` parameter values before calling `loadIncludes()`.
2. The Patient anchor is excluded from results when `_type` is specified and does not include `Patient`.
3. The `@OperationParam` description for the `_type` parameter in `BaseJpaResourceProviderPatient` is updated to document this behavior.
4. Three new tests cover the fixed behavior: single-type filter (`_type=Observation` — Patient and Encounter omitted), anchor-type filter (`_type=Patient` — only Patient returned), and multi-type filter (`_type=Observation,Condition` — Patient and Encounter omitted).
5. A new documentation page for the `$everything` operation is added under the JPA server docs.
6. A changelog YAML entry is included for the 8.12.0 release.

## Code Review Suggestions

- [ ] **`_type` filter scope change (Patient anchor)**: The Patient anchor resource is now subject to the `_type` filter. Previously it was always returned. Confirm this behavioral change is acceptable and well-documented.
- [ ] **`SearchBuilderLoadIncludesParameters` construction in `fetchNext()`**: Confirm that `desiredResourceTypes` is populated from the correct source (the `_type` parameter on the `$everything` operation request, not a search parameter). Verify the types are extracted as plain resource type strings (e.g., `"Observation"`) rather than canonical URLs or search token values.
- [ ] **Wildcard include scope**: `Include("Observation:*", true)` uses a wildcard over all Observation search parameters with `isRecurse=true`. Verify that setting `desiredResourceTypes` correctly constrains the SQL predicate without conflicting with recursive include logic for the allowed types.
- [ ] **Null / empty `_type` case**: When `_type` is absent entirely, `desiredResourceTypes` should be `null` or empty, preserving existing behavior (return all types including Patient anchor). Confirm this fallback is handled.
- [ ] **Documentation page**: The new `patient_everything.md` page should be cross-referenced or linked from the existing JPA server operation index so it is discoverable.

### Setup
- [ ] Deploy a JPA R4 server with the patched build.
- [ ] Create a Patient resource and link at least one Observation, one Encounter (referenced from the Observation), and one Condition to that patient.

### Test Cases
- [ ] **Single-type filter — Observation only**: Call `GET /Patient/{id}/$everything?_type=Observation`. Confirm the response bundle contains only Observation resources — **the Patient anchor and Encounter should NOT be present**.
- [ ] **Anchor type included**: Call `GET /Patient/{id}/$everything?_type=Observation,Patient`. Confirm the bundle contains both the Patient resource and Observation resources, but not Encounter.
- [ ] **Anchor type only**: Call `GET /Patient/{id}/$everything?_type=Patient`. Confirm only the Patient resource is returned.
- [ ] **Multi-type filter**: Call `GET /Patient/{id}/$everything?_type=Observation,Condition`. Confirm the bundle contains Observation and Condition resources but not Patient or Encounter.
- [ ] **No `_type` filter (regression)**: Call `GET /Patient/{id}/$everything` without `_type`. Confirm all linked resources (Patient, Observation, Encounter, Condition) are still returned, validating no regression in the default case.
- [ ] **Paging with `_type` filter**: If the server supports paged `$everything`, call with `_count=1&_type=Observation` across multiple pages and confirm all pages respect the type filter.

Closes #7847

---
> **Superseded by #7853** — rebased onto `rel_8_10` to target the 8.10 release branch (post-feature-freeze).

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7847-everything-type-filter-fix.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7847-everything-type-filter-fix.yaml
@@ -4,5 +4,5 @@ issue: 7847
 title: "Fixed a bug where the Patient $everything operation with a _type parameter would return
   linked resources of types not specified in _type. Forward-include expansion in
   IncludesIterator now respects the _type filter, so only resources matching the requested
-  types are included in the response bundle. The Patient anchor resource is always returned
-  regardless of the _type filter, as per the operation's documented contract."
+  types are included in the response bundle. When _type is specified, the Patient anchor
+  resource is also subject to the filter and is omitted unless Patient is listed in _type."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7847-everything-type-filter-fix.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7847-everything-type-filter-fix.yaml
@@ -2,7 +2,7 @@
 type: fix
 issue: 7847
 title: "Fixed a bug where the Patient $everything operation with a _type parameter would return
-  linked resources of types not specified in _type. Forward-include expansion now respects
-  the _type filter, so only resources matching the requested types are included in the
-  response bundle. Additionally, when _type is specified and Patient is not among the listed
-  types, the Patient anchor resource is no longer unconditionally included in the response."
+  linked resources of types not specified in _type. Forward-include expansion in
+  IncludesIterator now respects the _type filter, so only resources matching the requested
+  types are included in the response bundle. The Patient anchor resource is always returned
+  regardless of the _type filter, as per the operation's documented contract."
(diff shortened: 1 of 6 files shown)


tadgh:
Formatting check succeeded!

PR #7862, opened 2026-05-05
7861: Run PartitionLookupSvcImpl.listPartitions in a transaction

### LINKED ISSUE #7861: PartitionLookupSvcImpl.listPartitions() does not run in a transaction bound to the partition-owning EntityManager and can return empty results
PartitionLookupSvcImpl.listPartitions() calls myPartitionDao.findAll() directly, without wrapping in executeInTransaction(...), which then executes against whatever EntityManager is bound to the calling thread. In a MegaScale-style deployment where write transactions are bound to a tenant-shard DataSource (not the cluster-manager DataSource that owns the partition table), findAll() either fails silently or returns an empty list — even though getPartitionByName(...) correctly returns each partition individually.                                     

 This is inconsistent with the other read methods in the same class:                                                                                        
   
  - getPartitionByName(String) (line 267) → executeInTransaction(() -> myPartitionDao.findForName(theName))                                                  
  - getPartitionById(Integer) (line 276) → executeInTransaction(() -> myPartitionDao.findById(theId))
  - An adjacent findAll() call (line 283) → executeInTransaction(() -> myPartitionDao.findAll())

jmarchionatto:
In `PartitionLookupSvcImpl.listPartitions()` wrap call in transaction to avoid executing for EntityManager bound in current thread.

Fixes #7861

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7861-partition-lookup-list-partitions-transaction.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7861-partition-lookup-list-partitions-transaction.yaml
@@ -1,4 +1,5 @@
 ---
 type: fix
 issue: 7861
-title: "Fixed `listPartitions` method on the partition lookup service, which could return wrong empty results in some cases."
+title: "Previously, `PartitionLookupSvcImpl.listPartitions` could return an empty list when invoked without 
+an active transaction bound to the partition-owning `EntityManager`. This has been fixed."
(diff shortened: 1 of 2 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7861-partition-lookup-list-partitions-transaction.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+issue: 7861
+title: "Fixed `listPartitions` method on the partition lookup service, which could return wrong empty results in some cases."
```
codeforgreen: Can you make this more specific "some cases" in particular? Also mention the name of the service. Something like "Method `PartitionLookupSvcImpl.listPartitions` returned empty list when <scenario>. This has been fixed."

codeforgreen (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7861-partition-lookup-list-partitions-transaction.yaml
jmarchionatto: Done

PR #7864, opened 2026-05-05
GH-7851: Fix chained search params throwing HAPI-1322 in patient compartment mode

### LINKED ISSUE #7851: PatientIdPartitionInterceptor throws HAPI-1322 when chained param is combined with direct patient reference
In MegaScale patient-id partition mode, a query like `Encounter?subject=Patient/abc&subject.gender=female` fails with HTTP 405 / HAPI-1322. `getResourceIdsForSearchParam` iterates all values for a param name including chained `ReferenceParam` instances; Guard 1 throws on the `.gender` qualifier before the direct reference is reached. Fix: skip `ReferenceParam` values with a non-null chain during partition selection.

PR: https://github.com/hapifhir/hapi-fhir/pull/7850

epeartree:
Closes #7851

## Summary

Fixes HTTP 405 / HAPI-1322 thrown in MegaScale PATIENT_ID partition mode when a search combines a direct patient reference with a chained qualifier on the same `patient` or `subject` search parameter (e.g. `Observation?subject=Patient/abc&subject.gender=female`). Because `patient` and `subject` are 0..1 cardinality SPs that always resolve to a single Patient, the partition can be unambiguously derived from the direct reference — making the chained qualifier safe to apply without needing its own partition resolution.

1. **`PatientIdPartitionInterceptor.getResourceIdsForSearchParam`** — chained params on `patient`/`subject` are now collected rather than rejected. If at least one direct reference is also present, the chained qualifier is applied normally; if only chains are present with no direct reference, HAPI-2928 is thrown with an actionable message. Chaining on any other compartment SP continues to throw HAPI-1323 (unchanged).
2. **`ResourceLinkPredicateBuilder`** — passes the already-resolved `RequestPartitionId` down to `determineReadPartitionForRequestForSearchType` so the interceptor can merge it with its own derivation rather than starting from scratch.
3. **`IRequestPartitionHelperSvc`** — new `default` overload of `determineReadPartitionForRequestForSearchType` accepting a `RequestPartitionId` hint; sets it on `ReadPartitionIdRequestDetails` before delegating, preserving backward compatibility.
4. **`ResourceCompartmentUtil`** — `PATIENT_COMPARTMENT_SP_PATIENT` and `PATIENT_COMPARTMENT_SP_SUBJECT` constants added as single source of truth for the `"patient"` and `"subject"` SP names shared across the two classes above.
5. **Tests** — new integration tests in `PatientIdPartitionInterceptorR4Test` covering the happy path (direct ref + chain, including reversed ordering), the chain-only rejection (HAPI-2928), and the non-patient/subject chained param rejection (HAPI-1323). Broken test `testSearchObservation_MultipleCompartmentMembership` identified and tracked in https://github.com/hapifhir/hapi-fhir/issues/7882.

## Code Review Suggestions

- [ ] **Partition hint merge in `identifyForRead` Patient branch** — the new block merges `readRequestPartitionId` (from `ReadPartitionIdRequestDetails`) with the partition derived from `_id` params. If both are present and refer to different partitions, `mergeIds` returns a multi-partition result. Confirm this is the desired behaviour and won't cause a scatter query when only one shard should be targeted.
- [ ] **`needingAtLeastOneChainedSpToResolveSet` vs. `idParts` ordering** — the check `if (idParts.isEmpty() && !needingAtLeastOneChainedSpToResolveSet.isEmpty())` relies on all params being processed by `forEach` before either set is inspected. Since `flatMap().forEach()` is sequential (not parallel), this is safe — but worth confirming no caller wraps the search map in a way that reorders params before reaching this method.
- [ ] **`default` overload on `IRequestPartitionHelperSvc`** — the new overload sets `details.setRequestPartitionId(theRequestPartitionId)` and delegates. Confirm `setRequestPartitionId` is propagated through to the interceptor hook (`STORAGE_PARTITION_IDENTIFY_READ`) and that no existing implementation of the interface caches or copies `ReadPartitionIdRequestDetails` in a way that loses the hint.
- [ ] **`testSearchObservation_MultipleCompartmentMembership` commented-out `fail()`** — this is an acknowledged incidental finding (issue #7882), not a regression. The behaviour being tested (rejecting multi-patient searches) was removed in earlier PRs. Confirm reviewers are aware this test was already non-functional before this branch.

### Setup
- Deploy Smile CDR with MegaScale PATIENT_ID mode enabled and at least two shard databases configured.
- Configure one pre-resolve patient identifier system (e.g. `http://patient`).
- Load: one female Patient `PAT-001` (identifier `http://patient|P001`), one male Patient `PAT-002` (identifier `http://patient|P002`), one Observation per patient.

### Test cases
- [ ] **Happy path — direct ref + chained qualifier:** `GET /fhir/Observation?subject=Patient/PAT-001&subject.gender=female` → HTTP 200, returns the correct Observation (not HTTP 405).
- [ ] **Reversed parameter order:** `GET /fhir/Observation?subject.gender=female&subject=Patient/PAT-001` → same result as above; order must not matter.
- [ ] **Pre-resolve identifier + chained qualifier:** `GET /fhir/Observation?subject.identifier=http://patient|P001&subject.active=true` → resolves partition from identifier and applies the chained qualifier; returns matching Observation.
- [ ] **Chain-only, no direct reference:** `GET /fhir/Observation?subject.gender=female` → HAPI-2928 with message pointing the caller to include a direct reference.
- [ ] **Non-patient/subject chained param:** `GET /fhir/Observation?performer.identifier=http://foo|bar` → HAPI-1323 (unchanged behaviour).
- [ ] **Non-matching chained qualifier:** `GET /fhir/Observation?subject=Patient/PAT-001&subject.gender=male` → HTTP 200, empty bundle (no error).
- [ ] **Unaffected regression — plain compartment search:** `GET /fhir/Observation?subject=Patient/PAT-001` → works identically to pre-fix behaviour.


tadgh:
Formatting check succeeded!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -841,19 +861,34 @@ private List<String> getResourceIdsForSearchParam(SearchParameterMap theParams,
 		List<String> idParts = new ArrayList<>();
 		for (List<IQueryParameterType> iQueryParameterTypes : paramAndListForParamName) {
 			for (IQueryParameterType aParam : iQueryParameterTypes) {
+
+				if (aParam instanceof ReferenceParam referenceParam) {
```
codeforgreen: question: Since there is more branching added, is it possible to collapse branching around this code for readability? I would collapse the for loops and use lambdas.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -841,19 +861,34 @@ private List<String> getResourceIdsForSearchParam(SearchParameterMap theParams,
...
+					String chain = referenceParam.getChain();
+					if (nonNull(chain)) {
+						if (PATIENT_COMPARTMENT_SP_PATIENT.equals(theParamName)
+								|| PATIENT_COMPARTMENT_SP_SUBJECT.equals(theParamName)) {
+							// 'patient' and 'subject' SP have a 0..1 cardinality and will always refer to a Patient
```
codeforgreen: please condense this multi-line comment into 1-2 lines

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -841,19 +861,34 @@ private List<String> getResourceIdsForSearchParam(SearchParameterMap theParams,
...
+							// on their own, they can't be resolved and don't need to as they are used as additional
+							// 'filters'
+							// to another SP that resolves to a direct reference (Patient/abc).  Essentially, if we have
+							// one SP
+							// that resolves directly, we don't need to resolve other SP pointing to the same
+							// resource
+							needingAtLeastOneSpToResolveSet.add(chain);
+							continue;
+						} else {
+							throw new MethodNotAllowedException(Msg.code(1323) + "The parameter " + theParamName + "."
```
codeforgreen: nitpick: I saw there is an integration test for this - `testSearchObservation_ChainedSubjectParameter`. Probably a unit would be enough in `PatientIdPartitionInterceptorTest`. A unit test would detect regression faster and would also be targeted. But this is a nice to have change, it is fine.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -708,7 +707,7 @@ public void testSearchObservation_ChainedValue() {
 		try {
 			myObservationDao.search(SearchParameterMap.newSynchronous().add("subject", new ReferenceParam("identifier", "http://foo|123")), mySrd);
 		} catch (MethodNotAllowedException e) {
-			assertEquals(Msg.code(1322) + "The parameter subject.identifier is not supported in patient compartment mode", e.getMessage());
+			assertEquals("HAPI-2928: Could not resolve chained parameter(s) [identifier] on parameter subject. Consider adding a direct Patient reference to your search (?subject=Patient/abc&subject.identifier=...)", e.getMessage());
```
codeforgreen: I think we need to update this test because if this error is not thrown the test still succeeds. Please update the method name as well.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -667,9 +666,9 @@ public void testSearchObservation_ChainedSubjectParameter() {
 		// Multiple ANDs
 		assertThatThrownBy(() -> myObservationDao.search(SearchParameterMap
 				.newSynchronous()
-				.add("subject", new ReferenceParam("identifier", "http://patient|1")), mySrd))
+				.add("performer", new ReferenceParam("identifier", "http://patient|1")), mySrd))
 			.isInstanceOf(MethodNotAllowedException.class)
-			.hasMessageContaining(Msg.code(1322) + "The parameter subject.identifier is not supported in patient compartment mode");
+			.hasMessageContaining(Msg.code(1323) + "The parameter performer.identifier is not supported in patient compartment mode");
```
codeforgreen, line 706: We don't have a test for HAPI-1322 anymore. I suppose it's still needed. Unless we unify conditions and throw in a single place.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -841,19 +861,34 @@ private List<String> getResourceIdsForSearchParam(SearchParameterMap theParams,
...
+									+ chain + " is not supported in patient compartment mode");
+						}
+					}
+				}
+
 				String qualifier = aParam.getQueryParameterQualifier();
 				if (isNotBlank(qualifier) && !Constants.PARAMQUALIFIER_MDM.equals(qualifier)) {
 					throw new MethodNotAllowedException(Msg.code(1322) + "The parameter " + theParamName + qualifier
```
codeforgreen: question: HAPI-1322 and HAPI-1323 seem to talk about the same problem. Is it possible to merge into one condition to throw this error?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
```diff
@@ -403,7 +403,7 @@ private RequestPartitionId calculatePotentialCrossPartitionsForPredicateTarget(
 
 								RequestPartitionId nextPartitionId =
```
codeforgreen, line 409: We need to exercise this method using a chained search `Observation?subject=Patient/abc&subject.gender=female`. Looks like there is no test coverage for the parent method `calculatePotentialCrossPartitionsForPredicateTarget`. `PatientIdPartitionInterceptorR4Test` would probably be a good place to have it. I know this was not introduced by your change but would really be good to have it so the main scenario fails here in hapi-fhir.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -870,9 +905,29 @@ private List<String> getResourceIdsForSearchParam(SearchParameterMap theParams,
...
+		if (idParts.isEmpty() && !needingAtLeastOneSpToResolveSet.isEmpty()) {
+			String errorMsgString = buildErrorMsgForChainedParameters(theParamName, needingAtLeastOneSpToResolveSet);
+			throw new MethodNotAllowedException(Msg.code(2928) + errorMsgString);
+		}
+
 		return idParts;
 	}
 
+	private String buildErrorMsgForChainedParameters(String theParamName, @Nonnull Set<String> theChainedParameterSet) {
+		return new StringBuilder("Could not resolve chained parameter(s) ")
```
codeforgreen, line 916: `StringBuilder` adds verbosity without a performance benefit here (the method is called at most once per request, in an error path). Simple concatenation or `String.format` would be more readable

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -833,6 +851,8 @@ && isBlank(nextEntry.getConditionalUrl())
 
 	@SuppressWarnings("SameParameterValue")
 	private List<String> getResourceIdsForSearchParam(SearchParameterMap theParams, String theParamName) {
+		Set<String> needingAtLeastOneSpToResolveSet = new LinkedHashSet<>();
```
codeforgreen: I am not able to understand what this variable does/stores by reading the name. Please reconsider a more meaningful name.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -708,7 +707,7 @@ public void testSearchObservation_ChainedValue() {
 		try {
 			myObservationDao.search(SearchParameterMap.newSynchronous().add("subject", new ReferenceParam("identifier", "http://foo|123")), mySrd);
```
codeforgreen: please add a test with multiple chained parameters as well e.g. `subject.gender=female&subject.identifier=http://foo|bar` to see how this plays out.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -870,9 +905,29 @@ private List<String> getResourceIdsForSearchParam(SearchParameterMap theParams,
...
+		return new StringBuilder("Could not resolve chained parameter(s) ")
+				.append(theChainedParameterSet)
+				.append(" on parameter ")
+				.append(theParamName)
+				.append(". Consider adding a direct Patient reference to your search (?")
+				.append(theParamName)
+				.append("=Patient/abc&")
+				.append(theParamName)
+				.append(".")
+				.append(theChainedParameterSet.iterator().next())
```
codeforgreen, line 925: nitpick: We probably should list all the chained parameters that are included, or none.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -92,6 +93,9 @@
 import static ca.uhn.fhir.jpa.interceptor.ResourceCompartmentStoragePolicy.alwaysUseDefaultPartition;
 import static ca.uhn.fhir.jpa.interceptor.ResourceCompartmentStoragePolicy.mandatorySingleCompartment;
 import static ca.uhn.fhir.jpa.interceptor.ResourceCompartmentStoragePolicy.nonUniqueCompartmentInDefault;
+import static ca.uhn.fhir.jpa.util.ResourceCompartmentUtil.PATIENT_COMPARTMENT_SP_PATIENT;
+import static ca.uhn.fhir.jpa.util.ResourceCompartmentUtil.PATIENT_COMPARTMENT_SP_SUBJECT;
+import static java.util.Objects.nonNull;
```
codeforgreen, line 98: Replace with `Jackarta.annotation.Nonnull`

codeforgreen (changes requested):
I can see alot of work has been put into this PR. I appreciate the quality of the tests you added and the comment explanations.
One thing to note as a general comment. It looks like this PR does not contain (at least some) changes from https://github.com/hapifhir/hapi-fhir/pull/7487. Please check changes again after merging. 
Another thing which is a must is having enough integration test coverage in hapi-fhir for both success and error cases for chained search. 

nitpick: Formatting changes needed around a few if statements in `ResourceLinkPredicateBuilder` e.g. `if(ResourceCompartmentUtil.PATIENT_COMPARTMENT_SP_SUBJECT`

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -841,19 +861,34 @@ private List<String> getResourceIdsForSearchParam(SearchParameterMap theParams,
 		List<String> idParts = new ArrayList<>();
 		for (List<IQueryParameterType> iQueryParameterTypes : paramAndListForParamName) {
 			for (IQueryParameterType aParam : iQueryParameterTypes) {
+
+				if (aParam instanceof ReferenceParam referenceParam) {
+					String chain = referenceParam.getChain();
+					if (nonNull(chain)) {
+						if (PATIENT_COMPARTMENT_SP_PATIENT.equals(theParamName)
```
codeforgreen: question: out of curiosity, do we only support `*-subject` and `*-patient` search parameters with patient compartment mode? I see that there are other SPs that resolve to Patient resource (see search-parameters.json). If we don't support all, it's probably worth updating our documentation.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
epeartree: Added

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
epeartree: added tests asserting that error 1322 is thrown

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
epeartree: test added

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
epeartree: done

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: i'm using nonNull(Object) not as an annotation on method arguments or return value.  it has to stay

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: we store the name of chained SPs needing at least one resolution.  we do that to support SP multiple ordering like:
?subject=Patient/123&subject.gender=female
?subject.gender=female&subject=Patient/123

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: we only support SP chaining for 'subject' and 'patient' if and only if another SP pointer to the same resource resolves. take as an example:
Observation?subject=Patient/abc&subject.gender=female

subject has a cardinality of 0..1 so if subject points to Patient/abc, there is no other way around that subject.gender points to the same patient.  and this is the hole purpose of this PR, to allow, chained SP when the SP points to an object that is or will be resolved.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: done

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: different problem:
1323 is thrown when chaining is attempted on something else than 'subject' or 'patient' like Observation?performer.identifer.  
1322 is thrown for Patient?_id:exact=123

error messages are similar but different

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: all the chained parameters are used as the first argument. i'm using the first value of 'theChainedParameterSet' only because to build an example.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: good suggestion, done.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: tried to use String.format, turned out to be ugly so stringBuilder it is

epeartree (commented):
Thank you for the review!

**Re: PR #7487 changes** — confirmed: the merge commit from #7487 () is already in this branch (merged 2026-03-16, picked up via the rel_8_10 mergeback). No changes are missing.

**Re: integration test coverage** — added three integration tests in `PatientIdPartitionInterceptorR4Test`:
- `testSearchWithChainedAndResolvedReference` — verifies the reported scenario succeeds (`?subject=Patient/A&subject.active=true` and reversed ordering)
- `testSearchChainedValue_noResolvedReference_fails` — verifies HAPI-2928 is thrown for chain-only with no direct ref
- `testSearchObservation_ChainedSubjectParameterNoDirectReference` — verifies HAPI-1323 still thrown for non-patient/subject chained params

**Re: `if(ResourceCompartmentUtil...` formatting** — spotless was applied in a subsequent commit; no formatting issues remain.

**Re: for-loop branching** — collapsed to `flatMap().forEach()` stream in a subsequent commit.

epeartree (commented):
Thank you for the review!

**Re: PR #7487 changes** — confirmed: the merge commit from #7487 (faea3f636e56) is already in this branch (merged 2026-03-16, picked up via the rel_8_10 mergeback). No changes are missing.

**Re: integration test coverage** — added three integration tests in PatientIdPartitionInterceptorR4Test:
- testSearchWithChainedAndResolvedReference — verifies the reported scenario succeeds (subject=Patient/A&subject.active=true and reversed ordering)
- testSearchChainedValue_noResolvedReference_fails — verifies HAPI-2928 is thrown for chain-only with no direct ref
- testSearchObservation_ChainedSubjectParameterNoDirectReference — verifies HAPI-1323 still thrown for non-patient/subject chained params

**Re: if(ResourceCompartmentUtil... formatting** — spotless was applied in a subsequent commit; no formatting issues remain.

**Re: for-loop branching** — collapsed to flatMap().forEach() stream in a subsequent commit.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7851-fix-subject-chained-sp-megascale.yaml
```diff
@@ -0,0 +1,10 @@
+---
+type: fix
+jira: SMILE-11955
+title: "Previously, when operating in patient-id partition mode, the use of chained SP in searches would error out
+(HAPI-1322) even if the target of the chain is provided with the search (`Observation?subject=Patient/abc&subject.gender=female`).
+
```
codeforgreen, line 6: There is an extra line here and need a new line at the end of the file as well.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7851-fix-subject-chained-sp-megascale.yaml
```diff
@@ -0,0 +1,10 @@
...
+Because the `patient` and `subject` search parameters have 0..1 cardinality and always reference a single Patient
```
codeforgreen, line 7: Do we have 2 fixes in one issue? If so then each in a changelog file probably is more clear.

epeartree: nope, subject and patient are SPs pointing to the same target object and have a cardinality of 0..1

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
codeforgreen: I understood that part. What I am asking about is other SPs which point to Patient resources which would be valid to be resolved in Patient compartment. Probably @jamesagnew  knows since he added these 2. I am not sure if our documentation says we support only these 2 SPs in Patient compartment mode.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
```diff
@@ -403,7 +403,7 @@ private RequestPartitionId calculatePotentialCrossPartitionsForPredicateTarget(
...
 										myRequestPartitionHelperSvc.determineReadPartitionForRequestForSearchType(
-												theRequest, targetType, targetParamMap);
+												theRequest, targetType, targetParamMap, theRequestPartitionId);
```
codeforgreen: nitpick: I would have kept the same call but added the missing setRequestPartitionId call.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -684,6 +708,7 @@ public void testSearchObservation_MultipleCompartmentMembership() {
 					.add("subject", new ReferenceParam("Patient/A"))
 					.add("subject", new ReferenceParam("Patient/B"))
 				, mySrd);
+//			fail();
```
codeforgreen, line 721: What is this one about? Is this intended?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -694,25 +719,12 @@ public void testSearchObservation_MultipleCompartmentMembership() {
 				.add(
 					"subject", new ReferenceOrListParam().add(new ReferenceParam("Patient/A")).add(new ReferenceParam("Patient/B"))
 				), mySrd);
+//			fail();
```
codeforgreen, line 732: Same as above. Is it indended?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -685,9 +689,28 @@ public RequestPartitionId identifyForRead(
...
+						RequestPartitionId readRequestPartitionId = theReadDetails.getRequestPartitionId();
+
 						List<String> idParts = getResourceIdsForSearchParam(params, "_id");
-						return provideMultipleCompartmentPartition(theRequestDetails, idParts);
+						RequestPartitionId multiCompartmentRequestPartitionId =
+								provideMultipleCompartmentPartition(theRequestDetails, idParts);
+
+						if (nonNull(readRequestPartitionId) && readRequestPartitionId.hasPartitionIds()) {
+							if (multiCompartmentRequestPartitionId.hasPartitionIds()) {
+								readRequestPartitionId =
```
codeforgreen, line 706: question: Sorry, I must've missed this before. Does this use-case have a test? I suppose this is about resolving more than one patient?

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
epeartree: yes it is.  this test has silently been failing for about 3 years.  i've raised a ticket to have it addressed.  all details are in the PR description

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
epeartree: yes it is

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
codeforgreen: Sure.  I think I misread thinking you commented it out but actually it was what was signaled before that this test is not working properly. It would be good to resolve #7882 as a fast follow-up against master branch, otherwise it would be lost. Please the issue to code itself if you don't think it will be a fast-follow.

codeforgreen (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
epeartree: i currently can't see a scenario where this could be useful but it may come.  the purpose here is to preserve partitionIds as much as possible.  the issue is with the mergeIds operation as if one of the two operands is a 'allPartitions', then the merge result is all partitions.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java
```diff
@@ -401,9 +402,14 @@ private RequestPartitionId calculatePotentialCrossPartitionsForPredicateTarget(
 								SearchParameterMap targetParamMap = new SearchParameterMap();
 								targetParamMap.add(chain, targetParam);
 
+								ReadPartitionIdRequestDetails details =
```
michaelabuckley, line 405: suggestion: if it is easy, we could collect all the chains on the same singleton prefix into a single SearchParamMap, and that would avoid a bunch of complexity in the interceptor.  If that's easy to do here, you could back out some of the magic in the interceptor.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -658,18 +658,42 @@ void testSearchObservation_mixedCompartmentParam_findsAll() {
...
+		createObservationB();
+
+		// Chain
+		try {
+			myObservationDao.search(SearchParameterMap.newSynchronous().add("subject", new ReferenceParam("identifier", "http://foo|123")), mySrd);
+			fail();
+		} catch (MethodNotAllowedException e) {
+			assertEquals("HAPI-2928: Could not resolve chained parameter(s) [identifier] on parameter subject. Consider adding a direct Patient reference to your search (?subject=Patient/abc&subject.identifier=...)", e.getMessage());
+		}
+
```
michaelabuckley, line 693: suggestion: these look good.  Please add some singleton chains to non-compartment resources to make sure nothing breaks.  E.g. `Encounter?organzation.name=king&organization.active=true`

michaelabuckley (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
epeartree: tested added and passing

PR #7791, opened 2026-04-17
Add ordinal dates to non-unique combo params

jamesagnew:
This PR adds a new ordinal date column to the non-unique combo index table. This allows for high-efficiency queries following common patterns like: `Observation?subject=Patient/A&code=http://foo|123&date=gt2023`

PR #7866, opened 2026-05-05
GL-8702: Fix StringIndexOutOfBoundsException in combo-param extraction

fil512:
## Summary

This PR fixes a regression introduced in PR #7791 ("Add ordinal dates to non-unique combo params", merged 2026-04-25) where any FHIR resource write that triggered composite search-parameter extraction would throw `StringIndexOutOfBoundsException` and return HTTP 500 if a date component's `getValueAsQueryToken()` returned an empty string while its precision indicator was above `DAY`. The concrete trigger is a `Timing` element with only `repeat.bounds.boundsPeriod` set (no explicit event timestamps): the precision field defaults to `SECOND` while the query token is empty, causing the unguarded `substring(0, 10)` call in `BaseSearchParamExtractor.extractParameterCombinationsForComboParamComponent` to crash. The fix adds a length guard that skips the degenerate entry — consistent with how sub-day-precision dates are already handled in the same method — and emits a performance-tracer log at the skip site.

1. **Guard in `BaseSearchParamExtractor`** — a `paramValueAsString.length() < 10` check before the `substring(0, 10)` call skips the combo index entry and fires a performance-tracer log message when the date token is shorter than 10 characters.
2. **Reproduction test (degenerate case)** — `testCreateObservation_TimingWithOnlyBoundsPeriod_ShouldSucceed` in `FhirResourceDaoR4ComboNonUniqueParamTest` creates an `Observation` with a `Timing` effective containing only `boundsPeriod`, asserting the write succeeds without exception.
3. **Adjacent regression test (happy path)** — `testCreateObservation_TimingWithEvent_ShouldSucceed` confirms that a `Timing` with a well-formed event datetime still indexes correctly.
4. **Changelog entry** — YAML entry under `8_12_0` documenting the regression and fix.

## Code Review Suggestions

- [ ] **`continue` semantics in the combinations loop** — the outer loop builds all permutations of component values for the combo index. Verify that `continue`-ing on one component correctly skips the entire combination entry rather than just that component's iteration, so no partial index entry is written.
- [ ] **Guard placement vs. precision check order** — the new length guard sits inside the `> DAY` precision branch. Confirm there is no scenario where a string shorter than 10 chars could arrive in the `== DAY` branch (which uses `paramValueAsString` directly without a `substring` call) and silently produce an incorrect index entry.
- [ ] **`theRequestDetails` null-safety** — the new `firePerformanceInfo` call passes `theRequestDetails`, which can be `null` in batch/reindex code paths. Confirm `firePerformanceInfo` is null-safe for that argument, or that all callers of `extractParameterCombinationsForComboParamComponent` always supply a non-null value.
- [ ] **Unique combo SP coverage** — the fix is in the shared helper used by both unique and non-unique paths, but the tests only cover the non-unique path (`FhirResourceDaoR4ComboNonUniqueParamTest`). Reviewers familiar with the unique-combo path (`FhirResourceDaoR4ComboUniqueParamTest`) should confirm whether a parallel test case is needed there.
- [ ] **Changelog version bucket** — the entry is filed under `8_12_0`. Confirm this matches the intended release milestone for the fix given that PR #7791 (which introduced the regression) was merged to master on 2026-04-25.

### Setup

1. Deploy a HAPI-FHIR JPA server built from this branch.
2. Register a non-unique composite `SearchParameter` on `Observation` that includes an `effective` (date) component — for example, combining `subject` (reference) and `date` (date) with `type=composite`.

### Test Cases

- [ ] **Regression: Timing with only boundsPeriod** — `POST` an `Observation` with `effectiveTiming` set to a `Timing` containing only `repeat.bounds` (a `Period` with a start date, no `event` entries). Verify the response is `201 Created` (not `500 Internal Server Error`) and the resource is retrievable via `GET /Observation/<id>`.
- [ ] **Happy path: Timing with event** — `POST` an `Observation` with `effectiveTiming` set to a `Timing` with at least one explicit event datetime (e.g. `2021-01-02T12:00:01Z`). Verify `201 Created` and that a combo SP search returns the resource (confirms the normal indexing path is unaffected).
- [ ] **Normal dateTime effective still indexes** — `POST` an `Observation` with `effectiveDateTime` set to a full ISO datetime. Verify the resource is returned when searching via the combo SP (confirms the happy-path indexing is unbroken).
- [ ] **Unique combo SP variant** — repeat the Timing-with-only-boundsPeriod test using a *unique* composite search parameter. Verify `201 Created` and no constraint-violation error.
- [ ] **Reindex after fix** — store an `Observation` with a degenerate Timing via this fix, then trigger `$reindex` on that resource. Verify the operation completes without error and the resource remains retrievable.
- [ ] **Performance-tracer log emitted** — with the performance logger at `DEBUG`, reproduce the degenerate-Timing write and confirm a log line containing "no usable date portion" appears for the combo SP component, and that no `WARN` or `ERROR` is logged for the skip.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7866-fix-combo-param-substring-oob.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7866-fix-combo-param-substring-oob.yaml
@@ -1,0 +1,8 @@
+---
+type: fix
+issue: 7866
+title: "Fixed a regression introduced in 8.12.0 where FHIR resource writes would fail with an
+  HTTP 500 / StringIndexOutOfBoundsException when a composite search parameter included a date
+  component whose value produced an empty query token (e.g. a Timing resource with only
+  repeat.bounds.boundsPeriod set). The degenerate date value is now skipped during combo index
+  extraction, consistent with the existing behaviour for sub-day precision dates."
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/BaseSearchParamExtractor.java
@@ -545,6 +545,18 @@ public abstract class BaseSearchParamExtractor implements ISearchParamExtractor 
 					continue;
 				}
 				if (date.getPrecision().ordinal() > TemporalPrecisionEnum.DAY.ordinal()) {
+					if (paramValueAsString.length() < 10) {
+						// No usable date portion (e.g., a Timing with only repeat.bounds.boundsPeriod
+						// produces an empty getValueAsQueryToken() while precision still reads as
+						// the default SECOND). Skip — treat the same as "precision below DAY".
+						myPerformanceTracingLogger.firePerformanceInfo(
+								theRequestDetails,
+								"Not creating a " + theComboParam.getComboSearchParamType()
+										+ " combo index entry for index[" + theComboParam.getName()
+										+ "] because value for component " + theParamComponent.getParamName()
+										+ " has no usable date portion");
+						continue;
+					}
 					// Limit to date portion since that's what we index in the non-unique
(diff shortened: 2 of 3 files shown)


tadgh:
Formatting check succeeded!

jmarchionatto (approved):
Looks good. Approved.

PR #7873, opened 2026-05-06
Change datetimepicker format to use data attribute

tisoft:
The fhir testpage generates date/datetime input fields for search parameters depending on their name. It stores the format in the data-bs-toggledate-format attribute, but the datepicker is always initialized with a date without time.

This change corrects that, so that the datepicker respects the date format from the input field. This allows to enter dates with time.


jamesagnew (approved):
Looks good! Thanks for the contribution!

tadgh:
Formatting check succeeded!

PR #7879, opened 2026-05-08
Use better query plan in MSSQL DBPM

jamesagnew:
In MSSQL/SQL Server, when running in DB Partition Mode the query for fetching resource bodies during a search is inefficient because MSSQL does not properly support tuples. This query has been reworked for better performance.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
@@ -49,6 +49,7 @@ import ca.uhn.fhir.jpa.model.dao.JpaPid;
 import ca.uhn.fhir.jpa.model.entity.BaseResourceIndexedSearchParam;
 import ca.uhn.fhir.jpa.model.entity.ResourceHistoryTable;
 import ca.uhn.fhir.jpa.model.entity.ResourceLink;
+import ca.uhn.fhir.jpa.model.entity.ResourceTable;
 import ca.uhn.fhir.jpa.model.search.SearchBuilderLoadIncludesParameters;
 import ca.uhn.fhir.jpa.model.search.SearchRuntimeDetails;
 import ca.uhn.fhir.jpa.model.search.StorageProcessingMessage;
@@ -74,6 +75,7 @@ import ca.uhn.fhir.jpa.searchparam.util.PerformanceTracingLogger;
 import ca.uhn.fhir.jpa.util.BaseIterator;
 import ca.uhn.fhir.jpa.util.CartesianProductUtil;
 import ca.uhn.fhir.jpa.util.CurrentThreadCaptureQueriesListener;
+import ca.uhn.fhir.jpa.util.DialectSvc;
 import ca.uhn.fhir.jpa.util.QueryChunker;
 import ca.uhn.fhir.jpa.util.ScrollableResultsIterator;
 import ca.uhn.fhir.jpa.util.SqlQueryList;
@@ -110,9 +112,12 @@ import ca.uhn.fhir.util.StopWatch;
 import ca.uhn.fhir.util.StringUtil;
 import ca.uhn.fhir.util.UrlUtil;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ListMultimap;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Multimap;
 import com.google.common.collect.MultimapBuilder;
+import com.google.common.collect.Multimaps;
 import com.healthmarketscience.sqlbuilder.ComboCondition;
 import com.healthmarketscience.sqlbuilder.Condition;
 import com.healthmarketscience.sqlbuilder.SelectQuery;
@@ -128,6 +133,9 @@ import jakarta.persistence.Tuple;
 import jakarta.persistence.TypedQuery;
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
+import jakarta.persistence.criteria.Fetch;
+import jakarta.persistence.criteria.Join;
+import jakarta.persistence.criteria.JoinType;
 import jakarta.persistence.criteria.Predicate;
 import jakarta.persistence.criteria.Root;
 import jakarta.persistence.criteria.Selection;
@@ -163,6 +171,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import static ca.uhn.fhir.jpa.dao.index.IdHelperService.EMPTY_PREDICATE_ARRAY;
 import static ca.uhn.fhir.jpa.model.util.JpaConstants.NO_MORE;
 import static ca.uhn.fhir.jpa.model.util.JpaConstants.UNDESIRED_RESOURCE_LINKAGES_FOR_EVERYTHING_ON_PATIENT_INSTANCE;
 import static ca.uhn.fhir.jpa.search.builder.QueryStack.LOCATION_POSITION;
@@ -209,6 +218,7 @@ public class SearchBuilder implements ISearchBuilder<JpaPid> {
 	protected final IInterceptorBroadcaster myInterceptorBroadcaster;
 	protected final IResourceTagDao myResourceTagDao;
 	private final PerformanceTracingLogger myPerformanceTracingLogger;
+	private final DialectSvc myDialectSvc;
 	private String myResourceName;
 	private final Class<? extends IBaseResource> myResourceType;
 	private final HapiFhirLocalContainerEntityManagerFactoryBean myEntityManagerFactory;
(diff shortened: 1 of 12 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -1459,6 +1466,68 @@ private void doLoadPids(
...
+
+		// Split the resource PIDs into partitions for efficient querying. We use a tree
+		// for the partition IDs to make the SQL consistent for tests.
+		Multimap<Integer, Long> partitionIdToPid =
+				MultimapBuilder.treeKeys().arrayListValues().build();
+		for (JpaPid pid : theResourcePids) {
+			// The SearchBuilder pads the list with entries that will never match
+			// a real resource in order to create predictable numbers of parameters,
+			// but there is no reason to do that here
+			if (!JpaConstants.NO_MORE.equals(pid)) {
```
jmarchionatto: Not sure if we want to remove this "feature", because having queries be backed by a plan guide require exact query shape, including number of params. If we don't do that, last page will likely not use the plan.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
```diff
@@ -154,6 +155,14 @@ public class ResourceTable extends BaseHasResource<JpaPid> implements Serializab
 			valueBridge = @ValueBridgeRef(type = JpaPidValueBridge.class))
 	private JpaPid myPid;
 
+	/**
+	 * This is only here to support SearchBuilder#loadCurrentResourceVersionsForMsSqlDbpm
+	 */
+	@SuppressWarnings("unused")
+	@Basic(fetch = FetchType.LAZY)
+	@Column(name = "RES_ID", nullable = false, insertable = false, updatable = false)
+	private Integer myResourceId;
```
jmarchionatto: Shouldn't be Long?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/DialectSvc.java
```diff
@@ -0,0 +1,56 @@
+package ca.uhn.fhir.jpa.util;
+
+import ca.uhn.fhir.jpa.config.HibernatePropertiesProvider;
+import ca.uhn.fhir.jpa.migrate.DriverTypeEnum;
+import ca.uhn.fhir.jpa.model.dialect.IHapiFhirDialect;
+import graphql.VisibleForTesting;
```
jmarchionatto: Shouldn't be com.google.common.annotations?

hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/jpa/dao/r5/dbpartitionmode/DbpmEnabledTest.java
```diff
@@ -7,21 +7,31 @@
 import ca.uhn.fhir.jpa.model.dao.JpaPid;
 import ca.uhn.fhir.jpa.model.util.JpaConstants;
 import ca.uhn.fhir.jpa.provider.TerminologyUploaderProvider;
+import ca.uhn.fhir.jpa.searchparam.SearchParameterMap;
 import ca.uhn.fhir.jpa.term.api.ITermLoaderSvc;
+import ca.uhn.fhir.jpa.util.DialectSvc;
+import ca.uhn.fhir.rest.api.server.IBundleProvider;
 import ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException;
 import ca.uhn.fhir.util.ClasspathUtil;
+import net.sourceforge.plantuml.klimt.creole.Sea;
```
jmarchionatto: Unused imports. This one is only a transitive dependency.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/DialectSvc.java
```diff
@@ -0,0 +1,56 @@
...
+			return true;
+		}
+		return getDriverType().orElse(null) == DriverTypeEnum.MSSQL_2012;
+	}
+
+	/**
+	 * For unit testing only
+	 */
+	@VisibleForTesting
+	public static void setForceMsSqlMode(boolean theForceMsSqlMode) {
```
jmarchionatto, line 52: A test that throws before @AfterEach runs (or any test in the same JVM that doesn't reset the flag) leaves the entire test JVM permanently in "MSSQL mode," which silently changes the SQL emitted by SearchBuilder for every subsequent test. DbpmEnabledTest resets it in @AfterEach, but any future test author who copies the pattern and forgets to reset will introduce hard to debug bugs.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/DialectSvc.java
```diff
@@ -0,0 +1,56 @@
...
+import org.hibernate.dialect.Dialect;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Optional;
+
+public class DialectSvc {
+	private static final Logger ourLog = LoggerFactory.getLogger(DialectSvc.class);
+
+	@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
```
jmarchionatto: Meant to be for `myDriverType`?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -1459,6 +1466,68 @@ private void doLoadPids(
...
+				partitionIdToPid.put(pid.getPartitionId(), pid.getId());
+			}
+		}
+
+		CriteriaBuilder cb = myEntityManager.getCriteriaBuilder();
+		CriteriaQuery<ResourceHistoryTable> cq = cb.createQuery(ResourceHistoryTable.class);
+		Root<ResourceHistoryTable> from = cq.from(ResourceHistoryTable.class);
+
+		Join<?, ?> resourceTable = (Join<?, ?>) from.fetch("myResourceTable", JoinType.LEFT);
```
jmarchionatto: Why LEFT join? Original query is: `JOIN FETCH v.myResourceTable t … WHERE t.myVersion = v.myResourceVersion` -> INNER JOIN.

jmarchionatto (approved):
Looks good. Approved with a few comments

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
jamesagnew: Yeah, I think you're right... I was going with the thought that because we're restructuring the query to be grouped by partition, we'd be less likely to get reuse. But I guess "less likely" is still better than "never" so I'll bring this back.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/model/entity/ResourceTable.java
jamesagnew: Whoa! Good catch!!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/DialectSvc.java
jamesagnew: fixed

hapi-fhir-jpaserver-test-r5/src/test/java/ca/uhn/fhir/jpa/dao/r5/dbpartitionmode/DbpmEnabledTest.java
jamesagnew: fixed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/DialectSvc.java
jamesagnew: Hmm good point.... I'll put it in a try-finally, and also reset from BaseJpaTest

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/DialectSvc.java
jamesagnew: Oops yup

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
jamesagnew: Good point. I don't think it matters in the end since both sides will always be present, but we should be equivalent. I'll change it here.

PR #7880, opened 2026-05-08
cve dep bumps for rel 8 10 

jdar8:
**Notable changes:**
- Bump spring boot from 3.4 to 3.5
- Replaced `org.glassfish.jakarta.el` with `tomcat.embed.el` because they both implement `jakarta.el-api`, which caused duplicate dependency errors. `tomcat.embed.el` is the implementation of the `el-api` used by Spring boot, and now comes with spring-boot 3.5.x+. Additionally added an exclusion for jetty since it also brought in an `el-api ` implementation as well.


tadgh:
Formatting check succeeded!

epeartree (approved):

PR #7886, opened 2026-05-11
fix(#7885): Timing boundsPeriod start-only no longer incorrectly populates sp_value_high

### LINKED ISSUE #7885: Timing boundsPeriod with start-only incorrectly sets sp_value_high to start date instead of null
**Describe the bug**
When persisting a resource with a `Timing` search parameter whose `boundsPeriod` contains only a `start` date (no `end`), `BaseSearchParamExtractor.addDate_Timing()` incorrectly populates `sp_value_high` with the start date instead of leaving it null.

The root cause lies in `addDate_Timing()`, which collects all dates (Timing events, `boundsPeriod.start`, and `boundsPeriod.end`) into a single `TreeSet`, then uses `dates.first()` and `dates.last()` for the low and high bounds respectively. When only `boundsPeriod.start` is present, the set has one entry, so `first() == last()`, making `sp_value_high` a copy of `sp_value_low`. This conflates "open-ended period" with "period that ends at its own start", losing the semantic meaning of an absent `period.end`.

According to the [FHIR spec](https://hl7.org/fhir/R4/datatypes-definitions.html#Period.end), an absent `period.end` means the period has no upper bound. Setting `sp_value_high` to the start date misrepresents this as a zero-duration closed period.

This is a follow-on to #7601, which fixed the NPE triggered by the same scenario.

**To Reproduce**
1. POST the following `ServiceRequest`:
```json
{
  "resourceType": "ServiceRequest",
  "status": "active",
  "intent": "option",
  "occurrenceTiming": {
    "repeat": {
      "boundsPeriod": {
        "start": "2025-09-17T02:25:28-04:00"
      }
    }
  }
}
```
2. Query `hfj_spidx_date` for the created resource.
3. Observe that `sp_value_low` and `sp_value_high` are both populated with the start date.

**Expected behavior**
`sp_value_low` should be set to the start date and `sp_value_high` should be `null`, reflecting an open-ended period.

**Environment**
- HAPI FHIR Version: 8.10.0

jorgepinilla-smile:
## Summary

Fixes #7885 (follow-on to #7601).

- `addDate_Timing()` now tracks `boundsPeriod.end` separately from the shared `TreeSet`. When no explicit end date is present, `sp_value_high` is set to `null` rather than defaulting to the start date.
- Added unit tests in `SearchParamExtractorR4Test` covering: start-only `boundsPeriod` (the bug), `boundsPeriod` with both dates, multiple Timing events, and a single Timing event.
- Added changelog entry `8712-error-in-creating-search-parameter.yaml`.

## Test plan

- [x] `testBoundsPeriodStartOnlyProducesNullHighValue` — open-ended period yields null `sp_value_high`
- [x] `testBoundsPeriodWithStartAndEndPopulatesBothValues` — both bounds correctly populated
- [x] `testTimingEventsOnlyUseFirstAndLastEventDates` — existing event-date behaviour unchanged
- [x] `testSingleTimingEventProducesSameLowAndHighValue` — single point-in-time event unaffected

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8712-error-in-creating-search-parameter.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8712-error-in-creating-search-parameter.yaml
@@ -1,0 +1,7 @@
+---
+type: fix
+issue: 8712
+jira: SMILE-11050
+title: "Previously, when indexing a Timing search parameter whose boundsPeriod contained only a start
+  date, the date extractor would incorrectly populate sp_value_high with the start date instead
+  of null, violating the FHIR specification for open-ended periods. This has been fixed."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8712-error-in-creating-search-parameter.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8712-error-in-creating-search-parameter.yaml
@@ -1,7 +1,0 @@
----
-type: fix
-issue: 8712
-jira: SMILE-11050
-title: "Previously, when indexing a Timing search parameter whose boundsPeriod contained only a start
-  date, the date extractor would incorrectly populate sp_value_high with the start date instead
-  of null, violating the FHIR specification for open-ended periods. This has been fixed."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7885-error-in-creating-search-parameter.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7885-error-in-creating-search-parameter.yaml
@@ -1,0 +1,7 @@
+---
+type: fix
+issue: 7885
+jira: SMILE-11050
+title: "Previously, when indexing a Timing search parameter whose boundsPeriod
+  contained only a start date, the date extractor would incorrectly set the
+  upper bound to the start date instead of leaving it open-ended. This has been fixed."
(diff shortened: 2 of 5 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8712-error-in-creating-search-parameter.yaml
```diff
@@ -0,0 +1,7 @@
+---
+type: fix
```
epeartree: seems like 8712 is the gitlab ticket.  should be 7885.  
also, dont' forget to update the file prefix.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8712-error-in-creating-search-parameter.yaml
```diff
@@ -0,0 +1,7 @@
...
+issue: 8712
+jira: SMILE-11050
+title: "Previously, when indexing a Timing search parameter whose boundsPeriod contained only a start
```
epeartree: the description is hard to read in one breath.  consider refactoring removing  references to 'sp_value_high'.  it is also not required to mention that it was violating the fhir spec.

epeartree (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8712-error-in-creating-search-parameter.yaml
jorgepinilla-smile: oh good catch! I've correctly renamed the file and updated the ticket number to the GitHub ticket number.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/8712-error-in-creating-search-parameter.yaml
jorgepinilla-smile: Did some refactoring to the message. Removed the names of the fields and the fhir spec bit and used more user friendly verbose.

PR #7889, opened 2026-05-11
Use snomed type id instead of display label to determine hierarchy

### LINKED ISSUE #6492: Terminology Upload - SnomedCT - relationships lost without EN version

**Describe the bug**
The `hapi-fhir-cli` does no import relationships when loading the SnomedCT in a different language than English.

**To Reproduce**
1. Download the french edition of SnomedCT here: https://smt.esante.gouv.fr/catalogue-des-terminologies/
2. Import the ZIP file:  `hapi-fhir-cli upload-terminology -d ./SnomedCT_ManagedServiceFR_PRODUCTION_FR1000315_20240621T120000Z.zip -v r5 -t http://localhost:9090/server/DEFAULT -u 'http://snomed.info/sct'`
3. The table `TRM_CONCEPT_PC_LINK` is not populated with the links between SnomedCT concepts

**Expected behavior**
All the relationships must be imported.

HAPI FHIR: 7.2.2
OS: Windows

**Additional context**
In the `SctHandlerRelationship` class, we can find this condition that fails when the SCT is not in english:

```java
if (typeConcept.getDisplay().equals("Is a (attribute)"))
```

I don't why a string comparison is performed here since simply cheking the `typeId` should work the same way and be valid for other languages.

joel-bach:
Extensions for other languages do not always contain the "is-a" concept directly and if they do, it does not use the English label. Therefore, using the type id directly makes the import of the hierarchy possible for other languages while not changing the behavior for English.

I also took the liberty to remove some unused code I encountered in the process (`rootConcepts` being passed to the handler and `ignoredTypes`). Let me know if you want me to revert that.

Fixes #6492


jamesagnew (approved):
Looks good, thanks for the contribution!

tadgh:
Formatting check succeeded!

PR #7891, opened 2026-05-11
Use typed partition id in IN-clause padding sentinel

jmarchionatto:
The NO_MORE JpaPid sentinel used to pad IN-clause expansions now binds a concrete Integer (-1) to PARTITION_ID instead of null. This keeps MSSQL sp_executesql parameter declarations uniform between real and padded slots so plan-guide matching stays stable. JpaPidFk also gets explicit @JdbcTypeCode on its id/partition columns for the same reason.


tadgh:
Formatting check succeeded!

jamesagnew (approved):

PR #7895, opened 2026-05-11
7829 unique package dependencies name version should only be installed once

### LINKED ISSUE #7829: Unique package dependencies  (name, version) should only be installed once
**Describe the bug**
When running the package install operation with i`nstallDependencies=true` (the default) the same package is installed multiple times (e.g. `hl7.terminology.r4` and `hl7.fhir.r4.core` tend to be very common).

**To Reproduce**

**Expected behavior**

**Screenshots**

- HAPI FHIR Version 8.10

**Additional context**

JasonRoberts-smile:
This MR introduces new state into the package installer to track which dependency packages have already been processed, so that a package that appears multiple times in the dependency hierarchy will not get processed multiple times.

The synchronous processing handles this by passing a Set down the call hierarchy.

Since the asynchronous processing mode is multi-threaded, we need to store the state in the repository so that all worker threads can be given the id of a shared resource. This is modelled with a `Basic` resource with the modules processed so far stored as extensions.

This MR also fixes a related bug where dependent packages would be written to the package cache even if the install is running in INSTALL_ONLY or dry run modes.

Fixes #7829


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java
```diff
@@ -598,7 +613,7 @@ public NpmPackage installPackage(PackageInstallationSpec theInstallationSpec) th
 
 		return newTxTemplate().execute(tx -> {
 			try {
-				return loadPackageInner(theInstallationSpec.getName(), theInstallationSpec.getVersion());
+				return loadPackageInner(theInstallationSpec.getName(), theInstallationSpec.getVersion(), true);
```
tadgh, line 616: question (non-blocking): wonder why we built a manual tx template, instead of using the HapiTxManager here. Do you know?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -388,32 +397,45 @@ public PackageInstallationStatusJson checkInstallationStatus(String theJobId) {
...
+			PackageInstallationSpec theInstallationSpec,
+			PackageInstallOutcomeJson theOutcome,
+			DependencyRegister theDependencyRegister)
 			throws ImplementationGuideInstallationException {
 		List<PackageUtils.DependentPackage> dependentPackages =
 				PackageUtils.extractDependentPackages(npmPackage, theInstallationSpec, theOutcome);
 
 		for (PackageUtils.DependentPackage nextPackage : dependentPackages) {
+			if (theDependencyRegister.containsDependency(nextPackage.name(), nextPackage.version())) {
+				continue;
```
tadgh, line 441: nit: add a debug log here, since if this feature doesn't work as intended, debugging it will be tricky without.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -388,32 +397,45 @@ public PackageInstallationStatusJson checkInstallationStatus(String theJobId) {
...
 				if (theInstallationSpec.isDryRun()) {
 					theOutcome
 							.getMessage()
 							.add(String.format(
 									"Installation would install %s:%s", nextPackage.name(), nextPackage.version()));
 				} else {
+					boolean updateCache = theInstallationSpec.getInstallMode()
+									== PackageInstallationSpec.InstallModeEnum.STORE_AND_INSTALL
+							|| theInstallationSpec.getInstallMode()
+									== PackageInstallationSpec.InstallModeEnum.STORE_ONLY;
```
tadgh, line 456: question(non-blocking): is there a reason we aren't using the inversion? e.g. InstallModeEnum != INSTALL_ONLY. I get that this is safer in case we add a new enum value, but i'm wondering if any other enum values even exist. INSTALL, STORE, STORE_AND_INSTALL seems like a closed set to me.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -1141,4 +1163,21 @@ private boolean resourceHasElementNamed(IBaseResource theResource, String theEle
...
+
+		public boolean containsDependency(String thePackageName, String theVersion) {
+			return dependencies.contains(buildKey(thePackageName, theVersion));
+		}
+
+		@Nonnull
+		private static String buildKey(String thePackageName, String theVersion) {
+			return thePackageName + "#" + Objects.requireNonNullElse(theVersion, "current");
+		}
+	}
```
tadgh, line 1234: question: This smells like an in-memory-cache to me. Did you consider the design of just adding an in-memory-cache from the MemoryCacheService? Its totally possible its not suitable for this use case, but any time we can avoid writing a new class, we should consider it.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImplTest.java
```diff
@@ -528,9 +533,9 @@ void testInstall_existingSearchParamHasDomainResourceBase_splitsBaseForIncomingC
 			SearchParameter updatedExisting = iterator.next();
 			assertEquals("existing-sp", updatedExisting.getIdPart());
 			List<String> updatedBase = updatedExisting.getBase().stream().map(CodeType::getCode).toList();
-			assertThat(updatedBase).doesNotContain("Patient");
-			assertThat(updatedBase).doesNotContain("DomainResource");
-			assertThat(updatedBase).contains("Observation");
+			assertThat(updatedBase).doesNotContain("Patient")
+				.doesNotContain("DomainResource")
+				.contains("Observation");
```
tadgh, line 638: nice cleanup, much cleaner

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcR4Test.java
```diff
@@ -1982,12 +1982,89 @@ public void testInstallR4Package_doesOverwriteNonExternalCodeSystem(@TempDir Pat
...
+		);
+
+		// validate
+		assertThat(myFakeNpmServlet.getTimesCalled("/test.package.a/1.0")).isEqualTo(1);
+		assertThat(myFakeNpmServlet.getTimesCalled("/test.package.b/1.0")).isEqualTo(1);
+		assertThat(myFakeNpmServlet.getTimesCalled("/test.package.c/1.0")).isEqualTo(1);
+	}
+
+	@Test
+	public void testInstallR4Package_sharedDependencyWithSibling_installsDependencyOnce() throws Exception {
```
tadgh, line 2108: suggestion: have claude whip up an ASCII representation of the parent/child/sibling relationship and leave it as a javadoc on these tests, so the relationship is more easily viewable by future readers.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
```diff
@@ -0,0 +1,192 @@
...
+				basicDao.update(resource, srd);
+
+				return true;
+			} catch (ResourceNotFoundException | ResourceGoneException e) {
+				// This should never happen
+				// The lifespan of the resource is controlled by the root job.
+				// A dependency job should never find that the resource hasn't been created yet
+				//   or has already been destroyed.
+				// However, if it does happen, we should carry on to install the dependency anyway.
+				ourLog.warn(
```
tadgh: issue: if this should never be reached, this log should be error level as it should be floated back to the dev team

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
```diff
@@ -0,0 +1,192 @@
...
+						"Unable to determine whether the package dependency {}#{} has been processed.",
```
tadgh:
```suggestion
						"Unable to determine whether the package dependency {}#{} has been processed. Exptected to find Resource  `Basic/{}`, but this resource does not exist.",
```

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
```diff
@@ -0,0 +1,192 @@
...
+						thePackageName,
+						thePackageVersion,
+						e);
+				return true;
+			} catch (ResourceVersionConflictException e) {
+				ourLog.info(
```
tadgh: thought: this seems like a pretty likely scenario to occur when a job kicks off. Have you checked how often it happens? I wouldn't want to pollute info-level logs with something that isn't relevant in day-to-day operations. Maybe debug?

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
```diff
@@ -0,0 +1,192 @@
...
+		}
+
+		return extensions.stream().anyMatch(t -> isExtensionMatches(t, thePackageName, thePackageVersion));
+	}
+
+	@SuppressWarnings("unchecked")
+	private boolean isExtensionMatches(IBase theExtension, String thePackageName, String thePackageVersion) {
+		IPrimitiveType<String> url =
+				(IPrimitiveType<String>) TerserUtil.getFirstFieldByFhirPath(myFhirContext, FIELD_URL, theExtension);
+		if (url == null || !Strings.CS.equals(EXTENSION_URL, url.getValue())) {
```
tadgh, line 159: huh wow i've never seen Strings.CS.equals before. Makes it pretty clear.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
```diff
@@ -0,0 +1,192 @@
...
+			IPrimitiveType<String> subUrl =
+					(IPrimitiveType<String>) TerserUtil.getFirstFieldByFhirPath(myFhirContext, FIELD_URL, subExtension);
+			IPrimitiveType<String> subValue = (IPrimitiveType<String>)
+					TerserUtil.getFirstFieldByFhirPath(myFhirContext, FIELD_VALUE, subExtension);
+			match &= (subUrl != null && subValue != null)
+					&& ((Strings.CS.equals(subUrl.getValue(), SUBEXTENSION_NAME_URL)
+									&& Strings.CS.equals(subValue.getValue(), thePackageName))
+							|| (Strings.CS.equals(subUrl.getValue(), SUBEXTENSION_VERSION_URL)
+									&& Strings.CS.equals(subValue.getValue(), thePackageVersion)));
+		}
```
tadgh: This is some high cognitive load code. I understand its purpose, but recommend java-doccing this method to explain the matching rules.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
```diff
@@ -0,0 +1,192 @@
...
+		return myDaoRegistry.getResourceDao(implementingClass);
+	}
+
+	private RequestDetails createRequestDetails() {
+		SystemRequestDetails requestDetails = new SystemRequestDetails();
+		if (myPartitionSettings.isPartitioningEnabled()) {
+			requestDetails.setRequestPartitionId(myPartitionSettings.getDefaultRequestPartitionId());
+		}
+		return requestDetails;
+	}
```
tadgh, line 142: issue: I believe some hard work went in recently to ensure that partitioning with SRDs works. I believe theres a constructor we should be using to support the usage here. I'll go look for it.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
```diff
@@ -0,0 +1,192 @@
...
+		return outcome.getId().toUnqualifiedVersionless().getValueAsString();
+	}
+
+	public <T extends IBaseResource> boolean shouldProcessDependency(
+			String theId, String thePackageName, String thePackageVersion) {
+		IIdType idType = TerserUtil.newElement(myFhirContext, FIELD_ID, theId);
+		IFhirResourceDao<T> basicDao = getBasicDao();
+
+		RequestDetails srd = createRequestDetails();
+		while (true) {
```
tadgh: issue: unbounded infinite loop in case of multiply thrown ResourceVersionCOnflictExceptions, which _should_ be impossible, but will probably happen anyhow. Recommend a maximum iteration count as a fallback

tadgh (approved):
Approved pending minor changes. Also, I didnt see it in here, but do we exercise the Async Package Loading in an IT where we are _also_ in a partitioned environment? If not, that probably needs to be added, but can wait until this merges due to time constraints.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/JpaPackageCache.java
JasonRoberts-smile: No clue. I just left it as I found it

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
JasonRoberts-smile: I don't think the MemoryCacheService meets my needs. I really want this register to be per-call, but the cache service manages global caches. However, the value of this tiny class, in my intention, isn't the cachiness of it, it's the encapsulation of the formatting of the key. And that would still be needed even if I replaced the HashSet with a Cache.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
JasonRoberts-smile: I tend to prefer positive expressions over negative expressions, unless there's a significant difference in complexity or performance.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcR4Test.java
JasonRoberts-smile: I can't have Claude do anything. I didn't get in before the moratorium on new licenses. But this is a good suggestion, and I can try to write something up myself

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
JasonRoberts-smile: Fair point

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
JasonRoberts-smile: A bunch of stuff from StringUtils got depricated a couple of upgrades ago. I've been rewriting stuff to use the new patterns as I find them, but there's a lot out there.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
tadgh: Wow I made a typo in my suggestion, hopefully you typed it correctly (unlike me)

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
JasonRoberts-smile: Were you able to find an example of this? I went on a scavenger hunt but couldn't find any recent cases that are significantly different from what I'm doing here

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
JasonRoberts-smile: done

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
JasonRoberts-smile: I have extracted the logic to a well-documented method

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
JasonRoberts-smile: Limited the loop to 10 iterations, and log an error if that limit is ever reached.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/installpackage/DependencyManager.java
tadgh: Ah yeah. `SystemRequestDetails.forRequestPartition(myPartitionSetings.getDefaultPartition())` which should be safe

JasonRoberts-smile: Thanks. Done. I suppose the edge case where a caller has provided a non-null default partition id while leaving `partitioningEnabled == false` has been sufficiently guarded against downstream that it won't be a problem if we don't explicitly guard against it here.

PR #7900, opened 2026-05-12
Snomed CT import designations (synonyms, fully specified names)

joel-bach:
Currently, only one of the display values is imported for every Snomed concept.
This means that if a concept has multiple synonyms, only one of them is available in hapi-fhir after the import.
This is the case in the international edition but the impact is bigger for national editions with other languages as only one of the languages is imported (and it can be a different language for every concept).

This PR changes the behavior to
- always use English as the language for the `display` attribute of the concept (if available).
- also create designations for all synonyms as well as fully specified names, across all languages encountered.

This allows a requester to look up any synonyms and available translations for a concept. E.g. in the Swiss edition, we have three languages beside English (German, French and Italian) and currently we only get one for each concept, rendering the import relatively useless unless translations are imported as well.


tadgh:
Formatting check succeeded!

joel-bach:
@jamesagnew This is another issue we ran into with the Snomed import of non-English editions (although it also affects the English edition to some extent). We added it in our fork but supposed that it could be beneficial for other folks as well, hence the PR. Let me know what you think 🙏

PR #7902, opened 2026-05-12
[7897] fix history count cache stale after write

### LINKED ISSUE #7897: "_history" query omits version 1 temporarily after update
Run a `_history` query against a Resource, then immediately update the Resource and run the `_history` query again (within 60 seconds), the new version shows up, but not version 1.

YalingPeiS:
### Issue
Run a `_history` query against a Resource, then immediately update the Resource and run the `_history` query again (within 60 seconds), the new version shows up, but not version 1.

### Cause
The HISTORY_COUNT memory cache (60-second TTL) is populated by history reads but never invalidated on writes. When a stale total of 1 sits in the cache, `PersistedJpaBundleProvider.size()` returns 1, the page builder requests rows [0, 1), `HistoryBuilder.fetchEntities` runs `setMaxResults(1)` sorted by myUpdated DESC, and the database — which now correctly contains v1+v2 — returns only the newest row (v2). v1 is silently dropped from the page until the cache entry expires.

### Fix
Invalidate the affected HISTORY_COUNT cache entries whenever the underlying history changes.

Closes #7897

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7897-fix-history-count-cache-stale-after-write.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7897-fix-history-count-cache-stale-after-write.yaml
@@ -1,0 +1,10 @@
+---
+type: fix
+issue: 7897
+jira: SMILE-11997
+title: "Fixed the bug where a `_history` query immediately following a resource update could
+  return a stale total and omit older versions for up to 60 seconds. The HISTORY_COUNT
+  memory cache was populated by reads but never invalidated by writes, so a primed total
+  of 1 caused the next history page to be sized to a single row and return only the newest
+  version. Resource create/update/delete now invalidate the affected instance, type, and
+  system level HISTORY_COUNT cache entries after the transaction commits."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7897-fix-history-count-cache-stale-after-write.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7897-fix-history-count-cache-stale-after-write.yaml
@@ -6,5 +6,5 @@ title: "Fixed the bug where a `_history` query immediately following a resource 
   return a stale total and omit older versions for up to 60 seconds. The HISTORY_COUNT
   memory cache was populated by reads but never invalidated by writes, so a primed total
   of 1 caused the next history page to be sized to a single row and return only the newest
-  version. Resource create/update/delete now invalidate the affected instance, type, and
+  version. Resource create/update/delete now invalidates the affected instance, type, and
   system level HISTORY_COUNT cache entries after the transaction commits."
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
@@ -3208,6 +3208,39 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 		}
 	}
 
+	@Override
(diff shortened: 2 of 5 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7897-fix-history-count-cache-stale-after-write.yaml
```diff
@@ -0,0 +1,10 @@
+---
+type: fix
+issue: 7897
+jira: SMILE-11997
+title: "Fixed the bug where a `_history` query immediately following a resource update could
+  return a stale total and omit older versions for up to 60 seconds. The HISTORY_COUNT
+  memory cache was populated by reads but never invalidated by writes, so a primed total
+  of 1 caused the next history page to be sized to a single row and return only the newest
+  version. Resource create/update/delete now invalidate the affected instance, type, and
```
TipzCM: "invalidates"

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -3208,6 +3208,44 @@ protected void validatePartitionIdMatch(
...
+
+	@Override
+	protected void postUpdate(ResourceTable theEntity, T theResource, RequestDetails theRequestDetails) {
+		super.postUpdate(theEntity, theResource, theRequestDetails);
+		invalidateHistoryCountCacheAfterCommit(theEntity);
+	}
+
+	@Override
+	protected void postDelete(ResourceTable theEntity) {
+		super.postDelete(theEntity);
```
TipzCM, line 3230: I worry that these are all 'middle layer' overrides.

overrides just so happen to be calling 'super' as well, but that might not have been the case.

It's a bit moot though, because the documentation on BaseHapiFhirDao indicates that it's tech debt that `BaseHapiFhirDao` and this class should be merged (we should probably try and prioritize that)

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
```diff
@@ -277,6 +280,103 @@ public void testSystemHistory_CountCacheEnabled_Concurrent() throws ExecutionExc
 
 	}
 
+	@Test
+	public void testInstanceHistory_CacheInvalidatedAfterUpdate() {
```
TipzCM: fyi, test naming should be things like

`nameOrFeatureUnderTest_conditions_result`

example (for this one):

historyOnPatientInstant_afterUpdate_yieldsCorrectHistory

TipzCM (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7897-fix-history-count-cache-stale-after-write.yaml
YalingPeiS: Good catch. Updated.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/JpaHistoryR4Test.java
YalingPeiS: Updated to follow the same naming approach as those in this test class.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
YalingPeiS: The super method is empty in this case as shown below
```
/**
	 * Subclasses may override to provide behaviour. Called when a pre-existing resource has been updated in the database
	 *
	 * @param theEntity The resource
	 */
	protected void postDelete(ResourceTable theEntity) {
		// nothing
	}
```

PR #7904, opened 2026-05-13
Fix remote terminology property handling for Snowstorm

### LINKED ISSUE #7212: ClassCastException during snomed CodeSystem $lookup with RemoteTerminology server as Snowstorm Terminology Server
https://github.com/hapifhir/hapi-fhir-jpaserver-starter/issues/834

joel-bach:
Snowstorm returns properties where the `code` part uses `valueString` instead of `valueCode` (which is what was expected in hapi-fhir). This commit changes the behavior to also allow `valueString` and therefore allows the usage of Snowstorm as a remote terminology server in hapi-fhir.

Fixes #7212


tadgh:
Formatting check succeeded!

PR #7921, opened 2026-05-15
7920 fhir search involving AND joins where search param type is reference returns nothing

### LINKED ISSUE #7920: FHIR search queries involving AND joins where search param type is reference returns nothing
Repro steps:
- Create two new patients and keep track of their IDs
- Create an AuditEvent where `AuditEvent.entity[0].what=Patient/A` and `AuditEvent.entity[1].what=Patient/B`
- Execute a fhir search for the AuditEvent that contains references to Patient/A AND Patient/B in `AuditEvent.entity.what`

**Expected**

`http://localhost:8000/AuditEvent?entity=Patient/A&entity=Patient/B` <--- this should return a Bundle with a single entry containing the AuditEvent with those two references.

**Actual**

A bundle returned with no entry.

YalingPeiS:
### Issue
The search `http://localhost:8000/AuditEvent?entity=Patient/A&entity=Patient/B` returns empty bundle instead of the bundle with the 1 matching entry.

### Cause
The cache key only include the parameter name and the resource type with chained reference search. For the bug case `AuditEvent?entity=Patient/A&entity=Patient/B` that has direct references, the cache key `entity:Patient` is the same for both.

### Fix
Update the cache key construction so that **direct reference** parameters (no chain) contribute a per-target component (resource type + id part) to the key, while **chained** reference parameters keep contributing only the target resource type.

Add tests.

Closes #7920


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7920-fix-and-joins-on-direct-reference-search-return-nothing.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 7920
+title: "Previously, FHIR searches that AND together multiple clauses on the same reference
```
tadgh: Reword this to indicate that this was a temporary regression during development of 8.10, not a longstanding issue.

tadgh (approved):

tadgh (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_10_0/7920-fix-and-joins-on-direct-reference-search-return-nothing.yaml
YalingPeiS: Updated.

PR #7929, opened 2026-05-19
Update org.hl7.fhir.core to 6.7.8

dotasek:


tadgh:
Formatting check succeeded!

dotasek:
This will compile locally, but because this isn't configured for the same SNAPSHOT repo org.hl7.fhir.core distributes to, it won't pass on pipelines until an actual release comes out.

PR #7930, opened 2026-05-19
7933: Fix per-resource $expunge of resources with contained-resource cross-links

### LINKED ISSUE #7933: Per-resource $expunge fails with HAPI-2415 when index_contained_resources is enabled
With `index_contained_resources` enabled, calling `$expunge` per-resource on
a resource that is referenced via a contained resource in another resource
fails with `HAPI-2415` (referential integrity violation on `HFJ_RES_LINK`).

The bulk `$expunge everything` path is not affected; only the per-resource
path fails.

another-luis:
## Summary

This PR fixes a `HAPI-2415` constraint violation that occurs when calling per-resource `$expunge` on a resource that is the TARGET of an `HFJ_RES_LINK` row owned by a still-alive resource. The failure surfaces specifically when `index_contained_resources=true` (with recursive indexing enabled), because the contained-resource indexer creates cross-resource link rows where the parent resource is the SOURCE and a referenced real resource is the TARGET — leaving an orphaned FK reference that blocked expunge.

The root cause was asymmetric cleanup: `JpaResourceExpungeService.deleteAllSearchParams` deleted SOURCE-side link rows (where the expunged resource is the source) but never deleted TARGET-side rows (where the expunged resource is the target). `ExpungeEverythingService` already performed symmetric cleanup on the bulk path; this PR brings the per-resource path into alignment.

Key changes:

1. **New DAO method** — `IResourceLinkDao.deleteByTargetResourceId(JpaPid)` issues a `DELETE FROM HFJ_RES_LINK WHERE TARGET_RESOURCE_ID = ?` to remove inbound link rows before the resource row itself is deleted.
2. **Service call** — `JpaResourceExpungeService.deleteAllSearchParams` now calls the new method at the end of its cleanup sequence, symmetrically matching the existing source-side deletion.
3. **Regression tests** — `ExpungeContainedResourcesR4Test` covers the primary failure scenario plus two adjacent cases (expunge-everything and expunge on a non-contained-indexed resource).
4. **Changelog** — `8648-fix-expunge-contained-resources.yaml` recorded under `8_12_0` (`type: fix`).

---

## Code Review Suggestions

- [ ] **`isHasLinks()` flag is SOURCE-side only.** `ResourceTable.isHasLinks()` describes whether the resource has outbound link rows (SOURCE). There is no symmetric `isTargetOfLinks()` flag, so the new `deleteByTargetResourceId` call is intentionally unconditional — it always runs a single indexed lookup. Reviewers should confirm this is acceptable rather than expecting a guard flag.
- [ ] **Alive-source side-effect.** When a resource is expunged, any `HFJ_RES_LINK` rows where it is the TARGET (but whose SOURCE is still alive) will be deleted. The source resource's serialized JSON blob is unaffected; only the index row is removed. This matches `ExpungeEverythingService` behaviour — but reviewers should confirm no secondary index or search query relies on TARGET-side link rows for a live source resource remaining present after its target is expunged.
- [ ] **FK symmetry with `ResourceTableFKProvider`.** The change aligns cleanup with `ResourceTableFKProvider#getResourceForeignKeys` (lines 59–60), which already enumerates both `SRC_RESOURCE_ID` and `TARGET_RESOURCE_ID` columns. Confirm the ordering of deletions in `deleteAllSearchParams` is consistent with FK dependency order.
- [ ] **Performance — unconditional DELETE on every soft-delete path.** The new `DELETE FROM HFJ_RES_LINK WHERE TARGET_RESOURCE_ID = ?` executes on every `BaseHapiFhirDao#updateResource` call where `newParams == null` — which covers every soft-delete and every save under `RESOURCE_SKIP_INDEXING`. For resources with no inbound links this is a single indexed lookup returning zero rows: cheap but not free. On high-write deployments or tables with heavy link traffic this adds a round-trip per delete. **Persistence-team reviewers should assess whether this is acceptable or whether a pre-check (e.g. a count query or a flag) would be warranted.**
- [ ] **Test scope.** The reproduction test uses `index_contained_resources=true` with recursive indexing. Verify the test storage-settings setup accurately reflects how this feature is typically deployed.

---

### Setup
- [ ] Deploy a HAPI-FHIR JPA server (R4) with `daoConfig.setIndexContainedResources(IndexEnabledEnum.ENABLED)` and recursive contained indexing enabled.

### Integration test scenarios

- [ ] **Primary scenario — expunge of a cross-resource link target:** Create a parent resource (e.g. `Organization`) containing an inline `Practitioner`. Have that contained practitioner reference a standalone `Patient`. Verify the contained-resource indexer writes an `HFJ_RES_LINK` row from the `Organization` (SOURCE) to the `Patient` (TARGET). Delete the `Patient`. Call `$expunge` with `expungeDeletedResources=true` on the `Patient`. **Pre-fix:** fails with `HAPI-2415`. **Post-fix:** succeeds; confirm the `HFJ_RES_LINK` row pointing at the expunged `Patient` is gone.
- [ ] **Expunge-everything path still works:** Run `$expunge` with `expungeEverything=true` on a server containing contained-resource link rows. Confirm no constraint violations and all `HFJ_RES_LINK` rows are cleared.
- [ ] **Non-contained-indexed resource:** Expunge a resource that has never been the target of a contained-resource link row. Confirm the new `deleteByTargetResourceId` call completes without error (zero-row delete).
- [ ] **Live-source resource unaffected:** After expunging the TARGET resource, fetch the SOURCE resource (the one that contained the reference). Confirm it is still readable and its FHIR representation is unchanged — only the index row was removed.
- [ ] **Regression — standard expunge unaffected:** Verify existing expunge behaviour (resources with no inbound links) is unchanged in performance and correctness.

Closes #7933


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeContainedResourcesR4Test.java
```diff
@@ -0,0 +1,184 @@
...
+ * references another real resource, the contained-resource indexer creates extra
+ * {@code HFJ_RES_LINK} rows where the parent is the source and the externally referenced
+ * resource is the target. After deleting the target and asking for a per-resource
+ * {@code $expunge} of just the target, {@code JpaResourceExpungeService.deleteAllSearchParams}
+ * only deletes link rows where the target is the source; rows where it is the link target
+ * are left behind. The subsequent {@code DELETE FROM HFJ_RESOURCE} fails the
+ * {@code FK_RESLINK_TARGET} referential integrity check and is surfaced as
+ * {@code HAPI-2415}.
+ * <p>
+ * The customer's published stack shows {@code FK_RESLINK_SOURCE} (i.e. the symmetric
```
tadgh: issue: wow claude is yappin. Reduce the verbosity of this comment by at least 50%
issue: Do not refer to customers in fixes.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeContainedResourcesR4Test.java
```diff
@@ -0,0 +1,184 @@
...
+		super.before();
+		myStorageSettings.setExpungeEnabled(true);
+		myStorageSettings.setAllowMultipleDelete(true);
+		myStorageSettings.setIndexOnContainedResources(true);
+		myStorageSettings.setIndexOnContainedResourcesRecursively(true);
+		// The customer's reproduction uses a bundle-delete of a graph of resources;
+		// within a bundle, referential integrity on delete is not enforced between
+		// the just-deleted siblings. Simulate that here by disabling enforcement so
+		// we can target the per-resource expunge bug in isolation.
+		myStorageSettings.setEnforceReferentialIntegrityOnDelete(false);
```
tadgh: issue(blocking): if this is not what the original bug report showed, then this replication is unfaithful to the original report. Is it possible to replicate the actual failure exhibited by the bug report?

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8648-fix-expunge-contained-resources.yaml
```diff
@@ -0,0 +1,4 @@
+---
+type: fix
+jira: SMILE-12000
+title: "When index_contained_resources is enabled, performing a per-resource $expunge on a resource that is referenced from a contained resource in another resource failed with HAPI-2415. The per-resource $expunge now succeeds in this scenario."
```
tadgh:
```suggestion
title: "When `index_contained_resources` is enabled, performing a per-resource $expunge on a resource that is referenced from a contained resource in another resource failed with `HAPI-2415`. The per-resource `$expunge` now succeeds in this scenario."
```


also doesn't this failure mode require recursive indexing to be enabled?

tadgh (changes requested):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8648-fix-expunge-contained-resources.yaml
another-luis: Done.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeContainedResourcesR4Test.java
another-luis: Yeah, this is waaay too long and refers to what the main test point does. This comment doesn't belong at class level - it belongs at test level in simplified form.

another-luis: Done

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/expunge/ExpungeContainedResourcesR4Test.java
another-luis: Interestingly enough, there are three different scenarios in which $expunge can cause a FK constraint violation cause by either the src or target columns in  HFJ_RES_LINK. We now address them all.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IResourceLinkDao.java
```diff
@@ -36,6 +36,16 @@ public interface IResourceLinkDao extends JpaRepository<ResourceLink, Long>, IHa
...
 
+	/**
+	 * Deletes all {@code HFJ_RES_LINK} rows where {@code thePid} is the source or target.
+	 * Queries raw scalar columns to avoid the {@code JpaPid} null-partition comparison issue
+	 * ({@code partition_id = NULL} is always false in SQL).
+	 */
+	// Generated by claude-sonnet-4-6
+	@Modifying
+	@Query("DELETE FROM ResourceLink t WHERE t.mySourceResourcePid = :pid OR t.myTargetResourcePid = :pid")
+	void deleteAllLinksByResourcePid(@Param("pid") Long thePid);
```
jamesagnew: This won't perform well on Database Performance Mode because it's including the Long part of the PID but not the partition ID

We need to be passing in a `JpaPid` and not a `Long`, so the query needs to be something like
`WHERE t.mySourceResource.myId = :pid OR t.myTargetResource.myId = :pid`

another-luis: Alright. Done.

jorgepinilla-smile:
Creating a separate pull request for release branch rel_8_12

PR #7931, opened 2026-05-19
downgrade mssql version

jdar8:
downgrade back to old version (12.4.3.jre11) to resolve a startup error


tadgh:
Formatting check succeeded!

tadgh (approved):

PR #7934, opened 2026-05-19
7932 searchparamregistry rebuilds once per custom searchparameter installed at startup

### LINKED ISSUE #7932: SearchParamRegistry rebuilds once per custom SearchParameter installed at startup
When a server starts up with an NPM package containing N custom `SearchParameter` resources, the `SearchParamRegistry` cache is rebuilt **N times** during package installation — once per SP — on top of the initial bulk load that handles the built-in SPs.

The boot-time built-in load coalesces into **a single** rebuild via `handleInit()`. The per-resource path used by `PackageInstallerSvcImpl` does not coalesce: each SP write triggers its own rebuild. There is no public API for "install all these SPs, then rebuild once."

`rebuildActiveSearchParams()` re-reads every SearchParameter in the database, re-canonicalizes, re-overrides built-ins, repopulates the phonetic encoder map, and refreshes the identity cache. Every per-SP rebuild repeats that work in full.

jmarchionatto:
## Summary

Installing an NPM package with N custom `SearchParameter` resources triggered N full rebuilds of `SearchParamRegistry` at startup — each rebuild re-reads every SP (~1325 built-ins + customs) from the DB, recanonicalises, re-applies overrides, and refreshes the identity cache. Scales linearly with package size and is highly visible on cold start (issue #7932).

This PR adds a batch API on `ISearchParamRegistry` that defers rebuilds across a unit of work, and calls it from `PackageInstallerSvcImpl` around the install loop, so a package with N SPs now costs **one** rebuild instead of N.

- New default method `ISearchParamRegistry#withDeferredRebuild(Supplier<T>)`.
- `SearchParamRegistryImpl` gates both `handleChange()` and `forceRefresh()` on a defer-depth counter; sets a dirty flag while deferred and fires one coalesced rebuild when the outermost scope exits. Nested calls compose.
- `PackageInstallerSvcImpl` wraps `installPackage(...)` + the trailing `refreshCacheIfNecessary()` in `withDeferredRebuild`.

Both gating points were needed: `handleChange` covers the resource-change-listener path; `forceRefresh` covers the `JpaResourceDaoSearchParameter.afterCommit` callback that fires per-SP create.


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7932-searchparamregistry-rebuilds-once-per-custom-searchparameter-installed-at-startup.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: perf
+issue: 7932
+title: "Previously, installing an NPM package containing N custom SearchParameter resources
+  would trigger N full rebuilds of the SearchParamRegistry at startup, each one re-reading
+  every SearchParameter from the database. The package install path now coalesces these
+  into a single rebuild for the whole package, noticeably reducing cold-start time on
+  servers configured with NPM packages that include custom SearchParameters."
```
codeforgreen: suggestion: Maybe condense this description.  _each one re-reading every SearchParameter from the database_ and _noticeably reducing cold-start time on servers configured with NPM packages that include custom SearchParameters_ are probably redundant.  They both are implied from the rest of the description I believe from mentioning `SearchParamRegistry` and `startup` respectively.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -249,16 +253,22 @@ public PackageInstallOutcomeJson install(PackageInstallationSpec theInstallation
...
+
+							if (theInstallationSpec.getInstallMode()
+									== PackageInstallationSpec.InstallModeEnum.INSTALL_ONLY) {
+								retVal.getMessage()
+										.add(
+												"Resources have been successfully installed. This is INSTALL only, so there will be no NPM packages persisted.");
+							}
+						} finally {
+							// On the exception path, so partially persisted SPs reach the
+							// registry without waiting for the next polling refresh.
```
codeforgreen: question: by partially persisted you mean installs that result in failure? and next polling refresh means next search parameter reindex job trigger (via quartz)? Not sure if this comment is necessary. I agree with this being in the finally block.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
```diff
@@ -249,16 +253,22 @@ public PackageInstallOutcomeJson install(PackageInstallationSpec theInstallation
...
-
-					if (theInstallationSpec.getInstallMode() == PackageInstallationSpec.InstallModeEnum.INSTALL_ONLY) {
-						retVal.getMessage()
-								.add(
-										"Resources have been successfully installed. This is INSTALL only, so there will be no NPM packages persisted.");
-					}
-
-					// If any SearchParameters were installed, let's load them right away
-					mySearchParamRegistryController.refreshCacheIfNecessary();
+					mySearchParamRegistry.withDeferredRebuild(() -> {
```
codeforgreen: Consider moving `withDeferredRebuild` to `ISearchParamRegistryController` instead of `ISearchParamRegistry`

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
```diff
@@ -130,6 +130,11 @@ public class SearchParamRegistryImpl
 	private volatile RuntimeSearchParamCache myActiveSearchParams;
 	private boolean myPrePopulateSearchParamIdentities = true;
 
+	// Single-caller deferred-rebuild state for the package-install path.
+	// Not thread-safe; behaviour under concurrent callers is undefined.
+	private boolean myDeferRebuild;
```
codeforgreen: Why not make these volatile to be thread-safe? If the reindex quartz job triggers during a deferRebuild then will not see the changes for these flags.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
```diff
@@ -573,17 +578,45 @@ public void requestRefresh() {
...
+	 * synchronously so initial population always completes before {@code forceRefresh}
+	 * returns.
+	 */
 	@Override
 	public void forceRefresh() {
 		RuntimeSearchParamCache activeSearchParams = myActiveSearchParams;
 		myResourceChangeListenerCache.forceRefresh();
 
-		// If the refresh didn't trigger a change, proceed with one anyway
+		if (tryDeferRebuild()) {
```
codeforgreen, line 594: Is it deliberate to make this call even when no changes occur?
Mainly referring to condition `myActiveSearchParams == activeSearchParams` below.

codeforgreen (changes requested):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7932-searchparamregistry-rebuilds-once-per-custom-searchparameter-installed-at-startup.yaml
jmarchionatto: Shortened.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
jmarchionatto: Removed.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImpl.java
jmarchionatto: Good call! Moved

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
jmarchionatto: Making the property volatile alone doesn't make this code thread-safe. A whole thread-safe approach was considered and discarded because the added complexity wasn't warranted given that the race would be harmless (only adds an extra rebuild in an unlikely case).

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
jmarchionatto: Yes, it is. Outside a deferred scope, forceRefresh() already rebuilds even when no change events were fired. This is  to preserve that contract inside a deferred scope. 
If tryDeferRebuild() only marked pending=true when events had been observed, a forceRefresh() called inside withDeferredRebuild with no pending events would silently skip the rebuild at scope exit. Marking pending unconditionally is the same tradeoff the non-scoped path already accepts: one rebuild per explicit forceRefresh() regardless of whether change events were queued.

codeforgreen: I see. It comes from the nature of the action which is a force refresh.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
codeforgreen: Can you please update this comment then? To me it sounds concerning as if in threaded environment, there would be a major issue. 
I wonder if it's worth testing to make sure no inconsistencies are caused by the multi-threaded use-case. For that case no comment would be needed. Then we prove that not making this thread safe only adds processing (some performance impact) but does not cause major issues.
Maybe an integration test which which spins up two threads, one doing package install, one performing reindex through the same method that the quartz job uses.

codeforgreen (approved):

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/registry/SearchParamRegistryImpl.java
jmarchionatto: Updated the comment and added the test, but also made the properties volatile, as there was actually a window of uncertainty without that.
Thanks for bringing this up.

PR #7948, opened 2026-05-22
Add per-step progress metrics and OpenTelemetry observability to Batch2 jobs

RohithVangalla1:
## Summary

Enhances the Batch2 job progress calculator to track per-step throughput metrics,
enabling operators to identify bottleneck steps in long-running batch jobs
(bulk export, reindex, MDM matching).

## Motivation

Addresses the existing TODO in `JobInstanceProgressCalculator.java`:
> "wipmb separate status update from stats collection in 6.8"

Currently, operators have no visibility into which step of a multi-step batch job
is the bottleneck. This change provides per-step records/sec, completion %, and
elapsed time — both in logs and via OpenTelemetry.

## Changes

- **New `StepProgressData` class** — tracks per-step: chunk counts, records processed,
  throughput/sec, elapsed time, and completion percentage
- **Enhanced `InstanceProgress`** — collects per-step metrics and exposes via
  `getStepProgressMap()` / `getStepProgress(stepId)`
- **Refactored `JobInstanceProgressCalculator`** — clearly separates:
  1. Statistics collection from work chunks
  2. Instance update with collected stats  
  3. Status transition logic
- **Enhanced `BatchJobOpenTelemetryUtils`** — new `addProgressEventToCurrentSpan()`
  emits progress events with per-step throughput summaries to distributed tracing
- **Per-step throughput** logged at DEBUG level for operational visibility
- **Unit tests** for `StepProgressData` and `InstanceProgress` step tracking

## Non-breaking

All existing APIs remain unchanged. `StepProgressData` is purely additive.
No additional database queries are introduced.


hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/InstanceProgress.java
mrdnctrk, line 88: The switch here now duplicates the categorization logic in StepProgressData.updateStatusCounts(). Since every chunk already flows through StepProgressData via updateStepProgress(), we now have two independent switches that do very similar calculation one for  a step and the other for job instance. 

Rather than maintaining these similar switces, this method could derive the instance-level totals (incomplete/complete/errored/failed counts) by summing the values from myStepProgressMap. That way the status categorization logic lives in one place (StepProgressData), and the instance totals are always consistent with the per-step data by construction.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/6424-batch2-per-step-progress-metrics.yaml
```diff
@@ -0,0 +1,10 @@
+---
```
mrdnctrk: this changelog entry should go  into 8_12 folder which is the current development version,  instead of 8_8

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_8_0/6424-batch2-per-step-progress-metrics.yaml
```diff
@@ -0,0 +1,10 @@
...
+type: perf
+issue: 6424
```
mrdnctrk: issue [6424](https://github.com/hapifhir/hapi-fhir/issues/6424), which is a closed issue, is not related to this PR. I suggest you to create a new issue for the PR , if you aren't addressing an existing one.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/JobInstanceProgressCalculator.java
```diff
@@ -103,11 +133,11 @@ public InstanceProgress calculateInstanceProgress(String instanceId) {
 
 			// global stats
 			myProgressAccumulator.addChunk(next);
-			// instance stats
+			// instance stats (includes per-step tracking)
 			instanceProgress.addChunk(next);
 		}
 
-		// wipmb separate status update from stats collection in 6.8
```
mrdnctrk, line 110: does this PR really address this comment? I see the status update statement below still in this progress calculation function so this comment should stay, and new comment should not claim "Status calculation is now a separate phase"

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/util/BatchJobOpenTelemetryUtils.java
```diff
@@ -59,4 +76,59 @@ public static void addAttributesToCurrentSpan(
...
+			attBuilder.put(OTEL_JOB_THROUGHPUT_ATT_KEY, theJobInstance.getCombinedRecordsProcessedPerSecond());
+		}
+
+		// Build per-step summary
+		Map<String, StepProgressData> stepProgressMap = theInstanceProgress.getStepProgressMap();
+		if (!stepProgressMap.isEmpty()) {
+			StringBuilder summary = new StringBuilder();
+			for (Map.Entry<String, StepProgressData> entry : stepProgressMap.entrySet()) {
+				StepProgressData stepData = entry.getValue();
+				if (summary.length() > 0) {
```
mrdnctrk: SonarQube produces a warning on this line, suggesting to use `!summary.isEmpty()`

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/util/BatchJobOpenTelemetryUtils.java
```diff
@@ -59,4 +76,59 @@ public static void addAttributesToCurrentSpan(
...
+	 * via distributed tracing systems.
+	 *
+	 * @param theJobInstance the current job instance
+	 * @param theInstanceProgress the calculated progress including per-step data
+	 */
+	public static void addProgressEventToCurrentSpan(
+			JobInstance theJobInstance,
+			InstanceProgress theInstanceProgress) {
+
+		Span currentSpan = Span.current();
```
mrdnctrk: `LocalRootSpan.current();` could be used here instead of `Span.current()`. Currently they both return the scheduled maintenance job's root span (persistence.ca.uhn.fhir.batch2.maintenance.JobMaintenanceServiceImpl$JobMaintenanceScheduledJob) but in the future if another span gets added between the maintenance jobs root span and where this function is called, Span.current would start returning that new span. So  `LocalRootSpan.current();` guarantees that the events you are adding always end up in the root span, I am assuming that is what you are aiming for.

hapi-fhir-storage-batch2/src/test/java/ca/uhn/fhir/batch2/progress/InstanceProgressStepTrackingTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+import java.util.Date;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class InstanceProgressStepTrackingTest {
```
mrdnctrk: Consider renaming `InstanceProgressStepTrackingTest` to `InstanceProgressTest` which I noticed we didn't have already. The tests you are adding is mostly focusing on step progress so maybe those could be nested test class. When later someone else wants to add more test to InstanceProgress they will have class to put those into.

hapi-fhir-storage-batch2/src/test/java/ca/uhn/fhir/batch2/progress/InstanceProgressStepTrackingTest.java
```diff
@@ -0,0 +1,116 @@
...
+		assertNotNull(progress.getStepProgress("stepA"));
+	}
+
+	@Test
+	void testStepProgressMap_isUnmodifiable() {
+		InstanceProgress progress = new InstanceProgress();
+		progress.addChunk(createChunk("stepA", WorkChunkStatusEnum.COMPLETED, 10, 1000L, 2000L));
+
+		Map<String, StepProgressData> stepMap = progress.getStepProgressMap();
+		try {
```
mrdnctrk: This would be a cleaner assertion: ```assertThatThrownBy(() -> stepMap.put("illegal", new StepProgressData("illegal")))
    .isInstanceOf(UnsupportedOperationException.class);```

mrdnctrk (changes requested):
I added some comments, overall looks like a useful addition but I have a  broader remark on the approach of adding events to spans:

1. If you're already collecting logs with OpenTelemetry, traces and logs are automatically correlated — the OTel SDK injects the active trace ID and span ID into the logging MDC/context, so every log line emitted during a span carries those IDs. Backends like Grafana, Datadog, etc. use these IDs to link logs to their parent trace/span without any extra work. So emitting the same progress data as both a log line and a span event creates duplication — the log is already linked to the span. And switching between them back and forth (from trace to logs and from logs to trace) is super easy with one click on tools like Grafana. 
2. The OpenTelemetry community is working on deprecating the span events API for this reason: https://opentelemetry.io/blog/2026/deprecating-span-events/

Not blocking on this — if people find span events useful that's fine. But worth being aware that the direction of the OTel project is moving away from span events in favor of correlated logs.

RohithVangalla1:
Hi @mrdnctrk, Thank you for the thorough review! I've addressed all comments in 72cfb36:

1. **Removed span events**,  You're right that with OTel log correlation, the per-step data logged at DEBUG level is already linked to the trace via MDC. Removed `addProgressEventToCurrentSpan` entirely and reverted `BatchJobOpenTelemetryUtils`. Thanks for the link to the OTel deprecation blog post.

2. **Kept the instance-level switch** — I considered deriving totals from `myStepProgressMap`, but the switch also captures error messages (first ERRORED, last FAILED) which aren't tracked per-step. Added a comment explaining why it's intentionally kept. Cleaned up the map operations to use `computeIfAbsent`/`merge`.

3. **Moved changelog to `8_12_0`** and updated issue to reference this PR (#7948).

4. **Restored the TODO comment** — You're right, the status update still happens in the same function. Removed the incorrect "Phase" comments.

5. **Renamed test** to `InstanceProgressTest` with `@Nested StepTracking` class.

6. **Used `assertThatThrownBy`** for the unmodifiable map assertion.

Note: I didn't apply the `LocalRootSpan.current()` or `!summary.isEmpty()` suggestions since the OTel span event code was removed entirely. Let me know if anything else needs adjustment!

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/InstanceProgress.java
RohithVangalla1: You're right, I'll refactor to derive the instance-level counts (incomplete/complete/errored/failed) by summing from myStepProgressMap, keeping the status categorization logic in one place (StepProgressData). The only piece that needs to remain at the instance level is the error message capture (first ERRORED, last FAILED), which I'll handle separately without duplicating the switch.

mrdnctrk (approved):

tadgh:
Formatting check succeeded!

jamesagnew (approved):

jamesagnew:
Hi @RohithVangalla1 - Looks like one of the tests is failing, can you have a look?

RohithVangalla1:
Hi @jamesagnew , Thansk for looking into it. Fixed — `WorkChunkMetadata.setTargetStepId()` validates that the value is not blank, so calling it with `null` throws NPE. A new `WorkChunk` already has `null` targetStepId by default, so the test now simply skips the setter call. The test still verifies that chunks without a step ID don't get tracked in the step progress map.

jamesagnew:
Thanks for the contribution! This will be in the 8.12.0 (August) release.

PR #7952, opened 2026-05-24
[7950] Fix chained searches returning empty results with reference parameter is multi-valued

### LINKED ISSUE #7950: Chained search with multi-valued reference of the same resource type produces wrong result
### Issue

Chained search with multi-valued reference of the same resource type incorrectly collapses to single HFJ_RES_LINK join, producing wrong results.

Given the following search
```
GET /Patient?general-practitioner.family=Smith&general-practitioner.gender=female
```

Storage:

- `patA` → `generalPractitioner = [Prac1 (Smith/male), Prac2 (Jones/female)]`
- `patB` → `generalPractitioner = [Prac1 (Smith/male)]`

Expected: `patA` returned (Prac1 satisfies `family=Smith`, Prac2 satisfies `gender=female`). 
Actual: no results — the collapsed single-join SQL requires one practitioner to be both Smith AND female.

YalingPeiS:
### Issue
chained searches returning empty results when the reference parameter is muti-valued and the search contains multiple conditions on the same target resource type.

### Cause
The join-reuse optimization introduced in PR #7720 added a caching mechanism in `QueryStack.createPredicateReference` that reused a single `HFJ_RES_LINK` join across multiple AND-clauses for chained search when the search parameters had the same name and were on the same target resource type. It, however, caused empty search result in the case when the chained search has multiple references of the same target resource type. For example, when a patient has two Practitioner references, the following search
```
GET /Patient?general-practitioner.family=Apple&general-practitioner.family=Orange
```
would return the empty result instead of the patient because the cache key is the same for both condition, resulting  in a single `HFJ_RES_LINK` join in the search query.

### Fix
Instead of building a cache key, the fix adds logic to check the reference parameter's cardinality. The join-reuse optimization (single HFJ_RES_LINK join) applies only to the single-valued reference with cardinality <= 1. For multi-valued reference, each value binds to its own HFJ_RES_LINK row during chained searches to ensure correct search results.

Closes #7950


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/ChainingR4SearchTest.java
```diff
@@ -180,6 +180,34 @@ void testMixedTypeOrGroupAndClause_UseSeparateJoinsRegardlessOfElementOrder() {
...
+		myCaptureQueriesListener.clear();
+		IBundleProvider results = myPatientDao.search(map, newSrd());
+
+		// Only patA is returned — each filter satisfied independently by a different practitioner
+		assertThat(toUnqualifiedVersionlessIdValues(results)).containsExactlyInAnyOrder("Patient/PatA");
+
+		// 2 HFJ_RES_LINK joins required: multi-valued reference cannot collapse to a single join
+		List<SqlQuery> selectQueries = myCaptureQueriesListener.getSelectQueriesForCurrentThread();
+		String querySql = selectQueries.get(0).getSql(false, false);
+		assertThat(StringUtils.countMatches(querySql, "HFJ_RES_LINK")).isEqualTo(2);
```
jamesagnew: Might be nice to extend the search term from `"HFJ_RES_LINK"` to `"JOIN HFJ_RES_LINK"` to ensure it's there twice to support two joins

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -1469,25 +1470,19 @@ public Condition createPredicateReference(
...
-					.map(p -> buildReferenceCacheQualifier((ReferenceParam) p))
-					.filter(StringUtils::isNotBlank)
-					.distinct()
-					.sorted()
-					.collect(Collectors.joining("|"));
-			if (isNotBlank(typeQualifier)) {
-				cacheParamName = theParamName + ":" + typeQualifier;
+			ResourceLinkPredicateBuilder predicateBuilder;
+			if (isReferenceSingleValued(theResourceName, theSearchParam)) {
+				// Cardinality 0..1 reference — chained AND-clauses can share a single HFJ_RES_LINK join.
```
jamesagnew, line 1475: This case doesn't appear to have a test case in ChainingR4SearchTest - would be good to add one and verify that only one JOIN is used in that case

hapi-fhir-storage/src/test/java/ca/uhn/fhir/jpa/util/FhirPathUtilTest.java
```diff
@@ -20,4 +25,68 @@ public void cleansePath_errorCases_throws(String theStr) {
 			assertTrue(ex.getLocalizedMessage().contains("Cannot cleanse path"));
 		}
 	}
+
+	@Test
```
jamesagnew, line 29: Thought: Should probably add a test for functions which appear within the expression. E.g. Observation uses this one: `Observation.value.ofType(CodeableConcept).text`

jamesagnew (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
YalingPeiS: Updated.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/ChainingR4SearchTest.java
YalingPeiS: The sql search query for multiple joins looks like the following 
```
SELECT t0.SRC_RESOURCE_ID FROM HFJ_RES_LINK t0 
INNER JOIN HFJ_SPIDX_STRING t1 ON (t0.TARGET_RESOURCE_ID = t1.RES_ID) 
INNER JOIN HFJ_RES_LINK t2 ON (t0.SRC_RESOURCE_ID = t2.SRC_RESOURCE_ID) 
INNER JOIN HFJ_SPIDX_STRING t3 ON (t2.TARGET_RESOURCE_ID = t3.RES_ID) 
WHERE ...
```
where the 1st `HFJ_RES_LINK` is in the FROM clause and the subsequent ones are in JOIN clauses. With single-valued cardinality, only one `FROM HFJ_RES_LINK` is in the query and there won't be a `JOIN HFJ_RES_LINK`.

hapi-fhir-storage/src/test/java/ca/uhn/fhir/jpa/util/FhirPathUtilTest.java
YalingPeiS: Updated

PR #7955, opened 2026-05-25
Include additional resource types in the package install defaults.

### LINKED ISSUE #7954: Enable support for adding additional resource folders and document included resource types

codeforgreen:
Fixes #7954

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7954-expand-default-package-install-types.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/7954-expand-default-package-install-types.yaml
@@ -1,0 +1,8 @@
+---
+type: add
+issue: 7954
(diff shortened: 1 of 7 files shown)


tadgh:
Formatting check succeeded!

jkiddo:
@codeforgreen you sure you want `OperationDefinition` resources installed?

codeforgreen:
> @codeforgreen you sure you want `OperationDefinition` resources installed?

@jkiddo Do you see any issue with that?

jkiddo:
storing an OperationDefinition may end up leave the wrong impression as the logic that is behind it does not follow along

codeforgreen:
> storing an OperationDefinition may end up leave the wrong impression as the logic that is behind it does not follow along

Thanks for the feedback. I will revisit this. It is true that if it talks about an operation that the server does not implement or does not implement exactly as specified, it would be inaccurate.

codeforgreen:
Decided to cancel this PR and document how to add the additional resource types as a parameter. So I will raise a documentation only PR instead for this.

PR #7956, opened 2026-05-25
Add raw post ability to fhir client

jamesagnew:
The FHIR GenericClient `IRawHttp` interface now has an additional method `post()` which can be used to send a raw (potentially non-FHIR payload) HTTP POST through the client, leveraging interceptors and other functionality from the client.


tadgh:
Formatting check succeeded!

tadgh (approved):

PR #7958, opened 2026-05-26
Add unqualified path name support to FhirPatch

jamesagnew:
The FHIR Patch engine now supports untyped path expressions (e.g. `code` instead of the typed equivalent `Observation.code`). Previously only typed expressions were supported.

### DIFF
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java
@@ -163,6 +163,8 @@ public class FhirPatch {
 	private void handleAddOperation(@Nullable IBaseResource theResource, IBase theParameters, PatchOutcome theOutcome) {
 
 		String path = ParametersUtil.getParameterPartValueAsString(myContext, theParameters, PARAMETER_PATH);
+		path = defaultString(path);
+		path = prefixResourceTypeToPathIfneeded(path, theResource);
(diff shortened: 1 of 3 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java
```diff
@@ -347,6 +352,13 @@ private void handleReplaceOperation(
 		replaceValuesByPath(cd, theParameters, fhirPath, parsedFhirPath, theOutcome);
 	}
 
+	private String prefixResourceTypeToPathIfneeded(String thePath, IBaseResource theResource) {
+		if (isNotBlank(thePath) && Character.isLowerCase(thePath.charAt(0))) {
+			return myContext.getResourceType(theResource) + "." + thePath;
+		}
+		return thePath;
+	}
```
tadgh, line 360: How do we handle paths that start with a `.`? I imagine that `code` and `.code` are semantically the same for untyped paths, and if we allow `code` we should probably also allow `.code`, as that particular syntax is probably how I would write it, if I knew nothing about how fhirpaths were going to be evaluated.

tadgh (approved):

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/patch/FhirPatch.java
jamesagnew: FHIRPath expressions aren't allowed to start with a dot, so we'd error out as always

PR #7961, opened 2026-05-26
Improve loinc valueset fetching and expanding logic

jamesagnew:
When fetching and expanding valuesets, the system will try to use the marked current version of the codesystem when possible. Additionally, offset and count parameters will be honoured for expansions even if hibernate search is not enabled.


tadgh:
Formatting check succeeded!

PR #7965, opened 2026-05-27
7945 replace deprecated default partition selection

### LINKED ISSUE #7945: Replace deprecated default partition selection
RequestPartitionId exposes three methods named defaultPartition() with varying signatures. Two of them are deprecated. The deprecated signatures have, in total, nearly 100 calls throughout the code base. Some of these calls are in tests, and potentially benign, but many of them are in production code and represent potential bugs.
All uses of the deprecated methods should be rewritten.

The deprecated methods have been deprecated since 2025.08 release, and are candidates to be removed.

- HAPI FHIR Version
8.10.0

jmarchionatto:


tadgh:
Formatting check succeeded!

jmarchionatto:
## Summary

This branch removes the long-deprecated default-partition resolution API whose central flaw was treating a `null` partition ID as "the default partition." That assumption silently ignored any non-`null` configured default partition ID, so deployments that configured a real default partition were not honored at a number of call sites. The work splits into two distinct kinds of edits that reviewers must keep separate: call sites that *should* honor the configured default now route through `PartitionSettings.getDefaultRequestPartitionId()`, while call sites whose semantics must stay literally "null partition" were migrated to the behavior-preserving `RequestPartitionId.fromPartitionId(null)`.

1. **Removed deprecated API.** `RequestPartitionId.defaultPartition()`, `RequestPartitionId.defaultPartition(LocalDate)`, and the single-argument `PartitionablePartitionId.toRequestPartitionId(PartitionablePartitionId)` overload are deleted. Callers must now supply an `IDefaultPartitionSettings` (implemented by `PartitionSettings`).
2. **Behavior fix at real call sites.** Search-param matching (`InMemoryResourceMatcher` / `ResourceIndexedSearchParams`), terminology mapping translation (`TermConceptClientMappingSvcImpl`), bulk import (`ConsumeFilesStepV1/V2`), `SearchBuilder`, and `BaseHapiFhirResourceDao` now resolve the configured default partition. Several services gained an `@Autowired PartitionSettings`, and `matchParam`/`isMatchSearchParam` gained `PartitionSettings`-aware overloads (old overloads deprecated for removal).
3. **Subscription messaging no longer pre-seeds a null-partition sentinel.** `ResourceDeliveryMessage` and `ResourceModifiedMessage` constructors no longer set `myPartitionId` to the old sentinel; the field is left `null` until a producer/deserializer populates it, and `getPartitionId()` is now `@Nullable` (GL-8692). A `null` partition is resolved to the configured default downstream.
4. **Transaction-service default-partition helpers.** Added `HapiTransactionService.withSystemRequestOnDefaultPartition()` (override) and the instance method `executeWithConfiguredDefaultPartitionInContext(...)`, and deprecated the static `executeWithDefaultPartitionInContext(...)` for removal (its behavior is unchanged this release).
5. **Behavior-preserving swaps where null really means null.** Migration hash recalculators in `HapiFhirJpaMigrationTasks`, the deprecated `IResourceChangeListenerRegistry` overload, `JpaSearchParamCache`, and the deprecated transaction helper now use `RequestPartitionId.fromPartitionId(null)` to keep identical behavior rather than adopting the configured default.

## Code Review Suggestions

- [ ] **Verify each migrated call site got the *correct* replacement.** The crux of this MR is the choice between `getDefaultRequestPartitionId()` (honor configured default — a behavior change) versus `fromPartitionId(null)` (preserve literal-null semantics). Confirm no call site that should preserve null-partition behavior was switched to the configured default, and vice versa.
- [ ] **Migration tasks MUST stay null-partition.** `HapiFhirJpaMigrationTasks` hash recalculators were changed to `fromPartitionId(null)`. These compute `HASH_IDENTITY`/`HASH_SYS`/etc. for already-stored rows; adopting a non-null configured default here would recompute different hashes and corrupt existing indexes. Confirm these were intentionally left as null-partition (note: this file is migration code — review-only, not modifiable by the agent).
- [ ] **New `@Autowired PartitionSettings` fields resolve in every context.** `InMemoryResourceMatcher`, `TermConceptClientMappingSvcImpl`, and `ConsumeFilesStepV1/V2` now inject `PartitionSettings`. Confirm a `PartitionSettings` bean is present in every Spring context that instantiates these (e.g. subscription matching, terminology, batch2 import) — a missing bean would surface as a startup failure, but partial/standalone contexts are the risk.
- [ ] **`ResourceModifiedMessage`/`ResourceDeliveryMessage` now leave partition `null`.** Any consumer that previously relied on `getPartitionId()` being non-null after construction could now NPE or behave differently. Check serialized/deserialized message paths and any code that reads the partition before a producer sets it.
- [ ] **Deprecated `matchParam`/`isMatchSearchParam` overloads build a fresh `new PartitionSettings()`.** When `isIndexStorageOptimized()` is enabled and a deployment has a non-null configured default partition, the deprecated overload computes a hash against a fresh, unconfigured default. Confirm all internal callers were moved to the `PartitionSettings`-aware overload so the deprecated path is external-only.
- [ ] **`NonTransactionalHapiTransactionService` default constructor sets `new PartitionSettings()`.** This guards against NPE when instantiated without Spring (tests). Confirm this class is genuinely test/standalone-only and that the comment's claim "field injection still overrides this value" holds when it *is* wired as a bean.
- [ ] **`SubscriptionTopicRegisteringListener` default-partition reset.** The corrupt-default-partition cleanup now uses `RequestPartitionId.fromPartitionId(getDefaultPartitionId())` with a null-safe `getDefaultPartitionId()`. Confirm this still repairs the "list containing null values" case created by the package installer's STORE_AND_INSTALL path under partitioning.

tadgh (approved):

PR #8016, opened 2026-05-28
Fix package install overwriteContentNotPresentCodeSystems not propagating to dependencies in async path

### LINKED ISSUE #7968: Async package install does not propagate `overwriteContentNotPresentCodeSystems` to dependency packages
**Description:** 

When installing a package asynchronously (batch2 path) with `overwriteContentNotPresentCodeSystems=true` and `fetchDependencies=true`, the flag is not carried over to dependency package installations. The `PackageInstallationSpec` copy constructor in `InitializeDependenciesStep.buildStartRequest()` does not copy `myOverwriteContentNotPresentCodeSystems`, so it silently resets to false for all dependency installs.

**Impact**
A user who sets `overwriteContentNotPresentCodeSystems=true` expecting dependency `CodeSystem` resources to overwrite existing `content=not-present` `CodeSystem` will find that only the top-level package respects the flag. Dependencies will skip those overwrites, leaving stale `content=not-present` `CodeSystem` resources in place. This only affects the async (`batch2`) install path — the synchronous path passes the same spec object and is unaffected.

**Fix:** 
Replace the manual copy constructor with a copyOf() static factory method that uses JSON serialize/deserialize, eliminating the need to maintain field-by-field copying. 

**Side fix** — add missing error codes:
The following error messages in InitializeDependenciesStep are missing Msg.code() prefixes:
  - "Failed to launch child job for dependency package "
  - "Failed to process dependencies for package "
  - "Version conflict error "

codeforgreen:
Fixes #7968

### DIFF
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallationSpec.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallationSpec.java
@@ -20,6 +20,7 @@
 package ca.uhn.fhir.jpa.packages;
 
 import ca.uhn.fhir.model.api.annotation.ExampleSupplier;
+import ca.uhn.fhir.util.JsonUtil;
 import com.fasterxml.jackson.annotation.JsonAutoDetect;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonInclude;
@@ -28,7 +29,6 @@ import com.fasterxml.jackson.annotation.JsonPropertyOrder;
 import io.swagger.v3.oas.annotations.media.Schema;
 
 import java.util.ArrayList;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Supplier;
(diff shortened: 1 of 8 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/packages/PackageInstallationSpec.java
```diff
@@ -129,27 +129,9 @@ public class PackageInstallationSpec {
...
-			myDependencyExcludes = new ArrayList<>(theOriginalSpec.myDependencyExcludes);
-		}
-		if (theOriginalSpec.myAdditionalResourceFolders != null) {
-			myAdditionalResourceFolders = new HashSet<>(theOriginalSpec.myAdditionalResourceFolders);
-		}
-
-		// we don't copy myPackageContents because it is transient anyway
+	public static PackageInstallationSpec copyOf(PackageInstallationSpec theOriginalSpec) {
+		String json = JsonUtil.serialize(theOriginalSpec);
+		return JsonUtil.deserialize(json, PackageInstallationSpec.class);
```
JasonRoberts-smile, line 134: Very nice. Future-proof if we need to add more fields

JasonRoberts-smile (approved):
Looks good to me, but still needs a code owner review

tadgh (approved):

PR #8020, opened 2026-05-29
Optimize MSSQL resource fetches in DBPM

jamesagnew:
PR #7879 introduced an optimized SQL when fetching batches of resources on MSSQL in Database Partition Mode.

It turns out there is a second separate code path we also needed to optimize, in `BaseHapiFhirResourceDao#searchForResourceIds()`.

This PR makes the same fix for that second code path.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
@@ -2581,6 +2581,10 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 					theRequestPartitionId.getPartitionIds().get(0)));
 		}
 
+		/*
+		 * Partitioned PIDs have to parts (partition ID and PID), so we can only have half
+		 * as many parameters without exceeding the max parameters.
+		 */
 		int chunkSize = SearchBuilder.getMaximumPageSize();
 		if (myPartitionSettings.isPartitioningEnabled()) {
 			chunkSize = chunkSize / 2;
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
@@ -2581,16 +2581,12 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 					theRequestPartitionId.getPartitionIds().get(0)));
 		}
 
-
 		int chunkSize = SearchBuilder.getMaximumPageSize();
 		if (myPartitionSettings.isPartitioningEnabled()) {
 			chunkSize = chunkSize / 2;
 		}
 
-		return QueryChunker
-			.chunk(pidStream, chunkSize)
-			.flatMap(this::pidListToVersionedIdList);
-
+		return QueryChunker.chunk(pidStream, chunkSize).flatMap(this::pidListToVersionedIdList);
 	}
 
 	/**
@@ -2610,9 +2606,9 @@ public abstract class BaseHapiFhirResourceDao<T extends IBaseResource> extends B
 		if (myPartitionSettings.isDatabasePartitionMode() && myDialectSvc.isMssql()) {
 			Multimap<Integer, Long> partitionIdToPid = splitPidListByPartitionId(pidsNormalized);
 			List<Predicate> partitionAndPidPredicates =
-				new ArrayList<>(partitionIdToPid.keySet().size());
+					new ArrayList<>(partitionIdToPid.keySet().size());
 			for (Map.Entry<Integer, Collection<Long>> entry :
-				partitionIdToPid.asMap().entrySet()) {
+					partitionIdToPid.asMap().entrySet()) {
 				Integer partitionId = entry.getKey();
 				Collection<Long> pidsForPartitionId = entry.getValue();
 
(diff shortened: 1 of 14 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -2577,10 +2581,54 @@ private Stream<IIdType> pidsToIds(
...
 		}
 
-		return new QueryChunker<>()
-				.chunk(pidStream, SearchBuilder.getMaximumPageSize())
-				.flatMap(ids -> myResourceTableDao.findAllById(ids).stream())
-				.map(ResourceTable::getIdDt);
+		int chunkSize = SearchBuilder.getMaximumPageSize();
+		if (myPartitionSettings.isPartitioningEnabled()) {
+			chunkSize = chunkSize / 2;
+		}
```
tadgh, line 2591: I assume this is a shim with a future plan that isn't just `n/`2?

tadgh (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -2577,10 +2581,54 @@ private Stream<IIdType> pidsToIds(
...
+			wherePredicate = cb.or(partitionAndPidPredicates.toArray(EMPTY_PREDICATE_ARRAY));
+		} else {
+			wherePredicate = from.get("myPid").in(pidsNormalized);
+		}
+
+		cq.where(wherePredicate);
+
+		return myEntityManager.createQuery(cq).getResultStream().map(t -> {
+			String resourceType = t.get(0, String.class);
+			String fhirId = t.get(1, String.class);
```
jmarchionatto: Persisted resources should have RES_FHIR_ID populated, so this may be minor, but the old pidsToIds produced IDs via ResourceTable::getIdDt(), which falls back to the numeric resource id when myFhirId is null/empty. The new code unconditionally selects and uses the myFhirId column.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -2577,10 +2581,54 @@ private Stream<IIdType> pidsToIds(
 					theRequestPartitionId.getPartitionIds().get(0)));
 		}
 
-		return new QueryChunker<>()
-				.chunk(pidStream, SearchBuilder.getMaximumPageSize())
-				.flatMap(ids -> myResourceTableDao.findAllById(ids).stream())
-				.map(ResourceTable::getIdDt);
+		int chunkSize = SearchBuilder.getMaximumPageSize();
+		if (myPartitionSettings.isPartitioningEnabled()) {
+			chunkSize = chunkSize / 2;
```
jmarchionatto, line 2590: No comment explains *why* the chunk is halved (presumably the tuple `(partition,res_id)` IN clause doubles the bound-parameter count, so halving keeps it under the MSSQL ~2100 param ceiling). A magic `/2` on a hot path needs a one-line rationale.  

The halving is gated on the broad isPartitioningEnabled(), while the tuple-shaped query that motivates the halving is gated on the narrower isDatabasePartitionMode() (pidListToVersionedIdList). For partitioning that is not DBPM the halving is applied without the corresponding 2-column IN clause. If that's intentional/harmless, a comment would apply; otherwise align the two on isDatabasePartitionMode().

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/InClauseNormalizer.java
```diff
@@ -26,6 +26,8 @@
 import java.util.Collections;
 import java.util.List;
 
+import static ca.uhn.fhir.jpa.search.builder.SearchBuilder.getMaximumPageSize;
```
jmarchionatto: `InClauseNormalizer` (a generic util in `ca.uhn.fhir.jpa.util`) now statically depends on `SearchBuilder`, while `SearchBuilder` already statically depends on `InClauseNormalizer.normalizeIdListForInClause`. That's a class-level cycle and an inverted dependency (util depending on the search engine). The underlying value already lives in a neutral place — `SearchConstants.MAX_PAGE_SIZE` (hapi-fhir-storage). The only reason to route through `SearchBuilder` is its test-override hook (`myMaxPageSizeForTests`). Consider relocating the page-size accessor (with the test override) to a neutral class so the util doesn't reach into the search builder.

jmarchionatto (approved):
Looks good. Approved with a few comments.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
jamesagnew: I'll add a comment - The point here is that partitioned PIDs have to parts (partition ID and PID), so we can only have half as many parameters without exceeding the max parameters

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
jamesagnew: I've added a comment explaining - this applies to non-MSSQL as well, where we're using a normal partitionID/PID tuple, so every PID uses 2 parameters.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/util/InClauseNormalizer.java
jamesagnew: Makes sense - I'll rework this.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
jamesagnew: I don't know if it's possible for there to be any instances at this point with null FHIR_ID, but I'll add a fallback just in case..

PR #8026, opened 2026-06-01
Resolve "Unqualified _id chained search on multi-target reference parameter silently returns zero results"

### LINKED ISSUE #8025: Unqualified _id chained search on multi-target reference parameter silently returns zero results
**Summary**

When a chained _id search is issued against a reference-type SearchParameter that declares multiple target[] resource types, and the chain is unqualified (no :
qualifier), the search returns zero results even when matching resources exist. Adding the type qualifier restores correct behaviour.
Root cause is in HAPI's ResourceIdPredicateBuilder.createPredicateResourceId — it calls SearchQueryBuilder.setMatchNothing() when it cannot resolve the ID for a given target type,
but it is invoked inside an OR loop where other target types may still resolve. setMatchNothing() is a global, sticky kill-switch on the shared SearchQueryBuilder, so once any
iteration of the multi-target loop fails to find a matching ID, the entire outer query is short-circuited to return no results — discarding the work the later iterations did to build correct subqueries.

**Affected versions**

- Bug is present in HAPI master at time of writing — not yet fixed in a later release.

**Reproduction (FHIR R4)**

Minimal setup:

PUT /ServiceRequest/sr-1
{ "resourceType": "ServiceRequest", "id": "sr-1", "status": "active", "intent": "order", "subject": { "reference": "Patient/p-1" } }
PUT /DiagnosticReport/dr-1
{ "resourceType": "DiagnosticReport", "id": "dr-1", "status": "final", "code": { "text": "demo" },
"basedOn": [ { "reference": "ServiceRequest/sr-1" } ] }

Two searches:

Image

PUT /DiagnosticReport/dr-1
{ "resourceType": "DiagnosticReport", "id": "dr-1", "status": "final", "code": { "text": "demo" },
"basedOn": [ { "reference": "ServiceRequest/sr-1" } ] }

**Two searches:**

Image

The SearchParameter http://hl7.org/fhir/SearchParameter/DiagnosticReport-based-on declares target = [CarePlan, MedicationRequest, NutritionOrder, ServiceRequest, ImmunizationRecommendation].

**Root cause**

The flow in ResourceLinkPredicateBuilder.addPredicateReferenceWithChain (hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceLinkPredicateBuilder.java):

- Line 594-595: determineCandidateResourceTypesForChain returns all SP target types (here: 5 types).
- Line 622: a child QueryStack is built via myQueryStack.newChildQueryFactoryWithFullBuilderReuse(). Per QueryStack.newChildQueryFactoryWithFullBuilderReuse() (QueryStack.java:2371-2381) this child shares mySqlBuilder with the parent.
- Lines 642-707: the loop iterates each target type, calling childQueryFactory.searchForIdsWithAndOr(...) with paramName="_id", resourceName=, value sr-1. The resulting per-type condition is collected into orPredicates and OR'd at line 728.
- For chain on _id, searchForIdsWithAndOr (QueryStack.java:2384-2392) routes to ResourceIdPredicateBuilder.createPredicateResourceId.

Inside createPredicateResourceId (hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/predicate/ResourceIdPredicateBuilder.java):
- Lines 84-92: the raw value sr-1 is prefixed with the iteration's resource name → CarePlan/sr-1, MedicationRequest/sr-1, … one per iteration.
- Line 106-112: myIdHelperService.resolveResourceIdentities(...) is invoked. For any target type that has no resource with that FHIR id (i.e. all four iterations besides ServiceRequest), resolvedPids is empty, orPids is empty, allOrPids is set to the empty set.
- Line 123-125:
if (allOrPids != null && allOrPids.isEmpty()) {
setMatchNothing();
}
- setMatchNothing() flips myMatchNothing = true on the shared SearchQueryBuilder (SearchQueryBuilder.java:843-844).
- The flag is consulted at SQL generation (SearchQueryBuilder.java:598): new GeneratedSql(myMatchNothing, sql, bindVariables). The result executor checks isMatchNothing() and returns an empty bundle without executing the SQL.

setMatchNothing() is correct semantics when the predicate it represents is in an AND position at the top of the query — an empty IN clause genuinely means no row can match. It is incorrect when the predicate is one branch of an OR, because other branches may still match. The chain iteration in ResourceLinkPredicateBuilder collects per-type predicates into orPredicates (line 706) and OR's them (line 728), so each
per-type call to createPredicateResourceId is by construction in an OR position. The shared SearchQueryBuilder is the contamination vector — ResourceIdPredicateBuilder has no way to know it is being called as a child predicate.

**Why the qualified form works**

When the chain is qualified (based-on:ServiceRequest._id=sr-1), determineCandidateResourceTypesForChain (ResourceLinkPredicateBuilder.java:792-801) returns a single-element list containing only ServiceRequest. The loop runs once, resolveResourceIdentities returns a non-empty allOrPids, setMatchNothing() is never called, and the subquery is built correctly.
Impact / scope
Any unqualified chained _id search against a multi-target reference SearchParameter where the bare id is unique to one (or some subset) of the target types — i.e. the realistic case for _id, since FHIR ids are per-type. This includes common parameters:

- DiagnosticReport.based-on (5 targets)
- Observation.subject (Patient, Group, Device, Location — 4 targets)
- Observation.based-on (multiple)
- Task.focus (any resource)
- Communication.based-on, Provenance.target, ...

Symptom is silent: the user gets HTTP 200 with total: 0. No log warning. No OperationOutcome. The user has to discover the qualifier workaround empirically.

**Suggested fix**

Make ResourceIdPredicateBuilder.createPredicateResourceId join-context-aware:

```
if (allOrPids != null && allOrPids.isEmpty()) {
if (theSourceJoinColumn != null) {
// Child predicate: we're an OR branch — return an always-false
// condition for this iteration without poisoning the shared builder.
return ConstantCondition.FALSE;   // or equivalent "1=0" condition
}
setMatchNothing();
return null;
}
```

**Rationale:** theSourceJoinColumn != null is exactly the signal that this call is being made as a child join (line 149 already branches on this for the non-empty case to produce an IN predicate against the join column). The two empty/non-empty cases should be symmetric in their treatment of context.
Alternative shape: stop sharing mySqlBuilder across child query factories used for chain iteration, so each per-type iteration has its own myMatchNothing flag. This is a bigger change and probably the wrong choice — setMatchNothing() short-circuiting genuinely empty top-level queries is a valuable optimisation we want to keep.

**Test coverage**

Add tests in FhirResourceDaoR4SearchChainedTest (or equivalent):

- test_unqualifiedChainedIdOnMultiTargetReference_returnsMatchingResource
- test_unqualifiedChainedIdOnMultiTargetReference_returnsEmptyWhenIdMatchesNoTargetType
- test_unqualifiedChainedIdOnMultiTargetReference_returnsAllWhenIdAmbiguousAcrossTypes

The third case is interesting and a useful spec sanity-check: if both ServiceRequest/x and CarePlan/x exist (same FHIR id, different types), the unqualified chain should return all DiagnosticReports referencing either. The current code can't even reach that question.

**Workaround**

Always qualify multi-target _id chains:

`DiagnosticReport?based-on:ServiceRequest._id=`

KevinDougan:
See HAPI-FHIR Issue #8025 for a detailed description of the purpose of this Pull Request.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8025-fix-unqualified-id-chained-searches-on-multi-target-reference-parameter.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8025-fix-unqualified-id-chained-searches-on-multi-target-reference-parameter.yaml
@@ -1,7 +1,7 @@
 ---
 type: fix
 issue: 8025
-title: "Fixed: Unqualified _id chained searches on a multi-target reference parameter 
+title: "Fixed:  Unqualified _id chained searches on a multi-target reference parameter 
    (e.g. DiagnosticReport?based-on._id=sr-1) silently returned zero results when the SearchParameter declared 
    more than one target type. The per-target-type subqueries were OR'd together on a shared SearchQueryBuilder, 
    and the first iteration whose target type didn't contain the id called setMatchNothing(), short-circuiting 
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8025-fix-unqualified-id-chained-searches-on-multi-target-reference-parameter.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8025-fix-unqualified-id-chained-searches-on-multi-target-reference-parameter.yaml
@@ -1,0 +1,9 @@
+---
+type: fix
+issue: 8025
+title: "Fixed: Unqualified _id chained searches on a multi-target reference parameter 
+   (e.g. DiagnosticReport?based-on._id=sr-1) silently returned zero results when the SearchParameter declared 
+   more than one target type. The per-target-type subqueries were OR'd together on a shared SearchQueryBuilder, 
+   and the first iteration whose target type didn't contain the id called setMatchNothing(), short-circuiting 
+   the entire query — even when a sibling iteration would have matched. Empty per-branch lookups now emit 
+   a branch-local always-false condition instead, leaving the surrounding OR intact."
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8025-fix-unqualified-id-chained-searches-on-multi-target-reference-parameter.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8025-fix-unqualified-id-chained-searches-on-multi-target-reference-parameter.yaml
@@ -3,7 +3,4 @@ type: fix
 issue: 8025
 title: "Fixed:  Unqualified _id chained searches on a multi-target reference parameter 
    (e.g. DiagnosticReport?based-on._id=sr-1) silently returned zero results when the SearchParameter declared 
-   more than one target type. The per-target-type subqueries were OR'd together on a shared SearchQueryBuilder, 
-   and the first iteration whose target type didn't contain the id called setMatchNothing(), short-circuiting 
-   the entire query — even when a sibling iteration would have matched. Empty per-branch lookups now emit 
-   a branch-local always-false condition instead, leaving the surrounding OR intact."
+   more than one target type."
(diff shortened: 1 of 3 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8025-fix-unqualified-id-chained-searches-on-multi-target-reference-parameter.yaml
```diff
@@ -0,0 +1,9 @@
+---
+type: fix
+issue: 8025
+title: "Fixed:  Unqualified _id chained searches on a multi-target reference parameter 
+   (e.g. DiagnosticReport?based-on._id=sr-1) silently returned zero results when the SearchParameter declared 
+   more than one target type. The per-target-type subqueries were OR'd together on a shared SearchQueryBuilder,
```
jamesagnew: I'd recommend dropping the remaining text after "target type." That's too much detail for a changelog. :)

jamesagnew (approved):

PR #8028, opened 2026-06-01
feat: add baseline version option for existing schemas without migration history

### LINKED ISSUE #8027: Database migration can silently skip required tasks on existing schema without migration history (e.g. missing BT2_WORK_CHUNK.LAST_HEARTBEAT after 8.4.0 -> 8.10.0)
## Summary

When migrating an existing HAPI FHIR database (originally used by a running 8.4.0 server) to 8.10.0 with the CLI, the migration may complete “successfully” but still leave required columns missing.
At runtime, this causes SQL errors such as:
ERROR: column bwce1_0.last_heartbeat does not exist

## Environment

- HAPI FHIR CLI: 8.10.0
- Source DB: existing PostgreSQL schema from a running 8.4.0 server
- Driver: POSTGRES_9_4
- JDBC: jdbc:postgresql://localhost:5432/<db>

## Steps to Reproduce
1. Start from an existing DB schema previously used by HAPI FHIR 8.4.0.
2. Run CLI migration to 8.10.0.
3. Start server and trigger batch2/chunk processing queries.
4. Observe runtime SQL error for missing BT2_WORK_CHUNK.LAST_HEARTBEAT.

## Actual Behavior

- Migration reports success.
- FLY_HFJ_MIGRATION contains entries marked successful.
- Required schema change (LAST_HEARTBEAT on BT2_WORK_CHUNK) may still be missing.
- Server fails later with SQLGrammarException.

## Expected Behavior

- Migration must not silently skip required tasks on existing schemas.
- If schema exists but migration history is missing/incomplete, migration should fail fast with a clear message requiring an explicit baseline.
- Required columns/tables for target version must be guaranteed after migration.

## Root Cause (observed)
InitializeSchemaTask can mark schema as initialized when an indicator table exists, which leads to many tasks being skipped as “schema initialization mode”, even on non-fresh databases.

### Suggested Fix
1. Do not set “schema initialized” just because indicator tables already exist.
2. Add a guard: if existing schema is detected but no reliable migration history exists, abort with a clear error and require --baseline-version.
3. Record baselined tasks explicitly (e.g. result BASELINED) and only execute tasks newer than baseline.

patrick-werner:
This pull request addresses a critical issue with schema migrations in HAPI FHIR, ensuring that migrations on existing databases without migration history do not skip required tasks. It introduces a new `--baseline-version` option to explicitly record the current schema version for such cases, preventing missing schema elements and runtime errors. The implementation includes new logic in the migrator to enforce the presence of this option when necessary, enhanced error handling, and comprehensive tests to validate the new behavior.

**Migration safety and baseline enforcement:**
- Added a `--baseline-version` option to the CLI (`BaseFlywayMigrateDatabaseCommand`) to allow users to specify the current schema version when migration history is missing, ensuring all required migrations are tracked and applied correctly. [[1]](diffhunk://#diff-3d44ef245283713e78bd971cc6fd672c27fe97261eb2ded0a5e57bf49bd18e67R45) [[2]](diffhunk://#diff-3d44ef245283713e78bd971cc6fd672c27fe97261eb2ded0a5e57bf49bd18e67R102-R107) [[3]](diffhunk://#diff-3d44ef245283713e78bd971cc6fd672c27fe97261eb2ded0a5e57bf49bd18e67R156-R157)
- Updated `HapiMigrator` to detect existing schemas without migration history and require `--baseline-version`, throwing a clear error if not provided. When specified, the migrator records all tasks up to the baseline as already applied. [[1]](diffhunk://#diff-3f2a9b5c8143ebf895056d5e7952ff25d1aa27eb5cb4f54ed3290dac62617bb7R57) [[2]](diffhunk://#diff-3f2a9b5c8143ebf895056d5e7952ff25d1aa27eb5cb4f54ed3290dac62617bb7R111-R114) [[3]](diffhunk://#diff-3f2a9b5c8143ebf895056d5e7952ff25d1aa27eb5cb4f54ed3290dac62617bb7L152-L160) [[4]](diffhunk://#diff-3f2a9b5c8143ebf895056d5e7952ff25d1aa27eb5cb4f54ed3290dac62617bb7R224-R284)

**Migration tracking and storage:**
- Enhanced migration result tracking by introducing a new `BASELINED` result type and ensuring that baseline-applied migrations are stored distinctly. [[1]](diffhunk://#diff-b466472c385d5b198aeea411b2a54566a4768c6166768b5fc2c22bb21f8e4982R47-R51) [[2]](diffhunk://#diff-d7fc11150cc29d846a8be932c7cf9e6515b73952bc42b94d7dbb4b521409ce07R94-R101)
- Improved migration storage service (`HapiMigrationStorageSvc`, `HapiMigrationDao`) to better detect and handle migration history presence, and to avoid false positives when migration versions are missing. [[1]](diffhunk://#diff-d7fc11150cc29d846a8be932c7cf9e6515b73952bc42b94d7dbb4b521409ce07L60-R68) [[2]](diffhunk://#diff-a0a94bd6d41af85d800efef0cab78c3cdc23447dcc6199e86d066453c5f3581bR66)

**Testing and validation:**
- Added and updated integration and unit tests to verify that migrations on existing schemas without migration history now fail unless `--baseline-version` is specified, and that baseline logic works as intended. [[1]](diffhunk://#diff-475d6257be9ca77c0cc6f3a99bb45a0d2508e500a34c8db8c62e8b07c548fe05L229-R230) [[2]](diffhunk://#diff-475d6257be9ca77c0cc6f3a99bb45a0d2508e500a34c8db8c62e8b07c548fe05L255-R258) [[3]](diffhunk://#diff-08052aab7b7c5d0eb5b7b4330968f2b4bad063e34afd3ddf91708a754008972eL121-R132) [[4]](diffhunk://#diff-7d86798ff4c0c9fac4d39ad253e9a13518d3a29e00a6103db278b64a42ae2e00L24-R67) [[5]](diffhunk://#diff-7d86798ff4c0c9fac4d39ad253e9a13518d3a29e00a6103db278b64a42ae2e00L59-R82)

**Error handling and logging:**
- Improved error handling in the migrator to provide clearer exceptions when migration fails due to missing baseline information.

**Documentation and changelog:**
- Added a changelog entry describing the fix for incorrect migration skipping on existing databases without migration history.

fixes #8027


tadgh:
Formatting check succeeded!

patrick-werner:
updated Error Codes to reserved codes, added error if migration history was found &   --baseline-version is specified

hapi-fhir-cli/hapi-fhir-cli-api/src/main/java/ca/uhn/fhir/cli/BaseFlywayMigrateDatabaseCommand.java
```diff
@@ -98,6 +99,12 @@ public Options getOptions() {
 				SKIP_VERSIONS,
 				"Versions",
 				"A comma separated list of schema versions to skip.  E.g. 4_1_0.20191214.2,4_1_0.20191214.4");
+		addOptionalOption(
+				retVal,
+				null,
+				BASELINE_VERSION,
+				"Version",
+				"Record all migrations up to this version as already applied before migrating. This is intended for existing HAPI FHIR schemas that do not have migration history. E.g. 8.4.0");
```
dotasek: nitpick: This description is good, but should start with the format of the expected option ("The HAPI FHIR version to which this database has previously been migrated" or something similar)

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/HapiMigrator.java
```diff
@@ -206,6 +221,82 @@ public MigrationResult migrate() {
...
+		return retVal;
+	}
+
+	private MigrationVersion normalizeBaselineVersion(String theBaselineVersion) {
+		String baselineVersion = theBaselineVersion.trim().replace('_', '.');
+		if (baselineVersion.matches("\\d+\\.\\d+\\.\\d+")) {
+			// A product version like 8.4.0 was supplied. Pad it so that every schema version belonging to that
+			// release is treated as part of the baseline.
+			baselineVersion += ".99999999.999999";
+		} else if (!baselineVersion.matches("\\d+\\.\\d+\\.\\d+(\\.\\d+)+")) {
```
dotasek: thought: This regex gets flagged by Sonarqube as potentially having stack overflows for large inputs. This is a trusted input in my understanding, so isn't a worry, but perhaps an inline comment would speed up future reviewers that stumble upon it.

hapi-fhir-cli/hapi-fhir-cli-api/src/test/java/ca/uhn/fhir/cli/HapiFlywayMigrateDatabaseCommandTest.java
```diff
@@ -250,11 +251,39 @@ public void testMigrateFrom340_dryRun_whenNoMigrationTableExists() throws IOExce
...
+		String[] args = new String[]{
+			BaseFlywayMigrateDatabaseCommand.MIGRATE_DATABASE,
+			"-d", "H2_EMBEDDED",
+			"-u", url,
+			"-n", "",
+			"-p", "",
+			"--baseline-version", "3.4.0"
+		};
 
-		assertThat(JdbcUtils.getTableNames(connectionProperties)).doesNotContain("FLY_HFJ_MIGRATION");
```
dotasek, line 257: issue: This assertion is no longer present. It appears as though the two new tests only check where `--baseline-version` is valid, but no longer check a successful execution.

hapi-fhir-sql-migrate/src/test/java/ca/uhn/fhir/jpa/migrate/HapiMigratorIT.java
```diff
@@ -118,10 +119,17 @@ public void testInitializeSchema(boolean thePreCreateSchema) {
...
 		 */
 		migrator = buildMigrator(taskList.toTaskArray());
-		outcome = migrator.migrate();
 		if (thePreCreateSchema) {
-			assertThat(toTaskVersionList(outcome)).as(toTaskStatementDescriptions(outcome)).containsExactly("3");
+			assertThatThrownBy(migrator::migrate)
+					.isInstanceOf(HapiMigrationException.class)
+					.hasMessageContaining("--baseline-version");
+
+			migrator = buildMigrator(taskList.toTaskArray());
```
dotasek: issue: Is there an intent to expecting failure here, or could `setBaselineVersion` be set according to a condition earlier on and reduce the need for asserting an exception and then altering the migrator?

dotasek (changes requested):
This is a non-code-owner review. Please take all issues as non-blocking if code owners are in disagreement.

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/HapiMigrator.java
```diff
@@ -206,6 +221,82 @@ public MigrationResult migrate() {
...
+		}
+		return retVal;
+	}
+
+	private MigrationVersion normalizeBaselineVersion(String theBaselineVersion) {
+		String baselineVersion = theBaselineVersion.trim().replace('_', '.');
+		if (baselineVersion.matches("\\d+\\.\\d+\\.\\d+")) {
+			// A product version like 8.4.0 was supplied. Pad it so that every schema version belonging to that
+			// release is treated as part of the baseline.
+			baselineVersion += ".99999999.999999";
```
jamesagnew: This makes the assumption that our tasks will always use an 8 digit date (probably a safe assumption), and a <= 6 digit migration number (safe so far, but no guarantees in future). 

You should probably bake this assumption into `BaseTask#MIGRATION_VERSION_PATTERN` and add a javadoc link pointing between these two places so that a future person who wants to relax that rule knows better. Even better if you bake the length into a constant.

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/HapiMigrator.java
jamesagnew: Sonar's issue, which is probably overcautious but not wrong is that repeating groups like this are bad. Better to remove the repeating group and just explicitly use a pattern with 5 dot-separated numbers.

hapi-fhir-cli/hapi-fhir-cli-api/src/main/java/ca/uhn/fhir/cli/BaseFlywayMigrateDatabaseCommand.java
```diff
@@ -98,6 +99,12 @@ public Options getOptions() {
 				SKIP_VERSIONS,
 				"Versions",
 				"A comma separated list of schema versions to skip.  E.g. 4_1_0.20191214.2,4_1_0.20191214.4");
+		addOptionalOption(
```
jamesagnew, line 102: This new option should probably be documented in `upgrading.md`

hapi-fhir-sql-migrate/src/test/java/ca/uhn/fhir/jpa/migrate/HapiMigratorIT.java
```diff
@@ -34,6 +34,7 @@
 import java.util.stream.Collectors;
```
jamesagnew, line 34: Thought: There are negative tests for the new functionality ("you need to add --baseline-version") but there are no tests actually exercizing the new functionality that I can see (unless I missed it somehow?). Would be good to have a test that actually verifies that adding the given argument actually causes the migrator to skip migrations.

jamesagnew (changes requested):

hapi-fhir-cli/hapi-fhir-cli-api/src/main/java/ca/uhn/fhir/cli/BaseFlywayMigrateDatabaseCommand.java
patrick-werner: changed to: The HAPI FHIR version to which this database has previously been migrated. Before running migrations, all migrations up to and including this version are recorded as already applied. This is intended for existing HAPI FHIR schemas without migration history. Example: 8.4.0.

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/HapiMigrator.java
patrick-werner: changed to "\\d+\\.\\d+\\.\\d+\\.\\d+\\.\\d+"

hapi-fhir-cli/hapi-fhir-cli-api/src/test/java/ca/uhn/fhir/cli/HapiFlywayMigrateDatabaseCommandTest.java
patrick-werner: added a positive test

hapi-fhir-sql-migrate/src/main/java/ca/uhn/fhir/jpa/migrate/HapiMigrator.java
patrick-werner: added constants and pattern

hapi-fhir-cli/hapi-fhir-cli-api/src/main/java/ca/uhn/fhir/cli/BaseFlywayMigrateDatabaseCommand.java
patrick-werner: added documentation

hapi-fhir-sql-migrate/src/test/java/ca/uhn/fhir/jpa/migrate/HapiMigratorIT.java
patrick-werner: This is covered by testInitializeSchema_existingSchemaRequiresBaselineAndSkipsBaselinedMigrations(true).
The test sets --baseline-version through 20250101.1 and verifies that only migrations 20250101.2 and 20250101.3 execute. I renamed the test and added an explicit comment to make the baseline-skipping assertion easier to find.

hapi-fhir-sql-migrate/src/test/java/ca/uhn/fhir/jpa/migrate/HapiMigratorIT.java
patrick-werner: Yes, the failure is intentional — it's the behaviour under test. applyBaselineIfRequired() rejects an existing schema that has no successful migration history and points the operator at --baseline-version; setting the baseline up front would bypass that guard instead of covering it. The fail → set baseline → rerun sequence mirrors the documented upgrade path. Good catch on the migrator rebuild though: that's unnecessary, the exception is thrown before any task runs, so I've dropped it and just call setBaselineVersion on the existing instance.

patrick-werner:
@jamesagnew @dotasek 
i've addressed all requested changes/ comments

jamesagnew (approved):

jamesagnew:
@patrick-werner unfortunately in performing release QA this was causing pretty big migration issues so I've had to revert it.

The issue is the removal of [these 3 lines](https://github.com/hapifhir/hapi-fhir/pull/8028/changes#diff-a43a7140c555c856e76186e1e0e2dfa402cc66446edbf68c8c8a3ea600266720L70-L72) in `InitializeSchemaTask.java`.

On an actual newly instantiated schema (i.e. invoking the migrator for the first time on a completely empty database) because the `myInitializedSchema` flag no longer gets set when the schema is instantiated, the migrator then runs through all of the incremental migrations from the very start, instead of marking them all as done without executing them. In other words, one of the key bits of design in the migrator is that if the database is initialized via the DDL script, we consider it to be up-to-date with all past migrations up to the point where the DDL script was generated.

This doesn't seem to leave the database in a consistent state since the migrator starts applying migrations which assume a parevious version of the schema when in fact it has the current version. Do you think it would be possible to rework this change so that it preserves this behaviour?

PR #8031, opened 2026-06-02
GL-8718: Generalise patient SP Patient compartment membership beyond Device

### LINKED ISSUE #8168: "patient" SP not part of the patient compartment
**Describe the bug**
In most cases, the "patient" SP is an alias/convenience SP scoped on resourceType = Patient for some other "base" reference SP. (Eg. `Observation.patient` has the path `Observation.subject.where(resolve() is Patient)`. Logically, the "patient" SP should be part of the patient compartment when the "base" SP is part of the patient compartment, but more often than not, this is not the case in the FHIR spec.

jdar8:
Closes https://github.com/hapifhir/hapi-fhir/issues/8168

## Summary

`SearchParameterUtil.getMembershipCompartmentsForSearchParameter()` previously only added `Device.patient` to the Patient compartment via a hardcoded special case. As a result, `patient`-named reference SPs on other resources — such as `Observation.patient` (which narrows `Observation.subject` via `.where(resolve() is Patient)`) and R5 `Coverage.patient` (which is an exact alias for `Coverage.beneficiary`) — were not recognized as Patient compartment members. Any feature driven off that compartment map (compartment access control, MDM auto-expansion) silently ignored searches using the `patient` SP on these ~50 resources.

Key changes:

1. **`isCoveredByPatientCompartmentBaseSp()`** — new helper that checks whether a `patient`-named reference SP's path covers the same field as any other SP on the same resource that already declares Patient compartment membership. Two patterns are recognized: exact alias (e.g. `Coverage.patient` path = `Coverage.beneficiary`) and `.where(resolve() is Patient)` narrowing (e.g. `Observation.patient` path = `Observation.subject.where(resolve() is Patient)`). Pipe-delimited multi-resource paths are split by `ownPathSegments()` so only the segments belonging to the declaring resource are considered.
2. **Security exclusion guard** — the new rule is gated by `RESOURCE_TYPES_TO_SP_TO_OMIT_FROM_PATIENT_COMPARTMENT`. Deliberately excluded SPs such as `List.patient` (issue #7118) are unaffected.
3. **`getResourceTypeName()`** — uses `@ResourceDef` annotation to resolve the FHIR type name from the Java class (e.g. `ListResource` → `"List"`), falling back to `getSimpleName()`.
4. **Case-insensitive name comparisons** — all `patient` name checks now use `equalsIgnoreCase` for robustness.
5. **Comprehensive test suite in `SearchParameterUtilTest`**:
   - `testNoNewResourceTypesAddedToPatientCompartment()` — snapshot test pinning the exact set of resource types in the Patient compartment for R4 and R5.
   - `testPatientSp_PatientCompartmentMembership()` — example table with hand-derived positive (Observation R4/R5, Coverage R5, Device R4) and negative (SupplyRequest R5, List R4/R5, Sequence DSTU3) cases.
   - `testCompartmentMembershipMatchesAnnotationsForEverySearchParam()` — broad invariant test covering every SP across every buildable FHIR version: verifies no existing memberships are dropped, the omit map is respected, and only `patient` SPs are ever added to the Patient compartment.
   - `testPatientSpIsInPatientCompartmentForObservation()` — runtime compartment check using the full `FhirContext` for both R4 and R5.

---

## Code Review Suggestions

- [ ] **Correctness of `isCoveredByPatientCompartmentBaseSp()`**: The method scans `@SearchParamDefinition` annotations directly on class fields (not `RuntimeSearchParam`s) because `RuntimeSearchParam`s are still being constructed at call time. Confirm that reflection over `getFields()` finds all SP annotations for concrete FHIR resource classes (no SPs inherited from abstract superclasses are missed in the Patient compartment context).
- [ ] **`ownPathSegments()` filter**: The segment filter keeps only strings starting with `theResourceType + "."`. For `ListResource`, `getResourceTypeName()` returns `"List"` via `@ResourceDef`. Verify that the segment `"List.patient"` is correctly matched when `theResourceType = "List"`, and that this doesn't accidentally match a path that starts with a longer type name sharing the same prefix.
- [ ] **Omit-map key format**: `RESOURCE_TYPES_TO_SP_TO_OMIT_FROM_PATIENT_COMPARTMENT` is keyed by the FHIR resource type name (e.g. `"List"`). The lookup uses `getResourceTypeName(theResourceClazz)`, which now correctly returns `"List"` for `ListResource`. Confirm the omit-map definition uses the same string form (not the Java class name).
- [ ] **`Device.patient` special case is now redundant for the general rule**: `Device.patient`'s path is `Device.patient` (a direct reference field), which does not alias any other SP that provides Patient compartment membership. The two branches (old hardcode, new delegation check) are non-overlapping — Device continues to be handled by the pre-existing special case. Confirm this is intentional and that no Device path was accidentally omitted from the new logic.
- [ ] **Blast radius on `SearchNarrowingInterceptor`**: with additional `patient` SPs now in the Patient compartment map, compartment-based access control will enforce narrowing on queries that previously slipped through. This is the correct behavior, but operators upgrading should be aware that previously un-narrowed `?patient=` queries on resources like `Observation` will now be constrained by compartment rules.

---

### Setup
- [ ] Deploy a HAPI FHIR JPA server (R5) with MDM or compartment-based search narrowing enabled.
- [ ] Create two `Patient` resources (P1, P2) and establish an MDM link between them.
- [ ] Create one `Observation` with `subject = Patient/P1` and one with `subject = Patient/P2`.

### Core fix — MDM auto-expansion via `patient` SP
- [ ] `GET /Observation?patient=Patient/P1` **with** `FHIR_AUTO_MDM` — expect 2 results (P1's and MDM-linked P2's observation).
- [ ] `GET /Observation?patient=Patient/P1` **without** `FHIR_AUTO_MDM` — expect 1 result (P1's only; no expansion).
- [ ] `GET /Observation?subject=Patient/P1` **with** `FHIR_AUTO_MDM` — expect 2 results (regression check; `subject` expansion must continue to work).

### Exact-alias pattern (R5 Coverage)
- [ ] Create a `Coverage` resource with `beneficiary = Patient/P1`. Confirm `GET /Coverage?patient=Patient/P1` with `FHIR_AUTO_MDM` returns 2 results (P1's and P2's coverage).

### Security exclusions (omit map)
- [ ] Create a `List` resource referencing P1 via `List.subject`. Confirm `GET /List?patient=Patient/P1` does **not** expand to MDM-linked patients — `List.patient` must remain excluded.

### Negative cases (non-aliased patient SPs)
- [ ] `GET /SupplyRequest?patient=Patient/P1` with `FHIR_AUTO_MDM` — `SupplyRequest.patient` maps to `deliverFor`, which does not provide Patient compartment membership. Confirm no expansion occurs.

### Cross-version regression
- [ ] On an R4 server, confirm `GET /Observation?patient=Patient/P1` with `FHIR_AUTO_MDM` returns 2 results (R4 `Observation.patient` also narrows `subject` and should already work).
- [ ] On a DSTU3 server, confirm `Sequence?patient=` behaviour is unchanged (DSTU3 `patient` SP is a direct field with no compartment membership; expect no expansion).


tadgh:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
```diff
@@ -144,17 +145,91 @@ public static Set<String> getMembershipCompartmentsForSearchParameter(
 			validCompartments.add("Patient");
 		}
 
+		/*
+		 * In many cases, the "patient" SP is a "convenience" SP that resolves to a path to
+		 * another SearchParameter that  declares patient compartment membership. In this case,
```
epeartree: 'SearchParameter that  declares patient' - double space between 'that' and 'declares'

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
```diff
@@ -144,17 +145,91 @@ public static Set<String> getMembershipCompartmentsForSearchParameter(
...
+		Set<String> patientPathSegments = ownPathSegments(theResourceType, thePatientSp.path());
+		return Arrays.stream(theResourceClazz.getFields())
+				.map(f -> f.getAnnotation(SearchParamDefinition.class))
+				// Any other SP that is not the "patient" SP
+				.filter(spd -> spd != null && !thePatientSp.name().equalsIgnoreCase(spd.name()))
+				// That declares Patient compartment membership
+				.filter(spd -> Arrays.stream(spd.providesMembershipIn())
+						.anyMatch(c -> "Patient".equals(getCleansedCompartmentName(c.name()))))
+				// Get the path for theResourceType
+				.flatMap(spd -> ownPathSegments(theResourceType, spd.path()).stream())
```
epeartree: hard coding ' ".where(resolve() is Patient)"' is brittle.  works for now but if a space is added to patientPathSegments (which would be valid fhirPath), we're back to the original bug.

i'd suggest doing a spaceless  compare at line 217 and i'd add a comment above to alert that 'where(resolve()isPatient)'  is not a typo

epeartree (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/SearchParameterUtil.java
jdar8: Added the whitespace removal to `ownPathSegmentsNoWhitespace`, the comparison will now ignore whitespace.

Thanks for the review!

Issue #6885, opened 2025-04-21
JpaPackageCache#findPackageAsset with incorrect parameters should throw ResourceNotFound instead of returning null


lukedegruchy:
**Describe the bug**
See: https://github.com/hapifhir/hapi-fhir/issues/6865

When calling JpaPackageCache#findPackageAsset, if the resource is not found it returns null.

**To Reproduce**
1. Call JpaPackageCache#findPackageAsset with parameters that will not return a resource
2. Get back a null resource

**Expected behavior**
A ResourceNotFoundException with a clear error and a HAPI message code.

**Environment (please complete the following information):**
 - HAPI FHIR Version:  master

PR #8032, opened 2026-06-03
GL-6885: Honor _filter in subscription criteria matching

fil512:
## Summary

Fixes a long-standing bug where a FHIR `Subscription` whose `criteria` used the `_filter` search parameter would fire for **every** resource of the matching type instead of only those satisfying the filter. Root cause: `MatchUrlService.translateMatchUrl` silently discarded all `_`-prefixed parameters (except `_language`), leaving the `SearchParameterMap` empty so `InMemoryResourceMatcher` returned a universal match (failing open).

Key changes:

1. **`MatchUrlService.translateMatchUrl`** now retains `_filter` as a `StringParam`, so the DB search path / `QueryStack` evaluate it correctly. This also stops `_filter` being silently ignored in conditional create/update/delete and bulk-export `_typeFilter`.
2. **`InMemoryResourceMatcher`** adds `_filter` to `UNSUPPORTED_PARAMETER_NAMES`, so the composite matcher falls back to the DATABASE strategy (same as `_has`) rather than failing open. `_filter` subscriptions are stamped `EX_SUBSCRIPTION_MATCHING_STRATEGY = DATABASE`.
3. **`SubscriptionQueryValidator`** rejects subscriptions whose criteria contain `_filter` at submission time (`UnprocessableEntityException`, Msg 2792) when the server's filter search parameter is disabled — the default configuration.

## Behavior change / blast radius

Retaining `_filter` at the root of `translateMatchUrl` is shared by all match-URL consumers. As a result:
- A conditional match URL whose only parameter is `_filter` (e.g. `Patient?_filter=...`) now runs a real search instead of throwing Msg 518 ("URL has no search parameters"); with filter search disabled it throws Msg 1222 instead of silently ignoring `_filter`.
- `_filter` values in subscription/conditional match URLs must be URL-encoded (spaces as `%20`), consistent with normal REST search.

## Review notes

An independent code review ran against this branch; findings were addressed here:
- **(Fixed)** A duplicate `SubscriptionValidatingInterceptorTest` in `hapi-fhir-jpaserver-test-r4` was not initially updated and NPE'd under the new `JpaStorageSettings` injection — now wired with a mock (13/0/0/0).
- **(Fixed)** Added conditional-operation tests in `FhirResourceDaoR4FilterTest` covering the Msg 518→search flip (enabled resolves; disabled throws Msg 1222), TDD-red verified.
- **(Intentionally not changed)** The new `SubscriptionQueryValidator` constructor is left un-annotated to match the class's existing style; multi-value `_filter` joining is not a concern for the single-expression subscription/conditional criteria in scope.

### Setup
- [ ] Deploy a HAPI JPA server built from this branch.
- [ ] Filter-enabled tests: set `filter_search.enabled = true`. Filter-disabled tests: leave it at the default (`false`).

### Subscription delivery (filter search enabled)
- [ ] Create a rest-hook `Subscription` with `criteria = "Patient?_filter=family%20eq%20%22Smith%22"`. POST a `Patient` family=`Jones` → no notification. POST a `Patient` family=`Smith` → exactly one notification.
- [ ] Confirm the `Subscription` is accepted (HTTP 201) at submission when filter search is enabled.

### Subscription rejection (filter search disabled, default)
- [ ] POST a `Subscription` whose `criteria` contains `_filter` → HTTP 422 `OperationOutcome` referencing message code 2792.

### Conditional operations (filter search enabled)
- [ ] `PUT Patient?_filter=family%20eq%20%22Smith%22` with no match → creates (HTTP 201) rather than throwing Msg 518.
- [ ] Same `PUT` with a matching Patient (family=Smith) present → updates the existing resource (HTTP 200).

### Conditional operations (filter search disabled)
- [ ] `PUT Patient?_filter=...` → Msg 1222 (filter search not supported), not Msg 518.

### Regression
- [ ] A `Subscription` using a standard parameter (`Patient?family=Smith`) still delivers and is unaffected.
- [ ] A `Subscription` using `_has` still falls back to DATABASE strategy as before.

---

Resolves GitLab issue: https://gitlab.com/simpatico.ai/cdr/-/work_items/6885 (#6885)
Spec ticket (Jira): [SMILE-9308](https://simpaticois.atlassian.net/browse/SMILE-9308)


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa/search.md
```diff
@@ -20,7 +20,9 @@ Searching on Location.Position using `near` currently uses a box search, not a r
 
 ### _filter
 
-The special `_filter` is only partially implemented.
+The special `_filter` search parameter is only partially implemented, and is disabled by default. It must be explicitly enabled (via `JpaStorageSettings#setFilterParameterEnabled(true)`) before it can be used.
+
+`_filter` is honored anywhere a [match URL](https://hl7.org/fhir/http.html#cond-update) is parsed &mdash; including [Subscription](https://hl7.org/fhir/subscription.html) `criteria`, conditional create/update/delete, and the bulk export `_typeFilter` parameter. When a Subscription's `criteria` contains `_filter`, matching falls back to the database matching strategy (it cannot be evaluated by the in-memory matcher). If `_filter` is disabled on the server, such a Subscription is rejected at submission time with an `UnprocessableEntityException`, and other match URL contexts likewise return an error rather than silently ignoring the `_filter` clause.
```
TipzCM: 'criteria' only exists on dstu3 subscription objects.

But the link provided is for the latest release of fhir, which defaults to (right now) R5.

This (imo) makes this documentation a bit confusing ('criteria' doesn't exist on r4 or r5 subscription objects).

TipzCM (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa/search.md
fil512: Good catch — clarified in the latest push. The doc now notes that `criteria` is the pre-R5 Subscription element (DSTU2/DSTU3/R4), links to the DSTU3 Subscription page where `criteria` actually appears, and points to `Subscription.filterBy` as the R5 topic-based equivalent so the link no longer implies `criteria` exists on the R5 resource.

PR #8061, opened 2026-06-10
8037 searchurljobmaintenancesvcimplremovestaleentries cleanup delete exceeds database statement timeout once hfj res search url grows large should chunk

### LINKED ISSUE #8037: SearchUrlJobMaintenanceSvcImpl.removeStaleEntries() cleanup DELETE exceeds database statement timeout once HFJ_RES_SEARCH_URL grows large — should chunk
### Summary

`SearchUrlJobMaintenanceSvcImpl.removeStaleEntries()` runs a single unbounded JPQL DELETE against `HFJ_RES_SEARCH_URL` every 10 minutes:

```java
@Modifying
@Query("DELETE FROM ResourceSearchUrlEntity s WHERE (s.myCreatedTime < :cutoff)")
int deleteAllWhereCreatedBefore(@Param("cutoff") Date theCutoff);
```
(`IResourceSearchUrlDao.java:33-35`, called from `ResourceSearchUrlSvc.deleteEntriesOlderThan` → `SearchUrlJobMaintenanceSvcImpl.removeStaleEntries`)

When the table has accumulated a backlog of stale rows — easily triggered by leaky upstream paths like #8036, or simply by any workload that grows the table past the database's statement-timeout × per-row-delete-cost — that single DELETE runs longer than the configured database statement timeout, fails, and **rolls back the entire cycle, removing zero rows**. The next cycle finds the same backlog plus 10 minutes' worth of new entries, runs slower, fails again. The maintenance job ceases to make any progress at all and the table grows monotonically.

This is independent of #8036: even with that leak fixed, any future regression that creates orphans faster than this cleanup can clear them in a single statement will trip the same failure mode. The maintenance job should be resilient to backlog size regardless of how the backlog got there.

### Steps to reproduce

1. Run any deployment (single-node or MegaScale) with `SearchUrlJobMaintenanceSvcImpl` active and a configured database statement timeout (e.g. MSSQL default 30 s, set via JDBC `queryTimeout` or driver string).
2. Seed `HFJ_RES_SEARCH_URL` with ~1 M rows whose `CREATED_TIME` is older than `OUR_CUTOFF_IN_MILLISECONDS` (1 hour).
3. Trigger `removeStaleEntries()` either by waiting for the schedule or by directly invoking `ISearchUrlJobMaintenanceSvc.removeStaleEntries()` from a Spring bean.

### Expected behavior

The cleanup completes (regardless of backlog size) and removes all rows older than the cutoff.

### Actual behavior

On MSSQL with 1 M rows, the single DELETE statement takes ~20 s in our measurement; with the statement timeout fired the transaction rolls back and no rows are removed. A downstream customer report observed ~27 M accumulated rows in `HFJ_RES_SEARCH_URL` after the cleanup had been failing for weeks.

### Root cause

The DELETE is unbounded by row count. SQL Server (and behaviorally similar engines) acquire row locks proportional to the number of matched rows, escalate to a table lock around the ~5 000 lock threshold, hold those locks for the duration of the entire transaction, and write every deletion to the log before commit. With millions of matched rows, wall-clock easily exceeds typical statement-timeout settings (15-60 s).

There's also no per-chunk progress: a partial drain isn't possible because the whole statement rolls back together.

### Impact

- **High severity under any workload that builds a backlog.** Once the table crosses the per-deployment statement-timeout threshold, the maintenance job effectively stops working until the table is drained manually.
- **Positive feedback loop:** each failed cycle leaves the backlog plus new arrivals, making the next cycle slower. The system never self-recovers.
- **Affects MegaScale particularly hard** because the shared-DB topology lets orphans from many tenants pile up in one table.
- A documented case: 27 M rows accumulated over ~9 days of data ingestion before anyone noticed.

### Suggested fix

Replace the single unbounded DELETE with a chunked drain loop:

1. **Add a chunked DAO method** to `IResourceSearchUrlDao` that returns up to N rows older than the cutoff (JPQL with `Pageable` for portability across H2 / Postgres / MSSQL / Oracle — JPA `setMaxResults` works for SELECT but not for bulk DELETE on most dialects):
   ```java
   @Query("SELECT s FROM ResourceSearchUrlEntity s WHERE s.myCreatedTime < :cutoff")
   List<ResourceSearchUrlEntity> findOlderThan(@Param("cutoff") Date theCutoff, Pageable thePageable);
   ```

2. **Loop with bounded chunks** in `ResourceSearchUrlSvc.deleteEntriesOlderThan`:
   ```java
   public void deleteEntriesOlderThan(Date theCutoffDate) {
       int chunkSize = 10_000;
       while (true) {
           int removed = myTxTemplate.execute(status -> {
               List<ResourceSearchUrlEntity> batch =
                   myResourceSearchUrlDao.findOlderThan(theCutoffDate, PageRequest.of(0, chunkSize));
               batch.forEach(myEntityManager::remove);
               return batch.size();
           });
           if (removed < chunkSize) break;
       }
   }
   ```

3. **Each chunk should commit in its own transaction** (TransactionTemplate with `REQUIRES_NEW`) so:
   - Lock escalation doesn't span the whole drain
   - Partial progress survives if a later chunk fails
   - The cleanup can be cooperative with concurrent writers

Hibernate's JDBC batching (`hibernate.jdbc.batch_size`) keeps the per-row DELETEs cheap; on MSSQL each 10 k chunk completes in a few hundred ms — well under any reasonable statement timeout. A backlog of 27 M rows drains in a single 10-minute cycle.

A regression test that seeds N rows and asserts cleanup completes under a tight bound (with an explicit JDBC `queryTimeout` set so a regression to the unbounded statement actually fires the timeout) would be the right shape.

### Related

- #8036 — auto-created placeholder reference targets leak rows into `HFJ_RES_SEARCH_URL`. That's the primary cause of orphan accumulation; this ticket is the resilience gap that lets accumulation become unrecoverable. The two fixes are independent and both worth shipping.

### Files

- `hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/ResourceSearchUrlSvc.java`
- `hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/SearchUrlJobMaintenanceSvcImpl.java`
- `hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IResourceSearchUrlDao.java`

jmarchionatto:
Chunks the scheduled `SearchUrlJobMaintenanceSvc.removeStaleEntries` cleanup so it no longer exceeds the database per-statement timeout once `HFJ_RES_SEARCH_URL` grows large.

  Previously the cleanup issued a single unbounded `DELETE WHERE CREATED_TIME < cutoff`; on MegaScale this could take longer than MSSQL's per-statement budget, roll back, and leave stale entries to accumulate indefinitely. The cleanup now pages through stale rows in batches of 1,800 (well under MSSQL's 2,100-parameter-per-statement limit), with each page committed in its own short transaction.
  
  ## Changes

  - `ResourceSearchUrlSvc.deleteEntriesOlderThan`: replaced the single bulk DELETE with a paged loop using `findStaleIds` + `deleteByResIds`, each page in its own `TransactionTemplate` execution.
  - `IResourceSearchUrlDao`: replaced `deleteAllWhereCreatedBefore` with `findStaleIds(cutoff, Pageable)` returning a `Slice<Long>`.
  - `JpaConfig`: wires the `PlatformTransactionManager` into the new `ResourceSearchUrlSvc` constructor.
  - New integration test `SearchUrlMaintenanceSlowDeleteMsSqlIT` (Testcontainers MSSQL) seeds 100K stale + 10K recent rows, applies a 1 s per-statement timeout via the mssql-jdbc URL property, and asserts the maintenance call drains all stale rows without timing out. RED on the pre-fix unbounded DELETE, GREEN on the chunked fix.

  ## Test plan
  
  - Integration test runs against a Testcontainers MSSQL container.
  - Confirmed RED on `master` (pre-fix): `QueryTimeoutException` after ~3 s, 110K rows remain.
  - Confirmed GREEN on this branch: every stale row drained, recent rows kept, completes well under the wall-clock cap.


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/ResourceSearchUrlSvc.java
```diff
@@ -57,33 +71,62 @@ public class ResourceSearchUrlSvc {
...
 	 */
+	@Transactional(propagation = Propagation.NOT_SUPPORTED)
 	public void deleteEntriesOlderThan(Date theCutoffDate) {
-		ourLog.debug("About to delete SearchUrl which are older than {}", theCutoffDate);
-		int deletedCount = myResourceSearchUrlDao.deleteAllWhereCreatedBefore(theCutoffDate);
-		ourLog.debug("Deleted {} SearchUrls", deletedCount);
+		ourLog.debug("Deleting SearchUrls older than {}", theCutoffDate);
+		PageRequest page = PageRequest.of(0, DELETE_PAGE_SIZE);
+		AtomicLong totalDeleted = new AtomicLong();
+		while (true) {
```
codeforgreen: suggestion: I think we should add a safety cap to this loop, something along the lines of defining a maximum number of iterations to attempt this. Also wonder if it makes sense to log in debug level when it actually performs something.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/ResourceSearchUrlSvc.java
```diff
@@ -57,33 +71,62 @@ public class ResourceSearchUrlSvc {
...
 
 	/**
 	 * Perform removal of entries older than {@code theCutoffDate} since the create operations are done.
+	 *
+	 * <p>Pages the deletion in batches of {@link #DELETE_PAGE_SIZE} so no individual
+	 * statement scans the whole table. A single unbounded DELETE on millions of rows trips MSSQL's
+	 * per-statement timeout on MegaScale deployments (SMILE-12080); each paged DELETE is bounded.
+	 *
+	 * <p>Runs outside any caller transaction ({@link Propagation#NOT_SUPPORTED}) so each page commits
+	 * independently — otherwise the whole sweep would still be one giant transaction.
```
codeforgreen: Suggestion: I am not sure if this explanation is necessary once we've fixed the problem. I'd focus just on briefly describing what this method does as it is public. I am ok if you disagree in that case please revise from perspective that we should not explain a fix in a public javadoc.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IResourceSearchUrlDao.java
```diff
@@ -30,9 +32,8 @@
 
 public interface IResourceSearchUrlDao extends JpaRepository<ResourceSearchUrlEntity, Long>, IHapiFhirJpaRepository {
 
-	@Modifying
-	@Query("DELETE FROM ResourceSearchUrlEntity s WHERE (s.myCreatedTime < :cutoff)")
-	int deleteAllWhereCreatedBefore(@Param("cutoff") Date theCutoff);
+	@Query("SELECT s.myResourcePid FROM ResourceSearchUrlEntity s WHERE s.myCreatedTime < :cutoff")
+	Slice<Long> findStaleIds(@Param("cutoff") Date theCutoff, Pageable thePageable);
```
codeforgreen, line 36: Question: Could there could be duplicate RES_ID in a slice?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/ResourceSearchUrlSvc.java
```diff
@@ -57,33 +71,62 @@ public class ResourceSearchUrlSvc {
...
-		int deletedCount = myResourceSearchUrlDao.deleteAllWhereCreatedBefore(theCutoffDate);
-		ourLog.debug("Deleted {} SearchUrls", deletedCount);
+		ourLog.debug("Deleting SearchUrls older than {}", theCutoffDate);
+		PageRequest page = PageRequest.of(0, DELETE_PAGE_SIZE);
+		AtomicLong totalDeleted = new AtomicLong();
+		while (true) {
+			// Each page in its own transaction — the repository is @Transactional(MANDATORY), so it
+			// needs an enclosing tx, and we want that tx to commit between pages, not span the sweep.
+			Boolean shouldContinue = myPageTxTemplate.execute(status -> {
+				Slice<Long> stale = myResourceSearchUrlDao.findStaleIds(theCutoffDate, page);
```
codeforgreen: Comment: I guess another option would have been to use deleteAllByIdInBatch which is readily available in `JpaRepository` but that would have brought in too many changes I suppose because that requires an id of type Long.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/database/SearchUrlMaintenanceSlowDeleteMsSqlIT.java
```diff
@@ -0,0 +1,332 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowable;
+
+// Created by claude-opus-4-7
+
+/**
+ * TDD regression test for SMILE-12080 / work item 8819. Runs against a Testcontainers-managed
```
codeforgreen: Suggestion: We should probably remove these issue numbers from the hapi-fhir code, especially documentation. Maybe the whole content can be removed and replaced by a simple description of what the test class does.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/database/SearchUrlMaintenanceSlowDeleteMsSqlIT.java
```diff
@@ -0,0 +1,332 @@
...
+ * the budget. Assertion: every stale row drained, every recent row kept.
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@ExtendWith(SpringExtension.class)
+@ContextConfiguration(classes = SearchUrlMaintenanceSlowDeleteMsSqlIT.TestConfig.class)
+// Disable scheduler to avoid clashes
+@TestPropertySource(properties = {
+	UnregisterScheduledProcessor.SCHEDULING_DISABLED_EQUALS_TRUE,
+})
+public class SearchUrlMaintenanceSlowDeleteMsSqlIT {
```
codeforgreen, line 63: Nitpick: There is alot of code comments in this test. Consider trimming down the redundant / self-explanatory ones. e.g. 
// Trigger container start + run migrations against the raw container DataSource
// Return a fresh BasicDataSource whose URL has queryTimeout appended

codeforgreen (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/ResourceSearchUrlSvc.java
jmarchionatto: Both done.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/ResourceSearchUrlSvc.java
jmarchionatto: Done.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IResourceSearchUrlDao.java
jmarchionatto: Good observation! The schema doesn't forbid it, but the write path can't produce it — and even if a duplicate existed, the chunked loop handles it correctly.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/ResourceSearchUrlSvc.java
jmarchionatto: Agree.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/database/SearchUrlMaintenanceSlowDeleteMsSqlIT.java
jmarchionatto: Good point! Done.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/database/SearchUrlMaintenanceSlowDeleteMsSqlIT.java
jmarchionatto: Done.

PR #8067, opened 2026-06-10
8036 hfj res search url rows leaked on auto created placeholder reference targets never registered for deletebyresids cleanup

### LINKED ISSUE #8036: HFJ_RES_SEARCH_URL rows for auto-created placeholder resources are not cleaned up when the placeholder is updated
When `AutoCreatePlaceholderReferenceTargets` is enabled and an incoming resource references
a not-yet-existing target via a conditional URL (e.g.
`subject: { reference: "Patient?identifier=urn:hcn|123" }`), HAPI auto-creates a placeholder
for the target. The JPA hook (`JpaDaoResourceLinkResolver.verifyPlaceholderCanBeCreated`)
correctly inserts a concurrency-sentinel row into `HFJ_RES_SEARCH_URL` via
`ResourceSearchUrlSvc.enforceMatchUrlResourceUniqueness`, but — unlike the ordinary
conditional-create path (`BaseHapiFhirResourceDao.doCreateForPostOrPut`) — it never calls
`setSearchUrlPresent(true)` on the placeholder's `ResourceTable` entity.

That flag is what routes a sentinel row into update-time cleanup:
`BaseHapiFhirResourceDao.doUpdateForUpdateOrPatch` checks `isSearchUrlPresent()` on the
first update/delete of the resource and registers the PID (via
`TransactionDetails.addUpdatedResourceId`) so the row is deleted at the end of that update's
transaction. Because the flag is never set for placeholders, this check is always false for
them, with two consequences:

1. **Unnecessary retention.** In the common ingestion flow, the placeholder is overwritten
   in place seconds-to-minutes later when the real resource arrives as a conditional update
   (`PUT Patient?identifier=urn:hcn|123` finds the placeholder via its indexed identifier
   and updates it, same PID). The sentinel row *should* die at that update, as it does for
   ordinary conditional creates; instead it lingers until the hourly scheduled sweep
   (`SearchUrlJobMaintenanceSvcImpl`). At high ingestion rates this multiplies the standing
   size of `HFJ_RES_SEARCH_URL` by orders of magnitude (row lifetime of ~1 hour instead of
   seconds).
2. **Stale-sentinel correctness edge case.** If the placeholder is updated such that its
   identifier changes, or is deleted, within the hour, the sentinel row is now *wrong*: a
   subsequent conditional create for the same URL can collide with a row pointing at a
   changed or deleted resource, instead of legitimately creating a new one. For ordinary
   conditional creates this cannot happen because the row is removed in the same transaction
   as the update/delete that invalidates it.

## To reproduce

1. Enable `setAutoCreatePlaceholderReferenceTargets(true)` and
   `setPopulateIdentifierInAutoCreatedPlaceholderReferenceTargets(true)`.
2. Create an `Observation` whose `subject` is a conditional-URL reference, e.g.
   `Patient?identifier=http://example.org|abc`. A placeholder Patient is auto-created and a
   row appears in `HFJ_RES_SEARCH_URL`.
3. Perform a conditional update for the real Patient:
   `PUT Patient?identifier=http://example.org|abc` with full content. The placeholder is
   overwritten in place.
4. Inspect `HFJ_RES_SEARCH_URL`.

## Expected behavior

After step 3, the sentinel row is deleted at the end of the update transaction — identical
to the lifecycle of a row created by an ordinary conditional create that is subsequently
updated.

(Note the row is *expected* to still exist after step 2: sentinel rows must outlive the
transaction that created them, because they guard against concurrent transactions whose
match search ran before that transaction committed and which may insert at any later point.
This is why the cleanup contract is "flag at create, delete at first update/delete, hourly
sweep as backstop" — see the comment on
`SearchUrlJobMaintenanceSvcImpl.OUR_CUTOFF_IN_MILLISECONDS`: the retention "MUST be longer
that longest transaction possible to work". Any fix must preserve that property; deleting
the row in the transaction that created it would re-open the duplicate-creation race the
table exists to prevent.)

## Actual behavior

After step 3 the row is still present, and remains until the scheduled sweep removes it
(rows older than 1 hour, checked every 10 minutes). `isSearchUrlPresent()` is `false` on
the placeholder entity, so `doUpdateForUpdateOrPatch` never registers the row for cleanup.

## Proposed fix

In `JpaDaoResourceLinkResolver.verifyPlaceholderCanBeCreated`, alongside the existing
`enforceMatchUrlResourceUniqueness` call, set the flag on the placeholder entity:

```java
((ResourceTable) theStoredEntity).setSearchUrlPresent(true);
```

Nothing else. No cleanup registration at creation time (see the design constraint under
"Expected behavior"); the existing `doUpdateForUpdateOrPatch` →
`postUpdateTransaction`/`TransactionProcessor` chain then handles deletion at the
placeholder's first update or delete, exactly as for ordinary conditional creates.
Placeholders that are never touched again continue to fall to the hourly sweep, which is
the designed lifecycle.

## Test shape

Add to `FhirResourceDaoCreatePlaceholdersR4Test` (already injects `myResourceSearchUrlDao`):

1. Enable the two placeholder settings; create an Observation with a conditional-URL
   subject reference.
2. Assert the `HFJ_RES_SEARCH_URL` row **exists** after the create commits (guards the
   retention property).
3. Conditionally update the Patient with real content.
4. Assert the row is **gone** after the update commits (the fix; fails on current master).
5. Regression: `FhirResourceDaoR4ConcurrentCreateTest` must remain green.

## Related issues

- #8037 — chunked the scheduled sweep so it completes on very large `HFJ_RES_SEARCH_URL`
  tables (single unbounded DELETE previously exceeded MSSQL's per-statement timeout).
  #8037 bounds the table when cleanup falls behind; this issue restores prompt cleanup so
  far less falls to the sweep in the first place. A deployment doing sustained
  placeholder-producing ingestion observed ~27 M accumulated rows when both effects
  combined.
- #8036 — earlier report of the same accumulation symptom.

## Environment

- HAPI FHIR: master (applies to all storage engines; observed on MSSQL)
- Affects JPA server with `AutoCreatePlaceholderReferenceTargets` enabled

jmarchionatto:
Register search URL sentinel rows for auto-created placeholder resources so they are cleaned up on first update
  
  Fixes #8036

  When AutoCreatePlaceholderReferenceTargets is enabled, the HFJ_RES_SEARCH_URL row created for an auto-created placeholder (to guard against concurrent duplicate conditional creates) was never registered for deletion-on-update. Unlike rows created by ordinary conditional creates, it lingered until the scheduled cleanup job ran. Under sustained ingestion this caused significant table growth, and if the placeholder's identifier changed on update, the stale row could incorrectly block a later conditional create for the same URL.

The placeholder's search URL row is now registered exactly like a conditional create's.
  
  Changes:
  - DaoResourceLinkResolver now passes the placeholder's identifier-based match URL to the create flow via a userdata key
  (JpaConstants.PLACEHOLDER_RESOURCE_SEARCH_URL) instead of inserting the search URL row itself after the fact.
  - BaseHapiFhirResourceDao.create() registers the search URL through the same enforceMatchUrlResourceUniqueness path used by ordinary conditional creates,
  so the row is marked for deletion at the placeholder's first update (and avoids the extra update statement the old approach required).
  - Removed JpaDaoResourceLinkResolver, whose only purpose was the old post-create row insertion; JpaConfig now wires DaoResourceLinkResolver<JpaPid>
  directly. 
  - Tests: a JPA test verifying the sentinel row exists after placeholder creation and is deleted on first update, plus unit tests for the match URL userdata
  passing in DaoResourceLinkResolverTest.


tadgh:
Formatting check succeeded!

epeartree (approved):

PR #8068, opened 2026-06-10
Skip installing older package versions in the same package install call

### LINKED ISSUE #8058: Package installer should skip installing older package version if newer version is already installed
When calling `PackageInstallerSvcImpl.install` with `fetchDependencies=true` and `PackageInstallationSpec.reloadExisting=true` (the default) the package dependencies in the dependency tree have no specific ordering. As a result, the same package can appear with a newer version and an older version afterwards in the installation order. 

For example:
```
  hl7.fhir.us.carin-bb#2.1.0
  ├── hl7.fhir.us.core#6.1.0
  │   └── hl7.terminology.r4#6.2.0    ← newer
  └── hl7.terminology.r4#5.4.0        ← older (direct dep)
```

For `versionPolicy=SINGLE_VERSION`, that will end up affecting all canonical resource types. For `versionPolicy=MULTI_VERSION`, only `SearchParameter` resources are generally affected, and other types where the version of the canonical resource is declared the same across different package versions.

codeforgreen:
Fixes #8058 

When installing a FHIR package whose dependency tree pulls in multiple versions of the same package, resources could be silently downgraded. This PR adds two levels of protection:

**Solution**

Package-level: A new `PackageDependencyRegistry` tracks installed dependency packages during installation. In `SINGLE_VERSION` mode, it detects redundant dependencies (exact duplicates and older versions) and skips them entirely.

Resource-level: A new `PackageVersionStamper` reads the meta.source field (stamped as `packageName|packageVersion` during installation) to compare incoming vs existing package versions. In `MULTI_VERSION` mode (default), this protects `SearchParameter` resources which match by code and base rather than URL and version. In `SINGLE_VERSION` mode, all canonical resources are protected.

**Changes**
  
  - `PackageInstallerSvcImpl` — Added skip logic in install() and fetchAndInstallDependencies() using the two new classes
  - `PackageVersionStamper` (new) — Handles `meta.source` stamping, parsing (using existing `UrlUtil.parseCanonicalUrl())`, and version comparison
  - `PackageDependencyRegistry` (new) — Tracks highest installed version per package name, detects redundant dependencies
  - `packages.md `— Added "Version Policy" section documenting `MULTI_VERSION` vs `SINGLE_VERSION` behavior and redundant dependency handling

**Test plan**

  - `PackageVersionStamperTest` — 13 unit tests for stamping, parsing, and version comparison
  - `PackageDependencyRegistryTest` — 9 unit tests for dependency tracking and redundancy detection
  - `PackageInstallerSvcImplTest` — 8 integration tests covering `SearchParameter` skip, `CodeSystem` passthrough/skip, request-id in `meta.source`, and cross-package scenarios

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8058-skip-older-search-parameter-overwrite-multi-version.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8058-skip-older-search-parameter-overwrite-multi-version.yaml
@@ -1,0 +1,8 @@
+---
+type: fix
+issue: 8058
+title: "Previously, installing a package whose dependency tree pulls in multiple versions of the same
+package could silently downgrade resources. In `SINGLE_VERSION` mode, all canonical resources were
+affected. In `MULTI_VERSION` mode (the default), `SearchParameter` resources were affected since they
+match by `code` and `base` rather than URL and version. This has been fixed and the
+installer now detects redundant dependencies and skips older versions."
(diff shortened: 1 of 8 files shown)


tadgh:
Formatting check succeeded!

jamesagnew (approved):

PR #8070, opened 2026-06-10
Add support for ISNOTA filter operator for ValueSet expansion in the Hibernate search path

### LINKED ISSUE #8069: Add ISNOTA filter operator support to Hibernate Search `ValueSet` expansion path
The normative FHIR R4 ValueSet `http://hl7.org/fhir/ValueSet/patient-contactrelationship` (FMM 5) uses is-not-a to include all v2-0131 Contact Role codes except "O" (Other). It is bound to Patient.contact.relationship.

On a server with Hibernate Search enabled, this ValueSet cannot be expanded because it does not support `ISNOTA` — it throws `HAPI-0894`.

Example scenario:

A client creates a `Patient` with an emergency contact:
  
```
  {
    "resourceType": "Patient",
    "contact": [{
      "relationship": [{"coding": [{"system": "http://terminology.hl7.org/CodeSystem/v2-0131", "code": "C"}]}],
      "name": {"text": "Jane Doe"}
    }]
  }
```
  
Code "C" (Emergency Contact) is valid in this `ValueSet`. However, when the server validates this resource:

1. Pre-expansion of `patient-contactrelationship` failed with `HAPI-0894` because `ISNOTA` is unsupported.
2. Validation falls back to in-memory, which cannot access the persisted `CodeSystem` hierarchy. The in-memory expansion returns an empty result.
3. Validation rejects the valid code "C" as unknown with an error like `"Unknown code 'http://terminology.hl7.org/CodeSystem/v2-0131#C'" `— a false negative with no indication that the root cause is an unsupported filter operator.

codeforgreen:
Fixes #8069


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8069-fix-isnota-filter-hibernate-search.yaml
```diff
@@ -0,0 +1,6 @@
+---
+type: fix
+issue: 8069
+title: "Previously, `ValueSet` expansion with Hibernate Search enabled would fail with `HAPI-0894` when using
```
TipzCM, line 4: i'm not sure how helpful the number is to people.

Maybe more a "would fail because...."

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/BaseValueSetHSearchExpansionR4Test.java
```diff
@@ -1736,6 +1772,92 @@ public void testSearch() {
 			assertThat(codes).isEmpty();
 		}
 
+		// Created by Claude Opus 4.6
+		@Test
+		void testExpandWithIsAFilter() {
```
TipzCM: fyi, naming convention is: methodOrFeatureUnderTest_condtions_expectedResult

TipzCM (approved):

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/test/BaseValueSetHSearchExpansionR4Test.java
codeforgreen: I renamed tests on this topic and did some cleanup/simplificaton.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8069-fix-isnota-filter-hibernate-search.yaml
codeforgreen: HAPI errors are useful as that they're the way we uniquely identify issues, some of which can actually be troubleshooted and fixed by the client. This is one of those cases where the user would go and enable the setting `StorageSettings.setHibernateSearchIndexFullText(true)` which is disabled (false) by default.

PR #8072, opened 2026-06-11
Fix consent + bulk export IndexOutOfBoundsException in JpaPreResourceAccessDetails

### LINKED ISSUE #8071: Consent + bulk export: IndexOutOfBoundsException in JpaPreResourceAccessDetails.getResource when search PIDs don't all load
### Bug

`Patient/$export` (bulk export) throws `IndexOutOfBoundsException: Index 0 out of bounds for length 0` on the `STORAGE_PREACCESS_RESOURCES` pointcut when a Consent module is active, instead of producing a (consent-filtered) report.

```
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
	at java.util.ArrayList.get(ArrayList.java:427)
	at ca.uhn.fhir.jpa.interceptor.JpaPreResourceAccessDetails.getResource(JpaPreResourceAccessDetails.java:64)
	at ca.uhn.fhir.rest.server.interceptor.consent.ConsentInterceptor.interceptPreAccess(ConsentInterceptor.java:238)
	at ca.uhn.fhir.batch2.jobs.export.v3.ExpandResourceAndWriteBinaryStep.searchForResources(...)
```

### Root cause

`JpaPreResourceAccessDetails.size()` returns the PID count, but `getResource(idx)` indexes into a list that `SearchBuilder.loadResourcesByPid` may fill with **fewer** elements (it skips deleted/unloadable resources). `ConsentInterceptor.interceptPreAccess` loops over `size()` and calls `getResource(idx)`, overrunning the list when trailing — or all — PIDs fail to load. A naive `null` return alone is unsafe: `interceptPreAccess` has no null guard and passes the resource into `IConsentService#canSeeResource`, which dereferences it (NPE).

### Expected

The export completes; resources whose bodies can't be loaded are dropped from results, not consulted by consent.

### Fix

1. `JpaPreResourceAccessDetails.getResource(idx)` returns `null` when `idx >= myResources.size()`.
2. `ConsentInterceptor.interceptPreAccess` skips and `setDontReturnResourceAtIndex(idx)` when `getResource(idx)` is `null`.

### Environment

HAPI FHIR `8.8.0` (identical on `master`). Modules: `hapi-fhir-jpaserver-base`, `hapi-fhir-server`.

YalingPeiS:
## Summary

Fixes an `IndexOutOfBoundsException` that crashed `Patient/$export` (bulk export) when a Consent interceptor was active. `JpaPreResourceAccessDetails.size()` reports the search-result PID count, but `getResource(idx)` indexes into a list that `SearchBuilder.loadResourcesByPid` may fill with fewer elements (it skips deleted/unloadable resources). `ConsentInterceptor.interceptPreAccess` loops over `size()` and overran the list when trailing/all PIDs failed to load. Not a regression — identical on `8.8.0` and `master`.

## Changes

1. **`JpaPreResourceAccessDetails.getResource(int)`** — returns `null` when `idx >= myResources.size()` instead of letting `ArrayList.get` throw.
2. **`ConsentInterceptor.interceptPreAccess`** — null-guards `getResource(idx)`: skips consent and calls `setDontReturnResourceAtIndex(idx)`, so an unloadable resource is dropped rather than crashing the job or being passed as `null` into `canSeeResource` (which consent services dereference). Mirrors the existing guard in `interceptPreShow`.

Tests: `JpaPreResourceAccessDetailsTest` (zero/trailing/all/middle-null cases) and `ConsentInterceptorPreAccessNullResourceTest` (consent skipped, index marked, no NPE).

## Review notes

- `getResource(idx)` may now return `null` for a valid `idx < size()`; all hook consumers read only `isDontReturnResourceAtIndex(...)` afterward, so marking the index cleanly drops the PID.
- Negative index intentionally still fails fast (caller bug); only the upper bound returns `null`.

## QA

- With a Consent interceptor active and some search-matched resources unloadable (e.g. soft-deleted mid-page), run `Patient/$export` → job completes, no `IndexOutOfBoundsException`; loadable resources still returned and consent-evaluated; unloadable slots excluded.

Closes #8071


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/JpaPreResourceAccessDetails.java
```diff
@@ -61,6 +65,10 @@ public IBaseResource getResource(int theIndex) {
 					.call()
 					.loadResourcesByPid(myResourcePids, Collections.emptySet(), myResources, false, null);
 		}
+		// loadResourcesByPid skips unloadable PIDs, so myResources may be shorter than size().
```
codeforgreen: I won't be able to understand this comment without further diving deep. Comments should be easy to understand or not needed at all.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/JpaPreResourceAccessDetails.java
```diff
@@ -53,6 +53,10 @@ public int size() {
 		return myResourcePids.size();
 	}
 
+	/**
+	 * Returns the resource at the given index, or {@code null} if its PID had no loadable body.
```
codeforgreen, line 57: Does the loading happen here? If no then why is this relevant to understanding this method?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/consent/ConsentInterceptor.java
```diff
@@ -240,6 +240,11 @@ public void interceptPreAccess(
 				getAlreadySeenResourcesMap(theRequestDetails);
 		for (int resourceIdx = 0; resourceIdx < thePreResourceAccessDetails.size(); resourceIdx++) {
 			IBaseResource nextResource = thePreResourceAccessDetails.getResource(resourceIdx);
```
codeforgreen, line 242: question (probably not for this PR): I wonder if we can avoid iterating using an index in the first place. I tried to get rid of this once but I know there are other loops depending on it, I just wonder if there is an easier smarter way to do this. Maybe the scope for a larger refactoring.

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/consent/ConsentInterceptorPreAccessNullResourceTest.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * {@link ConsentInterceptor#interceptPreAccess} must safely handle the case where
+ * {@link IPreResourceAccessDetails#getResource(int)} returns {@code null} for a search PID that has
+ * no loadable resource body. The interceptor must not pass that {@code null} into a consent
+ * service's {@code canSeeResource}, which production services dereference.
+ */
+public class ConsentInterceptorPreAccessNullResourceTest {
```
codeforgreen: question: Does this scenario need it's own dedicated test? I am thinking we will probably be getting rid of it when the code is refactored and this scenario is not valid anymore?
suggestion: Another thing I notice is there are lots of comments that Claude adds to tests. Would be good to trim it down to necessary ones as code is self-explainatory.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/interceptor/JpaPreResourceAccessDetailsTest.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.Mockito.doAnswer;
+
+// Created by claude-opus-4-8
+@ExtendWith(MockitoExtension.class)
+class JpaPreResourceAccessDetailsTest {
```
codeforgreen, line 24: suggestion: Same as the other comment. This test may not be relevant in the future. It ties to an implementation detail which may go away. I think an integration test for the exact scenario in the bug is what we need.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/JpaPreResourceAccessDetails.java
codeforgreen: oh I see the `loadResourceByPid` above. 
It looks like the root cause may be in `SearchBuilder.doLoadPids`. It does padding in the loop however there are 2 skip (continue) conditions which never reach padding.

YalingPeiS: The method `loadResourceByPid` mentioned in the comment is right above the comment (line 66). I think this comment is necessary since it provides insight to the reader about what could be in `myResources` after `loadResourceByPid` was called.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/interceptor/JpaPreResourceAccessDetailsTest.java
YalingPeiS: I am not sure what you mean that the test would not be relevant in the future. If we change code in the future, we should change the associated tests accordingly. However, the goal of the unit tests here is to verify that the fix works as expected.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/consent/ConsentInterceptor.java
YalingPeiS: Agree that there is some tech debt. We probably need to do more investigation before code refactoring to avoid breaking existing functionalities. If you've already looked into this, you may keep track of your findings in an issue so it can be worked on later. Agree that it is out of scope of this issue.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/JpaPreResourceAccessDetails.java
YalingPeiS: Yes. This method calls `loadResourceByPid`, in which the resource was loaded if it had not been populated yet. This is a `public` method and I think it is good practice to add Javadoc/comment to it.

hapi-fhir-server/src/test/java/ca/uhn/fhir/rest/server/interceptor/consent/ConsentInterceptorPreAccessNullResourceTest.java
YalingPeiS: Good catch. I moved the tests to the existing `ConsentInterceptorTest` and deleted this test class.

YalingPeiS:
It appears that the issue could no long be reproduced after I checked out the latest version. 
Confirmed with QA that the issue has been resolved. Close this PR.

PR #8091, opened 2026-06-16
GL-8842: Widen MDM EID-match-only bulk export expansion to all partitions

epeartree:
**Companion CDR MR (integration tests):** https://gitlab.com/simpatico.ai/cdr/-/merge_requests/7286

## Summary

This fixes a regression introduced in PR #7487 ("Add Patient ID mode resource policies") where Group/Patient bulk `$export` with auto-MDM expansion in EID-match-only mode failed under `PATIENT_ID` partitioning with `HAPI-2001: Resource ... is not known`. The root cause was that `PatientIdPartitionInterceptor` began resolving Group reads to the default partition only, but `BulkExportMdmEidMatchOnlyResourceExpander` was never updated to widen member expansion to all partitions — so group members living on their own `PATIENT_ID` partitions were invisible during EID-match-only expansion.

1. `BulkExportMdmEidMatchOnlyResourceExpander` now calls `determineExpansionPartition()` before expanding group members and resolving PIDs: when `IMdmSettings.getSearchAllPartitionForMatch()` is true, it widens to `RequestPartitionId.allPartitions()` instead of using the request partition (which is the Group's partition, not the members').
2. `IBulkExportMdmEidMatchOnlyResourceExpander` is no longer a marker interface — it now declares `setMdmSettings(IMdmSettings)` so implementations can receive settings.
3. `MdmExpandersHolder.setMdmSettings` cascades settings into the EID-match-only expander, mirroring the existing pattern for the full-MDM search-expansion path in `MdmSearchExpansionSvc`.
4. Scope is intentionally narrow: the Group read itself stays on its resolved partition; only member link expansion and PID resolution widen. Full-MDM expansion is unaffected.
5. Integration tests (Group `$export` / `$davinci-data-export` auto-MDM expansion across `PATIENT_ID` partitions) live in the companion CDR MR and pass GREEN against this fix.

## Code Review Suggestions

- [ ] **`myMdmSettings` nullability in `determineExpansionPartition`**: The field is initialized to `null` and set later via `setMdmSettings`. The null guard is correct, but reviewers should confirm that `MdmExpandersHolder.setMdmSettings` is always called before any bulk export invocation. If MDM settings are not available (e.g., MDM not configured), the null path silently falls back to the request partition.
- [ ] **Interface change to `IBulkExportMdmEidMatchOnlyResourceExpander`**: Adding `setMdmSettings` is non-breaking within HAPI (the only known implementation is `BulkExportMdmEidMatchOnlyResourceExpander`), but downstream implementors outside the repo would require an update. Confirm this interface is considered internal/non-API.
- [ ] **`expandPatient` path also widened**: Confirm a Patient bulk export with EID-match-only should also cross partition boundaries for the same reasons as Group export.
- [ ] **Ordering in `setMdmSettings`**: The cascade is called before `determineBulkExportMDMResourceExpanderInstanceToUse()`. Verify the ordering does not matter.

### Setup
- [ ] HAPI-FHIR server with MDM in EID-match-only mode and `PATIENT_ID` partitioning active.
- [ ] Enable `searchAllPartitionForMatch` in MDM settings.
- [ ] Create a Group on partition A with Patient members assigned to their own `PATIENT_ID` partitions (B, C).

### Test Cases
- [ ] **Group bulk export finds cross-partition members**: Trigger a Group `$export`; confirm all members — including those on non-default partitions — appear in the output.
- [ ] **`HAPI-2001` no longer thrown** during a partition-spanning Group bulk export.
- [ ] **`searchAllPartitionForMatch = false` baseline**: confirm expansion does NOT widen — only same-partition members returned (fix gated correctly).
- [ ] **Patient bulk export (EID-match-only)**: confirm MDM-linked patients on other partitions resolve.
- [ ] **Full-MDM path unaffected**: existing Group/Patient export behavior unchanged.
- [ ] **Regression vs HAPI 8.9.6**: behavior matches pre-#7487; the companion CDR MR's integration tests are the primary signal.

Jira: SMILE-11727 · GitLab issue: simpatico.ai/cdr#8842

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/bulk/export/svc/BulkExportMdmEidMatchOnlyResourceExpander.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/bulk/export/svc/BulkExportMdmEidMatchOnlyResourceExpander.java
@@ -26,6 +26,7 @@ import ca.uhn.fhir.jpa.api.dao.IFhirResourceDao;
 import ca.uhn.fhir.jpa.api.svc.IIdHelperService;
 import ca.uhn.fhir.jpa.api.svc.ResolveIdentityMode;
 import ca.uhn.fhir.jpa.model.dao.JpaPid;
+import ca.uhn.fhir.mdm.api.IMdmSettings;
 import ca.uhn.fhir.mdm.svc.IBulkExportMdmEidMatchOnlyResourceExpander;
 import ca.uhn.fhir.mdm.svc.IBulkExportMdmResourceExpander;
 import ca.uhn.fhir.mdm.svc.MdmEidMatchOnlyExpandSvc;
@@ -57,6 +58,12 @@ public class BulkExportMdmEidMatchOnlyResourceExpander implements IBulkExportMdm
 	private final IIdHelperService<JpaPid> myIdHelperService;
 
 	/**
+	 * MDM settings. Mutable because settings are not available at construction time and are pushed in
+	 * later via {@link #setMdmSettings(IMdmSettings)} by {@code MdmExpandersHolder}.
+	 */
+	private IMdmSettings myMdmSettings;
+
+	/**
 	 * Constructor
 	 */
 	public BulkExportMdmEidMatchOnlyResourceExpander(
@@ -68,6 +75,24 @@ public class BulkExportMdmEidMatchOnlyResourceExpander implements IBulkExportMdm
 		myMdmEidMatchOnlyLinkExpandSvc = theMdmEidMatchOnlyLinkExpandSvc;
 		myFhirContext = theFhirContext;
 		myIdHelperService = theIdHelperService;
+	}
+
+	@Override
+	public void setMdmSettings(IMdmSettings theMdmSettings) {
+		myMdmSettings = theMdmSettings;
+	}
+
+	/**
+	 * Determines the partition to use when expanding group/patient members. When MDM is configured to
+	 * search all partitions for matching, members may live on partitions other than the request partition
+	 * (e.g. the Group's default partition), so expansion must widen to {@link RequestPartitionId#allPartitions()}.
+	 * Otherwise the original request partition is used.
+	 */
+	private RequestPartitionId determineExpansionPartition(RequestPartitionId theRequestPartitionId) {
+		if (myMdmSettings != null && myMdmSettings.getSearchAllPartitionForMatch()) {
+			return RequestPartitionId.allPartitions();
+		}
+		return theRequestPartitionId;
 	}
 
 	/**
@@ -96,6 +121,8 @@ public class BulkExportMdmEidMatchOnlyResourceExpander implements IBulkExportMdm
 		IFhirResourceDao<?> groupDao = myDaoRegistry.getResourceDao("Group");
 		IBaseResource groupResource = groupDao.read(groupId, srd);
 
+		RequestPartitionId expansionPartition = determineExpansionPartition(requestPartitionId);
+
 		Set<String> allResourceIds = new HashSet<>();
 		FhirTerser terser = myFhirContext.newTerser();
 		// Extract all member.entity references from the Group resource
@@ -106,7 +133,7 @@ public class BulkExportMdmEidMatchOnlyResourceExpander implements IBulkExportMdm
 			if (!entityRef.getReferenceElement().isEmpty()) {
 				IIdType memberId = entityRef.getReferenceElement();
 				Set<String> expanded =
-						myMdmEidMatchOnlyLinkExpandSvc.expandMdmBySourceResourceId(requestPartitionId, memberId);
+						myMdmEidMatchOnlyLinkExpandSvc.expandMdmBySourceResourceId(expansionPartition, memberId);
 				allResourceIds.addAll(expanded);
 			}
 		}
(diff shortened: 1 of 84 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/bulk/export/svc/BulkExportMdmEidMatchOnlyResourceExpanderTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+	@SuppressWarnings("rawtypes")
+	@Mock
+	private IFhirResourceDao myGroupDao;
+
+	private BulkExportMdmEidMatchOnlyResourceExpander myExpander;
+
+	@BeforeEach
+	void beforeEach() {
+		myExpander = new BulkExportMdmEidMatchOnlyResourceExpander(
```
TipzCM, line 76: could also just use @InjectMocks

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/bulk/export/svc/BulkExportMdmEidMatchOnlyResourceExpanderTest.java
```diff
@@ -0,0 +1,162 @@
...
+		when(myMdmEidMatchOnlyExpandSvc.expandMdmBySourceResourceId(any(), any()))
+				.thenReturn(Set.of("Patient/bulk-p1"));
+		when(myIdHelperService.resolveResourcePids(
+						any(RequestPartitionId.class), anyList(), any(ResolveIdentityMode.class)))
+				.thenReturn(List.of(JpaPid.fromId(1L)));
+	}
+
+	@ParameterizedTest
+	@ValueSource(booleans = {true, false})
+	void expandGroup_widensMemberExpandAndResolveOnlyWhenSearchAllPartitionForMatchEnabled(
```
TipzCM: naming convention is: methodOrFeatureUnderTest_conditions_expectedOutcome

So,

`expandGroup_withAllSearchPartitionsEnabled_widensMemberExpand` (or similar)

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/svc/MdmExpandersHolder.java
```diff
@@ -176,6 +176,9 @@ private IMdmLinkExpandSvc determineExpandSvsInstanceToUse() {
 	public void setMdmSettings(IMdmSettings theMdmSettings) {
 		myMdmSettings = theMdmSettings;
 		myLinkExpandSvcInstanceToUse = determineExpandSvsInstanceToUse();
+		if (myBulkExportMDMEidMatchOnlyResourceExpander != null) {
```
TipzCM: This looks like it's injected at construction time.

When would this be null?

If there is a possibility, it should be commented (imo); if not, we can remove the null check

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/bulk/export/svc/BulkExportMdmEidMatchOnlyResourceExpander.java
```diff
@@ -70,6 +77,24 @@ public BulkExportMdmEidMatchOnlyResourceExpander(
...
+	}
+
+	/**
+	 * Determines the partition to use when expanding group/patient members. When MDM is configured to
+	 * search all partitions for matching, members may live on partitions other than the request partition
+	 * (e.g. the Group's default partition), so expansion must widen to {@link RequestPartitionId#allPartitions()}.
+	 * Otherwise the original request partition is used.
+	 */
+	private RequestPartitionId determineExpansionPartition(RequestPartitionId theRequestPartitionId) {
+		if (myMdmSettings != null && myMdmSettings.getSearchAllPartitionForMatch()) {
```
TipzCM: Is there any chance this might be called before the settings are passed down?

Ie, a job resumes during startup before MDM has initialized?

TipzCM (approved):

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/bulk/export/svc/BulkExportMdmEidMatchOnlyResourceExpanderTest.java
epeartree: i'll keep the before each because ourFhirContext is not a mock and injectMock only stitches mocked object

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/svc/MdmExpandersHolder.java
epeartree: addressed.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/bulk/export/svc/BulkExportMdmEidMatchOnlyResourceExpanderTest.java
epeartree: addressed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/bulk/export/svc/BulkExportMdmEidMatchOnlyResourceExpander.java
epeartree: u are correct, that can't happen

PR #8094, opened 2026-06-16
Enable custom partition resolution for bulk operations $delete-expunge, $mdm-submit and $mdm-clear

### LINKED ISSUE #8083: Enable custom partition mapping for bulk operations $delete-expunge, $mdm-submit, and $mdm-clear
The purpose of this issue is to enable custom partition resolution for bulk operations `$delete-expunge`, `$mdm-submit` and `$mdm-clear`.

#### Scenario
When a user invokes the `$delete-expunge` operation with a URL like `Patient?active=false`. The partition resolver returns `RequestPartitionId.allPartitions()` for that URL. A custom `IJobPartitionProvider` is configured that expands `allPartitions` into individual per-partition work chunks (e.g. partition 1, 2, 3). 

#### Actual behavior
This expansion works for `$reindex`, but has no effect on `$delete-expunge`, `$mdm-submit`, or `$mdm-clear` — these operations bypass `IJobPartitionProvider` and pass the raw `allPartitions` token through as-is. 

#### Desired behavior
The job still executes correctly, but per-partition fan-out does not happen. The purpose of this issue is to enable custom partition resolution for these operations.

codeforgreen:
Fixes #8083


tadgh:
Formatting check succeeded!

PR #8096, opened 2026-06-17
Reduce id-sequence lock contention under high write concurrency (opt-in per-thread pooling) 

### LINKED ISSUE #8090: Global sequence-ID pool lock in HapiSequenceStyleGenerator serializes all writer threads during pool refill
`HapiSequenceStyleGenerator` configures Hibernate's `pooled` optimizer
(`StandardOptimizerDescriptor.POOLED`) for all JPA-assigned database IDs. Hibernate's
`PooledOptimizer.generate()` guards its entire body with a single shared lock, held for
the duration of the `nextval` database round-trip whenever the pool is exhausted. As a
result, on a busy node every thread that needs an ID is serialized behind whichever
thread is currently refilling the pool. Under high write concurrency this produces
periodic throughput stalls.

### Observed behavior

- Thread dumps under load show one thread inside
  `HapiSequenceStyleGenerator.generate` → `PooledOptimizer.generate` →
  `SequenceStructure...getNextValue` → JDBC `executeQuery`, waiting on the DB,
  while many other worker threads are `BLOCKED`/`TIMED_WAITING` on the same lock,
  all parked in `PooledOptimizer.generate`.
- On PostgreSQL backends, the blocked DB round-trips appear as `ClientRead` wait
  events in `pg_stat_activity`.
- Aggregate write throughput drops sharply at each pool refill.
- The effect is amplified on nodes with more worker threads and on deployments with
  higher DB latency (e.g. routed/sharded PostgreSQL such as Aurora Limitless), but it
  applies to any PostgreSQL/Aurora backend.

### Expected behavior

ID generation should not serialize unrelated writer threads. A thread refilling its pool
should not block other threads from allocating IDs.

### Steps to reproduce

1. Run a JPA server with a PostgreSQL backend.
2. Drive concurrent write traffic (e.g. ~40 concurrent clients submitting transaction
   bundles) with a load generator.
3. Capture thread dumps during load — observe many threads blocked in
   `PooledOptimizer.generate` behind a single lock holder doing a sequence fetch.
4. Observe `ClientRead` waits in `pg_stat_activity` and throughput dips correlated with
   pool refills.

### Steps to reproduce

1. Run a JPA server with a PostgreSQL backend.
2. Drive concurrent write traffic (e.g. ~40 concurrent clients submitting transaction
   bundles) with a load generator.
3. Capture thread dumps during load — observe many threads blocked in
   `PooledOptimizer.generate` behind a single lock holder doing a sequence fetch.
4. Observe `ClientRead` waits in `pg_stat_activity` and throughput dips correlated with
   pool refills.

### Proposed fix

Switch the optimizer to a per-thread pooling strategy so threads do not share a single
lock. Hibernate provides `PooledLoThreadLocalOptimizer`
(`StandardOptimizerDescriptor.POOLED_LOTL`, external name `"pooled-lotl"`), whose
`generate()` reads from a `ThreadLocal<GenerationState>` with no shared lock — each thread
fetches and pools its own range independently.

Constraints / things to validate before adopting:

- **Negative-value interaction.** `HapiSequenceStyleGenerator` deliberately moved off
  `pooled_lo` to `pooled` because the LO algorithm interacts badly with out-of-order
  sequence reads on Aurora Limitless and can yield negative values (see the comment at
  lines 124-125). `pooled-lotl` is the thread-local variant of the same LO algorithm, so
  this interaction must be re-tested on Limitless before switching. The `INITIAL_PARAM`
  guard may need re-examination under the LO variant.
- **Wasted ID ranges.** A thread-local pool means each thread reserves its own block of
  `incrementSize` IDs; idle/short-lived threads can leave gaps. IDs are opaque and gaps
  are already possible today, but the increment size (currently 50) should be reviewed in
  light of per-thread reservation.
- **No ordering assumptions on generated IDs.** With per-thread pools, IDs are no longer
  handed out in global creation order — a resource created later on one thread can receive
  a lower ID than one created earlier on another thread (the current `pooled` optimizer
  already allows some reordering, but the thread-local variant makes it routine). Before
  switching, audit the codebase for any logic that treats a generated PID as a creation
  sequence / monotonic ordering signal: e.g. ordering, paging, "last created", change-
  detection, or cursor logic keyed on the numeric resource PID. Such logic must use an
  explicit timestamp/version column instead, not the ID.
- **Backwards compatibility.** As with the existing comment's note about `pooled_hi`, any
  change must be safe for databases already advanced by the current `pooled` optimizer
  (continuity of allocated ranges, no collisions with already-issued IDs).

An alternative that needs no algorithm change is to increase `INCREMENT_PARAM` to reduce
refill frequency, but that only reduces — does not eliminate — the contention and widens
ID gaps; the thread-local optimizer is the structural fix.

### Acceptance criteria

- Concurrent ID allocation does not serialize on a single shared lock; a pool refill on
  one thread does not block ID allocation on others.
- No negative or colliding IDs on PostgreSQL/Aurora, including Aurora Limitless
  (out-of-order sequence reads).
- Confirmed (by code audit) that no code path relies on generated resource PIDs being
  monotonic / creation-ordered; any such logic is migrated to an explicit timestamp or
  version column.
- Existing sequence-reset test infrastructure
  (`hapi-fhir-jpaserver-test-utilities/.../ResetSequencesTestHelper.java`, which currently
  reflects into `PooledOptimizer` internal state) is updated for whatever optimizer is
  selected.
- A test demonstrating non-blocking concurrent allocation.

### Environment

- Hibernate 6.6.4.Final
- PostgreSQL / Amazon Aurora (incl. Aurora Limitless) backends
- High write concurrency

jmarchionatto:
Under high write concurrency, database id generation could bottleneck because all writer threads on a server shared a single id pool guarded by one lock, held across the nextval() round-trip during pool refills.

  This adds an option to allocate ids from a per-thread pool instead, so concurrent writers each draw from their own block and never serialize on that lock.

  Behavior / compatibility
  - Disabled by default — existing deployments are unaffected on upgrade. Enabled via JpaStorageSettings#setPerThreadIdSequencePoolingEnabled(true).
  - The two optimizers interpret the same DB sequence value differently, so per-thread pooling must be enabled cluster-wide in one coordinated restart, never node-by-node (documented in upgrade.md).
  - When enabled, ids are no longer handed out in strict creation order across threads. The package installer's "most recent existing version" selection now sorts by _lastUpdated (with internal id only as a tiebreaker) instead of by internal id.

  Tests
  - New HapiSequenceStyleGeneratorTest — guards optimizer selection and the reserved-value (HAPI-2791) guard.
  - ResetSequencesTestHelper handles both optimizers.
  - Retained the server-assigned reserved-value integration test in DeletedDatabaseRowsR5Test.


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/model/dialect/HapiSequenceStyleGeneratorTest.java
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	@Test
+	void perThreadPoolingDisabled_selectsLegacyPooledOptimizer() {
+		assertThat(HapiSequenceStyleGenerator.determineOptimizerExternalName(false))
+				.isEqualTo(StandardOptimizerDescriptor.POOLED.getExternalName());
+	}
+
+	@Test
+	void generateNonReservedValue_returnsValue_whenNotReserved() {
```
TipzCM: the naming convention is backwards:

methodOrFeature_conditions_result

TipzCM (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/model/dialect/HapiSequenceStyleGeneratorTest.java
jmarchionatto: Fixed

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/upgrade.md
```diff
@@ -26,3 +26,29 @@ callers are not broken.
...
+Database id generation can now be configured to allocate ids from a per-thread pool, instead of a single pool shared by
+all writer threads on a server. This removes lock contention between concurrent writers when the pool is refilled, and
+improves write throughput under high write concurrency.
+
+This behavior is **disabled by default**, so existing deployments are unaffected on upgrade. It can be enabled with
+`JpaStorageSettings#setPerThreadIdSequencePoolingEnabled(true)`.
+
+When per-thread pooling is enabled, ids are no longer handed out in strict creation order across threads (an id assigned
+later on one thread may be lower than one assigned earlier on another thread), so code must not treat the numeric
+internal id as a creation-order signal; use the last-updated time instead.
```
michaelabuckley: comment: this was never true between servers.  what has changed with this PR is that is no longer true even between threads on the same server.  I might simplify this to just:

```suggestion
Note that between different servers, ids have never  been allocated in strict creation order. So code must not treat the numeric
internal id as a creation-order signal; use the last-updated time instead.
```

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.java
```diff
@@ -1742,6 +1743,34 @@ public void setUniqueIndexesEnabled(boolean theUniqueIndexesEnabled) {
...
+	 * If set to <code>true</code> (default is <code>false</code>), the database sequence id generator
+	 * allocates ids from a pool kept per thread, so that concurrent writers do not serialize on a
+	 * single shared pool lock while the pool is refilled from the database. The default is
+	 * <code>false</code> (the legacy single shared pool) so that existing deployments are unaffected on
+	 * upgrade; enabling it should be done across an entire cluster at once (see the upgrade notes), never
+	 * node by node, because the two behaviors interpret the same database sequence differently.
+	 *
+	 * @since 8.12.0
+	 */
+	public boolean isPerThreadIdSequencePoolingEnabled() {
```
michaelabuckley: question: do we want a bare boolean here?  Should this be an enum? Hibernate has many allocators, and we may want to allow others.  Especially in AWS Limitless, we may want to do our own thing.

michaelabuckley (approved):

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.java
jmarchionatto: Changed as suggested

PR #8097, opened 2026-06-17
Improve progress calculation for Batch2 jobs

jamesagnew:
The mechanism for calculating Batch2 job progress has been improved to allow specific steps in the job to have different weights in the overall progress calculation. This allows jobs with unbalanced effort levels between steps to have more accurate progress returned to the user.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
@@ -101,7 +101,8 @@ public class ImportCustomTerminologyJobAppCtx {
 						importCustomTerminologyStepGenerateConceptClosures())
 				// This step doesn't gain any work chunks until the previous step, we want to give it
 				// a fixed portion of the overall progress
-				.setStepWeightForProgressCalculator(STEP_ID_GENERATE_CONCEPT_CLOSURES, STEP_WEIGHT_GENERATE_CONCEPT_CLOSURES)
+				.setStepWeightForProgressCalculator(
+						STEP_ID_GENERATE_CONCEPT_CLOSURES, STEP_WEIGHT_GENERATE_CONCEPT_CLOSURES)
 				.addFinalReducerStep(
 						STEP_ID_FINALIZE_IMPORT,
 						"Finalize ICD-10 Import",
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
@@ -97,12 +97,15 @@ public class ImportCustomTerminologyJobAppCtx {
 						"Generate concept closures",
 						TerminologyFileSetJson.class,
 						importIcdStepGenerateConceptClosures())
+				// This step doesn't gain any work chunks until the previous step, we want to give it
+				// a fixed portion of the overall progress
 				.setStepWeightForProgressCalculator(STEP_ID_GENERATE_CONCEPT_CLOSURES, 0.3)
 				.addFinalReducerStep(
 						STEP_ID_FINALIZE_IMPORT,
 						"Finalize ICD-10 Import",
 						ImportTerminologyResultJson.class,
 						importCustomTerminologyStepFinalize())
+				// This step takes very little time and shouldn't factor significantly into the progress
 				.setStepWeightForProgressCalculator(STEP_ID_FINALIZE_IMPORT, 0.01)
 				.build();
 	}
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
@@ -29,9 +29,11 @@ import ca.uhn.fhir.jpa.term.api.ITermCodeSystemStorageSvc;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
-import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_CHUNK_CONCEPTS_FOR_CLOSURE_GENERATION;
-import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_FINALIZE_IMPORT;
-import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_GENERATE_CONCEPT_CLOSURES;
+import static ca.uhn.fhir.jpa.batch2.jobs.term.base.TerminologyConstants.STEP_ID_CHUNK_CONCEPTS_FOR_CLOSURE_GENERATION;
+import static ca.uhn.fhir.jpa.batch2.jobs.term.base.TerminologyConstants.STEP_ID_FINALIZE_IMPORT;
+import static ca.uhn.fhir.jpa.batch2.jobs.term.base.TerminologyConstants.STEP_ID_GENERATE_CONCEPT_CLOSURES;
+import static ca.uhn.fhir.jpa.batch2.jobs.term.base.TerminologyConstants.STEP_WEIGHT_FINALIZE;
+import static ca.uhn.fhir.jpa.batch2.jobs.term.base.TerminologyConstants.STEP_WEIGHT_GENERATE_CONCEPT_CLOSURES;
 
 @Configuration
 public class ImportCustomTerminologyJobAppCtx {
@@ -99,14 +101,14 @@ public class ImportCustomTerminologyJobAppCtx {
 						importCustomTerminologyStepGenerateConceptClosures())
 				// This step doesn't gain any work chunks until the previous step, we want to give it
 				// a fixed portion of the overall progress
-				.setStepWeightForProgressCalculator(STEP_ID_GENERATE_CONCEPT_CLOSURES, 0.3)
+				.setStepWeightForProgressCalculator(STEP_ID_GENERATE_CONCEPT_CLOSURES, STEP_WEIGHT_GENERATE_CONCEPT_CLOSURES)
 				.addFinalReducerStep(
 						STEP_ID_FINALIZE_IMPORT,
 						"Finalize ICD-10 Import",
 						ImportTerminologyResultJson.class,
 						importCustomTerminologyStepFinalize())
 				// This step takes very little time and shouldn't factor significantly into the progress
-				.setStepWeightForProgressCalculator(STEP_ID_FINALIZE_IMPORT, 0.01)
+				.setStepWeightForProgressCalculator(STEP_ID_FINALIZE_IMPORT, STEP_WEIGHT_FINALIZE)
 				.build();
 	}
 
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
@@ -31,6 +31,7 @@ import org.springframework.context.annotation.Configuration;
 
 import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_CHUNK_CONCEPTS_FOR_CLOSURE_GENERATION;
 import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_FINALIZE_IMPORT;
+import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_GENERATE_CONCEPT_CLOSURES;
 
 @Configuration
 public class ImportCustomTerminologyJobAppCtx {
@@ -92,15 +93,17 @@ public class ImportCustomTerminologyJobAppCtx {
 						TerminologyFileSetJson.class,
 						importIcdStepChunkConceptsForClosureGeneration())
 				.addIntermediateStep(
-						"generate-concept-closures",
+						STEP_ID_GENERATE_CONCEPT_CLOSURES,
 						"Generate concept closures",
 						TerminologyFileSetJson.class,
 						importIcdStepGenerateConceptClosures())
+				.setStepWeightForProgressCalculator(STEP_ID_GENERATE_CONCEPT_CLOSURES, 0.3)
 				.addFinalReducerStep(
 						STEP_ID_FINALIZE_IMPORT,
 						"Finalize ICD-10 Import",
 						ImportTerminologyResultJson.class,
 						importCustomTerminologyStepFinalize())
+				.setStepWeightForProgressCalculator(STEP_ID_FINALIZE_IMPORT, 0.01)
 				.build();
 	}
 
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/snomedct/ImportSnomedCtJobAppCtx.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/snomedct/ImportSnomedCtJobAppCtx.java
@@ -111,7 +111,8 @@ public class ImportSnomedCtJobAppCtx {
 						importSnomedStep5GenerateConceptClosures())
 				// This step doesn't gain any work chunks until the previous step, we want to give it
 				// a fixed portion of the overall progress
-				.setStepWeightForProgressCalculator(STEP_ID_GENERATE_CONCEPT_CLOSURES, STEP_WEIGHT_GENERATE_CONCEPT_CLOSURES)
+				.setStepWeightForProgressCalculator(
+						STEP_ID_GENERATE_CONCEPT_CLOSURES, STEP_WEIGHT_GENERATE_CONCEPT_CLOSURES)
 				.addFinalReducerStep(
 						STEP_ID_FINALIZE_IMPORT,
 						"Finalize Snomed Import",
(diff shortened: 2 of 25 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/snomedct/ImportSnomedCtJobAppCtx.java
```diff
@@ -39,6 +39,7 @@
 
 import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_CHUNK_CONCEPTS_FOR_CLOSURE_GENERATION;
 import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_FINALIZE_IMPORT;
+import static ca.uhn.fhir.jpa.batch2.jobs.term.loinc.ImportLoincJobAppCtx.STEP_ID_GENERATE_CONCEPT_CLOSURES;
```
jmarchionatto: Not specific pf this enhancement, but SNOMED imports use loinc constants. Could those be moved up the hierarchy one step, to be generic?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
```diff
@@ -92,15 +93,20 @@ public JobDefinition<ImportTerminologyJobParameters> importCustomTerminologyJobD
...
 						importIcdStepChunkConceptsForClosureGeneration())
 				.addIntermediateStep(
-						"generate-concept-closures",
+						STEP_ID_GENERATE_CONCEPT_CLOSURES,
 						"Generate concept closures",
 						TerminologyFileSetJson.class,
 						importIcdStepGenerateConceptClosures())
+				// This step doesn't gain any work chunks until the previous step, we want to give it
+				// a fixed portion of the overall progress
+				.setStepWeightForProgressCalculator(STEP_ID_GENERATE_CONCEPT_CLOSURES, 0.3)
```
jmarchionatto: I imagine that weight is the result of an observation, and dependent of the specific version being uploaded, which means it could need adjustment in a next versions. If that is the case it would be better be a properly named constant in a config value to ease future adjustment. 

This applies to all `.setStepWeightForProgressCalculator` calls.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/JobInstanceProgressCalculator.java
```diff
@@ -63,7 +64,12 @@ public void calculateAndStoreInstanceProgress(String theInstanceId) {
 		InstanceProgress instanceProgress = calculateInstanceProgress(theInstanceId);
 
 		myJobPersistence.updateInstance(theInstanceId, currentInstance -> {
-			instanceProgress.updateInstance(currentInstance);
+			JobDefinition<?> jobDefinition = myJobDefinitionRegistry
```
jmarchionatto: This throws exception with no code and no message. Reuse L128 pattern instead

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/InstanceProgress.java
```diff
@@ -155,7 +174,10 @@ public void updateInstance(JobInstance theInstance) {
 	 *
 	 * @param theInstance the instance to update with progress statistics
 	 */
-	public void updateInstance(JobInstance theInstance, boolean theCalledFromReducer) {
+	public void updateInstance(
```
jmarchionatto, line 175: Weighted steps with no chunks (and chunks for unknown steps) silently cap progress below 100%.**
In `InstanceProgress.updateInstance`, explicit-weight steps with `conditionalChunkCount == 0` are `continue`d (skipped), and `getChunkCount(stepId)` only ever iterates step IDs that exist in the
job definition's weighting sets. Consequences:
- A weighted step that has not yet received chunks contributes 0 while its weight is reserved, so
  total progress is capped (e.g. `generate-concept-closures` at weight 0.3 → progress maxes at 0.7
  until that step starts). Meanwhile `calculateNewStatus` marks the job `COMPLETED` based purely on
  chunk counts, independent of the weighted total. A job can therefore complete reporting < 1.0.
- Any chunk whose `targetStepId` is not a defined step is silently dropped from the calculation.
  This may be acceptable/intended, but it should be a documented, deliberate decision — and ideally
  the final progress should be clamped to 1.0 on completion so the UI doesn't show a job "done at 79%".

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/JobDefinition.java
```diff
@@ -511,6 +530,26 @@ public Builder<PT, NIT> setInitialStatus(StatusEnum theInitialStatus) {
...
+		 * The value here will not affect the actual job performance in any way, it just affects
+		 * how we calculate the "job complete %" number returned to the client while the job executes.
+		 *
+		 * @param theStepId The step ID to assign a weight to.
+		 * @param theWeight A weight. Value must be <code>i >= 0.0</code> and <code>i < 1.0</code>. The
+		 *                  combined weight of all steps must not exceed 1.0. If a given step should always
+		 *                  count for 20% of the total completion status, assign it a weight of <code>0.2</code>.
+		 * @see StepWeightingForProgressCalculator
+		 */
+		public Builder<PT, NIT> setStepWeightForProgressCalculator(String theStepId, double theWeight) {
```
jmarchionatto, line 545: Inconsistent weight-range validation between the two builders.**
`JobDefinition.Builder.setStepWeightForProgressCalculator`:
```java
Validate.isTrue(theWeight >= 0.0d && theWeight < 1.0d, "theWeight must be >= 0.0 and < 1.0");
```
but `StepWeightingForProgressCalculator.Builder.setStepWeightForProgressCalculator`:
```java
Validate.isTrue(theWeight > 0.0, "... step weight can not be <= 0.0");
```
`0.0` passes the outer guard and is then rejected by the inner one with a different message — the outer `>= 0.0` bound is dead and misleading. Suggestion: pick one source of truth (the validation is duplicated across two classes — a DRY violation). Recommend the outer guard delegate to / mirror the inner exactly.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/StepWeightingForProgressCalculator.java
```diff
@@ -0,0 +1,122 @@
...
+	private final Set<String> myStepIdsWithoutExplicitWeight;
+	private final Map<String, Double> myStepIdToWeight;
+
+	StepWeightingForProgressCalculator(
+			Map<String, Double> theStepIdToWeight,
+			Set<String> theStepIdsWithoutExplicitWeight,
+			double theWeightForStepsWithoutExplicitWeight) {
+		Validate.notNull(theStepIdToWeight, "theStepIdToWeight must not be null");
+		myStepIdToWeight = Map.copyOf(theStepIdToWeight);
+		myStepIdsWithoutExplicitWeight = theStepIdsWithoutExplicitWeight;
```
jmarchionatto: Map is protected but set is stored by reference and `getStepIdsWithoutExplicitWeight` returns that mutable reference.  

Wrap with `Set.copyOf(...)` for consistency.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/InstanceProgress.java
```diff
@@ -21,41 +21,46 @@
...
+	private Map<String, Counter> myFailedChunkCount = new HashMap<>();
 	private int myErrorCountForAllStatuses = 0;
 	private Date myEarliestStartTime = null;
 	private Date myLatestEndTime = null;
 	private String myErrormessage = null;
 	private StatusEnum myNewStatus = null;
-	private final Map<String, Map<WorkChunkStatusEnum, Integer>> myStepToStatusCountMap = new HashMap<>();
-	private final Map<String, StepProgressData> myStepProgressMap = new HashMap<>();
-	private final Set<String> myWarningMessages = new HashSet<>();
+	private int myAllIncompleteChunkCount;
```
jmarchionatto: Are these four vars used to avoid deriving them from the corresponding maps for performance?
That could potentially allow drifts and increases calculations for a questionable performance gain.
Suggestion: use the map values additions instead.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/StepWeightingForProgressCalculator.java
```diff
@@ -0,0 +1,122 @@
...
+					theSteps.stream().map(JobDefinitionStep::getStepId).collect(Collectors.toSet());
+
+			double combinedWeightTotal = 0.0;
+			for (String stepId : myStepIdToWeight.keySet()) {
+				Validate.isTrue(
+						remainingJobStepIds.remove(stepId), "Step weight provided for invalid step ID: %s", stepId);
+				combinedWeightTotal += myStepIdToWeight.get(stepId);
+			}
+
+			Validate.isTrue(
```
jmarchionatto, line 114: Because double is IEEE 754 binary floating point, most "round" decimals like 0.1, 0.3, 0.4 cannot be represented exactly in binary — they're stored as the nearest representable value, slightly above or below. When you add several of them, the tiny representation errors accumulate, and the sum can land just above the true mathematical total. Weights such as 0.3 + 0.3 + 0.4 can sum to 1.0000000000000002 and spuriously fail the build.  

Suggestion: Compare against 1.0 with a small tolerance instead of an exact bound:

```java
private static final double WEIGHT_SUM_EPSILON = 1e-9;
...
Validate.isTrue(
combinedWeightTotal <= 1.0 + WEIGHT_SUM_EPSILON,
"Combined step weights can not be greater than 1.0, but was %s",
combinedWeightTotal);

```

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/JobDefinition.java
```diff
@@ -511,6 +530,26 @@ public Builder<PT, NIT> setInitialStatus(StatusEnum theInitialStatus) {
...
+			Validate.notBlank(theStepId, "theStepId must not be blank");
+			Validate.isTrue(theWeight >= 0.0d && theWeight < 1.0d, "theWeight must be >= 0.0 and < 1.0");
+			Validate.isTrue(
+					mySteps.stream().anyMatch(t -> t.getStepId().equals(theStepId)), "Unknown stepL: %s", theStepId);
```
jmarchionatto: StepL?

hapi-fhir-storage/src/main/java/ca/uhn/fhir/util/Counter.java
```diff
@@ -17,12 +17,16 @@
...
  * #L%
  */
-package ca.uhn.fhir.jpa.util;
+package ca.uhn.fhir.util;
 
 public class Counter {
 
 	private long myCount;
 
+	public long get() {
```
jmarchionatto: This class doesn't look like a great fit for this use case: It stores long, always cast to int, and uses `getThenAdd` just for the sake of the (side effect) addition, for which an `increment` method would show intent better. 

Suggestion: replace the use of this class by `Map.merge(stepId, 1, Integer::sum)` to increment map counters

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/StepWeightingForProgressCalculator.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * Given a Batch2 job with a series of steps, by default we calculate the job progress
+ * by counting all of the work chunks across all steps, dividing:
+ * <code>total incomplete chunks / total chunks</code>.
+ * However, it's also possible to assign weights to specific steps, if a specific
+ * step is expected to process steps faster than others, or if it won't get
+ * populated with steps until later in the process.
+ * An individual step may have an explicit weight assigned, which is a double between
+ * 0.0-0.999.
```
jmarchionatto: Mismatch with enforcement (L83: > 0.0))

hapi-fhir-storage-batch2/src/test/java/ca/uhn/fhir/batch2/progress/JobInstanceProgressCalculatorTest.java
```diff
@@ -0,0 +1,394 @@
...
+//		 * : Incomplete: 0, Complete: 185
+		chunks.addAll(createWorkChunks("import-hierarchy-concepts", 135, WorkChunkStatusEnum.COMPLETED));
+//		 * : Incomplete: 0, Complete: 1
+		chunks.addAll(createWorkChunks("import-ieee-medical-device-code", 1, WorkChunkStatusEnum.COMPLETED));
+//		 * : Incomplete: 0, Complete: 1
+		chunks.addAll(createWorkChunks("import-imaging-document-code", 1, WorkChunkStatusEnum.COMPLETED));
+//		 * : Incomplete: 0, Complete: 0
+		chunks.addAll(createWorkChunks("import-linguistic-variant", 0, WorkChunkStatusEnum.COMPLETED));
+//		 *  Incomplete: 0, Complete: 1
+		chunks.addAll(createWorkChunks("import-parent-group-file:", 1, WorkChunkStatusEnum.COMPLETED));
```
jmarchionatto: Has a trailing colon, so the chunk targets a step that doesn't exist in the definition and is silently excluded from the calculation. The test still passes — which both hides the typo and concretely demonstrates the finding in `InstanceProgress.updateInstance`

hapi-fhir-storage-batch2/src/test/java/ca/uhn/fhir/batch2/coordinator/JobDefinitionTest.java
```diff
@@ -35,4 +37,40 @@ public void builder_no_steps() {
...
+	void stepWeight_FailOnAllWeightAddsUpToMoreThanOne() {
+		assertThatThrownBy(()->{
+			super.createJobDefinition(b -> {
+				b.setStepWeightForProgressCalculator(STEP_1, 0.49);
+				b.setStepWeightForProgressCalculator(STEP_2, 0.49);
+				// Adds up to more than 1.0
+				b.setStepWeightForProgressCalculator(STEP_3, 0.2);
+			});
+		}).isInstanceOf(IllegalArgumentException.class)
+			.hasMessageContaining("Combined step weights can not be greater than 1.0, but was 1.18");
```
jmarchionatto: Assertion on `double` formatting is brittle.

jmarchionatto (approved):
Looks good. Approved with several improvement suggestions.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/snomedct/ImportSnomedCtJobAppCtx.java
jamesagnew: That makes sense. I've moved these constants to TerminologyConstants, where all the other import job constants live.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/custom/ImportCustomTerminologyJobAppCtx.java
jamesagnew: Good cal. Have made a constant, and docuented the logic behind the specfic value.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/JobDefinition.java
jamesagnew: Good call. The `> 0.0` is correct. I'm going to remove the outer validation and note that the inner validation will handle it.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/util/Counter.java
jamesagnew: The Counter class isn't actually used anywhere else, it was unused when I found it. I'm going to rename it to IntCounter so it makes sense here.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/StepWeightingForProgressCalculator.java
jamesagnew: Fixed

hapi-fhir-storage-batch2/src/test/java/ca/uhn/fhir/batch2/progress/JobInstanceProgressCalculatorTest.java
jamesagnew: Oops - This isn't actually a scenario that can happen for real since there will never actually be chunks assigned to steps that don't exist.  But this test was an experiement I intended to delete anyhow. I'm going to remove it.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/JobInstanceProgressCalculator.java
jamesagnew: done

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/InstanceProgress.java
jamesagnew: This is a great point. I've updated it to assign the full weight to the completed progress if we've already passed a given step

hapi-fhir-storage-batch2/src/test/java/ca/uhn/fhir/batch2/coordinator/JobDefinitionTest.java
jamesagnew: fixed

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/StepWeightingForProgressCalculator.java
jamesagnew: Done

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/JobDefinition.java
jamesagnew: fixed

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/model/StepWeightingForProgressCalculator.java
jamesagnew: Changed

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/progress/InstanceProgress.java
jamesagnew: Looking at the way these are used, we don't actually even care about the exact counts, we just care if there are "0 chunks" or "not 0 chunks". You're right, it's better to just calculate that on-demand from the maps.

PR #8106, opened 2026-06-18
8911 add job instance id to bulk export

ibaisi:


tadgh:
Formatting check succeeded!

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobService.java
```diff
@@ -87,6 +88,7 @@ public void startJob(
 		startRequest.setUseCache(useCache);
 		startRequest.setJobDefinitionId(Batch2JobDefinitionConstants.BULK_EXPORT);
 		Batch2JobStartResponse response = myJobCoordinator.startInstance(theRequestDetails, startRequest);
+		theRequestDetails.getUserData().put(JOB_INSTANCE_ID, response.getInstanceId());
```
codeforgreen: **question**: Why not write this to the response directly? The pattern seems to be a request invokes a job but does eventually does have a response. In this case we seem not to have a body for the response just yet. Would it make sense to add it?

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobServiceTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+		Batch2JobStartResponse response = new Batch2JobStartResponse();
+		response.setInstanceId("test-instance-id-123");
+		when(myJobCoordinator.startInstance(any(), any())).thenReturn(response);
+
+		// execute
+		svc.startJob(requestDetails, params);
+
+		// verify
+		String storedId = (String) requestDetails.getUserData()
```
codeforgreen: suggestion: This scenario is a good candidate for an integration test. The hapi-fhir ones are in `BulkDataExportProviderR4Test` and I am not sure that provides value in this case. A unit test is meant to test a specific method in isolation.

hapi-deployable-pom/pom.xml
```diff
@@ -5,7 +5,7 @@
 	<parent>
 		 <groupId>ca.uhn.hapi.fhir</groupId>
 		 <artifactId>hapi-fhir</artifactId>
-		 <version>8.11.13-SNAPSHOT</version>
+		 <version>8.11.14-SNAPSHOT</version>
```
codeforgreen: I don't think we need to do a version bump for this case. We're adding a new functionality not updating an existing behavior, and it does not break downstream as far as I can see. Let me know what you think.

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobServiceTest.java
```diff
@@ -0,0 +1,74 @@
...
+	private JpaStorageSettings myStorageSettings;
+
+	@Test
+	void startJob_storesJobInstanceIdInUserData() {
+		// setup
+		BulkExportJobService svc = new BulkExportJobService(
+			myInterceptorBroadcaster, myJobCoordinator,
+			myDaoRegistry, myRequestPartitionHelperSvc, myStorageSettings);
+
+		ServletRequestDetails requestDetails = spy(new ServletRequestDetails());
```
codeforgreen: Prefer to create an actual `RequestDetails` object. It is a pojo not a service. This makes tests easier to  write and maintain.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobService.java
codeforgreen: Looks like the jobId is present in the "Content-Location" response header and it is captured downstream as well using an interceptor. However the jobId (instanceId) is embedded into it as it is a URL containing the jobId. It would need it's own header as a simpler solution (fix) or adding a body to the response (needs design consultation). 
see: `writePollingLocationsToResponseHeaders`

ibaisi: This cannot be done like this as there is a known disconnection between the fhir request and the batch 2 job. That request cannot be captured in the batch job and that's why at the moment there's no way to tell which fhir request initiated which batch job by looking at transaction logs. We were doing it by looking at that header, but a better and simpler implementation was to set the jobInstanceId in both transaction logs.

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobServiceTest.java
ibaisi: done!

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobServiceTest.java
ibaisi: The unit test validates that BulkExportJobService.startJob() stores the job instance ID in RequestDetails user data. An integration test in BulkDataExportProviderR4Test wouldn't add much since the job coordinator is mocked there too. The meaningful integration test is on the CDR side in TransactionLogInterceptorTest, where we verify the user data key gets bridged into the transaction log's additionalJson — that's where the full flow is validated end-to-end.

hapi-deployable-pom/pom.xml
ibaisi: I haven't done a change that requires changes in both hapi and Cdr at the same time for a while. I thought I would require to bump the version so the changes and have the PR/MR linked. But if you're not sure this version bump is needed, I'll revert it.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobService.java
codeforgreen: Sorry but I don't understand your last comment. You seem to indicate there are different kids of requests and different kids of transaction logs. What is the difference because as I can see you're interested in the job instance id value which is unique. Can you be more specific? Do you only want the job instance id to be present for a certain type of request?

What I am suggesting (as a simple fix) is to simply add another header to the response with the job instance id. This is already done via another header `Constants.HEADER_CONTENT_LOCATION`. You will see that is populated in BulkExportJobServiceUtil#writePollingLocationsToResponseHeaders and then consumed by the transaction logs in `TransactionLogInterceptor#captureResponseHeaders` which leads to `TransactionLogRequestDetailsUtil#addAdditionalJsonProperty`.

I posted a question about this in the design channel, I'd appreciate if you can hold off the fix until then. 
https://hapi-2.slack.com/archives/C02PJQBV087/p1781903263147809

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobServiceTest.java
codeforgreen: if you'd like to have a unit test, then testing a public/protected method of a service is the way to go, not the entire flow which is an integration test scenario.

hapi-deployable-pom/pom.xml
codeforgreen: You will only need to make sure you merge hapi-fhir first, then build a snapshot that includes the commit (there is one automatically running every day at 8pm EDT) and run pipeline for downstream and merge downstream changes.

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobServiceTest.java
```diff
@@ -0,0 +1,76 @@
...
+import jakarta.servlet.http.HttpServletResponse;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import java.util.List;
+
+import static ca.uhn.fhir.batch2.jobs.export.BulkExportJobService.JOB_INSTANCE_ID;
+import static ca.uhn.test.util.AssertJson.assertThat;
```
codeforgreen: Need `org.assertj.core.api.Assertions.assertThat` instead.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobService.java
```diff
@@ -54,6 +54,7 @@ public class BulkExportJobService {
 	private final DaoRegistry myDaoRegistry;
 	private final IRequestPartitionHelperSvc myRequestPartitionHelperService;
 	private final JpaStorageSettings myStorageSettings;
+	protected static final String JOB_INSTANCE_ID = "jobInstanceId";
```
codeforgreen: If this constant is still needed, I'd like for it not to be declared both here and in downstream dependency, that would make it brittle.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobService.java
ibaisi: The key is in BulkExportJobService.startJob() itself. Look at the flow:

 1. The FHIR request arrives with a ServletRequestDetails that carries the transaction log context
 2. startJob() calls myJobCoordinator.startInstance(theRequestDetails, startRequest) — this submits the job and returns immediately
 3. The HTTP response is sent back to the client (with headers, body, etc.)
 4. The TransactionLogInterceptor fires and persists the transaction log for this request

 After that, the request is done. The batch job runs asynchronously in a completely separate thread. When it executes its steps, it creates fresh SystemRequestDetails:

 - ExpandResourceAndWriteBinaryV2Step.java:37: return new SystemRequestDetails().setRequestPartitionId(partitionId);
 - GenerateWorkPackagesV3Step.java:298: SystemRequestDetails requestDetails = new SystemRequestDetails().setRequestPartitionId(thePartitionId);
 - ExpandResourceAndWriteBinaryStep.java:771: RequestDetails requestDetails = myStepExecutionDetails.newSystemRequestDetails();

 These new SystemRequestDetails have no connection to the original ServletRequestDetails — no user data, no transaction log context, nothing. So even if you put the job instance ID in the HTTP
 response body, the batch job's own transaction logs would still have no reference back to the originating FHIR request.

 That's why your approach stores the jobInstanceId on the original ServletRequestDetails.getUserData() before the request completes — it's the only moment where the FHIR request and the job instance
 ID coexist in the same context.

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobServiceTest.java
ibaisi: good catch, done!

hapi-fhir-storage-batch2-jobs/src/test/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobServiceTest.java
ibaisi: Fair point — the test does exercise the full startJob() flow rather than a single method in isolation. However, the behaviour under test (storing the instance ID in user data) is a side effect of  startJob(), which is the public entry point. Extracting a protected method just to unit-test a single put() call feels like over-engineering. Happy to move this to BulkDataExportProviderR4Test if you prefer, though the setup would be similar since IJobCoordinator is mocked there as well.

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobService.java
ibaisi: moved to BulkExportJobServiceUtil where it can be accessed from Cdr as well

hapi-fhir-storage-batch2-jobs/src/main/java/ca/uhn/fhir/batch2/jobs/export/BulkExportJobService.java
codeforgreen: I was actually referring to the exact same code which manipulates the actual request and response, not the internal ones. Below the line where you populate the userData there is a line of code which feeds the `jobInstanceId` to the response headers as the polling location.

```
Batch2JobStartResponse response = myJobCoordinator.startInstance(theRequestDetails, startRequest);
BulkExportJobServiceUtil.writePollingLocationsToResponseHeaders(theRequestDetails, response.getInstanceId());
```

I don't have sufficient reason at this time to insist on having it in the response headers. That can be probably part of a larger feature to improve user experience. Because you're adding it for a non-functional reason (transaction logs), we can keep the userData solution for now.
Please make sure you refer to the same `BulkExportJobServiceUtil.JOB_INSTANCE_ID` from downstream.

codeforgreen (approved):

codeforgreen (approved):
Approved with one comment:
Please check if it make sense to move that line of code to the `startInstance` method as this will enable having it for all batch2 jobs. You mentioned offline that this is done for other s2s jobs, but I could not find the code.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobCoordinatorImpl.java
```diff
@@ -137,9 +138,17 @@ public Batch2JobStartResponse startInstance(
...
 		Batch2JobStartResponse response = new Batch2JobStartResponse();
 		response.setInstanceId(instanceAndFirstChunk.jobInstanceId);
+		storeJobInstanceIdInRequestDetails(theRequestDetails, response);
 		return response;
 	}
 
+	private static void storeJobInstanceIdInRequestDetails(
+			RequestDetails theRequestDetails, Batch2JobStartResponse theResponse) {
+		if (theRequestDetails != null) {
+			theRequestDetails.getUserData().put(USER_DATA_KEY_JOB_INSTANCE_ID, theResponse.getInstanceId());
```
michaelabuckley, line 148: question: will a single request ever start more than one job?  Should this be a collection?

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobCoordinator.java
```diff
@@ -34,6 +34,8 @@
 
 public interface IJobCoordinator {
```
michaelabuckley, line 36: I don't see a changelog.

michaelabuckley (approved):

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/api/IJobCoordinator.java
ibaisi: added!

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobCoordinatorImpl.java
ibaisi: Double checked and in the current codebase, a single FHIR request always starts exactly one batch job. The userData.put() will just overwrite the same key if it were ever called twice, keeping the last value. 
  But that scenario doesn't exist today.                                                                                                                                                                
                                                                                                                                                                                                        
  That said, if it ever did happen, the current implementation would silently keep only the last job instance ID. If you wanted to guard against that future possibility, you could store a List<String>
   instead — but that would change the contract and complicate the CDR-side bridge for a scenario that doesn't exist. Not worth it unless there's a concrete plan for multi-job operations.

bbergen:
@michaelabuckley if you have a moment, can you please merge this for @ibaisi 

I'm not sure anyone on the Prior Auth team has merge permissions here. Thanks.

hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobCoordinatorImpl.java
codeforgreen: @michaelabuckley There is one case where we spin multiple jobs with one request: package installation. And there could be more. But for that I think we need a proper design at which point we can replace this solution. This fix covers batch2 jobs started in hapi-fhir with the original `ServletRequestDetails`. We may need to cover cases where a single user request spins multiple underlying batch2 job requests with `SystemRequestDetails` and that does require a separate design and cannot be handled as a fix.

PR #8113, opened 2026-06-23
Fix package installer error message and update documentation

### LINKED ISSUE #8112: Fix package installer version conflict error message
`PackageInstallationSvcImpl` can emit the following error in the logs:

`ourLog.error("Version conflict error: This is possibly due to a collision between ValueSets from different IGs that are coincidentally using the same resource ID: \[{}\] and new resource URL: \[{}\], with the existing resource having URL: \[{}\]. Ignoring this update and continuing: The first IG wins. Cause: {}"`

The current message is obsolete after implementing `versionPolicy` parameter that defaults to `MULTI_VERSION` thus generating a client assigned id for creating new resources. See #7516.

codeforgreen:
Fixes #8112


tadgh:
Formatting check succeeded!

TipzCM (approved):

PR #8120, opened 2026-06-26
Allow search operations to use custom IBundleProvider as return type

bobvanderlinden:
Currently it is not possible to create a subclass of IBundleProvider and use that as the return type of a search operation. The return type has to be IBundleProvider. This could result in casting behavior in other places of the HAPI application.

In order to avoid casting, we can allow subclasses of IBundleProvider to be used as return type.

In our project, we have a IBundleProvider that holds additional information, like how the database is being queried. In places where we know the search operation beforehand, we would like to make use of the concrete return type instead of IBundleProvider without explicit casting.


jamesagnew (approved):
This looks good, thanks for the contribution!

tadgh:
Formatting check succeeded!

jamesagnew (approved):

PR #8134, opened 2026-07-02
Fix nightly SNAPSHOT central-publish job getting force-cancelled mid-…

### LINKED ISSUE #8133: SNAPSHOT central-publish job gets force-cancelled mid-upload
The snapshot-pipeline.yml → snapshot-deploy-job.yml nightly run gets killed by Azure while sitting silent inside central-publishing:0.8.0:publish (injected-central-publishing) @ hapi-fhir-jpaserver-uhnfhirtest:                                                                                           

  Root cause: pom.xml's central-repo profile sets waitUntil=uploaded on central-publishing-maven-plugin:0.8.0 but never overrides waitMaxTime (default 30 min), so the plugin polls Sonatype's Central Portal for up to 30 min after uploading the aggregated bundle. Neither snapshot-pipeline.yml nor snapshot-deploy-job.yml sets timeoutInMinutes on the DeploySnapshot job, so it inherits pipeline's default 60-minute job timeout. The observed gap (32m51s) lands just past the plugin's 30-min wait — Pipeline's job-level timeout is very likely winning a race against the plugin's own timeout before it can log a real error, plausibly because Sonatype's newer Central Portal is slower than the old OSSRH staging repo for large multi-module bundles.                    
                                                                                                                                                             
  Fix:                                                                                                                                                       
  1. Add timeoutInMinutes: 90 to the DeploySnapshot job in snapshot-deploy-job.yml (matches the existing convention in snapshot-release-pipeline.yml) so the 
  plugin's own timeout can surface instead of being pre-empted.                                                                                              
  2. Make waitMaxTime/waitPollingInterval explicit next to waitUntil in pom.xml's central-repo profile instead of relying on plugin defaults.

jmarchionatto:
…upload

Azure's default 60-minute job timeout was racing the central-publishing plugin's own 30-minute waitMaxTime, force-killing the job before the plugin could report a real timeout error. Set an explicit 90-minute job timeout (matching snapshot-release-pipeline.yml's convention) and make waitMaxTime/waitPollingInterval explicit in the central-repo profile instead of relying on plugin defaults.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3339,6 +3339,8 @@
 							<publishingServerId>central</publishingServerId>
 							<autoPublish>true</autoPublish>
 							<waitUntil>uploaded</waitUntil>
+							<waitMaxTime>1800</waitMaxTime>
+							<waitPollingInterval>10</waitPollingInterval>
 							<excludeArtifacts>
 								<excludeArtifact>hapi-fhir-base-test-jaxrsserver-kotlin</excludeArtifact>
 								<excludeArtifact>hapi-fhir-jpaserver-uhnfhirtest</excludeArtifact>
--- a/snapshot-deploy-job.yml
+++ b/snapshot-deploy-job.yml
@@ -3,6 +3,7 @@
 
 jobs:
   - job: DeploySnapshot
+    timeoutInMinutes: 90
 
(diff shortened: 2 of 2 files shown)


tadgh:
Formatting check succeeded!

jamesagnew (approved):

PR #8137, opened 2026-07-04
Fixes for package install non-canonical resources via additional resource folders

codeforgreen:
Fixes for #7954

Multiple issues were detected and fixed:
* installing example resources from IGs was failing with the referential integrity error HAPI-1286. For example Patient referencing Organization where Organization did not exist in package or was not installed yet. I recommended users add the `myAutoCreatePlaceholderReferenceTargets` setting. 
* non-conformance resources were receiving a new server-assigned in overriding the client-assigned id by default. That should only happen for canonical resources. Fixed such that `versionPolicy` parameter only applies to canonical resources. Had to introduce a utility to detect if a resource type is canonical - see `FhirTypeUtil`
* additional resource folder installation logic was not considering the `installResourceTypes` parameter if present and were still using the underlying default
* added more documentation about package installer features which was missing since hapi-fhir is where they are implemented

Note that for solving HAPI-1286, another solution could be to bundle resources in additional folders. This has been parked because it would require substantial refactoring.


tadgh:
Formatting check succeeded!

hapi-fhir-base/src/test/java/ca/uhn/fhir/util/FhirTypeUtilTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.stream.Stream;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+// Created by claude-sonnet-4-6
+class FhirTypeUtilTest {
+
+	// Stub hierarchy simulating DSTU3 / R4 / R4B: ConcreteType extends MetadataResource
+	abstract static class MetadataResource {}
```
JasonRoberts-smile: This approach of emulating the FHIR class hierarchies is a bit bizarre. Why not just use the actual DSTU3, R4, R5, etc. classes that we will be encountering at runtime?

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/packages/loader/PackageResourceParsingSvcTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+	@Test
+	void parseResourcesOfTypesFromFolder_specificType_returnsOnlyMatchingType() {
+		NpmPackage pkg = buildPackage();
+
+		List<IBaseResource> resources = mySvc.parseResourcesOfTypesFromFolder(List.of("Patient"), "example", pkg);
+
+		assertThat(resources).hasSize(1);
+		assertThat(resources.get(0).fhirType()).isEqualTo("Patient");
+	}
```
JasonRoberts-smile, line 106: I think this test is insufficient. How do we know, for example, that the Patient resource returned above is the one that was found in the "example" folder, and not the one from the root folder? All the resources have ids assigned, but they are never used in the asserts.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImplTest.java
```diff
@@ -733,6 +733,7 @@ void install_searchParameterWithDifferentPackage_allowsUpdate() throws IOExcepti
 		}
 	}
 
+<<<<<<< Updated upstream
```
JasonRoberts-smile: Looks like a merge went wrong here

JasonRoberts-smile (approved):
Approved with a few suggestions

hapi-fhir-base/src/test/java/ca/uhn/fhir/util/FhirTypeUtilTest.java
codeforgreen: I know. It's because doing that creates a circular dependency between modules. That is why alot of the unit tests from this module are actually in `hapi-fhir-structures-r4`. I can move the unit test there and use actual resources across the different FHIR versions, if that is less weird.

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/packages/PackageInstallerSvcImplTest.java
codeforgreen: Fixed

hapi-fhir-jpaserver-base/src/test/java/ca/uhn/fhir/jpa/packages/loader/PackageResourceParsingSvcTest.java
codeforgreen: Good point. I think I briefly noticed this but missed to correct it on the spot. The test data has Patient in both main folder (/package) and the additional folder (/example). It is missing an assertion for the id. I will add it. 
I probably need to rescan this class to make sure all edge cases are covered well.

PR #7939, opened 2026-05-20
Patient ID partition mode: support transaction bundles that create placeholder and conditional Patients

tynerguo:
Closes #8170                                                                                                                                                     
Closes #8171                                                                                                                                                     
Closes #8172                                                                                                                                                     
Closes #8173 

#### What Was Done
- Added TransactionBundleNormalizer: assigns every entry a placeholder fullUrl, binds inline match URL references to in-bundle entries by (type, identifier), and prepends synthetic conditional creates for the rest; synthetic response entries are stripped so the response aligns 1:1 with the request.
- Routing deferral: entries whose Patient compartment can't be resolved before pre-fetch (Msg 1321/1326) defer to allPartitions() — only when all-partition search is supported, otherwise is rejected as before. Create-time partition validation remains the same.
-  Modified Pointcut `STORAGE_TRANSACTION_WRITE_AFTER_PREFETCH`: `PatientIdPartitionInterceptor` resolves every Patient entry to a concrete id order-independently — matched conditionals reuse the existing id, unmatched ones get a minted UUID and stay conditional so in-bundle duplicates consolidate and the `HFJ_RES_SEARCH_URL` uniqueness guard protects against concurrent duplicates.
-   Added pointcut `STORAGE_TRANSACTION_RESPONSE_ASSEMBLED`: the interceptor restores each rewritten entry's OperationOutcome to what the caller's original request implied, by mutating the live outcome's primary issue in place — DAO-attached issues (e.g. auto-created placeholders) survive.
- TODO: version bump once approved.

#### Behaviour Changes Worth Noting
Some of the decisions below were made individually by me, so please feel free to post objections if you think it is not appropriate.
- this change only works when server Id strategy is UUID. On master, the synthea hack would create a UUID no matter the configuration.
- a transaction conditional update whose body id mismatches the conditional URL's match is now always rejected with HAPI-2279; the resolved-id leniency from #4602 is removed
- under REQUIRES_NEW partition-change propagation, writes join an enclosing all-partitions transaction instead of each committing separately (bundle atomicity restored; MegaScale-style implementations override isCompatiblePartition and are unaffected).
- Query count increase in some cases (see tests with modified query counts). Mostly due to the synthetic entries created by `TransactionBundleNormalizer` needing an extra read for the resource version in `preFetchResourceVersions`. However, this version is not used anywhere if the resource already exists in the database.

#### Limitations
[...]

PR #8147, opened 2026-07-07
Patient-ID partition mode: transaction conditional update by identifier should not duplicate

### LINKED ISSUE #8152: Patient ID partition mode: transaction conditional update by identifier does not update existing Patient (fails or duplicates)
### Describe the bug

In Patient ID partition mode, when the underlying storage supports an all-partition search, a FHIR transaction bundle containing a conditional update by identifier (`PUT Patient?identifier=sys|val`) for an **existing** Patient does not update that Patient. What the caller observes depends on the request body:

- **Body carries a client-assigned `id` that differs from the existing Patient's id → a duplicate is created.** A second Patient (with the body's id) is stored sharing the same identifier. No error is raised.
- **Body has no `id` (id-less) → the request fails with `HAPI-1321`.** Instead of updating the existing Patient, the request is rejected.

### To Reproduce

1. Enable Patient ID partition mode with an all-partition-search-capable backend.
2. Create a Patient with a server-assigned id and identifier `sys|val`.
3. Submit a transaction with a conditional update entry `PUT Patient?identifier=sys|val`:
   - with a body `id` that differs from the existing Patient's id → observe a **duplicate** Patient created (no error);
   - with an id-less body → observe the request **fail with HAPI-1321**.
4. In both of these cases the existing Patient is not updated.

### Expected behavior

The conditional update resolves to the existing Patient (across partitions) and updates it in place. No duplicate is created and no HAPI-1321 error is raised.

### Environment

- HAPI FHIR 8.12.0 (Patient ID partition mode)

mrdnctrk:
Closes #8152.

In Patient-ID partition mode with an all-partition-search-capable backend, a transaction `PUT Patient?identifier=...` against an existing Patient either created a silent duplicate or failed with HAPI-1321, because the conditional match search was pinned to one write-derived partition instead of fanning out.

### Fixes

- **TransactionProcessor** — when all-partition search is supported, don't collapse an unresolvable conditional match search onto the outer write partition. Letting it fan out finds an existing Patient in another partition, so a mismatched body id now yields HAPI-2279 instead of a silent duplicate.
- **PatientIdPartitionInterceptor** — an id-less conditional PUT (`PUT Patient?identifier=X` with no body id) is rewritten after pre-fetch to a direct `PUT Patient/<id>` using the already-resolved match, so it updates the existing Patient instead of failing with HAPI-1321.
- **BaseTransactionProcessor.shouldConditionalUpdateMatchId** — no longer strips a *resolved* body id (placeholder ids still stripped), so a same-id conditional update stays routable and updates in place. See below.

### Why remove the resolved-id branch in `shouldConditionalUpdateMatchId`

This method decides whether to keep or strip the id carried on the resource **body** before a conditional update runs. The old code had a branch that stripped the body id whenever that id was already "resolved" during the transaction (checked via `TransactionDetails.hasResolvedResourceId`, keyed by the id's value). Once the TransactionProcessor fix above lets the conditional match search succeed, that branch breaks a **same-id** conditional update in Patient-ID partition mode. Consider `PUT Patient?identifier=X` whose body carries the existing Patient's own id `A`:

1. Pre-fetch resolves the identifier to the existing Patient `A` and registers `A` in the transaction's resolved-id set.
2. At write time, the branch checks the **body id** (`A`) against that set. Because `A` is now resolved, the check matches and the body id is **stripped** — the resource becomes id-less.
3. The write then determines the storage partition for that now id-less Patient. In Patient-ID partition mode the partition is `hash(id)`, which needs an id, so the request fails with **HAPI-1321** instead of updating Patient `A`.

Removing the branch keeps the matching body id, so the Patient stays routable and updates in place. Placeholder ids (`urn:` / `#`) are still stripped by the remaining check.

(The branch only ever affected the same-id case: it keys on the body id's own value, and only the matched id — not a *different* client-provided id — is in the resolved set. The different-body-id → HAPI-2279 behavior comes from the TransactionProcessor fix above, not from this branch.)

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
@@ -763,7 +763,12 @@ public class TransactionProcessor extends BaseTransactionProcessor {
 						next.myResourceDefinition.getName(),
 						next.myMatchUrlSearchMap,
 						next.getAssociatedResource());
-				if (partition.isAllPartitions()) {
+				if (partition.isAllPartitions() && !myPartitionSettings.isAllPartitionSearchSupported()) {
+					// I couldn't determine from the history why allPartitions
+					// was originally defaulted to the outer/transaction partition; it may have been because
+					// all-partition search was not supported. Now that isAllPartitionSearchSupported() makes that
+					// explicit, we only apply this fallback when all-partition search is unsupported. Otherwise, we
+					// keep allPartitions so the pre-fetch can search across all partitions.
 					partition = theOuterRequestPartitionId;
 				}
 			}
--- a/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
+++ b/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
@@ -865,6 +865,7 @@ public class PatientIdPartitionInterceptor {
 	public void onTransactionWriteAfterPrefetch(
 			TransactionWriteAfterPrefetchDetails theDetails, TransactionDetails theTransactionDetails) {
 		rewriteInlinePatientMatchUrlReferences(theDetails.getEntries(), theTransactionDetails);
+		rewriteIdlessConditionalPatientUpdatesToDirect(theDetails.getEntries(), theTransactionDetails);
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/TransactionProcessor.java
```diff
@@ -763,7 +763,12 @@ private ListMultimap<RequestPartitionId, MatchUrlToResolve> groupMatchTargetList
 						next.myResourceDefinition.getName(),
 						next.myMatchUrlSearchMap,
 						next.getAssociatedResource());
-				if (partition.isAllPartitions()) {
+				if (partition.isAllPartitions() && !myPartitionSettings.isAllPartitionSearchSupported()) {
```
michaelabuckley, line 766: nice comment.  that makes sense

michaelabuckley (approved):
This makes sense.  We do need a better api instead of checking that msg code throw, but this will do for now.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -925,19 +926,71 @@ private static boolean isPatientInlineMatchUrl(@Nullable String theReferenceValu
...
-		// Reuse the FHIR id the pre-fetch already resolved for this PID (TransactionDetails' reverse map)
-		// If the reverse map has no entry, leave the reference untouched.
-		IIdType resolvedId = theTransactionDetails.getReverseResolvedId(resolved);
-		if (resolvedId != null) {
-			theReference.setReference(resolvedId.toUnqualifiedVersionless().getValue());
+			// Convert the conditional PUT into a direct update by rewriting the request URL to the matched Patient id
+			// (verb stays PUT). We only rewrite the URL. The transaction processor's direct-update path then stamps
+			// the body id from that URL before the write, so the resource routes to the matched Patient.
+			String reference = matchedId.get().toUnqualifiedVersionless().getValue();
+			terser.setElement(entry, "request.url", reference);
```
tynerguo, line 993: One problem this will bring is that the corresponding OperationOutcome returned to the user for this entry would become inconsistent with the user's input. Specifically, the value of the `code`, `display`, `diagnostics` element would change from the conditional ones to unconditional ones. For example, `code` would change from `SUCCESSFUL_UPDATE_WITH_CONDITIONAL_MATCH` to `SUCCESSFUL_UPDATE`, and `diagnostics` would lose the following information: `Existing resource matched URL \"Patient?identifier=sys|val\"`

That being said, I think it is probably not worth it to bring in the entire "fix operation outcome" logic into this PR because this is a narrower fix and my PR https://github.com/hapifhir/hapi-fhir/pull/7939 is also near to finish. Pulling in that somewhat heavy logic now would produce more pain in merging & refactoring later...  Also, since conditional update was not supported before this PR, it shouldn't break anything for the clients, either.

So I think it might be better to note this as a known issue for now, does this sound OK to you @mrdnctrk, @michaelabuckley?

tynerguo (approved):
There's a potential problem for modifying the entry to a direct PUT, please see my other comment for details. But otherwise LGTM!

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
mrdnctrk: Great catch, @tynerguo — really sharp observation. You're right: rewriting the entry to a direct PUT Patient/<id> means the response OperationOutcome loses its conditional flavor — code goes from SUCCESSFUL_UPDATE_WITH_CONDITIONAL_MATCH to SUCCESSFUL_UPDATE, and the Existing resource matched URL "Patient?identifier=sys|val" diagnostic drops off.

I'm happy to go with your suggestion and keep the full "fix operation outcome" logic out of this PR. It makes sense to keep this change narrow and let your #7939 handle that logic rather than pull it in early and create merge/refactor pain for you later — I'd rather not step on your toes there.

One reassuring detail that narrows the scope of the gap: the HTTP status is unaffected. The transaction entry's status is derived purely from created-vs-updated (outcome.getCreated() → 201/200), not from the StorageResponseCodeEnum, so a matched update returns 200 either way. So this is strictly the outcome's descriptive code/display/diagnostics — the status line, the updated resource, its version, and its Location are all unchanged. When your work lands it'll just restore the conditional outcome string; nothing else about the response shifts.

Thanks for the careful review!

PR #8158, opened 2026-07-14
allow for reduction step to be recoverable

### LINKED ISSUE #8157: Reduction steps can be killed midprocessing and this leaves them out of sync
Need a means to handle reduction steps that are killed mid-work, so that the job isn't left stalled in 'IN_PROGRESS'

TipzCM:
closes #8157

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
@@ -115,8 +115,6 @@ import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class JpaJobPersistenceImpl implements IJobPersistence {
 	public static final String CREATE_TIME = "myCreateTime";
-	// Special step id for reduction steps
-	public static final String REDUCTION_STEP_ID = "reduction-step";
 
 	private static final Logger ourLog = Logs.getBatchTroubleshootingLog();
 	public static final int DEFAULT_ATTACHMENT_CHUNK_SIZE = toIntExact(FileUtils.ONE_MB);
@@ -588,41 +586,6 @@ public class JpaJobPersistenceImpl implements IJobPersistence {
 				.fetchChunksForStep(theInstanceId, theStepId)
 				.map(this::toChunk);
 	}
-
-	//	@Override
-	//	public WorkChunk fetchWorkChunkById(String theWorkChunkId) {
-	//		return myWorkChunkRepository.findById(theWorkChunkId)
-	//			.map(this::toChunk)
-	//			.orElse(null);
-	//	}
-
-	//	@Override
-	//	public void createReductionStepChunk(JobInstance theInstance, Consumer<Integer> theCallback) {
-	////		myWorkChunkRepository.fetchChunksForStep(theInstance.getInstanceId(), REDUCTION_STEP_ID);
-	//
-	//
-	////		myWorkChunkRepository.getDistinctStatusesForStep(
-	////			theInstance.getInstanceId(),
-	////			REDUCTION_STEP_ID
-	////		);
-	//
-	//		Batch2WorkChunkEntity entity = new Batch2WorkChunkEntity();
-	//		entity.setId(UUID.randomUUID().toString());
-	//		entity.setSequence(0);
-	//		entity.setJobDefinitionId(theInstance.getJobDefinitionId());
-	//		entity.setJobDefinitionVersion(theInstance.getJobDefinitionVersion());
-	//		entity.setTargetStepId(REDUCTION_STEP_ID);
-	//		entity.setInstanceId(theBatchWorkChunk.instanceId);
-	//		entity.setSerializedData(theBatchWorkChunk.serializedData);
-	//		entity.setCreateTime(new Date());
-	//		entity.setStartTime(new Date());
-	//		entity.setStatus(WorkChunkStatusEnum.READY);
-	//
-	//		Batch2WorkChunkEntity chunk = myTransactionService.withSystemRequestOnDefaultPartition().execute(() ->
-	// myWorkChunkRepository.save(entity));
-	//
-	//
-	//	}
 
 	@Override
 	public Page<WorkChunkMetadata> fetchAllWorkChunkMetadataForJobInStates(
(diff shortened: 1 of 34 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -164,6 +166,17 @@ public void setMaxBytesPerAttachmentChunk(Integer theMaxBytesPerAttachmentChunk)
 	@Override
 	@Transactional(propagation = Propagation.REQUIRED)
```
jamesagnew: Something seems off about the transactionaity of this method. It's marked as `@Transactional(REQUIRED)` meaning it can only be called from within a transaction, but then it uses the TxManager to open a transaction below on line 174 (which should never be needed if this is only called from within a TX).

These annotations are slowly being replaced by the HapiTransactionService anyhow since it doesn't pollute the stack trace like the annotations do. Assuming this is actually always called from within a transaction, I'd recommend adding `HapiTransactionService.requireTransaction();` at the top of the method, dropping the annotation, and dropping the use of the HapiTransactionService on line 174. If it's called' from non-transactional contexts, just drop the annotation.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -857,11 +864,16 @@ private void invokePostStorageBatchHooks(JobInstance theJobInstance) {
 	@Transactional(propagation = Propagation.REQUIRES_NEW)
 	public boolean advanceJobStepAndUpdateChunkStatus(
 			String theJobInstanceId, String theNextStepId, boolean theIsReductionStep) {
+		AtomicReference<JobInstance> instanceRep = new AtomicReference<>();
```
jamesagnew: Thought: I find this logic pretty confusing to read. The cryptic name `instanceRep` doesn't help.

IMO this would be better to just drop the `AtomicReference` entirely and just fetch the instance entity again from the entity manager on line 903 below. It will still be in the L1 cache since this whole method has an open outer transaction, so this won't result in any extra database round trips.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
```diff
@@ -884,6 +896,34 @@ public boolean advanceJobStepAndUpdateChunkStatus(
 					theNextStepId,
 					List.of(WorkChunkStatusEnum.GATE_WAITING, WorkChunkStatusEnum.QUEUED),
 					nextStep);
+
+			/*
+			 * we add a data-free READY chunk that will be enqueued by the system
+			 */
+			JobInstance instance = instanceRep.get();
+			if (instance == null) {
```
jamesagnew: I don't get how this could ever be null if we get here. Probably moot since I recommended above to replace the AtomicReference entirely, but IMO this should be a simple assert as opposed to being a warning that should be impossible to hit but looks like an expected path.

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/AbstractIJobPersistenceSpecificationTest.java
```diff
@@ -99,6 +99,15 @@ public IJobPersistence getSvc() {
 		return mySvc;
 	}
 
+	public boolean inlineReduction() {
```
jamesagnew: Nitpick: This method's name suggests an action method as opposed to a state query. Recommend `isInlineReduction()`

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/AbstractIJobPersistenceSpecificationTest.java
```diff
@@ -174,6 +183,11 @@ public ITestFixture getTestManager() {
 		@Nested
 		class StateTransitions implements IWorkChunkStateTransitions {
 
+			@Override
```
jamesagnew, line 186: Nitpick: Having a chain where we change the name we're using for the same concept is confusing

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
TipzCM: updated

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
TipzCM: I didn't write this method, so i cannot comment as to why it's written the way it is. But i've updated it

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/JpaJobPersistenceImpl.java
TipzCM: updated

jamesagnew (approved):

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/AbstractIJobPersistenceSpecificationTest.java
TipzCM: updated

hapi-fhir-storage-batch2-test-utilities/src/main/java/ca/uhn/hapi/fhir/batch2/test/AbstractIJobPersistenceSpecificationTest.java
TipzCM: updated

PR #8160, opened 2026-07-14
8159 failed to get sql log file

### LINKED ISSUE #8159: Failed to get sql log file
When Hibernate SQL logging is enabled, a `FileNotFoundException` happens indicating hibernate-sql-log-filters.txt` could not be found, resulting in failure to load the SQL log filters.

YalingPeiS:
Closes #8159.

## Summary

Fixes a `FileNotFoundException` that occurred when enabling Hibernate SQL logging while running from a packaged JAR. `SqlLoggerFilteringUtil` loaded the SQL log filter file via `ClassPathResource.getFile()`, which fails for resources inside a JAR since they have no filesystem `File`.

## Changes

- Read the filter file as a classpath stream (`getInputStream()`) instead of resolving it to a `File`, so it works both on the filesystem and when packaged in a JAR.
- Added a regression test that packages the filter file inside a JAR and verifies the lines are read correctly.


tadgh:
Formatting check succeeded!

jmarchionatto (approved):
Looks good. Approved.

PR #8119, opened 2026-06-25
Update org.hl7.fhir.core to version 6.9.11

dotasek:

PR #8081, opened 2026-06-12
Bump org.hl7.fhir.core to 6.9.10

dotasek:

PR #8166, opened 2026-07-14
Update to org.hl7.fhir.core 6.9.12

dotasek:
This updates org.hl7.fhir.core to 6.9.12

Notes regarding these changes:

okhttp_version was updated to 5.3.2 for two reasons:
 - org.hl7.fhir.core uses that same version, and this prevents us from having to track two different versions in our code.
 - the okio library in the older versions of okhttp contained vulnerabilities (this is why org.hl7.fhir.core updated its version of the library)
 
There was a pinned okio dependency in dependencyManagement that is no longer needed, as the okio dependency used by the more current okhttp is more recent, and has no known vulnerabilities. 

kotlin.version was also updated in hapi-fhir-base-test-jaxrsserver-kotlin as this was old and incompatible with the kotlin used in the new okio

There are a number of added methods, altered signatures, and updated tests that correspond to the ones done in HAPI master in the following PRs:

https://github.com/hapifhir/hapi-fhir/pull/8119 <- simple dependency update
https://github.com/hapifhir/hapi-fhir/pull/8081 <- slightly more complex dependency updates to allow for org.hl7.fhir.core's usage of spotugs, which impacted JavaDoc generation
https://github.com/hapifhir/hapi-fhir/pull/7962 <- changes to pom.xml and to numerous classes to account for changed method signatures in org.hl7.fhir.core. All of the functional changes are very minimal, applying to methods that have always returned null, or have had no-op functionality. There is one change to validation output, which was accepted by reviewers in this pull as correct.

### DIFF
--- a/hapi-fhir-structures-hl7org-dstu2/pom.xml
+++ b/hapi-fhir-structures-hl7org-dstu2/pom.xml
@@ -272,6 +272,13 @@
 					<dependencySourceIncludes>
 						<include>ca.uhn.hapi.fhir:org.hl7.fhir.dstu2</include>
 					</dependencySourceIncludes>
+					<additionalDependencies>
+						<dependency>
+							<groupId>com.github.spotbugs</groupId>
+							<artifactId>spotbugs-annotations</artifactId>
+							<version>4.9.8</version>
+						</dependency>
+					</additionalDependencies>
 				</configuration>
 			</plugin>
 		</plugins>
--- a/hapi-fhir-structures-hl7org-dstu2/pom.xml
+++ b/hapi-fhir-structures-hl7org-dstu2/pom.xml
@@ -276,7 +276,7 @@
 						<dependency>
 							<groupId>com.github.spotbugs</groupId>
 							<artifactId>spotbugs-annotations</artifactId>
-							<version>4.9.8</version>
+							<version>${spotbugs_annotations_version}</version>
 						</dependency>
 					</additionalDependencies>
 				</configuration>
--- a/hapi-fhir-structures-r5/pom.xml
+++ b/hapi-fhir-structures-r5/pom.xml
@@ -265,7 +265,7 @@
 						<dependency>
 							<groupId>com.github.spotbugs</groupId>
 							<artifactId>spotbugs-annotations</artifactId>
-							<version>4.9.8</version>
+							<version>${spotbugs_annotations_version}</version>
 						</dependency>
 					</additionalDependencies>
 				</configuration>
@@ -325,7 +325,7 @@
 						<dependency>
 							<groupId>com.github.spotbugs</groupId>
 							<artifactId>spotbugs-annotations</artifactId>
-							<version>4.9.8</version>
+							<version>${spotbugs_annotations_version}</version>
 						</dependency>
 					</additionalDependencies>
 				</configuration>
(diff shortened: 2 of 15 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-structures-hl7org-dstu2/pom.xml
```diff
@@ -272,6 +272,13 @@
 					<dependencySourceIncludes>
 						<include>ca.uhn.hapi.fhir:org.hl7.fhir.dstu2</include>
 					</dependencySourceIncludes>
+					<additionalDependencies>
+						<dependency>
+							<groupId>com.github.spotbugs</groupId>
+							<artifactId>spotbugs-annotations</artifactId>
+							<version>4.9.8</version>
+						</dependency>
+					</additionalDependencies>
```
tadgh, line 281: nit: why do we need to add this?

hapi-fhir-structures-r5/pom.xml
```diff
@@ -317,6 +322,11 @@
 							<artifactId>okhttp</artifactId>
 							<version>${okhttp_version}</version>
 						</dependency>
+						<dependency>
+							<groupId>com.github.spotbugs</groupId>
+							<artifactId>spotbugs-annotations</artifactId>
+							<version>4.9.8</version>
+						</dependency>
```
tadgh, line 329: issue: if we do need this, move it up into depManagement in hapi-fhir pom.xml, and inherit in all needed places, to avoid version drift between child modules.

tests/hapi-fhir-base-test-jaxrsserver-kotlin/pom.xml
```diff
@@ -172,6 +172,6 @@
 
 	<properties>
 		<!--<kotlin.compiler.incremental>true</kotlin.compiler.incremental>-->
-		<kotlin.version>1.9.23</kotlin.version>
+		<kotlin.version>2.2.21</kotlin.version>
```
tadgh, line 175: This is possibly a fully breaking change for CQIS, will have to check with them.

tadgh (changes requested):

tests/hapi-fhir-base-test-jaxrsserver-kotlin/pom.xml
JPercival: Should be fine. We shade the Kotlin deps for CQIS.

hapi-fhir-structures-hl7org-dstu2/pom.xml
dotasek: This is for the JavaDoc plugin. org.hl7.fhir.core uses SpotBugs annotations internally to suppress issues, but doesn't include them in compiled code. JavaDoc needs them to to build from the org.hl7.fhir.core sources and will not compile if they're not provided.

hapi-fhir-structures-r5/pom.xml
dotasek: I'll try that.

dotasek: It looks like the javadoc plugin does not draw from dependencyManagement:

```
Execution default of goal org.apache.maven.plugins:maven-javadoc-plugin:3.11.2:jar failed: version can neither be null, empty nor blank
```

As an alternative, I could include it as a `spotbugs-annotations-version` property to prevent drift. Would that work?

tadgh (approved):

hapi-fhir-structures-r5/pom.xml
dotasek: Resolved after discussion with @tadgh with the property approach.

PR #8174, opened 2026-07-16
Update org.hl7.fhir.core to 6.9.12

dotasek:
This updates org.hl7.fhir.core to 6.9.12

Notes regarding these changes:

okhttp_version was updated to 5.3.2 for two reasons:

org.hl7.fhir.core uses that same version, and this prevents us from having to track two different versions in our code.
the okio library in the older versions of okhttp contained vulnerabilities (this is why org.hl7.fhir.core updated its version of the library)

There was a pinned okio dependency in dependencyManagement that is no longer needed, as the okio dependency used by the more current okhttp is more recent, and has no known vulnerabilities. 

kotlin.version was also updated in hapi-fhir-base-test-jaxrsserver-kotlin as this was old and incompatible with the kotlin used in the new okio. This was been approved by @JPercival in a related PR: https://github.com/hapifhir/hapi-fhir/pull/8166#discussion_r3589971408 

This also brings in a quality of life improvement from https://github.com/hapifhir/hapi-fhir/pull/8166, which manages the spotbugs-annotations version via a property instead of literal versions. See comment thread in another PR here: https://github.com/hapifhir/hapi-fhir/pull/8166#discussion_r3589940666

### DIFF
--- a/hapi-fhir-client-okhttp/pom.xml
+++ b/hapi-fhir-client-okhttp/pom.xml
@@ -42,7 +42,7 @@
 		</dependency>
 		<dependency>
 			<groupId>com.squareup.okhttp3</groupId>
-			<artifactId>okhttp</artifactId>
+			<artifactId>okhttp-jvm</artifactId>
 		</dependency>
 
 		<!-- conformance profile -->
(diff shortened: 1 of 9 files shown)


tadgh:
Formatting check succeeded!

tadgh (approved):

PR #7674, opened 2026-03-19
Support standard FHIR concept properties in ValueSet filter expansion

kathbigra:
Fixes: #7667
  - Handle 'child exists' filter to include/exclude leaf concepts
  - Handle 'parent exists' filter to include/exclude root concepts
  - Handle 'inactive exists' and 'notSelectable exists' filters
  - Pass through unknown properties with a warning instead of silently producing an empty ValueSet

PR #8181, opened 2026-07-21
Support child/parent "exists" filters and flat concept-property hierarchies in ValueSet expansion

### LINKED ISSUE #7667: ValueSet filter blocks property "child"
**Describe the bug**
We are using https://simplifier.net/KDL/kdl-vs-2025/~json, which filters out non-leaf concepts from https://simplifier.net/KDL/kdl-cs-2025/~json.

Before https://github.com/hapifhir/hapi-fhir/pull/6904
```json
"filter": [
  {
    "property": "child",
    "op": "exists",
    "value": "false"
  }
]
```
was ignored (so non-leaf codes valid), which is not ideal, but after this change the `ValueSet` is just empty without intervention.

https://github.com/hapifhir/hapi-fhir/blob/5fc1c76d43d49a4c16102aa74354f1c455b26e64/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValueSetExpansionFilterContext.java#L71-L72

`child` is not a custom property and not the only unsupported one: https://hl7.org/fhir/R4/codesystem-concept-properties.html#4.3.14.243.2 

**To Reproduce**
Expand https://simplifier.net/KDL/kdl-vs-2025/~json from https://simplifier.net/KDL/kdl-cs-2025/~json.

**Expected behavior**
Ideally the filters should be properly processed. Alternatively either a WARNING should be logged and/or the filter should pass when in doubt.

- HAPI FHIR 8.8.1

### LINKED ISSUE #8180: ValueSet $expand: support child/parent "exists" filters and flat concept-property hierarchies
### Problem
  
  `$expand` cannot handle a `compose.include.filter` that uses the hierarchical `child`/`parent`
  properties with `op = exists`, and it ignores CodeSystem hierarchies expressed **flat** — i.e.
  each concept references its parent via the canonical `parent` concept-property
  (`http://hl7.org/fhir/concept-properties#parent`) instead of nested `CodeSystem.concept` arrays.
  
  Example: the German KDL ValueSet selects leaf concepts via
  `filter: { property: "child", op: "exists", value: "false" }`, and its CodeSystem ships in both
  a flat and a nested representation. Neither expands today.
  
  ### Current behaviour
  
  - **In-memory:** `child`/`parent` filter properties are unsupported → empty expansion; flat
    parent-properties are never read → no hierarchy.                                                                                                                                                                                             
  - **JPA:** `child`/`parent` filters are rejected as LOINC-only for any other system                                                                                                                                                            
    (`HAPI-0895`) — this happens for **both** the nested and flat representations, since the                                                                                                                                                     
    filter property is validated before the hierarchy is consulted; additionally, flat                                                                                                                                                           
    parent-properties are ignored on store, so no parent/child links are created.     

  ### Expected                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                 
  Support, for **any** terminology and on both paths:                                                                                                                                                                                            
  - `property = child|parent` with `op = exists` (`child exists=false` → leaves,                                                                                                                                                                 
    `parent exists=false` → roots).                                                                                                                                                                                                              
  - Building the hierarchy from the canonical flat `parent`/`child` concept-properties (in                                                                                                                                                       
    addition to nested concepts). This also makes `is-a` / `descendent-of` work for flat                                                                                                                                                         
    CodeSystems.                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                 
  ### Environment                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                 
  HAPI FHIR 8.12.0-SNAPSHOT (master); affects in-memory validation-support expansion and the JPA                                                                                                                                                 
  terminology service.

patrick-werner:
Closes #8180
Closes #7667

## Summary

Enables expanding ValueSets that select concepts via the standard FHIR concept-properties with the `exists` operator — the hierarchical `child`/`parent`, the boolean `inactive`/`notSelectable`, and the date-valued `deprecated`/`deprecationDate`/`retirementDate` — for any terminology rather than only LOINC. It also builds the concept hierarchy from CodeSystems that express it FLAT (each concept referencing its parent/child via a `parent`/`child` concept-property) in addition to nested `CodeSystem.concept` arrays.

Motivating example: the German KDL terminology. Its ValueSet selects the leaf concepts via `filter: { property: "child", op: "exists", value: "false" }`, and its CodeSystem is published in both a flat and a nested representation. Neither expanded before this change (the child filter was rejected as LOINC-only — HAPI-0895 — and the flat hierarchy was ignored on store).

The standard properties are identified by their reserved code name. A property the CodeSystem declares with a non-canonical concept-properties URI (a different meaning) is not treated as the standard property.

### In-memory (`ValueSetExpansionFilterContext`)

- Build the concept hierarchy from flat `parent`/`child` concept-properties (matched by their reserved code name), not only from nested concepts. This makes the existing structural filters (`is-a`, `descendent-of`, …) work for flat CodeSystems too.
- Handle `child`/`parent`, the boolean `inactive`/`notSelectable` (flagged only when the value is `true`), and the date-valued `deprecated`/`deprecationDate`/`retirementDate` (presence check) with `op = exists`, via one `StandardConceptProperty` registry.
- Resolve the hierarchy case-insensitively when the CodeSystem is not case-sensitive, and guard descendant traversal against cycles — including not reporting a concept as its own descendant.
- A property declared with a non-canonical URI is reported as an error (meaning unknown); an undeclared property is matched by name, with an info log the first time it is matched for a filtered property.
- Report an error (rather than a silently empty expansion) for a filter it cannot evaluate — a custom concept property, or a non-`exists` operator on the standard properties — so the caller can surface it or delegate to another terminology service in the validation-support chain.

### JPA (`TermReadSvcImpl`, `TermCodeSystemStorageSvcImpl`, `ITermConceptParentChildLinkDao`)

- `child`/`parent` + `exists` is now generic (no longer LOINC-gated), resolved from the parent/child links. The `exists` value is validated as a strict boolean.
- `inactive`/`notSelectable` + `exists` resolves value-aware against the indexed property (matches value `true`), rather than a plain presence check.
- `deprecated`/`deprecationDate`/`retirementDate` + `exists` is a value-aware presence check (so `exists=false` selects the complement); the date value itself is not interpreted and no schema change is involved.
- Flat `parent`/`child` concept-properties are turned into parent/child links when a CodeSystem is persisted, matched by their reserved code name (so an undeclared flat hierarchy is linked); a property declared with a non-canonical URI is skipped. Existing structural filters (`is-a`, `descendent-of`, …) work for flat CodeSystems since they rely on the same links.
- Works for any terminology; existing LOINC-specific parent/child/ancestor/descendant filters are unchanged.

**Migration note:** existing stored CodeSystems must be re-persisted/reindexed to pick up links from previously-undeclared flat `parent`/`child` properties.

## Testing

- `ValueSetExpansionFilterContextTest` — child/parent `exists` on nested and flat hierarchies; flat hierarchy matched by name (undeclared) and skipped for a conflicting URI; `descendent-of` on a flat hierarchy (incl. case-insensitive); cycle termination and no self-match; `inactive`/`notSelectable` and `deprecated`/`deprecationDate`/`retirementDate` `exists`; and unsupported property/operator or conflicting-URI now throwing.
- `InMemoryTerminologyServerValidationSupportTest` — an unsupported filter surfaces as an expansion error rather than a silent empty result.
- `FhirResourceDaoR4TerminologyTest` — `child exists=false` → leaves and `parent exists=false` → roots on a linked hierarchy; the same on flat (declared, **undeclared**, and non-canonical-URI) and nested CodeSystem resources; codes containing special characters; `descendent-of` on a flat resource; `inactive`/`notSelectable`/`deprecated` `exists`; and rejection of a non-boolean `child exists` value.


tadgh:
Formatting check succeeded!

gM4n-sys:
IS this the official PR for the discovery made upon #7674 ?

> @gM4n-sys, thanks! You're right, this isn't fully fixed yet.
> 
> This PR only addressed the in-memory ValueSet expander (**ValueSetExpansionFilterContext** in hapi-fhir-validation), which is the path the original issue (#7667) was reported against.
> 
> Your setup persists the kdl CodeSystem in the JPA terminology database, so $expand goes through a different code path (TermReadSvcImpl) that still rejects child/parent exists filters for any non-LOINC system with HAPI-0895. That flow is not being covered by this PR.
> 
> Let me fix it. I will probably get some time tomorrow to work on this.
> 
> Thanks again for testing this.

https://github.com/hapifhir/hapi-fhir/pull/7674#issuecomment-4989158090

patrick-werner:
> IS this the official PR for the discovery made upon #7674 ?
> 
> > @gM4n-sys, thanks! You're right, this isn't fully fixed yet.
> > This PR only addressed the in-memory ValueSet expander (**ValueSetExpansionFilterContext** in hapi-fhir-validation), which is the path the original issue (#7667) was reported against.
> > Your setup persists the kdl CodeSystem in the JPA terminology database, so $expand goes through a different code path (TermReadSvcImpl) that still rejects child/parent exists filters for any non-LOINC system with HAPI-0895. That flow is not being covered by this PR.
> > Let me fix it. I will probably get some time tomorrow to work on this.
> > Thanks again for testing this.
> 
> [#7674 (comment)](https://github.com/hapifhir/hapi-fhir/pull/7674#issuecomment-4989158090)

Nothing official, just had the same issue with KDL. I've looked at https://github.com/hapifhir/hapi-fhir/pull/7674 and added more filters like inactive/notSelectable to this PR.

jamesagnew (approved):
This is a great change. Thanks for the contribution!

PR #8182, opened 2026-07-21
Rework DaoRegistry

jamesagnew:
This PR modifies the DaoRegistry so that it no longer relies on pulling in resource providers from the outer context (which violates the principle of Inversion of Control), in favour of a more predictable explicit registration step.


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderOnlySomeResourcesProvidedR4Test.java
```diff
@@ -70,14 +67,15 @@ public void start() {
 				myDaoRegistry.setSupportedResourceTypes("Patient", "Person", "SearchParameter");
 			}
 
-		}
+			@PreDestroy
+			public void stop() {
+				myDaoRegistry.setSupportedResourceTypes();
+			}
```
tadgh, line 73: nit: this indentation seems off in web view, worth a minor check (assuming spotless would catch this)

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
```diff
@@ -896,8 +887,8 @@ public HistoryBuilder newHistoryBuilderWithIds(
 
 	@Bean
 	@Primary
-	public ISearchParamProvider searchParamProvider() {
-		return new DaoSearchParamProvider();
+	public ISearchParamProvider searchParamProvider(DaoRegistry theDaoRegistry) {
+		return new DaoSearchParamProvider(theDaoRegistry);
```
tadgh, line 891: question: This seems to just remove an autowire in favour of a constructor argument. Is this part of the refactor? I don't see the purpose of the change here, unless it is that we n eed to construct DaoSearchParamProvider objects outside of a spring context?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/SearchConfig.java
```diff
@@ -62,33 +62,32 @@
...
-import org.springframework.transaction.PlatformTransactionManager;
 
 @Configuration
 public class SearchConfig {
 	public static final String SEARCH_TASK = "searchTask";
 	public static final String CONTINUE_TASK = "continueTask";
 
 	@Autowired
-	private DialectSvc myDialectSvc;
+	protected DialectSvc myDialectSvc;
```
tadgh: nit: having looked at the overriding implementation, this smells like we may have the wrong abstraction (inheritance and consumption of all autowires in a single method). I can't think of a better one offhand though, so take this with a grain of salt :salt:

tadgh (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/provider/r4/ResourceProviderOnlySomeResourcesProvidedR4Test.java
jamesagnew: Probably just a display artifact, indenting seems consistent here.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/JpaConfig.java
jamesagnew: This change wasn't strictly necessary but was a part of a broader change towards injecting DaoRegistry instances with constructor arguments. The change that was necessary was getting rid of `@Autowired DaoRegistry myDaoRegistry` field definitions inside of `@Configuration` classes. Having a field there turns out to make things really brittle so I did away with those.

tadgh (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/config/SearchConfig.java
jamesagnew: Good call. I've replaced this with more normal field autowiring like we do everywhere else.

PR #8184, opened 2026-07-22
Allow interceptors to select an alternate batch2 job definition for bulk export kickoff

maciej-ca:
Adds a `jobDefinitionIdOverride` property to `BulkExportJobParameters`, honored by `BulkExportJobService.startJob()`. When server-side interceptor code sets it (typically on `STORAGE_PRE_INITIATE_BULK_EXPORT`; `STORAGE_INITIATE_BULK_EXPORT` is honored equally), the bulk export kickoff (`$export` and `$davinci-data-export`) starts the named registered batch2 job definition instead of the default `BULK_EXPORT` job. Parameter parsing, permission checks, and `$export-poll-status` are unchanged.

The property can never be populated from the client request — it exists only on the server-side parameters object, and a test pins that a value smuggled in as a request parameter is ignored.

Changes:

- `BulkExportJobParameters`: new `jobDefinitionIdOverride` JSON property with javadoc'd accessors.
- `BulkExportJobService.startJob()`: uses the override when non-blank, else `BULK_EXPORT` as before.
- `Pointcut`: `STORAGE_PRE_INITIATE_BULK_EXPORT` / `STORAGE_INITIATE_BULK_EXPORT` javadoc mentions the field; fixed a stale `BulkDataExportOptions` parameter reference and a typo.
- 4 new tests in `BulkDataExportProviderR4Test`: override honored (incl. serialize round-trip), default job id when no hook (first test to pin it), client-supplied value ignored, override honored from `STORAGE_INITIATE_BULK_EXPORT` as well.
- Changelog entry for 8.12.0, including two operational caveats: stored parameters containing the new property don't deserialize on older binaries in a mixed-version cluster, and instances of an alternate job definition are not covered by the automatic bulk-export file purge (which only targets `BULK_EXPORT` instances).

Tracked internally as PRAR-4911.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/api/Pointcut.java
@@ -1167,13 +1167,15 @@ public enum Pointcut implements IPointcut {
 	 * Invoked when a Bulk Export job is being kicked off, but before any permission checks
 	 * have been done.
 	 * This hook can be used to modify or update parameters as need be before
-	 * authorization/permission checks are done.
+	 * authorization/permission checks are done. Among the parameters a hook may set is
+	 * <code>jobDefinitionIdOverride</code>, which routes the export to an alternate
+	 * registered batch2 job definition instead of the default <code>BULK_EXPORT</code> job.
 	 * <p>
 	 * Hooks may accept the following parameters:
 	 * </p>
 	 * <ul>
 	 * <li>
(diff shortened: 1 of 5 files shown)


tadgh:
Formatting check succeeded!

maciej-ca:
@jamesagnew  - is looking at a possible alternative.

maciej-ca:
Closing in favor of #8191 (merged as 8959a8bc8bf), which delivers the same capability through a general STORAGE_PRECREATE_BATCH_JOB_INSTANCE pointcut fired for every batch2 job start — interceptors can swap the job definition id and/or rewrite the serialized parameters on the JobInstanceStartRequest before the job-cache check. We have adopted that seam downstream; this PR's jobDefinitionIdOverride field approach is superseded.

PR #8185, opened 2026-07-22
Cross-partition Patient $merge / $undo-merge for multi db Patient ID mode partitioning

### LINKED ISSUE #8195: Improve Patient $merge and $undo-merge support when using the Patient ID Partitioning Interceptor
When using the Patient ID Partitioning Interceptor (`PatientIdPartitionInterceptor`), `Patient/$merge` and `$hapi.fhir.undo-merge` do not handle the following cases well:

- **Sharded deployments** — when partitions live in separate databases they cannot share a single transaction, so a failure partway through a merge leaves the earlier partitions committed. The operation should report what was committed rather than appearing to have rolled back.

- **Non-numeric resource ids** — resources whose ids are client-assigned, or otherwise not server-assigned numeric ids, are not always resolved to the correct partition during a merge.

mrdnctrk:
## What

Cross-partition `Patient/$merge` and `$hapi.fhir.undo-merge` are already supported when using the Patient ID Partitioning Interceptor. Three cases did not work correctly:

- **Partitions in separate databases** — separate databases cannot share a single transaction, so a merge that failed partway through left the earlier partitions committed, with nothing reporting what had actually landed.
- **Non-numeric resource ids** — resources whose ids are client-assigned, or otherwise not server-assigned numeric ids, were not always resolved to the correct partition.
- **Referrers in other partitions** — the lookup for resources referencing the source Patient was not scoped across partitions, so referrers stored in another partition were not discovered and kept pointing at the source after the merge.

This PR fixes all three, and reworks how a merge records its Provenance so that `$undo-merge` can operate partition by partition.

Fixes #8195

## Provenance grouping

When partitioning is enabled and a merge spans partitions, it no longer records a single Provenance.
(A merge within one partition, or on a server without partitioning, is unchanged and still records one.)
A cross-partition merge records:

- **one main Provenance**, carrying the operation's input parameters and outcome as contained resources, and
- **one member Provenance per partition and change type**, carrying no contained resources and listing only the resources it changed.

Every Provenance of one merge shares a group id, held in the `EXT_PROVENANCE_GROUP` extension. The main Provenance stores it bare; each member qualifies it with the partition and change type it records:

```
main:    merge|Patient|<sourceId>|<targetId>|<uuid>
member:  merge|Patient|<sourceId>|<targetId>|<uuid>;partition=0;changeType=update
```

In every Provenance, `target[0]` is the merge target and `target[1]` the source; the resources changed by that Provenance follow from index 2.

This is what allows `$undo-merge` to avoid cross-partition fan-out reads. It locates the main Provenance, collects the members sharing its group id, and restores each one in a transaction pinned to the partition that member names. Members are applied in change-type order — `delete`, then `update`, then `create` — so that resources deleted by the merge are restored before the copies it created are removed.

## Other changes

- **Partition resolution** — referrers are now discovered together with the partition they live in, read from the resource-link table rather than inferred from the resource id. This is what makes non server-assigned numeric ids resolve correctly.
- **Deletes** run in a transaction pinned to an explicit partition rather than as DELETE entries in a transaction bundle: a DELETE entry carries no resource body, so its partition cannot be resolved during a FHIR transaction.
- **Partial-failure reporting** — when partitions commit independently, a failure partway through is not rolled back; the ids already committed are reported in the OperationOutcome for manual reverting. `PartitionedTransactionPartialFailureException` carries the committed response entries per sub-bundle.

## Testing

- `PatientIdModePatientMergeR4Test` / `PatientIdModeNonPatientMergeR4Test` — merge and undo, undo restore ordering (referrer-before-referent, cyclic references), all-or-nothing undo rejection, full rollback when partitions share a transaction, and committed-work reporting when they do not.
- `ResourceMergeServiceTest`, `MergeProvenanceGroupValueTest`, `MergeChangeTypeTest`, `TransactionPartitionProcessorTest`, `FhirSystemDaoTransactionPartitionR5Test`.
- Integration tests against a real multi-database deployment live on the paired CDR branch.


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/merge/MergeOperationTestHelper.java
```diff
@@ -434,37 +457,229 @@ public void assertReferencesNotUpdated(
...
+				provenances,
+				theInputParams,
+				theExpectedSourceId,
+				theExpectedTargetId,
+				theExpectedProvenanceTargets,
+				theExpectedAgents);
+	}
+
+	// ================================================
+	// ================================================
```
volodymyr-korzh: nitpick: looks like these lines should be deleted (you can leave it as is to make sure it makes into relese, not an issue)

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/PartitionedTransactionPartialFailureException.java
```diff
@@ -0,0 +1,53 @@
...
+			String theMessage, List<List<IBase>> theResponseEntriesPerSubBundle, Throwable theCause) {
+		super(deriveStatusCode(theCause), theMessage, theCause);
+		myResponseEntriesPerSubBundle = theResponseEntriesPerSubBundle;
+	}
+
+	private static int deriveStatusCode(Throwable theCause) {
+		if (theCause instanceof BaseServerResponseException serverException) {
+			return serverException.getStatusCode();
+		}
+		return 500;
```
volodymyr-korzh: nitpick: `ca.uhn.fhir.rest.api.Constants.STATUS_HTTP_500_INTERNAL_ERROR` can be used instead of magic 500 number

volodymyr-korzh (approved):

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/merge/MergeOperationTestHelper.java
mrdnctrk: thanks, removed

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/PartitionedTransactionPartialFailureException.java
mrdnctrk: applied

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IResourceLinkDao.java
```diff
@@ -57,8 +56,8 @@ public interface IResourceLinkDao extends JpaRepository<ResourceLink, Long>, IHa
 	 * @return
 	 */
 	@Query(
-			"SELECT DISTINCT new ca.uhn.fhir.model.primitive.IdDt(t.mySourceResourceType, t.mySourceResource.myFhirId) FROM ResourceLink t WHERE t.myTargetResourceType = :resourceType AND t.myTargetResource.myFhirId = :resourceFhirId")
-	Stream<IdDt> streamSourceIdsForTargetFhirId(
+			"SELECT DISTINCT new ca.uhn.fhir.jpa.dao.data.ResourceIdWithPartition(t.myPartitionIdValue, t.mySourceResourceType, t.mySourceResource.myFhirId) FROM ResourceLink t WHERE t.myTargetResourceType = :resourceType AND t.myTargetResource.myFhirId = :resourceFhirId")
+	Stream<ResourceIdWithPartition> streamSourceIdsForTargetFhirId(
```
michaelabuckley: minor question: could this just be JpaPid?  You'd need a new constructor, but JpaPid already has a slot for FHIR id.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ResourceIdWithPartition.java
```diff
@@ -0,0 +1,39 @@
...
+import ca.uhn.fhir.interceptor.model.RequestPartitionId;
+import ca.uhn.fhir.model.primitive.IdDt;
+import jakarta.annotation.Nullable;
+
+/**
+ * A resource identifier (type and FHIR id) together with the partition it lives in. Currently produced by
+ * querying the resource-link table for the sources referencing a given target.
+ */
+// Created by claude-opus
+public record ResourceIdWithPartition(@Nullable Integer partitionId, String resourceType, String fhirId) {
```
michaelabuckley: question: is this a public api or a private detail?  If public, suggest we don't use a record.  Records are hard to evolve.  If this is a private detail, can we narrow visibility lower than public?

michaelabuckley (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/IResourceLinkDao.java
mrdnctrk: thanks. yes, changed to JpaId

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/data/ResourceIdWithPartition.java
mrdnctrk: got rid of this, as using JpaId now.

PR #8186, opened 2026-07-22
mergeback from 8 10 1

jdar8:


tadgh:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/VersionEnum.java
```diff
@@ -200,6 +200,7 @@ public enum VersionEnum {
 	V8_9_0,
 	V8_10_0,
 	V8_10_1,
```
epeartree, line 202: do we need V8_10_2?  

please make sure that we do.

epeartree (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/VersionEnum.java
jdar8: Yup, it's included to make the next mergeback (if needed) easier by eliminating the need for a version bump.

Thanks for the review!

PR #8194, opened 2026-07-24
Change Server to use an interface for method binding.

michaelabuckley:
Introduce an abstract IMethodBinding interface over BaseMethodBinding to allow handlers that are not based on annotated classes.  This is to start supporting programmatic configuration.

Most of the changes are mechanical replacement of BaseMethodBinding with IMethodBinding.

Some sharp edges:
- Conformance builders do raw class instanceof checks for Search and Operations, so the new interface is limited in what it can contribute to the conformance doc.  We will deepen this next release.
- Warnings and tests were using the Method as a key, but that has been replaced by an abstract BindingKey which is just Method.toString() for an annotation-based handler.


tadgh:
Formatting check succeeded!

volodymyr-korzh (approved):

PR #8203, opened 2026-07-29
GL-9002: Speed up subscription delivery by sending messages in batches

tadgh:
WIP: Resolving GL-9002 (SMILE-12408).

Amortizes the synchronous broker round-trip in subscription delivery by making the unit of synchronization a batch rather than a single message.

Companion CDR MR to follow (the broker producer implementations live in cdr-broker).


tadgh:
Formatting check succeeded!

PR #8209, opened 2026-08-01
GH-8208: JobDataSink must enqueue work chunks in their own transaction

### LINKED ISSUE #8208: JobSink mixes the step transaction context with the batch2 state machine.
@../hapi-fhir/hapi-fhir-storage-batch2/src/main/java/ca/uhn/fhir/batch2/coordinator/JobDataSink.java line 100, the call to  enqueueWorkChunkForProcessing does not establish a new transaction like we do on line 95 for the creation call.  Since we share the same connection pool between HAPI and Batch2, the Spring transaction manager will use an existing transaction.  If the step is in the middle of a tx when it sinks the child chunk, the state update of the chunk will be delayed until the step tx commits.  That could be after the child chunk has been delivered to a worker.

The fix is to use a new tx, like we do on line 95.

michaelabuckley:
## For Code Reviewer

`JobDataSink.acceptForStepId` used to call `enqueueWorkChunkForProcessing` (the READY→QUEUED transition) as a bare persistence call, which meant it joined whatever transaction the calling job step had open. For non-gated jobs whose step keeps a long-running transaction (`$mdm-submit`'s `ResourceIdListStep` is the only built-in example — it holds one transaction open across the whole ID stream), the chunk was published to the work channel while the state update was still uncommitted. Worker threads picking up the notification then blocked on the chunk row lock for the duration of the step, and chunks could be left stalled in `READY`.

The fix wraps the enqueue in its own `REQUIRES_NEW` transaction targeted at the default partition (`BT2_WORK_CHUNK` is unpartitioned), matching how chunk creation is already handled. The outbox callback that sends the work channel message is deliberately kept *inside* that new transaction rather than moved outside it — this preserves the transactional-outbox pattern (at-least-once delivery tied to the commit) rather than introducing a window where the message could be sent before the state change commits, or dropped if the send fails after an early commit.

## For QA

- [ ] Non-gated batch2 jobs (e.g. `$mdm-submit`) no longer leave work chunks stalled in `READY` state when a step holds a long-running transaction while creating chunks.
- [ ] Work chunks transition READY → QUEUED and are picked up by worker threads without lock contention delays.
- [ ] Work channel notifications are still sent exactly once per successful enqueue (no duplicate or dropped notifications) — verified by `testSimultaneousDeque_beforeEnqueCommit_doesNotDropChunk`.
- [ ] Gated job execution is unaffected (this change only applies to the non-gated enqueue path).
- [ ] Existing batch2 job processing (bulk export, reindex, etc.) continues to function normally.

Closes #8208

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8208-jobdatasink-enqueue-tx.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8208-jobdatasink-enqueue-tx.yaml
@@ -1,0 +1,8 @@
+---
+type: fix
+issue: 8208
+title: "Previously, Batch2 marked a newly created work chunk as QUEUED using the calling job step's
+  transaction rather than its own. For non-gated jobs whose step holds a long-running transaction,
+  the chunk was published to the work channel while that update was still uncommitted, so worker
+  threads blocked on the chunk's row lock and chunks could be left stalled in READY. The state
+  update now runs in its own transaction, so it commits independently of the calling step."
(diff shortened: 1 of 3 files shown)


tadgh:
Formatting check succeeded!

PR #8214, opened 2026-08-05
Fix trigger subscription partition isolation

volodymyr-korzh:
- Fixed `SubscriptionTriggeringSvcImpl#submitResource` to use `SystemRequestDetails.forRequestPartitionId` instead of `SystemRequestDetails.forAllPartitions()`
- IT test added
- change log

### DIFF
--- a/hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/triggering/SubscriptionTriggeringSvcImpl.java
+++ b/hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/triggering/SubscriptionTriggeringSvcImpl.java
@@ -50,6 +50,7 @@ import ca.uhn.fhir.rest.server.exceptions.InternalErrorException;
 import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import ca.uhn.fhir.rest.server.exceptions.PreconditionFailedException;
 import ca.uhn.fhir.rest.server.exceptions.ResourceGoneException;
+import ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException;
 import ca.uhn.fhir.util.ParametersUtil;
 import ca.uhn.fhir.util.StopWatch;
 import ca.uhn.fhir.util.UrlUtil;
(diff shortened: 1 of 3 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/triggering/SubscriptionTriggeringSvcImpl.java
```diff
@@ -542,7 +543,19 @@ private void submitResource(
 			String theSubscriptionId, RequestPartitionId theRequestPartitionId, String theResourceIdToTrigger) {
 		org.hl7.fhir.r4.model.IdType resourceId = new org.hl7.fhir.r4.model.IdType(theResourceIdToTrigger);
 		IFhirResourceDao dao = myDaoRegistry.getResourceDao(resourceId.getResourceType());
-		IBaseResource resourceToTrigger = dao.read(resourceId, SystemRequestDetails.forAllPartitions());
+		SystemRequestDetails requestDetails = SystemRequestDetails.forRequestPartitionId(theRequestPartitionId);
+		IBaseResource resourceToTrigger;
+		try {
+			resourceToTrigger = dao.read(resourceId, requestDetails);
+		} catch (ResourceNotFoundException e) {
```
codeforgreen, line 550: This is a change in behaviour. I think the operation would turn the exception wrapped into a `IBaseParameters` response. Is this change intentional?

codeforgreen (approved):

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/triggering/SubscriptionTriggeringSvcImpl.java
volodymyr-korzh: This is intentional. `submitResource` method runs inside the job. If `$trigger-subscription` is triggered for 100 resources and 1 of them does not exist or lives in different partition instead of failing the job we now will just ignore this resource. 

I was thinking of making the check if resource is exist before job submission, but this could impact performance of the  `$trigger-subscription` operation.

tadgh (approved):

PR #8217, opened 2026-08-05
Refactor Async Search Coordinator

jamesagnew:
The "async" search coordinator was reworked last year to no longer actually be asynchronous from the invoking HTTP client thread, but a lot of the surrounding infrastructure was not touched. This meant that a simple FHIR search had more database round trips than necessary (often quite a few for a simple search). This PR reworks the async search coordinator into a new "Cache Aware Search Service" which is very deliberate about avoiding unnecessary database round trips.


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
```diff
@@ -2398,26 +2396,8 @@ public IBundleProvider search(
 			cacheControlDirective.parse(theRequest.getHeaders(Constants.HEADER_CACHE_CONTROL));
 		}
 
-		IBundleProvider retVal = mySearchCoordinatorSvc.registerSearch(
+		return mySearchCoordinatorSvc.createNewSearch(
 				this, theParams, getResourceName(), cacheControlDirective, theRequest);
-
-		if (retVal instanceof PersistedJpaBundleProvider provider) {
```
jamesagnew, line 2404: Review note: This functionality has been moved to `BaseResourceReturningMethodBinding`

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/CacheAwareSearchSvcImpl.java
```diff
@@ -0,0 +1,158 @@
...
+import ca.uhn.fhir.rest.server.IPagingProvider;
+import jakarta.persistence.EntityManager;
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * This service performs cache-aware searches. In other words, when executing a search
+ * it will check the {@link ISearchCacheSvc} for any existing cached searches, and if appropriate will store any results it finds back in the
+ * search cache.
+ */
+public class CacheAwareSearchSvcImpl implements ICacheAwareSearchSvc {
```
jmarchionatto: Looks like CacheAwareJpsSearchSvcImpl would be more appropriate, given the delegations

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+
+	public static final SearchCacheStatus CACHE_STATUS_BYPASS = SearchCacheStatus.builder()
+			.withCacheName("HapiFhirQueryCache")
+			.setStatus(SearchCacheStatus.SearchCacheStatusEnum.FWD_BYPASS)
+			.build();
+	private static final int SEARCH_EXPIRY_OFFSET_MINUTES = 10;
+	private static final Logger ourLog = LoggerFactory.getLogger(BaseCacheAwareJpaSearchBundleProvider.class);
+
+	/**
+	 * Adjust this to raise the level of the debuga logs if you are
```
jmarchionatto: debug(a) logs

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+				/*
+				 * If we've already fetched the exact range we want, we are done
+				 */
+				if (myCachedPidsFromMatchesAndIncludes.toIndex() == theToIndex) {
+					return myCachedPidsFromMatchesAndIncludes;
+				}
+
+				/*
+				 * If we've finished the search, and the currently fetched range has the same
+				 * start as the one we want, and the end exceeds the end of the fetched results,
```
jmarchionatto: Comment says the opposite of the code: ...and the end exceeds the end of the fetched results... -> `theToIndex <= myCachedPidsFromMatchesAndIncludes.toIndex()`  

Also, (assuming the code is right, not the comment) it is not clear why that cached result would be OK. What is goint to remove the excess IDs (those between theToIndex and myCachedPidsFromMatchesAndIncludes.toIndex()) from the result?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+			}
+		}
+
+		/*
+		 * Ok, so we need to actually open a database transaction and perform the search. We do this in a loop to
+		 * defend against the case where multiple client threads are all fetching new pages of search results
+		 * concurrently (since they will all try to write new pages of results to the query cache, and only
+		 * one will succeed while the others fail with a constraint error.
+		 */
+		for (int i = 0; ; i++) {
```
jmarchionatto: Problems with this retry:  

1. It's a second retry loop stacked on an existing one. HapiTransactionService.doExecuteInTransaction already retries this, with proper backoff and jitter. Its default maxRetries is 0, so today the outer loop is normally the only active one — but register                             
   UserRequestRetryVersionConflictsInterceptor or send the retry header and you get nested retries: inner maxRetries × 6 outer attempts, each one re-executing a full search pass. The outer loop has no idea the inner one ran.

2. Fixed 500ms, no backoff, no jitter. This is precisely the lockstep-recollision problem jitter exists to prevent — N contending threads all wake at the same instant and collide again. The layer below it gets this right; this layer doesn't.

3. The retry path is untestable and untested. new SleepUtil() is constructed inline, so it can't be stubbed — SleepUtil's own javadoc says the instance methods exist "for easier mocking", and HapiTransactionService correctly holds it as an injectable field. Nothing in the repo greps for the "Constraint error while writing search results" message, so this branch has no coverage; exercising it costs 3 real seconds.

4. SleepUtil.sleepAtLeast busy-spins on interrupt: it restores the interrupt flag and then loops, so every subsequent Thread.sleep throws instantly. A cancelled request or a shutdown turns each retry into a 500ms hot spin plus an ERROR log per iteration — up to 2.5s of burn across 5   
   retries.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+					searchBuilder,
+					myCompositeBroadcaster,
+					myFetchedResources);
+			SearchProgressTracker outcome;
+			try {
+				outcome = searchBuilder.performSearchForPids(
+						consumer, myParams, searchDetails, myRequestDetails, myRequestPartitionId);
+				consumer.firePreAccessHooksOnNewPids();
+			} catch (Exception e) {
+				SearchCoordinatorSvcImpl.markSearchAsFailedWithExceptionDetails(mySearchEntity, e);
```
jmarchionatto, line 549: Shouldn't rethrow retriable exceptions and only mark FAILED for terminal ones?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+			myParams = extractSearchParameterMapFromSearchEntity();
+		}
+
+		/// If we have a `_count=summary` query, just calculate the count and return
+		if (SearchParameterMapCalculator.isWantOnlyCount(myParams)) {
+			if (mySearchEntity.getTotalCount() == null) {
+				Long countQuery = newSearchBuilder()
+						.createCountQuery(myParams, mySearchEntity.getUuid(), myRequestDetails, myRequestPartitionId);
+				mySearchEntity.setSearchParameterMap(myParams);
+				mySearchEntity.setTotalCount(Math.toIntExact(countQuery));
```
jmarchionatto: Same set has null check later(:595). Which is right?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+				.addArgument(mySearchEntity.getNumFound())
+				.addArgument(mySearchEntity.getNumBlocked())
+				.log();
+
+		mySearchCacheSvc.save(mySearchEntity, myRequestPartitionId);
+		mySearchResultCacheSvc.storeResults(
+				mySearchEntity, previouslyFoundPids, foundPidsToStore, myRequestDetails, myRequestPartitionId);
+
+		int numberToReturn = theToIndex - theFromIndex;
+		while (pidsToReturn.size() > numberToReturn) {
```
jmarchionatto: suggestion: 
```
  if (pidsToReturn.size() > numberToReturn) {
      pidsToReturn.subList(numberToReturn, pidsToReturn.size()).clear(); 
  } 
```

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+		boolean deduplicateInDatabase;
+
+		List<Integer> thresholds = myStorageSettings.getSearchPreFetchThresholds();
+		int firstThreshold = thresholds.get(0);
+		int lastThreshold = thresholds.get(thresholds.size() - 1);
+
+		/*
+		 * If we're searching for the first page of results:
+		 *
+		 * - If the requested count is greater than the first threshold, just search
```
jmarchionatto, line 664: Why is this the case?  

This comment points to design details worth to be documented along with the code. I suggest a README at the search package level, pointed from different sections of the code. It would save a lot of dev time figuring out the design from the code, with dubious outcome.

(*) This could be the most important comment of this review!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/CacheAwareJpaSearchBundleProviderFirstPage.java
```diff
@@ -0,0 +1,230 @@
...
+								myParams,
+								retVal.getResourceType(),
+								myRequestDetails,
+								myParams.toNormalizedQueryString(),
+								myRequestPartitionId);
+						if (cachedQueryOpt.isPresent()) {
+							retVal = cachedQueryOpt.get();
+
+							myCacheStatus = SearchCacheStatus.builder()
+									.withCacheName("HapiQueryCache")
```
jmarchionatto: Constant is warranted.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/CacheAwareJpaSearchBundleProviderFirstPage.java
```diff
@@ -0,0 +1,230 @@
...
+						}
+					}
+				}
+			} else {
+				myCacheStatus = CACHE_STATUS_BYPASS;
+			}
+
+		} else {
+
+			/// If we get here, we're in a second transaction on the same IBundleProvidr instance. This
```
jmarchionatto: Why does `myCandidateSearchEntity == null` mean we're in a second transaction on the same IBundleProvider instance? 

Add ref to design README section.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/StatelessJpaSearchSvcImpl.java
```diff
@@ -182,138 +171,157 @@ public IBundleProvider executeQuery(
...
 						compositeBroadcaster.callHooks(Pointcut.STORAGE_PREACCESS_RESOURCES, params);
-					}
 
-					for (int i = pids.size() - 1; i >= 0; i--) {
-						if (accessDetails.isDontReturnResourceAtIndex(i)) {
-							pids.remove(i);
+						for (int i = pids.size() - 1; i >= 0; i--) {
+							if (accessDetails.isDontReturnResourceAtIndex(i)) {
+								pids.remove(i);
+								loadedResources.remove(i);
```
jmarchionatto, line 206: How about an initial collections size match sanity check before indexed removal?

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8217-optimize-cache-aware-searching.yaml
```diff
@@ -0,0 +1,10 @@
+---
+type: perf
+issue: 8217
+title: "The 'async' search coordinator (used for cache-aware searching, which is generally
+   the default kind of search) was reworked last year to no longer actually be
```
jmarchionatto: I'm pretty sure this was several years ago, three at least.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8217-optimize-cache-aware-searching.yaml
```diff
@@ -0,0 +1,10 @@
+---
+type: perf
+issue: 8217
+title: "The 'async' search coordinator (used for cache-aware searching, which is generally
+   the default kind of search) was reworked last year to no longer actually be 
+   asynchronous from the invoking HTTP client thread, but a lot of the surrounding
+   infrastructure was not touched. This meant that a simple FHIR search had more database
+   round trips than necessary (often quite a few for a simple search). 
+   This coordinator has been reworked into into a new 'Cache Aware Search Service' which
```
jmarchionatto: into into

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java
jmarchionatto: Thanks for this comment! Good idea.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -1033,7 +1048,7 @@ private void createChunkedQueryForEverythingSearch(
 					theSearchQueryProperties.isDoCountOnlyFlag(),
 					false);
 			GeneratedSql allTargetsSql = fetchPidsSqlBuilder.generate(
-					theSearchQueryProperties.getOffset(), mySearchProperties.getMaxResultsRequested());
+					theSearchQueryProperties.getOffset(), theSearchQueryProperties.getMaxResultsRequested());
```
jmarchionatto, line 1052: May merit specific unit testing

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
```diff
@@ -3234,7 +3254,8 @@ private void fetchNext() {
 
 						if (nextPid != null) {
 							if (!myPidSet.contains(nextPid)) {
-								if (!mySearchProperties.isDeduplicateInDatabase()) {
+								if (!mySearchProperties.isDeduplicateInDatabase()
+										|| myFetchIncludesForEverythingOperation) {
 									/*
```
jmarchionatto, line 3258: Comment needs update to match new condition.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/DatabaseBackedPagingProvider.java
```diff
@@ -63,26 +55,24 @@ public DatabaseBackedPagingProvider(int theSize) {
...
-	 * Subclasses may override in order to modify the bundle provider being returned
+	 * Subclasses may override and modify or replace the bundle provider being returned
 	 */
 	@Nullable
-	protected PersistedJpaBundleProvider validateAndReturnBundleProvider(PersistedJpaBundleProvider theBundleProvider) {
-		if (!theBundleProvider.ensureSearchEntityLoaded()) {
-			return null;
-		}
-
+	protected BaseCacheAwareJpaSearchBundleProvider validateAndReturnBundleProvider(
```
jmarchionatto, line 70: Should it be renamed now that it validates nothing?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/PersistedJpaHistoryBundleProvider.java
```diff
@@ -19,61 +19,51 @@
...
 
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Optional;
-import java.util.Set;
 import java.util.function.Function;
 
+/**
```
jmarchionatto, line 64: Could the class be renamed accordingly?

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/PersistedJpaHistoryBundleProvider.java
```diff
@@ -127,40 +102,20 @@ public class PersistedJpaBundleProvider implements IBundleProvider {
...
-		return getSearchEntity();
-	}
-
-	protected Search getSearchEntity() {
-		return mySearchEntity;
-	}
-
 	// Note: Leave as protected, HSPC depends on this
 	@SuppressWarnings("WeakerAccess")
 	protected void setSearchEntity(Search theSearchEntity) {
```
jmarchionatto, line 121: This allows an external subclass to install a non-history entity. Validation is in order.

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.java
```diff
@@ -1804,13 +1804,15 @@ public List<Integer> getSearchPreFetchThresholds() {
 	 * </p>
 	 */
 	public void setSearchPreFetchThresholds(List<Integer> thePreFetchThresholds) {
-		Validate.isTrue(thePreFetchThresholds.size() > 0, "thePreFetchThresholds must not be empty");
+		Validate.isTrue(!thePreFetchThresholds.isEmpty(), "thePreFetchThresholds must not be empty");
 		int last = 0;
```
jmarchionatto: currentExceptLast, currentButNotLast?

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ServerInterceptorUtil.java
```diff
@@ -48,22 +48,36 @@ public static List<IBaseResource> fireStoragePreshowResource(
...
+	 * Fires {@link Pointcut#STORAGE_PRESHOW_RESOURCES} interceptor hook, and potentially remove resources
+	 * from the resource list
+	 */
+	@CheckReturnValue
+	public static List<IBaseResource> fireStoragePreshowResourcesToCompositeBroadcaster(
+			List<IBaseResource> theResources,
+			RequestDetails theRequest,
+			IInterceptorBroadcaster theCompositeBroadcaster) {
 		List<IBaseResource> retVal = theResources;
 		retVal.removeIf(Objects::isNull);
```
jmarchionatto: I know this is not new, but wondering if it is correct to have this side effect here, happening independently of the hook existing or not.

hapi-fhir-jpaserver-test-dstu2/src/test/java/ca/uhn/fhir/jpa/search/SearchCoordinatorSvcImplTest.java
```diff
@@ -299,8 +230,11 @@ public void testAsyncSearchLargeResultSetSameCoordinator() {
 		assertEquals("10", resources.get(0).getIdElement().getValueAsString());
 		assertEquals("39", resources.get(29).getIdElement().getValueAsString());
 
+		assertNotNull(result.getUuid());
+		assertNull(result.size());
+
 		verify(mySearchCacheSvc, atLeastOnce()).save(mySearchCaptor.capture(), any());
-		assertEquals(790, mySearchCaptor.getValue().getNumFound());
+		assertEquals(31, mySearchCaptor.getValue().getNumFound());
```
jmarchionatto, line 237: Niiiiiice!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/JpaPreResourceAccessDetails.java
```diff
@@ -38,14 +35,20 @@ public class JpaPreResourceAccessDetails implements IPreResourceAccessDetails {
...
-	private final ICallable<ISearchBuilder> mySearchBuilderSupplier;
 	private List<IBaseResource> myResources;
 
-	public JpaPreResourceAccessDetails(
-			List<JpaPid> theResourcePids, ICallable<ISearchBuilder> theSearchBuilderSupplier) {
+	/**
+	 * Constructor
+	 */
+	public JpaPreResourceAccessDetails(List<JpaPid> theResourcePids, List<IBaseResource> theUnsyncedResources) {
+		Validate.isTrue(
```
jmarchionatto, line 44: `SearchBuilder.doLoadPids` skips any PID whose current version is a delete marker, and only pads `theResourceListToPopulate` up to the highest index it actually writes. So whenever the trailing PID(s) of the batch resolve to deleted resources, the returned list is *shorter* than the PID list and `Validate.isTrue` throws `IllegalArgumentException` for
the whole search.

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/triggering/SubscriptionTriggeringSvcImpl.java
```diff
@@ -364,35 +363,23 @@ private void processAsynchronous(
...
 				theJobDetails.getRequestPartitionId());
 
-		List<? extends IResourcePersistentId<?>> allResourceIds;
-		RequestPartitionId requestPartitionId = theJobDetails.getRequestPartitionId();
+		List<IBaseResource> allResources;
 		try {
-			allResourceIds = mySearchCoordinatorSvc.getResources(
-					theJobDetails.getCurrentSearchUuid(), fromIndex, toIndex, null, requestPartitionId);
+			IBundleProvider search = mySearchCoordinatorSvc.continueExistingSearch(
+					theJobDetails.getCurrentSearchUuid(), new SystemRequestDetails());
```
jmarchionatto: Using new SystemRequestDetails() looses theJobDetails partition info

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+
+		public void firePreAccessHooksOnNewPids() {
+
+			// Interceptor call: STORAGE_PREACCESS_RESOURCES
+			// This can be used to remove results from the search result details before
+			// the user has a chance to know that they were in the results
+			if (myCompositeBroadcaster.hasHooks(Pointcut.STORAGE_PREACCESS_RESOURCES) && !myNewPidsThisPass.isEmpty()) {
+				Set<JpaPid> blockedPids = new HashSet<>();
+
+				List<IBaseResource> newResources =
```
jmarchionatto: This is full resource bodies for *every* PID found in the pass, not just the requested page. `myNewPidsThisPass` is bounded only by `SearchThreshold.threshold()`, and the default thresholds are
`[13, 503, 2003, 1000003, -1]`. Reaching the 4th tier loads ~1,000,003 resource bodies in one call; the `-1` tier makes `threshold` `null`, so the
consumer never returns `STOP` and the pass accumulates every matching PID — unbounded.

Every one of those resources is also put into `myFetchedResources`, a `HashMap` field that is never cleared for the life of the provider.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/entity/Search.java
```diff
@@ -28,6 +28,8 @@
 import ca.uhn.fhir.rest.param.HistorySearchStyleEnum;
 import ca.uhn.fhir.rest.server.util.ICachedSearchDetails;
 import ca.uhn.fhir.system.HapiSystemProperties;
+import ca.uhn.fhir.util.HapiToStringBuilder;
+import graphql.VisibleForTesting;
```
jmarchionatto: Wrong import

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/ResponsePage.java
```diff
@@ -141,6 +147,10 @@ public List<IBaseResource> getResourceList() {
 		return myResourceList;
 	}
 
+	public boolean isHasNextPage() {
+		return myHasNextPage;
```
jmarchionatto: Unboxing nullable could NPE

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/DatabaseBackedPagingProvider.java
```diff
@@ -63,26 +55,24 @@ public DatabaseBackedPagingProvider(int theSize) {
 	}
 
 	@Override
-	public synchronized IBundleProvider retrieveResultList(RequestDetails theRequestDetails, String theId) {
-		PersistedJpaBundleProvider provider = myPersistedJpaBundleProviderFactory.newInstance(theRequestDetails, theId);
-		return validateAndReturnBundleProvider(provider);
+	public IBundleProvider retrieveResultList(RequestDetails theRequestDetails, @Nonnull String theId) {
+		BaseCacheAwareJpaSearchBundleProvider retVal = (BaseCacheAwareJpaSearchBundleProvider)
```
jmarchionatto, line 59: mySearchCoordinatorSvc.continueExistingSearch returns `IBundleProvider` so any alternate implementation yields ClassCastException here.  
Also, method previously returned null when the search entity could not be loaded; now it always returns a provider and the failure surfaces later as a ResourceGoneException from getResources()/size()

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/StatelessSearchSvcImpl.java
```diff
@@ -339,7 +347,7 @@ public IBundleProvider executeQuery(
 						}
 					}
 
-					bundleProvider.setPreferredPageSize(theParams.getCount());
+					bundleProvider.setPreferredPageSize(resources.size());
```
jmarchionatto: resources includes _include/_revinclude resources. When the client sends no _count, BaseResourceReturningMethodBinding.buildResponseBundleRequest uses preferredPageSize() as the effective count, which then feeds offsetCalculatePageSize and the _count value written into the paging links. A search returning 10 matches plus 40 includes now advertises _count=50 on the next link.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/warm/CacheWarmingSvcImpl.java
```diff
@@ -95,7 +96,10 @@ private void refreshNow(WarmCacheEntry theCacheEntry) {
 		String queryPart = parseWarmUrlParamPart(nextUrl);
 		SearchParameterMap responseCriteriaUrl = myMatchUrlService.translateMatchUrl(queryPart, resourceDef);
 
-		callingDao.search(responseCriteriaUrl);
+		IBundleProvider search = callingDao.search(responseCriteriaUrl);
+
+		// Fetch the first 50 resources to get them into the cache.
+		search.getResources(0, 50);
```
jmarchionatto: Why 50? Shouldn't come from thresholds definition?
If not it should be a constant with a comment explaining the choice.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+		// we will send the resource list to our interceptors
+		// this can (potentially) change the results being returned.
+		int precount = retVal.size();
+		retVal = ServerInterceptorUtil.fireStoragePreshowResourcesToCompositeBroadcaster(
+				retVal, myRequestDetails, myCompositeBroadcaster);
+
+		// we only care about omitted results from this page
+		theResponsePageBuilder.setOmittedResourceCount(precount - retVal.size());
+		theResponsePageBuilder.setResources(retVal);
+		theResponsePageBuilder.setIncludedResourceCount(retVal.size());
```
jmarchionatto: retVal is not included resources

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+		theResponsePageBuilder.setTotalRequestedResourcesFetched(mySearchEntity.getNumFound());
```
jmarchionatto, line 294: Should be this page's only, but set to cumulative count

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
```diff
@@ -0,0 +1,1013 @@
...
+ * <ul>
+ *     <li>{@link CacheAwareJpaSearchBundleProviderFirstPage}</li> is used for the initial search
+ *     <li>{@link CacheAwareJpaSearchBundleProviderSubsequentPage}</li> is used for the following page loads from the query cache
+ * </ul>
+ *
+ * @see IStatelessSearchSvc The Synchronous Search Service is used instead of this class for searches which don't use the search cache
+ * @see ca.uhn.fhir.jpa.search.PersistedJpaBundleProvider The search result for <b>FHIR History</b> operations.
+ * @since 8.14.0
+ */
+public abstract class BaseCacheAwareJpaSearchBundleProvider implements IBundleProvider {
```
jmarchionatto, line 100: Class merits its own unit test

jmarchionatto (approved):
Looks good. Approved with comments and suggestions

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/CacheAwareSearchSvcImpl.java
jamesagnew: Makes sense. Changing.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: fixed

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: Good call... The whole `theResponsePageBuilder` is pretty awful API, it has no documentation at all. I didn't realize that was the intent. Fixing this will probably fix another issue I had to work around with paging....

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: Good call. Also fixed, and also added documentation to the page builder

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/warm/CacheWarmingSvcImpl.java
jamesagnew: Good call - using the first threshold makes sense.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/StatelessSearchSvcImpl.java
jamesagnew: Good point, will subtract the includes from the count

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: This is a great point about the potential nested retries. I've reworked this to just leverage the HapiTransactionService's retry functionality instead, and added a utility to that class to allow callers to specify a specific retry count.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: Added

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: This is a great point. I might as well just rely on the HapiTransactionService's retry functionality. I'm adding a method to allow a fixed retry count to a HapiTransactionService execution, and getting rid of the loop here.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: You're right that this isn't clear. I'm rewriting the comment and the logic:
```java
/*
 * We can reuse the previously fetched matches if the range being
 * requested now is the same as the one we already fetched. We can
 * also reuse if the end of the requested range (theToIndex) is
 * >= the total number of matches for the search, and <= the
 * upper end of the range (since this means that no matter how
 * high theToIndex is, we already have all the matches we need)
 */
```
The logic is sound though - If we're fetching past the end of the total number of results, then there is nothing to remove from those cached PIDs. They contain all the results they need to, and there won't be any extras to purge.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: I don't actually think it makes any sense here... We're not writing at this point, so any exceptions we catch are fatal (likely bugs) in our search execution

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: I guess we should defend here. The current implementation can never return null, but the API does use a `Long` and not a `long` so someone could change that in the future..

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: Makes sense, replacing as suggested

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: Have added a package-info.java, and beefed up this comment to include rationale.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/CacheAwareJpaSearchBundleProviderFirstPage.java
jamesagnew: Added

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/CacheAwareJpaSearchBundleProviderFirstPage.java
jamesagnew: Added more docs to explain this.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/StatelessJpaSearchSvcImpl.java
jamesagnew: Good call. Added

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8217-optimize-cache-aware-searching.yaml
jamesagnew: changed

hapi-fhir-jpaserver-model/src/main/java/ca/uhn/fhir/jpa/api/config/JpaStorageSettings.java
jamesagnew: How about "previousThreshold"

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/ServerInterceptorUtil.java
jamesagnew: Yeah I think you're right, I'll move it down to the `if` so it only happens if we're actually calling the interceptor

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/interceptor/JpaPreResourceAccessDetails.java
jamesagnew: Wow, great catch. FIxed by making sure we always increase the resource list to the appropriate size.

hapi-fhir-jpaserver-subscription/src/main/java/ca/uhn/fhir/jpa/subscription/triggering/SubscriptionTriggeringSvcImpl.java
jamesagnew: Good point. I just checked and there are tests for this but they were passing only by fluke. Have beefed up the tests and fixes this.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/exec/BaseCacheAwareJpaSearchBundleProvider.java
jamesagnew: Great point. I've reworked this to not keep more than what's actually needed for the current page in the local cache

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/DatabaseBackedPagingProvider.java
jamesagnew: I'll add a comment - nothing is done in the default implementation, but the purposes is for overriders and megascale does override it.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/PersistedJpaHistoryBundleProvider.java
jamesagnew: Done

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/PersistedJpaHistoryBundleProvider.java
jamesagnew: Added validation

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/entity/Search.java
jamesagnew: Fixed

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/method/ResponsePage.java
jamesagnew: I'm removing this method - your suggestions around the ResponsePageBuilder made this unnecessary anyhow

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/SearchBuilder.java
jamesagnew: Yup, this is tested in PatientEverythingPaginationR4Test. The previous use was wrong but just happened to work because we were incorrecrly ignoring the offset in the same SearchProperties object.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/DatabaseBackedPagingProvider.java
jamesagnew: These are both deliberate. The cast is totally safe - that interface is defined in the hapi-fhir-storage project, but we already know that the JPA server implementation of that interface will always return `BaseCacheAwareJpaSearchBundleProvider` so this class (which is also in the JPA server project) can make that assumption. I'll add a comment here explaining that.

The reason we no longer eagerly throw an exception is to avoid opening a database transaction and closing it again just to confirm whether any search matches the UUID. Delating that search means we only open a single DB transaction, which is more efficient. The end result to the client is the same. I'll add a comment about that too.

PR #8219, opened 2026-08-05
Fix Changelog action  bug

tadgh:
Correctly determine which folder should be used, fixes a bug with snapshots expecting the snapshot folder (should be N+1 on snapshot, not N)


tadgh:
Formatting check succeeded!

PR #8222, opened 2026-08-05
Make migrator check optional

jamesagnew:
#8028 introduced a new baseline capability. It looks like this is interfering with the way that Smile handles migrations, so this PR adds a new flag to explicitly enable the baseline check.


tadgh:
Formatting check succeeded!

Issue #8200, opened 2026-07-28
Patient/$merge fails with HTTP 412 (HAPI-2476) when source and target Patients resolve to the same partition in Patient ID partitioning mode


mrdnctrk:
**Describe the bug**

A `Patient/$merge` is rejected with **HTTP 412 / `HAPI-2476`** when all of the following hold:

- `PatientIdPartitionInterceptor` is in use (Patient ID partitioning mode), and
- the source and target Patients happen to hash to the **same partition**, and
- a patient-compartment resource (e.g. an `Observation` with `subject` = source Patient) references the source, and
- `PatientCompartmentEnforcingInterceptor` is registered.

The merge should succeed: the referencing resource does not change partition, it only needs its `subject` rewritten to the target Patient.

**Root cause**

Two behaviours combine:

1. `PatientIdPartitionInterceptor.provideSingleCompartmentPartition(...)` returns
   `RequestPartitionId.fromPartitionIdAndName(partitionId, thePatientIdPart)` — it puts the **patient's ID part into the partition name**.
2. `PatientCompartmentEnforcingInterceptor.storagePreStorageResourceUpdated(...)` compares the resource's old and new partition using `RequestPartitionId.toJson()`, and `partitionNames` is a serialized field of `RequestPartitionId`.

So the "did this resource change compartment?" check degenerates into "did the referenced Patient ID change?" — which is **always true during a merge**, even when the resource does not actually move partitions.

The interceptor's own warning shows it clearly — the partition ID is identical on both sides, only the name differs:

```
Resource compartment for Observation/1004 changed
  from {"allPartitions":false,"partitionIds":[14829],"partitionNames":["src-4963"]}
  to   {"allPartitions":false,"partitionIds":[14829],"partitionNames":["tgt-1"]}
```

Resulting error:

```
HAPI-2476: Resource compartment for Observation/1004 changed. Was a referenced Patient changed?
```

thrown from `PatientCompartmentEnforcingInterceptor.storagePreStorageResourceUpdated`.

**To Reproduce**

A failing reproduction test is on branch **`emre_same-partition-merge-compartment-referrer_repro`** (single commit, based on `master`):

- Branch: https://github.com/hapifhir/hapi-fhir/tree/emre_same-partition-merge-compartment-referrer_repro
- Commit: https://github.com/hapifhir/hapi-fhir/commit/556deacdfbb12cca4e3fa0586338ed64d3ba79c2
[...]

PR #8231, opened 2026-08-07
Fix same-partition Patient $merge blocked by compartment enforcement

mrdnctrk:
Fixes https://github.com/hapifhir/hapi-fhir/issues/8200.

### The problem

In Patient ID partitioning mode, a Patient `$merge` failed with HTTP 412 / `HAPI-2476` when the source and
target Patients resolved to the **same** partition id and a patient-compartment resource referenced the
source.

The rejection comes from **`PatientCompartmentEnforcingInterceptor`**, which blocks any update that would
move a resource out of the patient compartment it was stored under. It and the merge asked different
questions:

- **The enforcer** compares the whole `RequestPartitionId`. `PatientIdPartitionInterceptor` puts the patient
  id part in the partition *name*, so the enforcer correctly sees the compartment would change, and throws
  `Msg.code(2476)`.
- **The merge** compared only the numeric partition id. Two patient ids can hash to the same partition, and
  when they do it found no difference — so it concluded the resource did not need to move and updated it in
  place, which is exactly the write the enforcer rejects.

### The fix

- `ResourceMergeService` (routing) and `PartitionAwareReplaceReferencesSvc` (per-resource classification)
  now compare the full `RequestPartitionId`, recomputed from resource content — the same question the
  enforcer asks.
- That computation moved to `IRequestPartitionHelperSvc`, and `PatientCompartmentEnforcingInterceptor` was
  refactored onto it, so the two cannot drift apart.

A compartment resource whose compartment changes is therefore **moved** rather than updated in place, even
when both patients hash to the same partition, and undo restores it under its original id.

### Tests

`PatientIdModePatientMergeR4Test.SamePartitionMergeWithCompartmentEnforcement` uses two patient ids that
collide onto one partition. It fails on the base branch and passes here.

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8200-same-partition-merge-compartment-enforcement.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_12_0/8200-same-partition-merge-compartment-enforcement.yaml
@@ -1,0 +1,4 @@
+---
+type: fix
+issue: 8200
+title: "Previously, when the Patient ID Partitioning Interceptor was used together with the Patient Compartment Enforcing Interceptor, a Patient $merge could fail with an HTTP 412 error, because the merge tried to update patient compartment resources in place. This is now fixed by creating copies of those resources in the target Patient's compartment during the merge operation, similar to a cross-partition merge."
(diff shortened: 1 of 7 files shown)


tadgh:
Formatting check succeeded!

volodymyr-korzh (approved):

PR #8244, opened 2026-08-11
Fix for $meta-add and $meta-delete operations within transaction bundles

### LINKED ISSUE #8243: $meta-add / $meta-delete instance operations in a transaction Bundle are rejected by AuthorizationInterceptor (HAPI-1742)
### Issue

When an instance-level operation such as `$meta-add` or `$meta-delete` is invoked as an entry inside a `transaction` Bundle, the request fails with `HAPI-1742: Invalid FHIR resource URL` whenever an `AuthorizationInterceptor` is registered on the server. The operation is never dispatched.

The transaction processor itself supports these operations (`BaseTransactionProcessor.parseUrlForOperationInvocation` / `invokeOperationInTransaction`, added in #7734). The failure happens earlier, during authorization: `RuleImplOp.applyRuleToTransaction` calls `UrlUtil.parseUrl(entry.request.url)` on each entry to determine the target resource type/id for rule evaluation. `UrlUtil.parseUrl("Patient/p1/$meta-add")` throws `HAPI-1742` because the trailing `$meta-add` segment is neither a resource id nor `_history`.

Because it reproduces only when an `AuthorizationInterceptor` is in the interceptor chain, the DAO-level tests (`BaseTransactionProcessorTest.testOperation_MetaAdd_*`) pass while a real secured server fails.

### Steps to reproduce

1. `PUT /Patient/p1`
2. `POST` the following transaction Bundle to the base endpoint of a server that has an `AuthorizationInterceptor` registered:

```json
{
  "resourceType": "Bundle",
  "type": "transaction",
  "entry": [
    {
      "resource": {
        "resourceType": "Parameters",
        "parameter": [
          {
            "name": "meta",
            "valueMeta": {
              "tag": [
                { "system": "http://foo", "code": "bar1" }
              ]
            }
          }
        ]
      },
      "request": {
        "method": "POST",
        "url": "Patient/p1/$meta-add"
      }
    }
  ]
}
```

### Expected behavior

`200` — the `$meta-add` (and likewise `$meta-delete`) operation is applied to `Patient/p1`, consistent with invoking the operation outside of a transaction.

### Actual behavior

`400 HAPI-1742: Invalid FHIR resource URL: Patient/p1/$meta-add`. Same result when the entry URL is changed to `$meta-delete`.

### Environment

- FHIR version: R4
- Reproduced on Smile CDR 2026.05.PRE-48 (HAPI FHIR-based).

jorgepinilla-smile:
Fixes #8243

### Root Cause

RuleImplOp.applyRuleToTransaction parsed each entry's request.url with UrlUtil.parseUrl(...) to derive the target resource for rule evaluation. For Patient/p1/$meta-add, the trailing $meta-add segment isn't a resource id or _history, so parseUrl threw HAPI-1742 — before the transaction processor (which already supports these operations, #7734) was ever reached.

### Changes:

- RuleImplOp: detect an operation URL (trailing segment starting with $) and parse only the ResourceType/id prefix, so authorization resolves against Patient/p1.
- Added isMetaOperation(...) and exempted $meta-add / $meta-delete from the nested-Parameters guard (Msg.code(339)), since these legitimately carry a Parameters body via POST.

### DIFF
--- a/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
+++ b/hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
@@ -60,7 +60,6 @@ import java.util.stream.Collectors;
 
 import static org.apache.commons.lang3.StringUtils.defaultString;
 import static org.apache.commons.lang3.StringUtils.isBlank;
-import static org.apache.commons.lang3.StringUtils.isEmpty;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 import static org.hl7.fhir.instance.model.api.IAnyResource.SP_RES_ID;
 
@@ -753,13 +752,13 @@ class RuleImplOp extends BaseRule /* implements IAuthRule */ {
 
 				IBaseResource inputResource = nextPart.getResource();
 				IIdType inputResourceId = null;
-				String operationName = null;
+				String instanceOperation = null;
 				if (isNotBlank(url)) {
 
 					int lastSlashIdx = url.lastIndexOf('/');
 					if (lastSlashIdx != -1) {
-						operationName = url.substring(lastSlashIdx + 1);
-						if (operationName.startsWith("$")) {
+						instanceOperation = url.substring(lastSlashIdx + 1);
+						if (instanceOperation.startsWith("$")) {
 							url = url.substring(0, lastSlashIdx);
 						}
 					}
@@ -806,7 +805,8 @@ class RuleImplOp extends BaseRule /* implements IAuthRule */ {
 				// This is done because a Bundle can contain a nested PATCH with Parameters,
 				// which is supported but a non-PATCH nested Parameters resource may be problematic.
 				// The exception being meta operations such as $meta-add and $meta-delete
-				if (!isMetaOperation(operationName) && isInvalidNestedParametersRequest(ctx, nextPart, operation)) {
+				if (!RestOperationTypeEnum.isMetaOperation(instanceOperation)
+						&& isInvalidNestedParametersRequest(ctx, nextPart, operation)) {
 					throw new InvalidRequestException(Msg.code(339)
 							+ String.format("Can not handle nested Parameters with %s operation", operation));
 				}
@@ -848,15 +848,6 @@ class RuleImplOp extends BaseRule /* implements IAuthRule */ {
 		} else {
 			return null;
 		}
-	}
-
-	private boolean isMetaOperation(String theOperationName) {
-		if (isEmpty(theOperationName)) {
-			return false;
-		}
-
-		return theOperationName.equals(RestOperationTypeEnum.META_ADD.getCode())
-				|| theOperationName.equals(RestOperationTypeEnum.META_DELETE.getCode());
 	}
 
 	@Nullable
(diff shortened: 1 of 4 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
```diff
@@ -795,7 +805,8 @@ private Verdict applyRuleToTransaction(
 
 				// This is done because a Bundle can contain a nested PATCH with Parameters,
 				// which is supported but a non-PATCH nested Parameters resource may be problematic.
-				if (isInvalidNestedParametersRequest(ctx, nextPart, operation)) {
+				// The exception being meta operations such as $meta-add and $meta-delete
+				if (!isMetaOperation(operationName) && isInvalidNestedParametersRequest(ctx, nextPart, operation)) {
```
codeforgreen: question: Can't we use RestOperationTypeEnum to create the condition for $meta-add and $meta-delete? It contains those entries already.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
```diff
@@ -744,16 +745,25 @@ private Verdict applyRuleToTransaction(
...
+				if (isNotBlank(url)) {
+
+					int lastSlashIdx = url.lastIndexOf('/');
+					if (lastSlashIdx != -1) {
+						operationName = url.substring(lastSlashIdx + 1);
+						if (operationName.startsWith("$")) {
+							url = url.substring(0, lastSlashIdx);
+						}
+					}
+					UrlUtil.UrlParts parts = UrlUtil.parseUrl(url);
```
codeforgreen, line 765: question: I see the url manipulation logic changes. Is this a different bug being fixed? if so please add a different changelog and test it separately. Also it would be nice to have this parse logic added to UrlUtil as well for reuse and separation of responsabilities.

codeforgreen: Probably this logic could not handle having operations within a transaction. That sounds like a separate bug if true. My point is, there is a separate branch - operation within a transaction that is not meta-add or meta-delete. If that is a real use-case we should test it separately as well.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
codeforgreen: I see that we process bundle entries and we don't have this for each entry.

codeforgreen (approved):

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
jorgepinilla-smile: no, logic change is part of the bug fix. Main issue was that the url contained an operation for the instance which meant that this would error when it tried to parse it.

The changes mentioned in the comment above are a side effect of fixing this bug. Because the meta operation is allowed at the DAO for transaction bundles but it was never enabled at the authorization level.

hapi-fhir-server/src/main/java/ca/uhn/fhir/rest/server/interceptor/auth/RuleImplOp.java
jorgepinilla-smile: not sure what you mean here, this particular block of code is executed per entry.

jorgepinilla-smile: Great suggestion! I've moved the "isMetaOperation" method to RestOperationTypeEnum.

PR #8246, opened 2026-08-12
Jr 20260806 manage subscription permission

JasonRoberts-smile:
This MR introduces a hook in the `SubscriptionValidatingInterceptor` to allow an implementer to control which users are able to create or update Subscription resources. 

Since subscriptions run with system privileges, they can access resources that the user would not normally be allowed to read, and route them to arbitrary endpoints. Depending on the permission to write Subscriptions is not enough, since that permission could be granted through a wildcard. The default behaviour is unchanged for backward compatibility, but the new hook gives implementers the option of writing fine-grained security rules that take into account the user session, the contents of the Subscription, the partition being written to and whether the create or update pointcut was called.

In the edge case where the system itself is updating a Subscription, this must not be denied or the Subscription functionality will break. In this specific case, the user-provided rules will not be invoked. This does leave a residual risk where a user who does not have permission to write Subscriptions can trick a system process into writing the Subscription for them. However, this is a much narrower attack surface and harder to attack.

### DIFF
--- a/hapi-fhir-docs/src/main/java/ca/uhn/hapi/fhir/docs/SubscriptionInterceptors.java
+++ b/hapi-fhir-docs/src/main/java/ca/uhn/hapi/fhir/docs/SubscriptionInterceptors.java
@@ -1,0 +1,50 @@
+/*-
+ * #%L
+ * HAPI FHIR - Docs
+ * %%
+ * Copyright (C) 2014 - 2026 Smile CDR, Inc.
+ * %%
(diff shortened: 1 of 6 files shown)


tadgh:
Formatting check succeeded!

JasonRoberts-smile:
Abandoning this PR. During review, test failure patterns were observed that called into question the initial design. Work will start over on a new design under a separate PR.

PR #8254, opened 2026-08-13
Read authorize patch and conditional create responses

### LINKED ISSUE #8253: `AuthorizationInterceptor` does not read-authorize write-operation response bodies (PATCH, conditional create)
A client with write but no read access could read protected resource content through a write response. `PATCH` responses should be read-authorized, and a conditional create that matches an existing resource should not return the resource body unless read authorized.

codeforgreen:
fixes #8253


tadgh:
Formatting check succeeded!

jamesagnew (approved):

PR #8255, opened 2026-08-14
8247 with postgres db chainned or filters causing performance issues

### LINKED ISSUE #8247: With Postgres db, chainned OR filters causing performance issues
### Description

Search queries generated by the application build their token-matching conditions as a long chain of OR comparisons. When the list is large (113 values in the observed case), PostgreSQL cannot use the index efficiently and the query can hang indefinitely.

YalingPeiS:
Close #8247

### Problem

An unqualified chained token search generated one OR'ed equality per candidate target type. `Provenance.target` is a `Reference(Any)`, so `Provenance?target.identifier=system|value` produced over a hundred `HASH_SYS_AND_VALUE = ?` comparisons. On a large token index PostgreSQL discards the index and reads the whole table, and the query can hang.

### Root cause

Two layers each produced an equality rather than an `IN (...)`:

- `ResourceLinkPredicateBuilder.addPredicateReferenceWithChain` ran a separate `searchForIdsWithAndOr` per candidate target type and OR'ed the results — for a `Reference(Any)` chain, one sub-predicate per resource type in the model.
- Each call reached `TokenPredicateBuilder.createPredicateOrList` with a single code, and the existing `IN (...)` collapse only applied when one call had multiple codes sharing a hash column.

Token hashes already include the resource type, so the per-type fan-out was never necessary.

### Changes

The token predicate builders now accept a list of resource types instead of one, so a chained search resolves every candidate target type in a single predicate and matches the resulting hashes with one `IN (...)` clause per hash column. This path is taken only when all targets index the chained parameter as a plain token; nested chains, `_meta` and non-token parameters, and modifiers such as `:not`/`:text`/`:missing` keep the previous per-type OR behaviour. Search results are unchanged.

Grouping hashes by column also collapses plain token or-lists that share a column, which is why one existing SQL assertion now expects a single `HASH_VALUE` occurrence. New tests in `ChainingR4SearchTest` and `PartitioningSqlR4Test` cover the generated SQL for chained searches, on both unpartitioned and partitioned servers.

### Note
This change applies to all databases (not just Postgres) for performance improvement.


tadgh:
Formatting check succeeded!

volodymyr-korzh (approved):

PR #8257, opened 2026-08-14
8252 - Fix NPE that could occur in findContainedResource

### LINKED ISSUE #8252: Null pointer exception raised during SearchParameter evaluation.
**Describe the bug**
SearchParamExtractorService#findContainedResource throws a NullPointerException when it encounters a candidate resource that has no id, while resolving a reference-type search parameter. This is permitted by the spec. (ex: for .contained resources)

**To Reproduce**
Make sure myStorageSettings.setIndexOnContainedResources is set to true.
Create a Bundle of type document with two entries, each identified only by fullUrl:
  1. A Composition entry, whose subject references the other entry's fullUrl.
  2. A Patient entry, which is the target of that subject reference.

Register a reference-type SearchParameter with base Bundle whose expression resolves to that Composition's subject field, and refresh the search parameter registry so it's active.

Create the document Bundle as a resource on the server.

Expected: the Bundle is created successfully.
Actual: creation fails with HAPI-2223: Cannot invoke "String.equals(Object)" because the return value of "IIdType.getValue()" is null.

**Expected behavior**
findContainedResource should just skip resources with no id, and not throw 

- HAPI FHIR Version
8.13.0-SNAPSHOT

maxwellfrattolin:
Added two tests to catch Null Pointer Exception.
Added fix to findContainedResource to stop NPE.

### DIFF
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
@@ -498,12 +498,12 @@ public class SearchParamExtractorService implements ISearchParamExtractorSvc {
 		}
 	}
 
-	private IBaseResource findContainedResource(Collection<IBaseResource> resources, IBaseReference reference) {
+	IBaseResource findContainedResource(Collection<IBaseResource> resources, IBaseReference reference) {
 		for (IBaseResource resource : resources) {
 			String referenceString = reference.getReferenceElement().getValue();
 			if (referenceString != null && referenceString.length() > 1) {
 				referenceString = referenceString.substring(1);
-				if (referenceString.equals(resource.getIdElement().getValue())) {
+				if (Strings.CS.equals(referenceString, resource.getIdElement().getValue())) {
 					return resource;
 				}
 			}
--- a/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
+++ b/hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
@@ -503,7 +503,7 @@ public class SearchParamExtractorService implements ISearchParamExtractorSvc {
 			String referenceString = reference.getReferenceElement().getValue();
 			if (referenceString != null && referenceString.length() > 1) {
 				referenceString = referenceString.substring(1);
-				if (resource.getIdElement().getValue().equals(referenceString)) {
+				if (referenceString.equals(resource.getIdElement().getValue())) {
 					return resource;
 				}
 			}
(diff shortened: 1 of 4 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
```diff
@@ -503,7 +503,7 @@ private IBaseResource findContainedResource(Collection<IBaseResource> resources,
 			String referenceString = reference.getReferenceElement().getValue();
 			if (referenceString != null && referenceString.length() > 1) {
 				referenceString = referenceString.substring(1);
-				if (resource.getIdElement().getValue().equals(referenceString)) {
+				if (referenceString.equals(resource.getIdElement().getValue())) {
```
tadgh: nit: recommend apache StringUtil.equals() as it will protect from the NPE from either side.

hapi-fhir-jpaserver-searchparam/src/test/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorServiceTest.java
```diff
@@ -126,4 +128,61 @@ void testFindMatchingResourceLink_whenTargetResourceIdMapReturnsNull_shouldNotTh
...
+	private static IBaseResource invokeFindContainedResource(
+			SearchParamExtractorService theSvc, Collection<IBaseResource> theResources, IBaseReference theReference) throws Exception {
+		Method method = SearchParamExtractorService.class.getDeclaredMethod(
+			"findContainedResource",
+			Collection.class,
+			IBaseReference.class
+		);
+		method.setAccessible(true);
+		return unsafeCast(method.invoke(theSvc, theResources, theReference));
+	}
```
tadgh: issue(blocking): do not use reflection in tests. This is brittle, and prone to breaking. Find a different way to exercise this code path.

tadgh (changes requested):
changes requested, and also this is missing a changelog.

hapi-fhir-jpaserver-searchparam/src/main/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorService.java
maxwellfrattolin: Ended up using Strings.CS.equals() because apparently StringUtil.equals is deprecated.

hapi-fhir-jpaserver-searchparam/src/test/java/ca/uhn/fhir/jpa/searchparam/extractor/SearchParamExtractorServiceTest.java
maxwellfrattolin: Done, thank you!

tadgh (approved):

PR #8259, opened 2026-08-14
Migrate org.fusesource.jansi to org.jline:jansi

jdar8:
### For Code Reviewer

Migrates the CLI's console colour library off the deprecated, vulnerable `org.fusesource.jansi` (CVE-2026-8484) to its maintained successor `org.jline:jansi:3.29.0`, managed in the root pom `dependencyManagement`. CDR inherits this managed version, so the version pick here (`3.29.0` rather than the latest `4.3.1`) is deliberately aligned with `org.jline:jline:3.29.0`, which reaches CDR's classpath transitively via Spark Connect and bundles a byte-identical copy of the overlapping `org.jline.*` classes.

One behavioural note: `logback-cli-on.xml`'s `<withJansi>true</withJansi>` is removed. Logback's `wrapWithJansi()` reflectively loads the hardcoded literal class name `org.fusesource.jansi.AnsiConsole`, so once that artifact is gone the setting can only fail with a warning message in the CLI (it degrades gracefully — logback catches the `ClassNotFoundException` and falls back to the plain stream, it does not crash). There is no equivalent hook for the new library. The only observable regression is on legacy Windows consoles that relied on that wrapping for colour translation — called out in the changelog.

### For QA

- [ ] Full reactor build succeeds — package relocation (`org.fusesource.jansi.*` → `org.jline.jansi.*`) is compiler-enforced.
- [ ] `hapi-fhir-cli-api` test suite passes unchanged.
- [ ] CLI commands that print coloured output (e.g. `validate`) still render ANSI colour correctly on a standard terminal.
- [ ] On a legacy Windows console (cmd.exe without native VT support), confirm log output falls back to plain (uncoloured) text rather than raw escape codes — this is an accepted, documented regression, not a defect.
- [ ] `mvn dependency:tree` confirms `org.fusesource.jansi` is absent from `hapi-fhir-cli-api`.

**Companion Smile CDR MR**: https://gitlab.com/simpatico.ai/cdr/-/merge_requests/7443


tadgh:
Formatting check succeeded!

pom.xml
```diff
@@ -1960,10 +1963,16 @@
 				<version>1.1.0</version>
 				<scope>test</scope>
 			</dependency>
+			<!--
+				org.jline:jansi and org.jline:jline ship overlapping copies of org.jline.* classes. Downstream
+				builds that also resolve org.jline:jline must keep the two artifacts at the same version - bump
+				them together via the jline_version property (inherited by child builds, so it is the single
+				bump point). See GL-9148.
```
tadgh: issue: no referrals to private trackers. There is no issue #9148 in github, so you can't use this number.

hapi-fhir-cli/hapi-fhir-cli-api/src/main/resources/logback-cli-on-debug.xml
```diff
@@ -1,7 +1,6 @@
 <configuration>
 
 	<appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-		<useJansi>false</useJansi>
```
tadgh, line 4: question: intended removal?

tadgh (approved):

pom.xml
jdar8: Missed that! Fixed.

hapi-fhir-cli/hapi-fhir-cli-api/src/main/resources/logback-cli-on-debug.xml
jdar8: Yup. For a few reasons: 
- the default is false
- `<useJansi>` wasn't even the right tag (correct tag is `<withJansi>`)

thanks for the review!

TipzCM (approved):

PR #8275, opened 2026-08-20
[8271] Fix ConcurrentModification error that happens randomly during …

### LINKED ISSUE #8271: ConcurrentModificationError in getting batch request
### Issue

When using a batch request, it randomly throws a concurrentmodiciationexception. The error is difficult to reproduce due to the randomness.

It is found that the following  code is not thread safe
```
public Set<String> getOmittedSPNamesForResourceType(String theResourceType) {
       return myOmittedResourceTypeToParameterCodeMap.computeIfAbsent(
             theResourceType.toLowerCase(), (key) -> new HashSet<>());
    }
}
```

YalingPeiS:
Closes #8271

### Issue

`CompartmentSearchParameterModifications` is populated once when an `AuthorizationInterceptor` rule list is built, then read concurrently by every thread evaluating that list. Both getters used `computeIfAbsent` on a plain `HashMap`, so a *read* for an unregistered resource type structurally modified the shared map — causing intermittent `ConcurrentModificationException` during batch requests. It also leaked one empty entry per resource type queried.

### For Code Reviewer

- Getters are now pure reads (`getOrDefault` → `Collections.emptySet()`), returning `unmodifiableSet`.
- Backing maps are `ConcurrentHashMap`; writes are copy-on-write inside `compute(..)`, so readers holding a previously returned `Set` never see it mutated. Chosen over a synchronized `Set` to keep reads lock-free; writes are build-time only and rare.
- Resource-type lookup normalizes with `toLowerCase(Locale.ROOT)` to avoid locale-dependent casing (e.g. Turkish `I`). Search parameter names stay case-sensitive, matching FHIR.

**Behaviour change:** the getters previously returned a live, mutable `Set` that callers could add to. That now throws `UnsupportedOperationException`. No such caller exists in this repo, but it is public API — downstream callers must use `addSPToIncludeInCompartment(..)` / `addSPToOmitFromCompartment(..)`. Noted in the changelog.

### For QA

1. Batch/transaction requests against a rule list using compartment SP modifications no longer intermittently fail with `ConcurrentModificationException` — verify under concurrent load, repeated runs.
2. Compartment authorization decisions unchanged: SPs added via `addSPToIncludeInCompartment(..)` still include, SPs omitted via `addSPToOmitFromCompartment(..)` still exclude.
3. Resource types with no registered modifications still use the stock compartment definition.
4. Resource type names match case-insensitively; search parameter names remain case-sensitive.
5. Repeated authorization evaluations do not grow memory over time.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
@@ -32,7 +32,7 @@ import java.util.concurrent.ConcurrentHashMap;
  * This class is used in RuleBuilder, as a way to allow adding or removing certain Search Parameters
  * to the compartment.
  * For example, if you were to add as additional SPs
- * [device -> ["patient", "subject"]
+ * [device -> ["patient", "subject"]]
  * and apply it to compartment Patient/123, then any device with Patient/123 as its patient would be considered "in"
  * the compartment, despite the fact that device is technically not part of the compartment definition for patient.
  * <p>
@@ -53,7 +53,7 @@ public class CompartmentSearchParameterModifications {
 	 * @param theResourceType the resource type the SPs are based on
 	 * @param theAdditionalSPs the additional SP names
 	 * @param theOmittedSps the omitted SP names
-	 * @return
+	 * @return a new instance with the given SP names registered against the given resource type
 	 */
 	public static CompartmentSearchParameterModifications fromAdditionalAndOmittedSPNames(
 			@Nonnull String theResourceType,
@@ -69,8 +69,9 @@ public class CompartmentSearchParameterModifications {
 		return modifications;
 	}
 
+	@Nonnull
 	public static CompartmentSearchParameterModifications fromAdditionalCompartmentParamNames(
-			String theResourceType, @Nonnull Set<String> theAdditionalCompartmentParamNames) {
+			@Nonnull String theResourceType, @Nonnull Set<String> theAdditionalCompartmentParamNames) {
 		return fromAdditionalAndOmittedSPNames(theResourceType, theAdditionalCompartmentParamNames, Set.of());
 	}
 
@@ -89,7 +90,7 @@ public class CompartmentSearchParameterModifications {
 	 * @param theResourceType the resource type on which the SP exists
 	 * @param theSPName the name of the search parameter
 	 */
-	public void addSPToOmitFromCompartment(String theResourceType, String theSPName) {
+	public void addSPToOmitFromCompartment(@Nonnull String theResourceType, @Nonnull String theSPName) {
 		addSPName(myOmittedResourceTypeToParameterCodeMap, theResourceType, theSPName);
 	}
 
@@ -98,7 +99,7 @@ public class CompartmentSearchParameterModifications {
 	 * @param theResourceType the resource type on which the SP exists
 	 * @param theSPName the name of the search parameter
 	 */
-	public void addSPToIncludeInCompartment(String theResourceType, String theSPName) {
+	public void addSPToIncludeInCompartment(@Nonnull String theResourceType, @Nonnull String theSPName) {
 		addSPName(myAdditionalResourceTypeToParameterCodeMap, theResourceType, theSPName);
 	}
 
@@ -140,7 +141,6 @@ public class CompartmentSearchParameterModifications {
 	@Nonnull
 	private static Set<String> getSPNames(
 			Map<String, Set<String>> theResourceTypeToParameterCodeMap, String theResourceType) {
-		// can be called concurrently by every thread evaluating a shared authorization rule list.
 		String normalizedResourceType = normalizeResourceType(theResourceType);
 		return theResourceTypeToParameterCodeMap.getOrDefault(normalizedResourceType, Collections.emptySet());
 	}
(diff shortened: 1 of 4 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
```diff
@@ -77,30 +90,62 @@ public CompartmentSearchParameterModifications() {
...
+	 *
+	 * @param theResourceType the resource type to look up, matched case-insensitively
+	 * @return an immutable Set, empty if no omitted SPs are registered for this resource type
+	 */
+	@Nonnull
+	public Set<String> getOmittedSPNamesForResourceType(@Nonnull String theResourceType) {
+		return getSPNames(myOmittedResourceTypeToParameterCodeMap, theResourceType);
+	}
+
+	private static void addSPName(
```
codeforgreen, line 130: I would pass in the entire set of SPs to reduce complexity

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
```diff
@@ -67,8 +80,8 @@ public static CompartmentSearchParameterModifications fromAdditionalCompartmentP
 	private final Map<String, Set<String>> myOmittedResourceTypeToParameterCodeMap;
 
 	public CompartmentSearchParameterModifications() {
-		myAdditionalResourceTypeToParameterCodeMap = new HashMap<>();
-		myOmittedResourceTypeToParameterCodeMap = new HashMap<>();
+		myAdditionalResourceTypeToParameterCodeMap = new ConcurrentHashMap<>();
+		myOmittedResourceTypeToParameterCodeMap = new ConcurrentHashMap<>();
```
codeforgreen, line 84: nitpick: consider consolidating into one map to simplify. It would also serve scenario where atomicity of reads and writes would be required. It is not right now, but it would still read better.

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
```diff
@@ -77,30 +90,62 @@ public CompartmentSearchParameterModifications() {
 	 * @param theResourceType the resource type on which the SP exists
 	 * @param theSPName the name of the search parameter
 	 */
-	public void addSPToOmitFromCompartment(String theResourceType, String theSPName) {
-		myOmittedResourceTypeToParameterCodeMap
-				.computeIfAbsent(theResourceType.toLowerCase(), (key) -> new HashSet<>())
-				.add(theSPName);
+	public void addSPToOmitFromCompartment(@Nonnull String theResourceType, @Nonnull String theSPName) {
```
codeforgreen, line 93: nitpick: I am not a fan of using acronyms in the method names for Prod code. So I would replace SP with SearchParameter even if it makes names longer.

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
```diff
@@ -77,30 +90,62 @@ public CompartmentSearchParameterModifications() {
...
+	 * @param theResourceType the resource type to look up, matched case-insensitively
+	 * @return an immutable Set, empty if no omitted SPs are registered for this resource type
+	 */
+	@Nonnull
+	public Set<String> getOmittedSPNamesForResourceType(@Nonnull String theResourceType) {
+		return getSPNames(myOmittedResourceTypeToParameterCodeMap, theResourceType);
+	}
+
+	private static void addSPName(
+			Map<String, Set<String>> theResourceTypeToParameterCodeMap, String theResourceType, String theSPName) {
```
codeforgreen, line 131: nitpick: same thing `theResourceTypeToSearchParameterCodeMap`, `theSearchParameterName'

codeforgreen (approved):

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
YalingPeiS: These method names are existing ones. Since they are public methods that could be called from external sources, I think it is better to not change them unless it is absolutely necessary.

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
YalingPeiS: Interesting idea. These two maps are of <String, Set<String>>, see line 78, 80 in the existing code. Consolidating them into one map will require new map keys to differentiate `additional/omitted`, which would result a map of map. I think it is clearer and easy to maintain by keeping these maps separate as is.

hapi-fhir-base/src/main/java/ca/uhn/fhir/interceptor/auth/CompartmentSearchParameterModifications.java
YalingPeiS: As mentioned above, the two maps are of <String, Set>. Consolidating them into one map will require new map keys to differentiate additional/omitted, which would result a map of map. I think it is clearer and easy to maintain by keeping these maps separate as is.

PR #8278, opened 2026-08-21
Make the HTTP test harness work with both old and new Apache HTTP library versions

codeforgreen:
Included in #8277 

Changes:
* make the public APIs for `FhirHttpRequest` / `FhirHttpResponse` Apache httpclient version agnostic
* integrate into `RestfulServerExtension`
* update some of the existing tests (trivial replacements) to use the harness


Description:
                                                                                                                                                                                                    
`hapi-fhir-test-utilities` used to be tied to one specific version of the Apache HTTP client library. Different tests across the codebase use either the older 4.x version or the newer 5.x version, which made it hard to share test helper code between them.                                                                                                                               
                                                                                                                                                                                                    
This change adds a small interface that hides which HTTP library version is actually being used underneath, so test helper code can work with either one interchangeably:                         
                                                                                                                                                                                                    
  - New interface `IHttpTestTransport` with two implementations — one for the old library version, one for the new one.                                                                             
  - Renamed the main test request/response helper classes `FhirHttpRequest`/`FhirHttpResponse` are renamed to `HttpTestRequest` / `HttpTestResponse` since they don't actually do anything FHIR-specific.                
  - Added a few missing request types to the helper (`OPTIONS`, `HEAD`, raw `POST`/`PUT`).                                                                                                                  
  - Added tests that run the same 12 scenarios against both library versions to prove they behave identically.                                                                                      
  - Updated several existing tests to use the new shared helper instead of calling the old library directly.                                                                                        
                                                                                                                                                                                                    
Why it matters: this removes the version mismatch, where two different test helpers (one on each HTTP library version) needed to be merged into one.


tadgh:
Formatting check succeeded!

codeforgreen:
I need to create another PR as I've switched branches and changed the scope.

PR #8268, opened 2026-08-18
Patient ID partition mode: fix conditional update outside a transaction bundle

tynerguo:
### What was done

Where storage supports an all-partition search, partition selection is deferred to allPartitions and settled once the conditional URL has resolved — a matched update writes to the partition its match already lives in, and doCreateForPostOrPut re-derives and validates the target partition 

Note: Two disabled characterization tests were included to record a separate pre-existing bug: the REST layer's `validateResourceIdAndUrlIdForNonConditionalOperation` discards the user provided body id even for conditional update, making the behaviour violates the [FHIR spec on conditional update](https://www.hl7.org/fhir/http.html#cond-update).

Issue #8266, opened 2026-08-18
In Patient ID Partition Mode, Matched Conditional Update Without Logical ID in Resource Body Throws Hapi-1321


tynerguo:
**Describe the bug**
In Patient ID Partitioning, matched conditional updates outside a transaction bundle without logical ID in resource body throws `Hapi-1321`.

**To Reproduce**

1. setup server with Patient ID Partitioning
2. POST /Patient

```json
{
  "resourceType": "Patient",
  "identifier": [
    {
      "system": "http://patient",
      "value": "1"
    }
  ]
}
```

3. PUT localhost:8000/Patient?identifier=[http://patient|1](http://patient%7C1)

```json
{
  "resourceType": "Patient",
  "identifier": [
    {
      "system": "http://patient",
      "value": "1"
    }
  ],
  "name": [ {
    "family": "Smith",
    "given": [ "John" ]
  } ]
}
```

#### Observed Results

response 405, `ExceptionHandlingInterceptor - Failure during REST processing: ca.uhn.fhir.rest.server.exceptions.MethodNotAllowedException: HAPI-1321: Patient resource IDs must be client-assigned in patient compartment mode, or server id strategy must be UUID`

change Server ID Mode to UUID, re-do step 4, same issue

#### Expected Results

Resource is updated.

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**

- HAPI FHIR Version 8.11.21

**Additional context**

PR #8279, opened 2026-08-22
Add tests for conditional creates and updates in patient id partition mode

nathandoef:


tadgh:
Formatting check succeeded!

epeartree:
Review notes on the new `setIdOnIdlessConditionalPatientUpdate` no-match branch. The approach looks right to me overall — I traced the `RESOURCE_ID_SERVER_ASSIGNED` flag through `isClientAssignedId` and confirmed it correctly suppresses `validateResourceIdCreation`, that the resolved-match-URL keys need no normalization, and that the entry staying conditional means the DAO still reports `SUCCESSFUL_UPDATE_NO_CONDITIONAL_MATCH` natively. Two things I think are worth a second look.

---

### 1. The minted id isn't undone when the transaction rolls back, and the method can't correct it on a retry

[`PatientIdPartitionInterceptor.java:1139`](https://github.com/hapifhir/hapi-fhir/blob/d6db66a1ecd8a21fd88cecb752c2d8b0a69645fe/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java#L1139) writes the freshly minted UUID onto the `Patient` object that lives in the caller's bundle. That object outlives a single attempt at the transaction, because HAPI retries transactions on version conflicts — [`HapiTransactionService.doExecuteInTransaction`](https://github.com/hapifhir/hapi-fhir/blob/d6db66a1ecd8a21fd88cecb752c2d8b0a69645fe/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java#L375) is a retry loop that re-runs the whole callback, so pre-fetch runs again and this hook fires again on the same entry objects.

The framework's mechanism for undoing body mutations before a retry is `TransactionDetails.addRollbackUndoAction`; the rollback path [runs those undo actions in reverse](https://github.com/hapifhir/hapi-fhir/blob/d6db66a1ecd8a21fd88cecb752c2d8b0a69645fe/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/tx/HapiTransactionService.java#L671), and its comment describes exactly this situation ("the resource ID may get modified twice…"). The established pattern is at [`BaseTransactionProcessor.java:1498`](https://github.com/hapifhir/hapi-fhir/blob/d6db66a1ecd8a21fd88cecb752c2d8b0a69645fe/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/dao/BaseTransactionProcessor.java#L1498):

```java
theTransactionDetails.addRollbackUndoAction(() -> res.setId(previousResourceId));
```

The new branch doesn't register one, so the minted UUID survives the rollback into the retry. And on that second pass the method returns immediately at its own entry guard:

```java
if (resource == null
        || !PATIENT_STR.equals(myFhirContext.getResourceType(resource))
        || resource.getIdElement().getIdPart() != null) {   // ← the id we just stamped
    return;
}
```

which means the `matchedId != null` branch just below — the one that would stamp the *correct* id — never gets a chance to run.

Concretely: patient-id partitioning, UUID server id strategy, retry-on-version-conflict enabled (`X-Retry-On-Version-Conflict` or `UserRequestRetryVersionConflictsInterceptor`). Two concurrent transactions each send an id-less, fullUrl-less `PUT Patient?identifier=sys|A`.

1. Both pre-fetch to no match; each mints its own UUID onto its body.
2. The first transaction commits and writes the `HFJ_RES_SEARCH_URL` row.
3. The second collides on that row, throws `ResourceVersionConflictException`, rolls back (nothing clears its minted id) and retries.
4. On the retry, pre-fetch now resolves `Patient?identifier=sys|A` to the winner's real id — but the guard above returns before that id can be stamped.
5. `doUpdate` finds `match.size() == 1`, compares the body id against the matched id, and throws [HAPI-2279](https://github.com/hapifhir/hapi-fhir/blob/d6db66a1ecd8a21fd88cecb752c2d8b0a69645fe/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirResourceDao.java#L2875).

So the retry that should have quietly succeeded as a conditional update against the existing Patient fails instead. Worth noting the sibling placeholder path a couple hundred lines up doesn't have this problem: it keys on the `urn:uuid` fullUrl, which nothing modifies, so re-entering it on the retry hits `resolution.matched()` and overwrites the stale id — it self-heals, whereas this branch locks itself out.

Adding a rollback undo action that clears both the id and the userData flag would fix it, or tracking "minted here" separately so the guard doesn't skip re-evaluation.

### 2. `POST` + `If-None-Exist` is left out, and still fails with HAPI-1321

The new branch is reachable only when `TransactionBundleNormalizer` is a no-op — i.e. when `autoCreatePlaceholderReferenceTargets` is disabled, which is the `JpaStorageSettings` default. In that configuration no fullUrls get assigned, so entries land here rather than in `resolvePlaceholderPatientEntry`.

But the method filters on `PUT` at [line 1122](https://github.com/hapifhir/hapi-fhir/blob/d6db66a1ecd8a21fd88cecb752c2d8b0a69645fe/hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java#L1122), so the conditional-*create* form of the same request — `POST Patient` with `If-None-Exist: identifier=http://ids|A`, no body id, no fullUrl — falls straight through, reaches `identifyForCreate` with a null id part, and throws HAPI-1321. That's the same failure this PR fixes for the PUT form, and it's a slightly odd outcome given the error message itself says "or server id strategy must be UUID" while the UUID strategy is exactly what's configured.

`resolvePlaceholderPatientEntry` handles both verbs (the `"POST".equals(method)` branches around lines 889 and 932); this path handles only one. I realize the conditional-create case may be deliberately out of scope for this fix — if so it might be worth a line in the javadoc saying which forms are covered, since the method name reads more general than the behaviour.

---

Minor, not blocking: the two new tests use `myResourceTableDao.findAll().iterator().next()`, which assumes exactly one row in the table. It matches existing precedent in the file, but `assertResourceIsInPartition(int, IIdType)` already exists in the same class and doesn't depend on the table holding a single resource.

epeartree (approved):
good work Nathan.  

i did claude it and verified the findings.  the first one is legit.  second one is up to u.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
```diff
@@ -1123,6 +1131,13 @@ private void setIdOnIdlessMatchedConditionalPatientUpdate(
 		IIdType matchedId = getPreFetchResolution(url, theTransactionDetails).matchedId();
 		if (matchedId != null) {
 			resource.setId(matchedId.toUnqualifiedVersionless().getValue());
+			return;
```
tynerguo: Suggestion: This case has been fixed at the DAO level in #8266 so there is no need to assign ids this early anymore. Please see [this method](https://github.com/hapifhir/hapi-fhir/pull/8268/changes#diff-522de37509c439428e5f5a5215a5a160b27f63f028e7b51581d2fef2b573b302R2797).

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -1133,6 +1137,45 @@ void testTransaction_CreateObservationWithPatientInlineMatchUrl_multipleMatches_
 		}
 	}
 
+	@Test
+	public void testTransaction_IdlessConditionalUpdatePatient_noMatch_uuidStrategy_createsPatient() {
```
tynerguo, line 1146: Nitpick: Great test addition! I think it would fit nicely into `PatientIdPartitionNormalizerOffScenarios.Accepted` as a test case. But feel free to ignore this as there is no difference on correctness.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
```diff
@@ -1133,6 +1137,45 @@ void testTransaction_CreateObservationWithPatientInlineMatchUrl_multipleMatches_
...
+		runInTransaction(() -> {
+			ResourceTable pt = myResourceTableDao.findAll().iterator().next();
+			assertEquals(
+					PatientIdPartitionInterceptor.defaultPartitionAlgorithm(patientId.getIdPart()),
+					pt.getPartitionId().getPartitionId());
+		});
+	}
+
+	@Test
+	public void testTransaction_IdlessConditionalUpdatePatient_noMatch_defaultStrategy_fails() {
```
tynerguo, line 1186: Nitpick: This test is duplicate with [this test](https://github.com/hapifhir/hapi-fhir/pull/8268/changes#diff-4d62fdcc7ba64e0ed1979f88e9dc0391215a3258290097ab9bd1f888234f08cdR1257) in 8266. But your naming is much clearer! Might be better to remove that one once the merge back is done.

tynerguo (commented):
Hi Nathan, I believe this issue is resolved by #8266 at the DAO level, so there is no need to add a specific fix for the transaction path.

8266 is currently in the [third merge back](https://github.com/hapifhir/hapi-fhir/pull/8290) and hasn't landed in master yet. I think that might be why this was failing. I have verified your tests in the release branch `rel_8_12` and both of them passed.

That being said, I think your tests are great additions as #8266 does not cover the no match case very well. My suggestion is to keep the tests but remove the change in `PatientIdPartitionInterceptor` to lower redundancy.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptor.java
nathandoef: Amazing, thanks @tynerguo! 

I'll wait for the mergeback, revert my `PatientIdPartitionInterceptor` changes, and keep the tests

nathandoef:
> testTransaction_IdlessConditionalUpdatePatient_noMatch_defaultStrategy_fails

Thanks for the review @epeartree!

I've addressed all your comments in c7c7154fadb584f58a880d754e8dc5dacb498259

1. Rollbacks and minted ids - https://github.com/hapifhir/hapi-fhir/issues/8266 resolved this and I've added tests in `FhirResourceDaoR4ConcurrentWriteTest`
2. `POST` + `If-None-Exist` - https://github.com/hapifhir/hapi-fhir/issues/8266 resolved this and I've added tests in `PatientIdPartitionInterceptorR4Test`
3. `myResourceTableDao.findAll().iterator().next()` - strengthened this assertion in `PatientIdPartitionInterceptorR4Test#assertResourceIsInPartition`

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
nathandoef: Thanks - I've kept this outside `PatientIdPartitionNormalizerOffScenarios.Accepted` because this test uses `JpaStorageSettings.ClientIdStrategyEnum.NOT_ALLOWED` and the fixture creates patients with client assigned ids

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/PatientIdPartitionInterceptorR4Test.java
nathandoef: Going to keep this test because it covers the transaction path whereas `testConditionalUpdate_patientIdlessBodyNoMatch_rejectedWith1321` covers the the regular update (non-transaction) path

PR #8284, opened 2026-08-24
R4 validation issue text should not contains code+inferSystem

### LINKED ISSUE #8282: R4 validation issue text should not contains code+inferSystem
**Describe the bug**
  When a Remote Terminology Service is configured (RemoteTerminologyServiceValidationSupport) and ValueSet/$validate-code is called with only code + url (no system), the remote server may reject the request with an HTTP 422 whose message enumerates `code+inferSystem` as a supported parameter combination.

InferSystem is an R5+ $validate-code parameter 
                                                                                                                                                                                                                                  
This happens because RemoteTerminologyServiceValidationSupport#invokeRemoteValidateCode only catches ResourceNotFoundException (404) and InvalidRequestException (400) from the remote call:                                           

**To Reproduce**
1. Configure a Remote Terminology Service against an R4 endpoint:
```
module.persistence.config.remote_term_svc.enabled=true
module.persistence.config.remote_term_svc.base_url=<R4 terminology server>
```
2. `GET [base]/ValueSet/$validate-code?code=male&url=http://hl7.org/fhir/ValueSet/administrative-gender `(no system parameter)
3. If the remote server can't resolve a system on its own, it responds 422 with:
```
{

  "issue": [{
    "severity": "error",
    "code": "invalid",
    "details": { "text": "Unable to find code to validate (looked for coding | codeableConcept | code+system | code+inferSystem in parameters" }
  }]
}
```

**Expected behavior**
As above, except:
- 422 is caught
- there's no `code+inferSystem` in the error message

- HAPI FHIR Version
8.13.1-SNAPSHOT

maxwellfrattolin:
- Added failing test.  
- Added fix (exception is caught and doesn't mention inferSystem anymore in R4-).  
- Added changelog.

### DIFF
--- a/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
+++ b/hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
@@ -16,6 +16,7 @@ import ca.uhn.fhir.rest.gclient.IQuery;
 import ca.uhn.fhir.rest.gclient.StringClientParam;
 import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException;
+import ca.uhn.fhir.rest.server.exceptions.UnprocessableEntityException;
 import ca.uhn.fhir.util.BundleUtil;
 import ca.uhn.fhir.util.Logs;
 import ca.uhn.fhir.util.ParametersUtil;
@@ -687,9 +688,19 @@ public class RemoteTerminologyServiceValidationSupport extends BaseTerminologySe
 					.withParameters(input)
 					.execute();
 			return createCodeValidationResult(output, errorMessageBuilder, theCode, theDisplay);
-		} catch (ResourceNotFoundException | InvalidRequestException ex) {
+		} catch (ResourceNotFoundException | InvalidRequestException | UnprocessableEntityException ex) {
 			ourLog.error(ex.getMessage(), ex);
-			String errorMessage = errorMessageBuilder.buildErrorMessage(ex.getMessage());
+			// A remote server can reference a parameter that doesn't apply to this FHIR version
+			// (e.g. the R5-only code+inferSystem) even when talking to an R4/DSTU3 endpoint that
+			// never received it. In that specific case, describe our own supported parameter
+			// combinations instead of relaying the remote server's message verbatim. Any other
+			// failure (e.g. a genuine "code not found") still surfaces the remote's real message.
+			boolean referencesUnsupportedInferSystem =
+					!shouldInferSystem(getFhirContext()) && containsInferSystemWord(ex.getMessage());
+			String errorMessage = referencesUnsupportedInferSystem
+					? "Unable to find code to validate " + theCode + " (looked for "
+							+ getSupportedValidateCodeParameterCombinations(getFhirContext()) + " in parameters)"
+					: errorMessageBuilder.buildErrorMessage(ex.getMessage());
 			CodeValidationIssueCode issueCode = ex instanceof ResourceNotFoundException
 					? CodeValidationIssueCode.NOT_FOUND
 					: CodeValidationIssueCode.CODE_INVALID;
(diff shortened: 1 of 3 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -687,9 +688,19 @@ protected CodeValidationResult invokeRemoteValidateCode(
 					.withParameters(input)
 					.execute();
 			return createCodeValidationResult(output, errorMessageBuilder, theCode, theDisplay);
-		} catch (ResourceNotFoundException | InvalidRequestException ex) {
+		} catch (ResourceNotFoundException | InvalidRequestException | UnprocessableEntityException ex) {
```
tadgh, line 691: question: why add this exception subtype?

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -687,9 +688,19 @@ protected CodeValidationResult invokeRemoteValidateCode(
...
 			ourLog.error(ex.getMessage(), ex);
-			String errorMessage = errorMessageBuilder.buildErrorMessage(ex.getMessage());
+			// A remote server can reference a parameter that doesn't apply to this FHIR version
```
tadgh, line 693: question: what does this mean? a remote server? do you mean that the remote R4 server will somehow return the inferSystem message itself? I don't understand really any part of this comment paragraph. If the remote server indicates it should support inferSystem, I don't think its our job to "fix" a claim made by the remote terminology server.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
```diff
@@ -687,9 +688,19 @@ protected CodeValidationResult invokeRemoteValidateCode(
...
+			// (e.g. the R5-only code+inferSystem) even when talking to an R4/DSTU3 endpoint that
+			// never received it. In that specific case, describe our own supported parameter
+			// combinations instead of relaying the remote server's message verbatim. Any other
+			// failure (e.g. a genuine "code not found") still surfaces the remote's real message.
+			boolean referencesUnsupportedInferSystem =
+					!shouldInferSystem(getFhirContext()) && containsInferSystemWord(ex.getMessage());
+			String errorMessage = referencesUnsupportedInferSystem
+					? "Unable to find code to validate " + theCode + " (looked for "
+							+ getSupportedValidateCodeParameterCombinations(getFhirContext()) + " in parameters)"
+					: errorMessageBuilder.buildErrorMessage(ex.getMessage());
```
tadgh, line 703: nit: avoid ternary operator when building out strings. Tough to parse. Recommend refactoring into a method, or using a standard if condition.

tadgh (changes requested):
I don't think we should necessarily fix this. This is a problem of the upstream server, and us mucking the response to "fix" it for them is not the correct spot. Submit a bug with the offending upstream validation server.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/RemoteTerminologyServiceValidationSupport.java
maxwellfrattolin: Alright so I've done a bunch of investigating and I'm pretty sure you're right and I'm just going to be scrapping this code and pushing back on the ticket.  That said, still going to document this in case someone looks into it in the future.

Here's what happened:
- QA flagged an error message they were getting in R4 as having `inferSystem` in the response.  (The actual error is "Unable to find code to validate (looking for coding | codableConcept | code+system | code_inferSystem in parameters")  So far so good, inferSystem isn't supposed to be in R4, makes sense that they'd flag it.
- This error **isn't** actually anywhere in HAPI.  I think maybe because it looks like it's built dynamically that I made a mistake here.
- They had `module.persistence.config.remote_term_svc.base_url=https://tx.fhir.org/r4` set, and I've confirmed the actual error is in that remote terminology server.  (100% confirmed, I reproduced the error without even using HAPI, it's not our error).

So basically, I'm pretty sure you're right on all counts, we shouldn't be touching anyone else's error messages.  The only question left is should I catch that UnprocessableEntityException?  I don't know if there's a reason to leave it uncaught, my brain immediately thinks Uncaught Exception = bad, but I wanna make sure before adjusting things.

Issue #4843, opened 2023-05-09
Possible data consistency bug performing quick updates to same resource via dao


G-2-Z:
**Describe the bug**
Quick updates to the same resource via direct dao calls on the jpa server can lead to inconsistent database state.

We have some business logic performing resource transformations directly via the dao objects. In one case we update the same resource very quickly via the same thread and jpa transaction This may result in the second update being only partially applied to the resource database table (see example tables below). 
In this example the server will then answer GET calls with version 24 and will reject PUT with a conflict error, because it tries to insert version 25 again, causing `ERROR: duplicate key value violates unique constraint "idx_resver_id_ver"`. Note the update timestamp is correct.

In this particular case, we were able to get rid of one of the update calls, so it is not a problem for us anymore. Nevertheless, this seems to be an unexpected behavior or even a bug.

HFJ_RES_VER

```
 res_id |       res_updated       | res_ver
--------+-------------------------+---------
   1675 | 2023-05-04 19:21:47.047 |      23
   1675 | 2023-05-04 19:21:51.783 |      24
   1675 | 2023-05-04 19:21:51.795 |      25
```

HFJ_RESOURCE

```
 res_id |       res_updated       | res_ver
--------+-------------------------+---------
   1675 | 2023-05-04 19:21:51.795 |      24
```

**Expected behavior**
HFJ_RESOURCE should have correct res_ver.

**Environment (please complete the following information):**
 - HAPI FHIR Version: 6.4.4
 - OS: Debian
 - DB: PostgreSQL

**Additional context**
The server was set up using a previous version of HAPI FHIR. The official schema migration tool was used during upgrading.

jamesagnew:
I've committed a test which attempts but does not replicate this issue, as in the branch `ja_20230908_4843_resource_update_contention_issue` in test `testHighConcurrencyUpdatesToSameResource()`.

If anyone can provide reliable replication steps against 6.8.x or master, or even better can implement them in this test that would be great.

I vaguely remember seeing this same behaviour once a long time ago but haven't seen it recently.

jamesagnew:
One other thought is that it's possible that this is a Postgres-specific thing (I'm assuming everyone here is using PG?) I only tried to replicate in H2.

XcrigX:
[...]

PR #8288, opened 2026-08-25
GL-8721: Fix re-entrant interceptor update corrupting resource version state

### LINKED ISSUE #8319: Re-entrant interceptor write during a transaction Bundle leaves RES_VER out of sync with resource history
### Describe the bug

When a transaction Bundle updates a resource and a storage interceptor re-entrantly updates that
same resource while processing another entry in the same Bundle, both writes commit but the
resource's current-version pointer is left disagreeing with its history:

```
HFJ_RESOURCE.RES_VER = 2
HFJ_RES_VER rows     = [1, 2, 3]
```

The Bundle response also advertises a version the database does not hold.

`HFJ_RESOURCE.RES_VER` is a Hibernate `@Version` column, set to *loaded value + 1* on every `UPDATE`
Hibernate emits. History rows are written by HAPI and numbered from its own in-memory version field.
The invariant is one emitted `UPDATE` per history row created. A second write to the same resource
within one transaction breaks it: the flag guarding HAPI's in-memory increment is cleared by a
callback annotated `@PostPersist` only, so it never fires after an `UPDATE`. The second write finds
the flag still set, skips its increment, and the history row instead takes its number from the
compensating bump in `ResourceTable.toHistory(boolean)` — while Hibernate emits a single `UPDATE`.

### To Reproduce

1. Create a Patient (v1).
2. Register a server interceptor on `STORAGE_PRESTORAGE_RESOURCE_CREATED` that calls
   `dao.update(...)` on the Patient referenced by an incoming `Flag`.
3. POST a transaction Bundle containing a `PUT` of that Patient plus a conditional `PUT` of a `Flag`
   that references it.
4. Inspect `HFJ_RESOURCE` and `HFJ_RES_VER` for the Patient — the pointer does not name the newest
   history row.

### Expected behavior

Either the version state is repaired so that `HFJ_RESOURCE.RES_VER` matches the newest
`HFJ_RES_VER` row and the response reports the version actually stored, or the transaction is
rejected outright rather than committing an inconsistent state.

### Environment

- HAPI FHIR 8.6.7, confirmed still present on current master.
- Not a regression — the affected version bookkeeping has been in place since 8.0.0.

epeartree:
# GL-8721 / SMILE-12076 — re-entrant interceptor write corrupts resource version state

## What was wrong

A transaction Bundle updates a Patient. An interceptor on `STORAGE_PRESTORAGE_RESOURCE_CREATED`,
firing while another entry in the same Bundle is processed, calls `dao.update()` on that same
Patient. Both writes commit, and the database is left inconsistent:

```
HFJ_RESOURCE.RES_VER = 2
HFJ_RES_VER rows     = [1, 2, 3]
```

The inconsistency happens because the version artefacts are populated by different parties.

`HFJ_RESOURCE.RES_VER` is annotated `@Version`, which hands off the version bump to Hibernate: on
every `UPDATE` it emits, Hibernate writes *loaded value + 1*, discarding whatever HAPI put in
the field.

`HFJ_RES_VER` holds one row per version. HAPI writes those rows itself through `updateEntity` on
every write meant to produce a new version.

On a normal write the two move in lockstep: HAPI increments its in-memory version, stamps a history
row with that number, and sets a flag on the entity — `myVersionUpdatedInCurrentTransaction` —
recording that a bump is pending and guarding against a second increment. At flush, Hibernate emits
an `UPDATE` to `HFJ_RESOURCE.RES_VER`, setting the column to *loaded value + 1*, which matches the
history version. Each `UPDATE` produces one `HFJ_RESOURCE.RES_VER` increment and one row
in `HFJ_RES_VER`.

The invariant is therefore **one emitted `UPDATE` per history row created**, and it breaks when a
second write lands on the same resource before that flush. The first write comes from the
interceptor invocation of `dao.update()`, which mints an entry in `HFJ_RES_VER` and sets
`entity.myVersionUpdatedInCurrentTransaction`. The second write is performed by the Bundle `PUT` on
the same resource. Since the flag is set on the entity, version bumping is skipped at
`ResourceTable.java:457-476` but still happens in `ResourceTable.toHistory()`, which exists to
support delete-then-update in the same transaction. Another entry is added to `HFJ_RES_VER` with the
right version as a total fluke. When the transaction ends, a flush is generated and Hibernate sets
`HFJ_RESOURCE.RES_VER` to *loaded value + 1*. One flush, one increment, with two new entries in
`HFJ_RES_VER`.

## The fix, in two halves

**Version repair.** Three changes, each carrying a distinct part of the solution.

1. **`@PostUpdate` alongside `@PostPersist` on `ResourceTable.postPersistOrUpdate()`.** The flag
   means "a version bump is pending, unflushed", but it was only ever cleared on INSERT, so on an
   UPDATE it latched `true` for the rest of the transaction.

2. **Added `BaseHapiFhirDao.flushPendingResourceVersionUpdate(ResourceTable)`.** A single named
   expression of "make the pending increment physical before doing anything else".

3. **Call `flushPendingResourceVersionUpdate` in `updateEntity` when creating a new history
   entry.** The repair itself. `updateEntity` is the single choke point every physical
   resource-version write passes through. A second write now forces the first write's `UPDATE` out
   before the second history row is minted, so Hibernate's loaded version advances and the second
   write claims an increment of its own, restoring one emitted `UPDATE` per `HFJ_RES_VER` row.

**`If-Match` re-validation at the real write.** In a transaction the precondition is checked while
pass 1 walks the entries against the version the resource holds at that moment. The real write
happens later, in pass 2. A second evaluation is necessary to guard against a storage interceptor
firing on another entry calling `dao.update()` on this same resource and bumping its version. The
pass-1 check has then already passed against a version that is no longer current and the write goes
ahead ignoring the client's `If-Match` precondition.

**Public API change.** Carrying the expected version into the write needs a parameter the old
signature has no slot for, so `IJpaDao.updateInternal(...)` — the ten-argument variant — is now
deprecated and replaced with an overload taking a parameter object.

## Behaviour change reviewers should weigh

A Bundle whose `PUT` carries `If-Match` **and** whose resource is written re-entrantly by an
interceptor now **always** returns 409 and rolls back, where it previously appeared to succeed
while silently discarding the interceptor's version. Affected sites must either stop sending
`If-Match` for those resources, or move the interceptor's write to a `STORAGE_PRECOMMIT_*`
pointcut, which runs after the Bundle's own writes and is unaffected.

A Bundle with no interceptor cannot reach this — an ordinary `PUT` + `PATCH` pair on one resource
still succeeds, and there is a test asserting exactly that.

## Test coverage

`TransactionReentrantUpdateR4Test` — 15 tests in three groups named for what they prove:

- **`CurrentVersionMatchesHistory`** (10) — the version repair, across the reported route, a
  same-verb Bundle, an interceptor firing from inside pass 2, three writes to one resource, a
  no-op update, an interceptor-driven update-then-delete, a no-interceptor control, and the case
  that pins `@PostUpdate` specifically (see below).
- **`IfMatchRevalidatedAtTheWrite`** (4) — the new 409 on all three routes, plus a negative
  control proving a legitimate interceptor-free double write still succeeds.
- **`IfMatchRejectedDuringPassOne`** (1) — pre-existing `Msg.code(989)` behaviour, kept as
  contrast. It is deliberately *not* counted as coverage of the re-check.

Each half is independently pinned: reverting the flush reds the first group, reverting
`@PostUpdate` reds exactly one test, and removing the re-check reds the three refusal cases. All
three were verified by actually reverting and running, not by inspection.

## Regression suites run

`FhirSystemDaoR4Test` (147), `FhirResourceDaoR4Test` (130), `FhirResourceDaoR4QueryCountTest`
(134), `FhirSystemDaoTransactionR5Test` (58), `PatientMergeStandardEndpointR4Test`,
`MergeEdgeCaseR4Test`, `CascadingDeleteInterceptorTest`, `DeleteExpungeDaoTest`,
`FhirResourceDaoCreatePlaceholdersR4Test`, `FhirResourceDaoR4VersionedReferenceTest`,
`BulkImportR4Test`, `SystemProviderR4Test`, `PatchProviderR4Test` — 0 failures.

The delete path and the query-count assertions were the two highest-ranked regression risks: the
new flush reaches `updateEntityForDelete`, and an extra flush can fragment the configured
`hibernate.jdbc.batch_size=50` batches. Neither materialised.

## For QA

**To reproduce the original defect** (on a build without this change): register an interceptor on
`STORAGE_PRESTORAGE_RESOURCE_CREATED` that calls `dao.update()` on a Patient, then POST a
transaction Bundle containing a `PUT` of that Patient plus a conditional `PUT` of a `Flag`
referencing it. Inspect `HFJ_RESOURCE.RES_VER` against `HFJ_RES_VER` for that Patient —
they disagree.

**What to check on this build:**

- The same Bundle **without** `If-Match` now succeeds with `RES_VER` matching the newest history
  row and history gapless.
- The same Bundle **with** `If-Match` now returns 409 and rolls back completely — no `HFJ_RES_VER`
  row survives from the interceptor's write.
- A Bundle writing one resource twice with no interceptor still succeeds.
- Interceptors writing from `STORAGE_PRECOMMIT_*` are unaffected.
- Retrying a 409 from this cause with `X-Retry-On-Version-Conflict` will not clear it — the
  interceptor re-fires deterministically. Retries are bounded, so it terminates.

**Areas worth exercising beyond the automated suites:** `$merge` with delete-source,
delete-expunge, cascading delete, and bulk import — all reach the changed flush path.

Closes #8319


tadgh:
Formatting check succeeded!

jamesagnew (approved):

steveswinsburg:
This looks like it might resolve #4843 also?

PR #8274, opened 2026-08-20
8228: restrict package loader fetches to a configured URL allowlist

TipzCM:
closes #8228

PR #8292, opened 2026-08-25
Fix Upload Terminology intermittent test failures

### LINKED ISSUE #8291: Solve intermittent test failures related to Terminology Upload
`TerminologySvcImplCurrentVersionR4Test` fails intermittently. Terminology upload leaves async work behind: deferred concept/ValueSet storage, deletion of the `CodeSystem` version it replaces, and `ValueSet` pre-expansions. The tests asserted against that state before it had settled, relying on the 5-second deferred-storage scheduler and scattered per-test `awaitNoJobsRunning()` calls to get there first.

codeforgreen:
Fixes #8291

**Root Cause**

The new terminology upload uses a `CodeSystem` staging mechanism which uses a placeholder `TermCodeSystemVersion` which is replaced with the actual `TermCodeSystemVersion` once all concepts and all related information to a CodeSystem are persisted. This adds records to the delete deferred queue which queues a delete job for the obsolete `TermCodeSystemVersion`. The terminology can also contain ValueSets which can trigger `ValueSet` async pre-expansion jobs. 

Upload terminology tests do not wait for the async processes to finish before doing the assertions. This results in intermittent test failures.

**Changes**

- `TerminologyTestHelper` successful imports now settle their own async work via a new private `awaitImportDeferredTerminologyWork()`: re-enable deferred processing, `saveAllDeferred()`, then await the version-delete, CodeSystem-delete, and pre-expansion jobs. Callers no longer need to know what an import leaves behind.                                                                                                                                       
                                                                                                                                                                                                              
- `TerminologySvcImplCurrentVersionR4Test` — fixed the self-comparing assertion in matchUnqualifiedIds and gave it the ID prefix as a parameter so both call sites assert against the right set.        

- replaced `validateExpandedTermConceptsForVersion` with `validateTermConcepts`, which keys concepts by `CodeSystem` version and asserts exactly one per version, so a leftover `DELETED_` version shows up as a readable map diff instead of an exception. Removed the `@AfterEach` job wait, dead helpers, and unused fields now that the helper handles the waiting.                                                                      
                                                                                                                                                                                                               
- updated the version-count assertions across the terminology upload tests e.g. `TerminologyLoaderSvcLoincJpaTest` such that each upload now leaves exactly one `TermCodeSystemVersion`. That is the correct final number once the replaced one is deleted, so counts drop from 2 per upload to 1. The reasoning is     captured once in `TermTestUtil.MSG_ONE_CODE_SYSTEM_VERSION_PER_UPLOAD` and used as the assertion message. 
- `TermCodeSystemDeleteJobTest.runDeleteJobMultipleVersions` now genuinely loads two versions instead of relying on the placeholder version to make up the count.     

**Note**

Another cause for intermittence was related to the scenario with multiple upload terminology operations executed sequentially. The tests relied on the fact that upload order is the same as persistence order for `TermCodeSystemVersion` records which is not necessarily the case. The queries in the test assertions were fixed already as part of #8274.
  
**Known issue**
  
Awaiting the pre-expansion jobs exposed a pre-existing production race condition in `TermDeferredStorageSvcImpl.saveAllDeferred`. `ValueSet` pre-expansion can start while the replaced `CodeSystem` version is still being deleted, and an expansion that runs in that window falls back to an unvalidated in-memory path. Filed as #8321. Worked included here is a second `saveAllDeferred()` call after the delete jobs complete, commented and linked to the issue.

### DIFF
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologyLoaderSvcLoincJpaTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologyLoaderSvcLoincJpaTest.java
@@ -65,7 +65,8 @@ public class TerminologyLoaderSvcLoincJpaTest extends BaseJpaR4Test {
 			assertEquals(1, myTermCodeSystemDao.count());
 			assertEquals(82, myTermConceptDao.count());
 			assertEquals(8, myTermConceptParentChildLinkDao.count());
-			assertEquals(2, myTermCodeSystemVersionDao.count());
+			assertEquals(1, myTermCodeSystemVersionDao.count(),
+				"the placeholder version superseded during activation must have been deleted");
 			assertEquals(10, myTermValueSetDao.count());
 			assertEquals(5, myTermConceptMapDao.count());
 			assertEquals(16, myResourceTableDao.count());
@@ -109,7 +110,7 @@ public class TerminologyLoaderSvcLoincJpaTest extends BaseJpaR4Test {
 			assertEquals(1, myTermCodeSystemDao.count());
 			assertEquals(82 * 2, myTermConceptDao.count());
 			assertEquals(8 * 2, myTermConceptParentChildLinkDao.count());
-			assertEquals(2 * 2, myTermCodeSystemVersionDao.count());
+			assertEquals(2, myTermCodeSystemVersionDao.count());
 			assertEquals(10 * 2, myTermValueSetDao.count());
 			assertEquals(5 * 2, myTermConceptMapDao.count());
 			assertEquals(16 * 2, myResourceTableDao.count());
@@ -135,7 +136,7 @@ public class TerminologyLoaderSvcLoincJpaTest extends BaseJpaR4Test {
 			assertEquals(1, myTermCodeSystemDao.count());
 			assertEquals(82 * 3, myTermConceptDao.count());
 			assertEquals(8 * 3, myTermConceptParentChildLinkDao.count());
-			assertEquals(2 * 3, myTermCodeSystemVersionDao.count());
+			assertEquals(3, myTermCodeSystemVersionDao.count());
 			assertEquals(10 * 3, myTermValueSetDao.count());
 			assertEquals(5 * 3, myTermConceptMapDao.count());
 			assertEquals(16 * 3, myResourceTableDao.count());
@@ -191,7 +192,7 @@ public class TerminologyLoaderSvcLoincJpaTest extends BaseJpaR4Test {
 
 		runInTransaction(() -> {
 			assertEquals(1, myTermCodeSystemDao.count());
-			assertEquals(4, myTermCodeSystemVersionDao.count());
+			assertEquals(2, myTermCodeSystemVersionDao.count());
 			TermCodeSystem myTermCodeSystem = myTermCodeSystemDao.findByCodeSystemUri("http://loinc.org");
 
 			TermCodeSystemVersion myTermCodeSystemVersion_new =
--- a/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologyLoaderSvcLoincJpaTest.java
+++ b/hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologyLoaderSvcLoincJpaTest.java
@@ -57,7 +57,6 @@ public class TerminologyLoaderSvcLoincJpaTest extends BaseJpaR4Test {
 		files = new ZipCollectionBuilder(true);
 		TermTestUtil.addLoincMandatoryFilesWithPropertiesFileToZip(files, "v267_loincupload.properties");
 		String instanceId = myTerminologyTestHelper.startImportLoincJobAndWaitForCompletion("2.66", files);
-		myBatch2JobHelper.awaitNoJobsRunning();
 
 		logAllValueSets();
 
@@ -65,8 +64,7 @@ public class TerminologyLoaderSvcLoincJpaTest extends BaseJpaR4Test {
 			assertEquals(1, myTermCodeSystemDao.count());
 			assertEquals(82, myTermConceptDao.count());
 			assertEquals(8, myTermConceptParentChildLinkDao.count());
-			assertEquals(1, myTermCodeSystemVersionDao.count(),
-				"the placeholder version superseded during activation must have been deleted");
+			assertEquals(1, myTermCodeSystemVersionDao.count(), TermTestUtil.MSG_ONE_CODE_SYSTEM_VERSION_PER_UPLOAD);
 			assertEquals(10, myTermValueSetDao.count());
 			assertEquals(5, myTermConceptMapDao.count());
 			assertEquals(16, myResourceTableDao.count());
@@ -101,7 +99,6 @@ public class TerminologyLoaderSvcLoincJpaTest extends BaseJpaR4Test {
 		files = new ZipCollectionBuilder(true);
 		TermTestUtil.addLoincMandatoryFilesWithPropertiesFileToZip(files, "v267_loincupload.properties");
 		myTerminologyTestHelper.startImportLoincJobAndWaitForCompletion("2.67", files);
-		myBatch2JobHelper.awaitNoJobsRunning();
 
 		logAllCodeSystemsAndVersionsCodeSystemsAndVersions();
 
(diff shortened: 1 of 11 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologySvcImplCurrentVersionR4Test.java
```diff
@@ -508,34 +493,16 @@ private void validateValueExpandLoincAllVsForVersion(String theVersion) {
...
-			@SuppressWarnings("unchecked")
-			List<TermConcept> termConceptWithVerList = (List<TermConcept>) myEntityManager.createQuery(
-				"from TermConcept where myCode = '" + VS_VERSIONED_ON_UPLOAD_FIRST_CODE + "' order by myId.myId").getResultList();
-			assertEquals(theExpectedVersions.size(), termConceptWithVerList.size());
-			for (int i = 0; i < theExpectedVersions.size(); i++) {
-				assertEquals( prefixWithVersion(theExpectedVersions.get(i), VS_VERSIONED_ON_UPLOAD_FIRST_DISPLAY),
-					termConceptWithVerList.get(i).getDisplay(), "TermCode with id: " + i + " display");
-			}
-		});
+		theExpectedVersions.forEach(this::validateTermConceptForVersion);
```
jmarchionatto: 1. This validation is now duplicate. Every caller already ran validateOperations() with same version list, which reaches this method.  

2. Assertion was removed for the number of TermConcept rows across all code system versions against expected versions, which was not the source of flakiness. A a surplus row belonging to a version that should not exist will go undetected (the leak a DELETED_* version whose concepts were never cleaned up would produce).

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologySvcImplCurrentVersionR4Test.java
```diff
@@ -216,28 +194,35 @@ private void validateValueExpand(String currentVersion, Collection<String> theAl
...
+	 * displays match that version. Concepts are matched to a version via
+	 * {@code tcsv.myCodeSystemVersionId} rather than by insertion order (e.g. {@code myId.myId}),
+	 * since concept persistence goes through an asynchronous deferred-storage queue and is not
+	 * guaranteed to happen in upload order.
+	 */
+	private void validateTermConceptForVersion(String theVersion) {
 		runInTransaction(()->{
 			TermConcept termConceptNoVer = (TermConcept) myEntityManager.createQuery(
 				"select tc from TermConcept tc join fetch tc.myCodeSystem tcsv where tc.myCode = '" +
 					VS_NO_VERSIONED_ON_UPLOAD_FIRST_CODE + "' and tcsv.myCodeSystemVersionId = '" + theVersion + "'").getSingleResult();
```
jmarchionatto: `getSingleResult()` won't help diagnose cause of potential intermittency. Would be better to use getResultList() and assert on it.

jmarchionatto (changes requested):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologySvcImplCurrentVersionR4Test.java
codeforgreen: Good points. Thanks for seeing this. The scope for this fix was supposed to be only the query itself which contained the wrong assumption - the order by clause since the order of upload is not the order of insertion in the database necessarily because expansion is async.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologySvcImplCurrentVersionR4Test.java
codeforgreen: Good point. I will address this.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologyLoaderSvcLoincJpaTest.java
```diff
@@ -224,17 +219,7 @@ public void testValueSetExpansion() throws IOException {
...
-		assertThat(expansionMessage).contains("has not yet been pre-expanded");
-
-		// Now run the pre-expansion
-
-		logAllValueSets();
-		myBatch2JobHelper.awaitNoJobsRunning();
-
-		outcome = myValueSetDao.expand(new IdType("ValueSet/LL1001-8-2.67"), options, newSrd());
-		expansionMessage = outcome.getMeta().getExtensionString(EXT_VALUESET_EXPANSION_MESSAGE);
 		assertThat(expansionMessage).contains("using an expansion that was pre-calculated");
```
jmarchionatto, line 237: suggestion: assert stored state also to verify that the pre-expansion actually ran: 

```
runInTransaction(() -> assertEquals(
    TermValueSetPreExpansionStatusEnum.EXPANDED,
    myTermValueSetDao.findByResourcePid(...).getExpansionStatus()));
```

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/term/TerminologyTestHelper.java
```diff
@@ -89,21 +92,45 @@ public class TerminologyTestHelper {
...
+	 * delete jobs first would find no instances and return immediately. Deferred processing is re-enabled
+	 * before draining because {@link ITermDeferredStorageSvc#isStorageQueueEmpty(boolean)} reports non-empty
+	 * while processing is paused, which would leave {@link ITermDeferredStorageSvc#saveAllDeferred()} looping
+	 * forever.
+	 * <p>
+	 * Only successful imports call this. A failed import leaves whatever it queued behind for the per-test
+	 * terminology cleanup to discard.
+	 */
+	private void awaitDeferredTerminologyWork() {
+		myTermDeferredStorageSvc.setProcessDeferred(true);
```
jmarchionatto: suggestion: add javadoc semtence indicating this is left enabled after this call

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/term/TermTestUtil.java
```diff
@@ -58,6 +58,14 @@ public final class TermTestUtil {
 	public static final String URL_MY_CODE_SYSTEM = "http://example.com/my_code_system";
 	public static final String URL_MY_VALUE_SET = "http://example.com/my_value_set";
 
+	/**
+	 * Message for assertions on the surviving {@code TermCodeSystemVersion} count. An upload creates a
+	 * placeholder version and then supersedes it when the real version is activated, so each upload leaves
+	 * exactly one version behind once the batch2 deletion of the superseded one has run.
+	 */
+	public static final String MSG_ONE_CODE_SYSTEM_VERSION_PER_UPLOAD =
```
jmarchionatto: It is used in assertions expecting more than one version. Make it a formatter accepting the count or reword using something like: "...one CodeSystem version per upload..."

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/job/TermCodeSystemVersionDeleteJobTest.java
```diff
@@ -89,7 +90,7 @@ public void runDeleteJobDeleteOneVersion() throws Exception {
 		runInTransaction(() -> {
 			assertEquals(1, myTermCodeSystemDao.count());
 			assertNotNull(myTermCodeSystemDao.findByCodeSystemUri("http://loinc.org"));
-			assertEquals(3, myTermCodeSystemVersionDao.count());
+			assertEquals(1, myTermCodeSystemVersionDao.count());
```
jmarchionatto, line 93: suggestion: this is the post-delete counterpart of the pre-delete assertion at :74, which did get a message. Add one describing the post-condition ("one of two versions deleted, the other survives") so a failure distinguishes "delete did nothing" from "delete took too much".

jmarchionatto (approved):
Looks good. Approved with a few suggestions

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/term/TermTestUtil.java
codeforgreen: I think the failure message would already suggest the actual number it expects if it's multiple versions. No need to put the assertion in the message. I will simplify the message and leave as is.

hapi-fhir-jpaserver-test-utilities/src/main/java/ca/uhn/fhir/jpa/term/TerminologyTestHelper.java
codeforgreen: Looks like `myProcessDeferred` is always true. I removed this line.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/term/TerminologyLoaderSvcLoincJpaTest.java
codeforgreen: added

PR #8302, opened 2026-08-26
Add a transport-agnostic HTTP test client and consolidate test HTTP construction

codeforgreen:
Part of #8277.

#### Why

* hapi-fhir has about 1,300 raw Apache HttpClient call sites across about 250 test files which all contain the same coding pattern: build the request, execute, close the client and assert steps. There is no shared test framework integrated into the test harnesses we have that makes it easier to run http requests and assert based on the response. 
* most of the tests are directly dependent on imports from Apache HttpClient 4.x. If we wanted to migrate to migrate to 5.x then all test files would need to be touched since there are major API changes between 4.x and 5.x. 

#### What this does

* Introduced `HttpTestRequest`/`HttpTestResponse` which are Apache HttpClient version agnostic. The version is chosen via `IHttpTestTransport`. Tests can now use these to perform test steps and assertions against an http client. To do this they go through `TestHttpClientFactory` to client an http client. Existing test harnesses such as `RestfulServerExtension` have been integrated with `TestHttpClientFactory`.
* Applied a uniform configuration of connection pool size of 99 and timeout of 30s which was present for majority of tests prior to these changes.

!cdr-branch=9263-consolidate-http-test


tadgh:
Formatting check succeeded!

hapi-deployable-pom/pom.xml
```diff
@@ -96,6 +96,7 @@
 						</dependency>
 					</ignoredDependencies>
 					<ignoredResourcePatterns>
+						<ignoredResourcePattern>org/publicsuffix/list/effective_tld_names\.dat$</ignoredResourcePattern>
```
codeforgreen, line 99: This is to solve this error that occurs in the maven build once apache httpclient5 is added to hapi-fhir alongside docker library:
```
[INFO] --- duplicate-finder:2.0.1:check (default) @ hapi-fhir-test-utilities ---
[INFO] Checking compile classpath
[WARNING] Found duplicate and different resources in [com.github.docker-java:docker-java-transport-zerodep:3.7.0, org.apache.httpcomponents.client5:httpclient5:5.4.3]:
[WARNING]   org/publicsuffix/list/effective_tld_names.dat
[WARNING] Found duplicate classes/resources in compile classpath.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  8.051 s
[INFO] Finished at: 2026-09-03T11:21:55-07:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.basepom.maven:duplicate-finder-maven-plugin:2.0.1:check (default) on project hapi-fhir-test-utilities: Found duplicate classes/resources! -> [Help 1]
[ERROR] 
```

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/HttpTestRequest.java
```diff
@@ -0,0 +1,361 @@
...
+		return method("POST", encodeResource(theBody), Constants.CT_FHIR_JSON_NEW);
+	}
+
+	/**
+	 * POSTs the body as the given MIME type, with a UTF-8 charset.
+	 *
+	 * @param theContentType the MIME type, e.g. {@literal "text/plain"}
+	 */
+	public HttpTestResponse post(String theBody, String theContentType) {
+		return method("POST", theBody.getBytes(StandardCharsets.UTF_8), withUtf8Charset(theContentType));
```
jmarchionatto, line 226: (also :263 put, :288 patch, helper at :349)

The String-bodied overloads always encode with theBody.getBytes(StandardCharsets.UTF_8), but withUtf8Charset returns the MIME type unchanged when it already contains charset.

So post("héllo", "text/plain; charset=ISO-8859-1") puts UTF-8 bytes on the wire labelled ISO-8859-1, and the server decodes mojibake. There is no warning and no assertion catches it: the one test that exercises this path, HttpTestRequestTest.java:90 (post_contentTypeAlreadyHasCharset_isNotAppendedTwice), sends the pure-ASCII body "hello" and asserts only on the echoed rawContentType header, never on a round-trip of the body.

This is the kind of bug the framework exists to prevent — the whole point of withUtf8Charset is charset correctness, and it is exactly wrong in the one case where the caller cared enough to name a charset.

Fix: encode with the charset actually named in the content type (ContentType.parse(theContentType).getCharset(), defaulting to UTF-8), or — simpler and matching the class's stated contract that these overloads "POST the body ... with a UTF-8 charset" — reject a content type naming a charset other than UTF-8 with an IllegalArgumentException that points the caller at post(byte[], String). Either way, add a test with a non-ASCII body that asserts the server read the original string back.

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/server/BaseJettyServerExtension.java
```diff
@@ -254,10 +258,9 @@ public void onClosed(Connection connection) {
 
 		myPort = JettyUtil.getPortForStartedServer(myServer);
 		ourLog.info("Server has started on port {}", myPort);
-		PoolingHttpClientConnectionManager connectionManager = new PoolingHttpClientConnectionManager(5000, TimeUnit.MILLISECONDS);
-		HttpClientBuilder builder = HttpClientBuilder.create();
-		builder.setConnectionManager(connectionManager);
-		myHttpClient = builder.build();
+		// No read timeout: before this client moved onto TestHttpClientFactory it inherited Apache's
+		// unbounded default, and tests driving slow endpoints through this extension rely on that.
+		myHttpClient = TestHttpClientFactory.create(true, TestHttpClientFactory.NO_SOCKET_TIMEOUT);
```
jmarchionatto, line 263: Roughly 200 tests moved from HttpClientExtension — which builds its client through TestHttpClientFactory.create(boolean) and therefore gets the 30-second DEFAULT_SOCKET_TIMEOUT_MILLIS — onto RestfulServerExtension.fhirRequest() / HttpServletExtension.request(), which use this unbounded client. A server that stops responding now hangs the surefire fork until CI kills it, instead of failing in 30 seconds with a SocketTimeoutException naming the test.

The in-line comment justifies this as preserving what this extension's own client previously inherited, which is accurate for the extension's pre-existing callers — but it silently applies to every newly migrated caller too, and I found no test in the branch that actually needs an unbounded read.

Fix: give this client a bounded-but-generous timeout (say 120s, still far under any CI fork timeout) and have any genuinely slow-endpoint test opt out explicitly, rather than making unbounded the default for every consumer of these two extensions.

hapi-fhir-test-utilities/src/test/java/ca/uhn/fhir/test/utilities/HttpTestTransportContractTest.java
```diff
@@ -0,0 +1,413 @@
...
+	void put_withStringBody_sendsBodyAndContentType(String theTransport) {
+		HttpTestResponse response = request(theTransport, "/foo").put("hello", "text/plain");
+
+		assertThat(response.getBody()).contains("method=PUT").contains("body=hello");
+		assertThat(response.getBody()).contains("contentType=text/plain");
+	}
+
+	@ParameterizedTest
+	@ValueSource(strings = {APACHE_4, APACHE_5})
+	void put_withByteArrayBody_sendsBytesUnaltered(String theTransport) {
```
jmarchionatto, line 288: The name claims the bytes survive; the assertions check the verb and the content type and nothing else. Its POST sibling at :107 does assert on the body, but sends "bytes-payload" — pure ASCII, which round-trips through any encoding — so neither test would fail if a transport mangled a non-ASCII byte on the request side. Request-side binary fidelity is currently unverified, and it is precisely what the byte[] overloads exist for.

Fix: have EchoServlet render the raw request bytes as hex or base64 (read from getInputStream() before any parameter access) and assert the echoed encoding equals the encoding of PNG_MAGIC.

hapi-fhir-test-utilities/src/test/java/ca/uhn/fhir/test/utilities/HttpTestTransportContractTest.java
```diff
@@ -0,0 +1,413 @@
...
+		assertThat(body).contains("params=scope=openid,patient/*.read");
+	}
+
+	@ParameterizedTest
+	@ValueSource(strings = {APACHE_4, APACHE_5})
+	void withFormParam_nullValue_sendsTheNameWithNoValue(String theTransport) {
+		String body =
+				request(theTransport, "/foo").withFormParam("client_id", null).postForm().getBody();
+
+		assertThat(body).contains("params=client_id=");
```
jmarchionatto, line 171: "params=client_id=" is a prefix of "params=client_id=anything", so the test passes whether or not a value was sent — it asserts nothing about the behaviour it names.

Fix: pin the end of the segment. EchoServlet writes \nbody= immediately after the parameter list, so assertThat(body).contains("params=client_id=\nbody=") is exact for the single-parameter case.

hapi-fhir-test-utilities/src/test/java/ca/uhn/fhir/test/utilities/HttpTestRequestTest.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Covers the parts of {@link HttpTestRequest} that depend on a {@link FhirContext} or on
+ * FHIR-specific headers. Behaviour that is purely about issuing a request — verbs, headers,
+ * bodies, status and response parsing — lives in {@link HttpTestTransportContractTest}, which
+ * exercises it against every transport rather than just one.
+ */
+// Created by claude-sonnet-5
+class HttpTestRequestTest {
+
+	private static final FhirContext ourFhirContext = mock(FhirContext.class);
```
jmarchionatto, line 27: It is re-stubbed inside individual test methods (:38-40 and :60-62 both stub newJsonParser() on it) and never reset between them. It works today only because each test re-stubs before use; add a test that reads the context without stubbing it and it inherits whatever the previous test left behind, with the failure depending on method order.

Fix: make it a non-static field created in @BeforeEach, or drive it with @ExtendWith(MockitoExtension.class) and @Mock so Mockito resets it per test.

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/TestHttpClientFactory.java
```diff
@@ -0,0 +1,107 @@
...
+				new PoolingHttpClientConnectionManager(CONNECTION_TIME_TO_LIVE_MILLIS, TimeUnit.MILLISECONDS);
+		connectionManager.setMaxTotal(MAX_CONNECTIONS);
+		connectionManager.setDefaultMaxPerRoute(MAX_CONNECTIONS);
+		connectionManager.setDefaultSocketConfig(SocketConfig.copy(SocketConfig.DEFAULT)
+				.setSoTimeout(theSocketTimeoutMillis)
+				.build());
+
+		HttpClientBuilder builder = HttpClientBuilder.create()
+				.setConnectionManager(connectionManager)
+				.setMaxConnPerRoute(MAX_CONNECTIONS);
```
jmarchionatto, line 101: HttpClientBuilder.build() only applies maxConnPerRoute on the branch where it constructs its own PoolingHttpClientConnectionManager (confirmed in the httpclient 4.5.14 bytecode: the setDefaultMaxPerRoute(maxConnPerRoute) call sits inside the connManagerCopy == null branch). Because setConnectionManager(...) was called, this line does nothing. The value is already applied correctly two lines earlier via connectionManager.setDefaultMaxPerRoute(MAX_CONNECTIONS).

It was carried over verbatim from the old HttpClientExtension.initialize(), where it was equally dead. Fix: delete the .setMaxConnPerRoute(...) call.

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/BaseRestServerHelper.java
```diff
@@ -151,7 +174,16 @@ public String getSecureBase() {
 	}
 
 	public void stop() throws Exception {
-		JettyUtil.closeServer(myListenerServer);
+		try {
+			if (myHttpClient != null) {
+				myHttpClient.close();
+				myHttpClient = null;
```
jmarchionatto, line 180: myHttpClient = null only runs when close() returns normally. If it throws, the field keeps a half-closed client and the next fhirRequest(...) — the lazy initializer only allocates when the field is null — hands that client back out. BaseJettyServerExtension.stopServer() gets this right in the same branch by nulling inside its finally.

Fix: null the field before closing, or move both statements into the finally.

hapi-fhir-test-utilities/src/test/java/ca/uhn/fhir/test/utilities/HttpTestTransportContractTest.java
```diff
@@ -0,0 +1,413 @@
...
+
+		assertThat(response.getBodyBytes()).containsExactly(EchoServlet.PNG_MAGIC);
+	}
+
+	@ParameterizedTest
+	@ValueSource(strings = {APACHE_4, APACHE_5})
+	void getContentType_responseHasCharsetParameter_stripsIt(String theTransport) {
+		HttpTestResponse response = request(theTransport, "/foo").get();
+
+		assertThat(response.getContentType()).startsWith("text/plain");
```
jmarchionatto, line 245: Subsumed by next

hapi-deployable-pom/pom.xml
jmarchionatto, line 99: suggestion: add a comment identifying the reason of the need so it can be removed once that reason is gone

jmarchionatto (approved):
Approved with some findings

hapi-deployable-pom/pom.xml
codeforgreen: Thanks — recorded that reason in the pom next to the exclusion (see Juan's note below), so whoever finds it later knows it can go once a module no longer needs docker-java and httpclient5 on one classpath.

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/HttpTestRequest.java
codeforgreen: Good catch. Fixed. The `String` overloads now encode with the charset the content type actually names.

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/server/BaseJettyServerExtension.java
codeforgreen: Fixed. The client is bounded by default now to 30s as well. 30s is what `HttpClientExtension` already uses so it is consistent with the rest of the bounded test setup we have in hapi-fhir. For the record, there are only 2 test classes which used to have timeouts unbounded and now have them bounded to 30s `BulkExportWithPatientIdPartitioningTest` and `PatientIdPartitionInterceptorR4Test`.

hapi-fhir-test-utilities/src/test/java/ca/uhn/fhir/test/utilities/HttpTestTransportContractTest.java
codeforgreen: Fixed as suggested. `EchoServlet` now reads the raw request bytes from `getInputStream()` before any parameter access and echoes them as `bodyHex=`. The `byte[]` tests assert that hex equals the hex of `PNG_MAGIC`, so a transport mangling a non-ASCII request byte now fails.

hapi-fhir-test-utilities/src/test/java/ca/uhn/fhir/test/utilities/HttpTestTransportContractTest.java
codeforgreen: Fixed. The test asserts `"\nparams=client_id=\nbody="` now, which pins the end of the value as you suggested. Left a comment in the code to say why, so nobody trims it back to the prefix.

hapi-fhir-test-utilities/src/test/java/ca/uhn/fhir/test/utilities/HttpTestRequestTest.java
codeforgreen: Fixed. `ourFhirContext` is an instance field now, so each test gets a fresh mock and stubbing can't leak between them. Added a note on the field explaining both that and why it's a mock rather than a real `FhirContext`.

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/TestHttpClientFactory.java
codeforgreen: You're right, it's dead once `setConnectionManager(...)` has been called. I deleted it since the value is already applied via `connectionManager.setDefaultMaxPerRoute(MAX_CONNECTIONS)` two lines up.

hapi-fhir-test-utilities/src/main/java/ca/uhn/fhir/test/utilities/BaseRestServerHelper.java
codeforgreen: Fixed. `myHttpClient = null` is inside the `finally` now, so a `close()` that throws can't leave a half-closed client for the lazy initializer to hand back out. Matches what `BaseJettyServerExtension.stopServer()` already did.

hapi-fhir-test-utilities/src/test/java/ca/uhn/fhir/test/utilities/HttpTestTransportContractTest.java
codeforgreen: Covered by the binary-fidelity thread below.

hapi-deployable-pom/pom.xml
codeforgreen: Added. The exclusion now says docker-java-transport-zerodep and httpclient5 each ship a copy of the public suffix list and `hapi-fhir-test-utilities` is on the classpath of both, plus a note to remove it once no module needs both.

fil512 (approved):
nice to have this cleaned up

tadgh (approved):

PR #8309, opened 2026-08-27
Adding additional SQL statement to tag query that loads tags up front

### LINKED ISSUE #8308: `_tag`/`_security`/`_profile` searches join HFJ_TAG_DEF, causing query-planner cardinality underestimates and slow queries
### Version / Environment
- HAPI FHIR: 8.13.x (JPA server)
- Database: PostgreSQL (observed); the generated SQL shape is DB-agnostic
- FHIR version: R4

### Describe the bug
A search filtered by `_tag` (or `_security` / `_profile`) generates SQL that reaches the tag through a join chain:

```
HFJ_RESOURCE → HFJ_RES_TAG → HFJ_TAG_DEF
```

with the tag system/code/type applied as predicates on `HFJ_TAG_DEF` inside the same statement. Because the selective `TAG_ID` is only reachable via the join (never as a literal/bound value on `HFJ_RES_TAG`), the query planner cannot consult per-value (MCV) statistics for it and falls back to an average rows-per-tag estimate.

For a high-cardinality tag (e.g. an MDM `GOLDEN_RECORD` status tag matching millions of rows), the estimate is off by orders of magnitude (observed: **estimated ~246 vs actual ~2.16M**). The planner then chooses a nested-loop plan that probes the companion index millions of times, and the query takes tens of seconds.

### Generated SQL (simplified)
```sql
SELECT t1.PARTITION_ID, t1.RES_ID
FROM hfj_resource t1
  INNER JOIN hfj_spidx_token t0 ON t1.RES_ID = t0.RES_ID
  INNER JOIN hfj_res_tag     t2 ON t1.RES_ID = t2.RES_ID
  INNER JOIN hfj_tag_def     t3 ON t2.TAG_ID = t3.TAG_ID
WHERE t0.HASH_SYS_AND_VALUE = ?
  AND t3.TAG_TYPE = 0
  AND t3.TAG_SYSTEM = 'http://hapifhir.io/fhir/NamingSystem/mdm-record-status'
  AND t3.TAG_CODE   = 'GOLDEN_RECORD'
FETCH FIRST 10000 ROWS ONLY;
```

In the `EXPLAIN (ANALYZE)` output, the `hfj_res_tag` join has a badly wrong row estimate (est. hundreds vs actual millions), driving a nested loop over `hfj_spidx_token`.

### To Reproduce
1. Load a dataset where one tag is applied to a very large number of resources (hundreds of thousands to millions) so the estimate is materially off.
2. Issue a `_tag`-filtered search combined with another indexed parameter, e.g.:
   ```
   GET /Patient?_tag=http://hapifhir.io/fhir/NamingSystem/mdm-record-status|GOLDEN_RECORD&<another indexed param>
   ```
3. Capture the generated SQL and run `EXPLAIN (ANALYZE, BUFFERS)`.
4. Observe the estimate-vs-actual mismatch on the `hfj_res_tag`/`hfj_tag_def` join and the resulting nested-loop plan.

The same shape occurs for plain `_tag` searches without MDM, and for `_security`/`_profile` (they share the code path, differing only by `TAG_TYPE`).

### Expected behavior
The resource search should not hide the selective `TAG_ID` behind a join. The tag id(s) should be resolved up front and applied as a predicate on the indexed `HFJ_RES_TAG.TAG_ID` column, so the planner can use per-value statistics and choose an appropriate plan.

### Root cause
Tag filtering is built in `QueryStack.createPredicateTag(...)` via `TagPredicateBuilder`, which joins `HFJ_TAG_DEF` and filters on `TAG_TYPE`/`TAG_SYSTEM`/`TAG_CODE`. The `TAG_ID` therefore only appears as a join column, invisible to the planner as a value.

jorgepinilla-smile:
Context:
1. Resolve the matching TAG_ID(s) from HFJ_TAG_DEF (by type/system/code) before building the resource search.
2. In the resource search, filter HFJ_RES_TAG.TAG_ID IN (...) directly and drop the HFJ_TAG_DEF join.

NOTE: Applies uniformly to _tag, _security, and _profile (they share the code path, differing only by TAG_TYPE).

-- Step 1: resolve tag_id(s)
SELECT tag_id
FROM hfj_tag_def
WHERE tag_system = 'http://hapifhir.io/fhir/NamingSystem/mdm-record-status'
  AND tag_code   = 'GOLDEN_RECORD'
  AND tag_type   = 0;

-- Step 2: inject as literal
SELECT t1.PARTITION_ID, t1.RES_ID
FROM hfj_resource t1
  INNER JOIN hfj_spidx_token t0 ON t1.RES_ID = t0.RES_ID
  INNER JOIN hfj_res_tag     t2 ON t1.RES_ID = t2.RES_ID
WHERE t0.PARTITION_ID = 1
  AND t0.HASH_SYS_AND_VALUE = -8958345584868966345
  AND t2.PARTITION_ID = 1
  AND t2.TAG_ID IN (...from Step 1...)
FETCH FIRST 10000 ROWS ONLY;{code}


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8308-tag-search-filter-on-resolved-tag-id.yaml
```diff
@@ -0,0 +1,8 @@
+---
+type: fix
```
michaelabuckley: This is not a fix.  The behaviour was correct.  

```suggestion
type: performance
```

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8308-tag-search-filter-on-resolved-tag-id.yaml
```diff
@@ -0,0 +1,8 @@
...
+issue: 8308
+title: "Previously, searches using the `_tag`, `_security`, or `_profile` parameters resolved the tag
+  through a join to `HFJ_TAG_DEF`, hiding the selective tag id from the database query planner. For
+  high-cardinality tags this led to cardinality underestimates and very slow query plans. This has been
+  fixed by resolving the matching tag id(s) up front and filtering on the indexed `HFJ_RES_TAG.TAG_ID`
+  column directly."
```
michaelabuckley:
```suggestion
title: "Faster queries using _tag, and _security.
Previously, searches using the `_tag`, or `_security` parameters resolved the tag
  through a join to `HFJ_TAG_DEF`, hiding the selective tag id from the database query planner. For
  high-cardinality tags this led to cardinality underestimates and very slow query plans. This has been
  fixed by resolving the matching tag id(s) up front and filtering on the indexed `HFJ_RES_TAG.TAG_ID`
  column directly."
```

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchOptimizedTest.java
```diff
@@ -1039,7 +1039,7 @@ public void testSearchOnUnderscoreParams_AvoidHFJResourceJoins() {
 		searchParameter1.setExpression("Procedure.extension('Procedure#focalAccess')");
 		searchParameter1.setXpathUsage(SearchParameter.XPathUsageType.NORMAL);
 		searchParameter1.setStatus(Enumerations.PublicationStatus.ACTIVE);
-		IIdType sp1Id = mySearchParameterDao.create(searchParameter1).getId().toUnqualifiedVersionless();
```
michaelabuckley, line 1042: suggestion: I'm surprised to see no changes in FhirResourceDaoR4QueryCountTest. Please add a couple of tests using _tag and _security so we can track the performance there.  We try to minimize the number of round trips.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -2149,7 +2176,9 @@ public Condition createPredicateTag(
 								theParamName,
 								() -> mySqlBuilder.addTagPredicateBuilder(theSourceJoinColumn))
 						.getResult();
-				tagPredicate = tagJoin.createPredicateTag(tagType, tokens, theParamName, theRequestPartitionId);
+				tagPredicate = resolvedTagIds != null
+						? tagJoin.createPredicateTagIds(resolvedTagIds)
```
michaelabuckley, line 2185: performance question: Does this resolve each SP separately?  E.g. how many database queries are issued for the query
```
Patient?_tag=abc&_tag=def&_security=ghi
```
Optimal is 2.  I think this solution makes 4.

If so, design suggestion:
- collect all the tags to resolve up front, resolve them in one round trip, and store a small data structure in RequestDetails userData.
- then use this data here when building the predicates

michaelabuckley (changes requested):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
jorgepinilla-smile: It does resolve each tag separately. I've refactored the code to resolve the tag definitions in batch in order to avoid doing single lookups for each tag.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4SearchOptimizedTest.java
jorgepinilla-smile: Added a couple unit tests that cover the multi tag scenario

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -2159,6 +2188,46 @@ public Condition createPredicateTag(
...
+		}
+
+		List<Long> tagIds = new ArrayList<>();
+		for (Triple<String, String, String> next : theTokens) {
+			String system = next.getLeft();
+			String qualifier = next.getMiddle();
+			String code = next.getRight();
+
+			// A left-match (:below) can expand to an unbounded set of tag ids; keep the legacy join.
+			if (Objects.equals(qualifier, UriParamQualifierEnum.BELOW.getValue())) {
```
michaelabuckley, line 2213: sharp choice.  Agree.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -2159,6 +2188,46 @@ public Condition createPredicateTag(
...
+	 * Resolves the {@code HFJ_TAG_DEF.TAG_ID} value(s) matching the given tag tokens so a
+	 * {@code _tag}/{@code _security}/{@code _profile} search can filter on {@code HFJ_RES_TAG.TAG_ID}
+	 * directly, keeping the selective tag id visible to the query planner instead of hiding it behind a join.
+	 *
+	 * @return the matching tag ids (an empty list means none exist), or {@code null} when
+	 *     pre-resolution should not be applied — either no {@link ITagDefinitionDao} is wired, or a
+	 *     token uses the {@code :below} qualifier whose left-match must stay in the legacy join path.
+	 */
+	@Nullable
+	private List<Long> resolveTagIds(TagTypeEnum theTagType, List<Triple<String, String, String>> theTokens) {
```
michaelabuckley, line 2201: minor: we could combine all three types (_tag, _security, _profile) in one round trip.  So the params would look more like `List<Quad<>>` instead of `TagTypeEnum, List<Triple<>>`.  Maybe add that as a perf-todo comment so you can get this merged.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -2111,6 +2129,11 @@ public Condition createPredicateTag(
 			throw new IllegalArgumentException(Msg.code(1217) + "Param name: " + theParamName); // shouldn't happen
 		}
 
+		// Batch-resolve the tag definitions for every tag token in this parameter with a single
+		// HFJ_TAG_DEF lookup, so we do not issue one lookup per token. Null means pre-resolution is
+		// unavailable (no DAO wired) and the legacy HFJ_TAG_DEF join is used.
+		List<TagDefinition> resolvedDefinitions = batchResolveTagDefinitions(tagType, theList);
```
michaelabuckley, line 2135: performance: this is still making a separate round trip for each _tag param.  A single query might have multiple AND lists.  I suggest you add a field to QueryStack tracking this data.  A Map<tag-def,List<tag-ids>> that you populate much earlier than this with a GLOBAL look at the entire query.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
```diff
@@ -2159,6 +2193,84 @@ public Condition createPredicateTag(
...
+		if (theDefinitions == null) {
+			return null;
+		}
+
+		List<Long> tagIds = new ArrayList<>();
+		for (Triple<String, String, String> next : theTokens) {
+			String system = next.getLeft();
+			String qualifier = next.getMiddle();
+			String code = next.getRight();
+
```
michaelabuckley, line 2254: minor: a record type would make this much easer/safer than a Tripple.

michaelabuckley (approved):

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
```diff
@@ -5205,4 +5206,48 @@ public List<IAuthRule> buildRuleList(RequestDetails theRequestDetails) {
...
+		myCaptureQueriesListener.clear();
+		SearchParameterMap map = SearchParameterMap.newSynchronous()
+			.add(Constants.PARAM_TAG, new TokenParam("http://sys", "code-1"))
+			.add(Constants.PARAM_TAG, new TokenParam("http://sys", "code-2"))
+			.add(Constants.PARAM_TAG, new TokenParam("http://sys", "code-3"));
+		IBundleProvider outcome = myPatientDao.search(map, mySrd);
+		assertThat(toUnqualifiedVersionlessIdValues(outcome)).containsExactly(id.getValue());
+
+		myCaptureQueriesListener.logSelectQueriesForCurrentThread();
+		assertEquals(4, myCaptureQueriesListener.logSelectQueries().size());
```
michaelabuckley, line 5228: question: is this linear in the number of _tag params?  Ideally, this should be 2.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
jorgepinilla-smile: Good suggestion! I refactored triple to be a TokenTag record instead.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
jorgepinilla-smile: yes it is linear.  The refactor made was meant to resolved N number of tags under a single query. So the number of queries doesn't change.

Before my changes the number of queries was 3 where the tag definition table was inner joined with the resource table. Which we want to avoid in order to get a better estimate of rows per tag ID.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
jorgepinilla-smile: I've updated code to resolve all types (_tag, _security, _profile) under a single lookup. Created a hash map that would keep track of the tag codes and updated the sql query to search for tag types and tag codes in tag definition table.

Update / added unit tests to reflect this.

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/search/builder/QueryStack.java
jorgepinilla-smile: Updated this to do a single lookup for all tag type parameters by creating a hash map that keeps track of the tag codes as well as the tag types in order to pass them to the SQL query.

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/dao/r4/FhirResourceDaoR4QueryCountTest.java
michaelabuckley: Looks good.

michaelabuckley (approved):

PR #8320, opened 2026-08-28
GL-9258: Reindex v3 fails permanently on retryable TRM_VALUESET lock conflict

fil512:
### For Code Reviewer

`RetryChunkLaterException` (a `RuntimeException`) thrown by a bulk-modify step's pre-flight guard (`processPidsOutsideTransaction`) was being caught by the generic `catch (Throwable)` in `processPids` and converted into a permanent per-resource failure instead of deferring the chunk. This is what caused Reindex v3 to fail with `HAPI-2790` instead of retrying when a concurrent `ValueSet` pre-expansion holds the same `TRM_VALUESET` rows.

The fix hoists the `processPidsOutsideTransaction` call above the `try` block and adds `RetryChunkLaterException` to the rethrow arm (so it's also caught if thrown from inside the transaction). One side effect worth flagging: hoisting the pre-flight call above the `try` also changes how *generic* (non-retry) exceptions from that hook are handled — they now propagate out of `run()` too, so `StepExecutor` marks the chunk a retriable `ERRORED` chunk (capped at 3 attempts via `MAX_CHUNK_ERROR_COUNT`, then `FAILED`) instead of attributing a whole-chunk guard failure to every individual resource in the chunk. This was a deliberate design choice — per-resource attribution was misleading for a chunk-level guard failure — not an accidental side effect.

Javadoc on both `processPidsOutsideTransaction` and `processPidsInTransaction` now documents the deferral contract, including that a deferred chunk restarts from the beginning (so PIDs already committed earlier in the chunk get reprocessed).

### For QA

- [ ] `mvn verify -pl hapi-fhir-storage-batch2-jobs` passes (226 tests, 0 failures), including the new `BaseBulkModifyResourcesStepTest` (6/6) and the existing `BaseBulkModifyResourcesIndividuallyStepTest` (5/5, unchanged).
- [ ] A `$reindex` (Reindex v3) job whose pre-flight guard throws `RetryChunkLaterException` (e.g. a concurrent `ValueSet` pre-expansion holding `TRM_VALUESET` rows) now defers the affected chunk (`POLL_WAITING` → retried) instead of the job ending `FAILED` with `HAPI-2790`.
- [ ] A generic (non-retry) exception from the pre-flight guard now fails the whole chunk as a retriable `ERRORED` chunk (capped at 3 attempts) rather than marking every resource in the chunk as individually failed — confirm this is the intended, reviewed behavior change, not a regression.
- [ ] Support/QA note: unlike the generic-exception path, a chunk deferred via `RetryChunkLaterException` has no attempt cap — it cycles `POLL_WAITING` → `READY` for as long as the guard condition holds true. If that condition became permanently true, `$reindex` would poll roughly every 30 seconds indefinitely with no error surfaced, instead of failing fast with `HAPI-2790`. So the visible symptom of a stuck reindex changes from "job FAILED with HAPI-2790" to "job stuck in progress in `POLL_WAITING`, no error surfaced" — when investigating a reindex that appears hung, check job status and how long it's been `POLL_WAITING`, not just jobs that ended `FAILED`. This is the existing, long-standing batch2 `POLL_WAITING` contract (Reindex v2 has behaved this way for years) — it is not new behavior introduced by this fix.


tadgh:
Formatting check succeeded!

codeforgreen:
@fil512 the intermittent test failure with `ValueSet` expansion is addressed. Please merge changes from master to retry. Also, is this ready for review?

PR #8334, opened 2026-08-31
Support multiple EID systems per resource type in MDM

### LINKED ISSUE #8317: MDM - Support Multiple EIDs for a Single Resource Type
Enterprise Identifiers (EIDs) are unique keys (such as Medical Record Numbers [MRN] for hospitals or National Provider Identifiers [NPI] for providers) that are almost always present and unique to a person or resource.

Currently, the system is restricted to implementing only _one_ EID per resource type. This causes limitations in specific regions and contexts:

* In some contexts, there are two distinct, strong candidates for EIDs.
* Evaluative matching would be much faster for high-volume data loads if the system could check multiple candidate EIDs and match if either is present and matches.

### Acceptance Criteria

* Update the MDM configuration schema to support a list of multiple EIDs per resource type rather than a single string.
* The matching algorithm must evaluate multiple EIDs (using logical "OR" conditions) and register a match if any configured, unique EID matches.
* Verify that matching performance is optimized for high-volume loads when multiple EIDs are evaluated.
* Complete and document the architectural design session with James and Michael, ensuring resolution of any potential EID collision scenarios.

Design notes from 2026/08/17:

* Modify the mappings on the rules JSON to allow an array of urls per resource type, and to escalate a singleton to an array for backward compatibility with existing rule definition files
* Link to the golden resource if any of the EID identifiers match
* For the case where there are identifiers from multiple EID systems assigned to a single person, but due to the order of resources arriving, we have already created golden resources for each of the identifiers before we learn that they both apply to a single real-world person (the diamond pattern): If there are multiple EID identifiers on a single incoming resource, and they match different golden resources, create a POSSIBLE_MATCH link to both of them, and then mark the golden resources themselves as potential duplicates

jmarchionatto:
`eidSystems` now accepts either a single system URI, as before, or an array of them. An incoming resource links to a golden resource if any of its EIDs match. Where its EIDs resolve to different golden resources, it is linked to each as a POSSIBLE_MATCH and those golden resources are flagged as possible duplicates of one another.

Two supporting behaviour changes, both no-ops under any single-system configuration:

- `preventMultipleEids` is rescoped from at most one EID per resource to at most one EID per configured system, so the default configuration accepts the payload the feature exists to support. HAPI-0766 now names the offending system.
- `eidMatchExists` compares EIDs on system and value rather than value alone, restoring a check dropped in 77149d66b90 that was a tautology while only one system could be configured.

EID candidate lookup now issues a single OR query across all of a resource's EIDs rather than one search apiece, and de-duplicates by golden resource so that several EIDs resolving to the same golden take the plain-match path.

Reviewer note: This feature widens the known gap for message concurrency, which is being addressed by other issues.

!cdr-branch=9155-smile-12697-mdm-support-multiple-eids-for-a-single-resource-type


tadgh:
Formatting check succeeded!

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/util/GoldenResourceHelper.java
```diff
@@ -209,10 +213,22 @@ private void validateContextSupported() {
 	}
 
 	/**
-	 * Updates EID on Golden Resource, based on the incoming source resource. If the incoming resource has an external EID, it is applied
-	 * to the Golden Resource, unless that golden resource already has an external EID which does not match, in which case throw {@link IllegalArgumentException}
+	 * Updates EIDs on a Golden Resource, based on the incoming source resource.
+	 * <ul>
+	 * <li>Where the Golden Resource has no external EID, or multiple EIDs are permitted, the incoming EIDs
+	 * are added to it.</li>
+	 * <li>Where the two share an EID, they are the same entity, so the incoming EIDs belonging to an EID
```
mrdnctrk: maybe this case can be divided into 2 sub cases with examples for readibility. I had to read it several times, and ask AI to give me an example to grasp it. This example helped me: `Golden Resource has MRN 1 and NPI 9; incoming has MRN 1 and NPI 7. The shared MRN settles who this is, so NPI 7 is simply not copied and the drop is logged. If Golden Resource has MRN 1, incoming has MRN 2, no overlap, so it throws.`

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/util/GoldenResourceHelper.java
```diff
@@ -244,13 +260,31 @@ public IAnyResource updateGoldenResourceExternalEidFromSourceResource(
...
-									.map(CanonicalEID::toString)
-									.collect(Collectors.joining(","))
-							+ " does not need to overwrite the EID in the Golden Resource, as this EID is already present in the Golden Resource");
+			// The resources agree on at least one EID, so they are the same entity. Any EID the incoming
+			// resource carries from an EID system the Golden Resource has no EID in is new information and
+			// is applied; an EID from a system the Golden Resource already uses is not, because a Golden
+			// Resource holding two EIDs of one system would be rejected by the "prevent multiple EIDs"
+			// safeguard on the next write.
+			if (addCanonicalEidsForEidSystemsAbsentFromGoldenResource(
+					theGoldenResource, goldenResourceOfficialEid, incomingSourceEid)) {
```
mrdnctrk: minor: I know you didn't introduce them in this PR but changing the variable names `goldenResourceOfficialEid `and `incomingSourceEid` to plural `goldenResourceOfficialEids` and `incomingSourceEids` would increase readibility.

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/FindCandidateByEidSvc.java
```diff
@@ -60,31 +62,37 @@ protected List<MatchedGoldenResourceCandidate> findMatchGoldenResourceCandidates
...
+				eidFromResource,
+				theIncomingResource.getIdElement().getResourceType(),
+				myMdmPartitionHelper.getRequestPartitionIdFromResourceForSearch(theIncomingResource));
 
-					retval.add(mpc);
-				}
+		// Several of the incoming EIDs may resolve to the same golden resource. That is one candidate, not
+		// several - reporting it more than once would send the resource down the multiple-candidate path
+		// and flag a duplicate that does not exist.
+		Set<IResourcePersistentId<?>> seenGoldenResourcePids = new LinkedHashSet<>();
```
mrdnctrk, line 79: If I am understanding right, we run a single search query with eids joined through logical OR. Could dao really return the same resource multiple times for that search? Or are we being too defensive here? I am not suggesting to remove it, as it looks harmless, but i wasn't expecting this behaviour.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/util/MdmSearchParamBuildingUtils.java
```diff
@@ -49,16 +56,66 @@ public static SearchParameterMap buildSearchParameterForBlockedResourceCount(Str
...
 	 * by EID specifically.
+	 *
+	 * @deprecated use {@link #buildEidSearchParameterMap(Collection)}, which matches on the EID system as
+	 * well as the value and can search several EIDs at once.
 	 */
+	@Deprecated(since = "8.14.0", forRemoval = true)
 	public static SearchParameterMap buildEidSearchParameterMap(
 			String theEid, String theResourceType, MdmRulesJson theMdmRules) {
-		SearchParameterMap map = buildBasicGoldenResourceSearchParameterMap(theEid);
+		SearchParameterMap map = buildBasicGoldenResourceSearchParameterMap(theResourceType);
```
mrdnctrk: I noticed that buildBasicGoldenResourceSearchParameterMap doesn't use its  parameter at all. I think we should remove the parameter completely, unless the intention was to use it.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/util/MdmSearchParamBuildingUtils.java
```diff
@@ -49,16 +56,66 @@ public static SearchParameterMap buildSearchParameterForBlockedResourceCount(Str
...
+	 * Translates EIDs into a single token OR query. Each EID is matched on its own system as well as its
+	 * value, so that the same value issued by two different EID systems is not conflated.
+	 * <p>
+	 * EIDs with no value are dropped. Such an identifier is valid FHIR and identifies nobody, but a token
+	 * search with a blank value matches on the system alone
+	 * </p>
+	 * <p>
+	 * This is the EID half of an EID search on its own, with no restriction on which resources are
+	 * searched. Use {@link #buildEidSearchParameterMap(Collection)} to search golden resources; use this
+	 * directly to search source resources.
```
mrdnctrk: I think this comment needs some clarification.

- it says `This is the EID half of an EID search on its own, with no restriction on which resources are searched`. But it doesn't clarify which other half it is referring to. Is the other half the golden resource tag? 
- Also when i read `with no restriction on which resources are searched`. I thought of FHIR resource types at first rather than golden resources vs source resources. 
- Also I think this search param map generated by this would return the golden resources as well, right, not just the source resources? `use this directly to search source resources.` gives the impression that this only returns source resource excluding the golden ones.

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/broker/MdmMessageKeySvc.java
```diff
@@ -24,23 +24,28 @@
...
+
+	public MdmMessageKeySvc(EIDHelper theEidHelper) {
+		myEIDHelper = theEidHelper;
+	}
 
+	/**
+	 * The broker routes messages sharing a key to the same consumer, which is what keeps changes to one
+	 * patient in order when several MDM consumers are running. The key must therefore not depend on the
+	 * order identifiers happen to appear in the payload. Where a resource type is identified by several
+	 * EID systems, the primary one - the first configured for that resource type - decides the key.
```
mrdnctrk: if prevent multiple eids is turned off, and the resource has multiple eids for a system, then the value would depend on the body ordering. say MRN is the first configured eid system and resource has 2 of those, then the value would be the first from the payload. i think this is an edge case, I am not saying we should do sth about it just thought about it, since comment said not to rely on body ordering.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8317-mdm-support-multiple-eid-systems-per-resource-type.yaml
```diff
@@ -0,0 +1,14 @@
+---
+type: add
+issue: 8317
+jira: SMILE-12697
+title: "MDM rule definitions can now map a resource type to more than one Enterprise Identifier (EID)
+  system: each entry in `eidSystems` accepts an array of system URIs as well as a single one, and a
+  resource links to a golden resource if any of its EIDs match. Note these changes to existing
```
mrdnctrk: `Note these changes to existing behaviour.` Something is off grammatically in this sentence. Should it be `Note this changes existing behaviour.` or `This is a change to the existing behaviour`?

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_mdm/mdm_eid.md
```diff
@@ -1,18 +1,53 @@
...
+## How a Golden Resource Accumulates EIDs
+
+A Golden Resource holds the EIDs of the source resources linked to it, and gains them as those resources
+arrive. A Golden Resource created from a resource carrying only a medical record number acquires the
+national provider identifier as soon as a resource carrying both is matched to it - and from that point a
+later resource carrying only the national provider identifier resolves to the same Golden Resource. This
+applies on create and on update alike.
+
+<p class="helpInfoCalloutBox">
+    There is one exception. Where the Golden Resource already holds an EID from a given system, an incoming EID from that <i>same</i> system with a different value is not copied while <b>Prevent multiple EIDs</b> is enabled, since the Golden Resource would then be ambiguous within that system. Such an EID is a contradiction between two issuing authorities rather than new information; it is recorded in the MDM transaction log so that it can be found, but no link or duplicate is raised for it.
```
mrdnctrk: It is not immediately obvious from this callout box that the case we are talking about is that the golden resource and the incoming resource match on one eid system and disagree on another. Stating that explicitly would make it read easier. An example could help as well, like "Where several EID systems are configured, a Golden Resource holding MRN 1 and NPI 9 that matches a resource carrying MRN 1 and NPI 7 keeps its existing NPI 9 and discards NPI 7, and logs this discrepancy in transaction log. "

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmEidUpdateServiceSurvivorshipOrderTest.java
```diff
@@ -0,0 +1,121 @@
...
+		myCandidate = new MatchedGoldenResourceCandidate(goldenPid, MdmMatchOutcome.EID_MATCH);
+
+		when(myMdmGoldenResourceFindingSvc.getGoldenResourceFromMatchedGoldenResourceCandidate(
+						eq(myCandidate), any()))
+				.thenReturn(myGoldenResource);
+
+		// The incoming resource carries an EID that the Golden Resource already has, which is the branch
+		// that merges EIDs into the Golden Resource.
+		when(myEIDHelper.hasEidOverlap(myGoldenResource, myTargetResource)).thenReturn(true);
+		when(myEIDHelper.getExternalEid(myTargetResource)).thenReturn(Collections.emptyList());
```
mrdnctrk: There is an inconsistency: the comment above says the target resource has an eid matching golden but getExternalEid is setup to return empty list on targetresource.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchFinderSvcEidR4Test.java
```diff
@@ -0,0 +1,84 @@
...
+		Patient incoming = createPatient(addExternalEID(buildFrankPatient(), eidSystem, "eid-1"));
+
+		assertThat(eidMatchedIds(incoming)).containsExactly(versionlessId(jane));
+	}
+
+	/**
+	 * An identifier carrying an EID system but no value is legal FHIR and identifies nobody. Searching on
+	 * it would fall back to matching the EID system alone and return every resource in that system, each
+	 * as a full MATCH at score 1.0.
+	 */
```
mrdnctrk, line 41: This comment describes the buggy behaviour as if it is what is expected to happen. There should be a statement that this is the undesidered behaviour that this test verifies that cannot happen.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchLinkSvcMultiEidSystemTest.java
```diff
@@ -0,0 +1,228 @@
...
+ * Linking behaviour when a resource type is identified by more than one EID system. Runs at the default
+ * {@code prevent_multiple_eids=true}, which is scoped per EID system and so permits a Patient carrying
+ * one MRN and one NPI.
+ */
+@TestPropertySource(properties = {"module.mdm.config.script.file=classpath:mdm/mdm-rules-multi-eid-systems.json"})
+// Created by claude-opus-5
+public class MdmMatchLinkSvcMultiEidSystemTest extends BaseMdmR4Test {
+
+	private String mrnSystem() {
+		return patientEidSystems().get(0);
```
mrdnctrk: would it make sense to move these methods to base class, mrnSystem() and npiSystem()? I think i have seen them defined, used several times in different classes.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/json/EidSystemListDeserializer.java
```diff
@@ -0,0 +1,98 @@
...
+
+		if (token == JsonToken.VALUE_STRING) {
+			return Collections.singletonList(theParser.getText());
+		}
+
+		if (token == JsonToken.START_ARRAY) {
+			List<String> retVal = new ArrayList<>();
+			while (theParser.nextToken() != JsonToken.END_ARRAY) {
+				if (theParser.currentToken() != JsonToken.VALUE_STRING) {
+					throw invalidValue(theParser);
```
mrdnctrk: I think in this line the invalid value message does not produce the field name, as we are in the middle of the parsing array element `theParser.currentName()` returns `null`.  The test `deserialize_arrayContainingNonString_throwsNamingTheResourceType` still passes  

```
		assertThatThrownBy(() -> deserialize("""
			{"eidSystems": {"Patient": ["http://example.com/mrn", 42]}}"""))
			.isInstanceOf(JsonMappingException.class)
			.hasMessageContaining(Msg.code(3046))
			.hasMessageContaining("eidSystems")
			.hasMessageContaining("Patient");
```

but the message contains Patient because of Jackson  adds the reference chain rather than `theParser.currentName()` returning `'Patient'`. It might be worth checking the exact error message for this case in the test.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/json/EidSystemListDeserializer.java
```diff
@@ -0,0 +1,98 @@
...
+	 * An explicit {@literal null} is treated as "no EID systems configured for this resource type"
+	 * rather than as a null list, so that callers never have to null-check.
+	 */
+	@Override
+	public List<String> getNullValue(DeserializationContext theContext) {
+		return Collections.emptyList();
+	}
+
+	private JsonMappingException invalidValue(JsonParser theParser) throws IOException {
+		return JsonMappingException.from(
```
mrdnctrk: I think it is worth checking what kind of exception the user might get in this case, claude tells me since JsonMappingException is an IOException, we reflect that to the client as an InternalErrorException in this case rather than a user error, like a configuration error. It might be worth adding a bad json rule test for this case to see that the user gets a ConfigurationException or sth that indicate user error, instead of InternalError.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/json/EidSystemListDeserializer.java
```diff
@@ -0,0 +1,98 @@
...
+		throw invalidValue(theParser);
+	}
+
+	/**
+	 * An explicit {@literal null} is treated as "no EID systems configured for this resource type"
+	 * rather than as a null list, so that callers never have to null-check.
+	 */
+	@Override
+	public List<String> getNullValue(DeserializationContext theContext) {
+		return Collections.emptyList();
```
mrdnctrk: this returns an unmodifiable list but serialize returns a modifiable one in one case and unmodifiable one in another case, is that a problem? probably not, as we create copies of this to return to caller but just wanted to raise it in case you think otherwise.

mrdnctrk (approved):

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/broker/MdmMessageKeySvc.java
jmarchionatto: Made the comment explicit about it. Note that this case is expected to be changed as part of the MDM concurrency fixing about to take place.

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8317-mdm-support-multiple-eid-systems-per-resource-type.yaml
jmarchionatto: Referred to the changes described after. Changed to: "Note the following changes to existing
  behaviour."

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/server_jpa_mdm/mdm_eid.md
jmarchionatto: Rewritten and example added.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchFinderSvcEidR4Test.java
jmarchionatto: Fixed

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/util/GoldenResourceHelper.java
jmarchionatto: Done

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/util/GoldenResourceHelper.java
jmarchionatto: Done

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/FindCandidateByEidSvc.java
jmarchionatto: Clarified the comment

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/util/MdmSearchParamBuildingUtils.java
jmarchionatto: Right. Removed.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/util/MdmSearchParamBuildingUtils.java
jmarchionatto: All good points, specially the third. Fixed.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchLinkSvcMultiEidSystemTest.java
jmarchionatto: Absolutely! Done.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmEidUpdateServiceSurvivorshipOrderTest.java
jmarchionatto: Right. Fixed

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/json/EidSystemListDeserializer.java
jmarchionatto: Great catch! Fixed.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/json/EidSystemListDeserializer.java
jmarchionatto: Yes. Good point!                                                                                                                                    
                                                                                                                                                             
Fixed only the eidSystems case: the deserializer throws InvalidEidSystemsException (a JsonMappingException subclass), and setScriptText reads via the InputStream overload and converts just that type to ConfigurationException. Every other malformed rules document behaves as before. A distinct type rather than a message check because InternalErrorException is built as Msg.code(2060) + e with no cause, so the alternative was string-matching on HAPI-3046.     
                                                                                                                                                             
Covered by MdmRuleValidatorTest.eidSystems_withNonStringInTheArray_throwsConfigurationException.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/json/EidSystemListDeserializer.java
jmarchionatto: Both getters wrap in unmodifiableList, but consistency doesn't hurt. Changed so MdmRulesJson holds one kind of list either way.

tadgh (approved):

PR #8127, opened 2026-06-30
Convert ValueSet pre-expansion to Batch2 job

jamesagnew:
The terminology service ValueSet pre-expander has been converted from an anonymous scheduled background job into a Batch2 job. This makes the process more stable, easier to monitor, and means that the `$invalidate-expansion` operation now returns ongoing status information and produces a report when the pre-expansion is complete.

PR #8335, opened 2026-08-31
Fix ValueSet pre-expansion running before terminology is ready

### LINKED ISSUE #8321: ValueSet pre-expansion runs before terminology is ready (deferred concepts; import finalize ordering)
**Describe the bug**

Two defects cause `ValueSet` pre-expansion to run before the terminology it depends on is ready. This used to work in hapi-fhir 8.12 but is now broken post migration of terminology related operations to batch2. See: #8127, #7953, #8040, #8077

### D1 — pre-expansion starts while CodeSystem concepts are still deferred

`TermValueSetStorageSvcImpl` registers a `TransactionSynchronization.afterCommit()` that starts a `PRE_EXPAND_VALUESET` batch2 job as soon as an active `ValueSet` commits, with no check that terminology storage is complete. When the referenced `CodeSystem` is large enough to defer concept indexing (`deferIndexingForCodesystemsOfSize`), the concepts still queued in `TermDeferredStorageSvcImpl` are not yet in `TRM_CONCEPT`, so the expansion written to `TRM_VALUESET_CONCEPT` is missing codes. The `ValueSet` is then marked `EXPANDED`, so nothing retries it and every later `$expand` returns the undercount.

`TermReadSvcImpl.isNotSafeToPreExpandValueSets()` (`TermReadSvcImpl:2495`) is the readiness check for this, but it is no longer reachable from the pre-expansion trigger — its only caller is a `@VisibleForTesting` helper on the reindex path. `hapi-fhir-docs/.../server_jpa/valueset_expansion.md` still documents the guard as current behaviour and needs correcting alongside the fix.

### D2 — import finalize activates ValueSets before the CodeSystem version

`ImportTerminologyStepFinalize.run()` loops `updateResourceStatusToActive(...)` over `myResourcesToActivate` — patching the `ValueSet`s to `ACTIVE`, which fires the trigger above on commit — and only afterwards calls `activateStagingCodeSystemVersion` / `makeCodeSystemCurrent`. Every SNAPSHOT-mode import starts pre-expansion against a `CodeSystem` version that is not yet current.

**To Reproduce**

Both tests are committed on branch `8321-race-condition-terminology` (commit `a7c5f045c83`, tests only) and fail:

1. `ValueSetExpansionR4Test#preExpand_CodeSystemConceptStorageDeferred_ExpandsAllConcepts` — `expected 150, was 100`. Deterministic: the `CodeSystem` is sized past the defer threshold, so the queue cannot be empty when the trigger fires.
2. `ImportTerminologyStepFinalizeTest#run_valueSetsToActivate_activatesCodeSystemVersionBeforeValueSets` — Mockito `InOrder` finds `activateStagingCodeSystemVersion` but no `patch` after it.

**Expected behavior**

1. A pre-expansion job does not start until the terminology it depends on is fully stored, applying the `isNotSafeToPreExpandValueSets()` condition rather than a second copy of it.
2. `ImportTerminologyStepFinalize` activates the `CodeSystem` version before it activates the `ValueSet`s.
3. Both tests above pass.

- HAPI FHIR Version 8.13.4-SNAPSHOT

codeforgreen:
Fixes #8321

A `ValueSet` could be pre-expanded before the `CodeSystem` it references was fully stored, and the partial result was then marked `EXPANDED`. Nothing re-runs it, so `$expand` and `$validate-code` silently omit valid codes. Two independent defects caused this.

## 1. Pre-expansion ran against concepts still on the deferred queue

Storing an active `ValueSet` starts the `PRE_EXPAND_VALUESET` job on commit, with no check that the terminology it needs is stored. Concepts past `deferIndexingForCodesystemsOfSize` (default 100) are still on `TermDeferredStorageSvcImpl`'s in-memory queue at that point, so the expansion is written from whatever is already in the database. The equivalent check used to live in `TermReadSvcImpl.preExpandDeferredValueSetsToTerminologyTables()`, removed by #8127 along with the scheduled task that called it.

**Fix** — `Step1InitiateJob.run()` checks `isStorageQueueEmpty(false)` before staging anything. If concepts are still queued it drains them with `saveDeferred()`; if any remain it throws `RetryChunkLaterException`, parking the work chunk in `POLL_WAITING` until they are stored. The check has to precede `startStagingVersion()`, which inserts a `TermValueSet` staging row on every call.

## 2. Import finalize activated ValueSets before the CodeSystem version

`ImportTerminologyStepFinalize.run()` patched `ValueSet`s to `ACTIVE` — firing the trigger above — and only then made the new `CodeSystem` version current. Pre-expansion therefore resolved the empty placeholder on a first upload, or the previous release on an upgrade. A full LOINC release starts one such job per `ValueSet`, all inside that window.

**Fix** — the activation block moves above the `updateResourceStatusToActive` loop.

## Two defects found while investigating a CI failure

- `TermDeferredStorageSvcImpl.saveAllDeferred()` spun forever while deferred processing was paused. `isStorageQueueEmpty()` reports a paused queue as non-empty — that is how a terminology reindex holds off pre-expansion — and `saveDeferred()` is a no-op in that state, so the loop had no exit condition. It now returns.
- `ReductionStepExecutorServiceImpl` left a job `FAILED` with a blank error message unless the reduction worker threw `ReductionStepFailureException`. Every other exception was logged but never recorded on the instance, so the cause was invisible to the admin API and to tests. The generic catch now records it.

## Tests

- `TermValueSetPreExpansionLifecycleR4Test#preExpansion_onCodeSystemConceptStorageDeferred_expandsAllConcepts` — a `CodeSystem` larger than `deferIndexingForCodesystemsOfSize`, so the queue cannot be empty when the job starts. Fails with `expected 150, was 100` before the fix.
- `TermValueSetPreExpansionLifecycleR4Test#preExpansion_onDeferredStorageProcessingPaused_waitsForTheQueueThenExpandsAllConcepts` — pauses deferred processing so the step cannot drain the queue itself, and asserts the chunk parks in `POLL_WAITING` with the `ValueSet` still `NOT_EXPANDED`, then completes once processing resumes. This is what exercises `RetryChunkLaterException` end to end.
- `ImportTerminologyStepFinalizeTest` — Mockito `InOrder` on the two activation calls, for both SNAPSHOT and ADD modes.
- `Step1InitiateJobTest` — `RetryChunkLaterException` with the configured delay and `Msg.code(3047)` when the queue does not empty, and no call to `startStagingVersion` in that case.
- `TerminologyLoaderSvcLoincJpaTest#importLoinc_twoVersions_PreExpansionsContainOnlyCodesStoredByTheImport` — asserts a pre-expansion holds only codes the import actually stored.
- `TermDeferredStorageSvcImplTest` and `ReductionStepExecutorServiceImplTest` cover the two defects above.

Also updated `server_jpa/valueset_expansion.md` and the `TermValueSetPreExpansionStatusEnum` descriptions, which still described a scheduled task.

## Known limitation, not fixed here

The deferred queue is in-memory and per-node, so the readiness check only describes the node it runs on; a job dispatched elsewhere still sees an empty queue. Not a regression — the check removed by #8127 read the same in-memory queue — and fixing it needs a `MigrationTask` for a load-state column on `TRM_CODESYSTEM_VER`, after which it is a one-line change at the same call site. Tracked in #8354.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/valueset/preexpand/Step1InitiateJob.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/valueset/preexpand/Step1InitiateJob.java
@@ -22,11 +22,14 @@ package ca.uhn.fhir.jpa.batch2.jobs.term.valueset.preexpand;
 import ca.uhn.fhir.batch2.api.IJobDataSink;
 import ca.uhn.fhir.batch2.api.IJobStepWorker;
 import ca.uhn.fhir.batch2.api.JobExecutionFailedException;
+import ca.uhn.fhir.batch2.api.RetryChunkLaterException;
 import ca.uhn.fhir.batch2.api.RunOutcome;
 import ca.uhn.fhir.batch2.api.StepExecutionDetails;
 import ca.uhn.fhir.batch2.api.VoidModel;
 import ca.uhn.fhir.context.support.IValidationSupport;
+import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.jpa.entity.TermValueSetConcept;
+import ca.uhn.fhir.jpa.term.api.ITermDeferredStorageSvc;
 import ca.uhn.fhir.jpa.term.api.ITermValueSetStorageSvc;
 import ca.uhn.fhir.util.IntCounter;
 import ca.uhn.fhir.util.UrlUtil;
@@ -38,6 +41,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.List;
 
 public class Step1InitiateJob
@@ -52,8 +57,16 @@ public class Step1InitiateJob
 	 */
 	private static final int MAX_CONCEPTS_PER_COMPOSE = 1_000_000;
 
+	/**
+	 * The delay before the next poll for the deferred terminology storage queue to empty.
+	 */
+	private static final Duration RETRY_DELAY = Duration.of(30, ChronoUnit.SECONDS);
+
 	@Autowired
 	private ITermValueSetStorageSvc myTermValueSetStorageSvc;
+
+	@Autowired
+	private ITermDeferredStorageSvc myDeferredStorageSvc;
 
 	@Autowired
 	private VersionCanonicalizer myVersionCanonicalizer;
@@ -71,6 +84,25 @@ public class Step1InitiateJob
 		PreExpandValueSetParameters parameters = theStepExecutionDetails.getParameters();
 		String url = parameters.getUrl();
 		String version = parameters.getVersion();
+
+		/*
+		 * Expanding while concepts are still queued would resolve the compose rules against a
+		 * partially stored CodeSystem. Process the deferred entities here rather than waiting for the
+		 * scheduled task, which may be suppressed on this node. Must precede startStagingVersion(..),
+		 * which creates a new staging TermValueSet row on every call.
+		 */
+		if (!myDeferredStorageSvc.isStorageQueueEmpty(false)) {
+			myDeferredStorageSvc.saveDeferred();
+
+			if (!myDeferredStorageSvc.isStorageQueueEmpty(false)) {
+				ourLog.info(
+						"Deferred terminology storage is still in progress, delaying pre-expansion of ValueSet[url={}, version={}]",
+						url,
+						version);
+				throw new RetryChunkLaterException(Msg.code(3043), RETRY_DELAY);
+			}
+		}
+
 		String stagingVersion = myTermValueSetStorageSvc.startStagingVersion(url, version);
 
 		UrlUtil.CanonicalUrlParts canonicalUrl = UrlUtil.parseCanonicalUrl(url, version);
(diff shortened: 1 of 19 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/valueset/preexpand/Step1InitiateJob.java
```diff
@@ -71,6 +87,25 @@ public RunOutcome run(
...
+		 */
+		if (!myDeferredStorageSvc.isStorageQueueEmpty(false)) {
+			myDeferredStorageSvc.saveDeferred();
+
+			if (!myDeferredStorageSvc.isStorageQueueEmpty(false)) {
+				ourLog.info(
+						"Deferred terminology storage is not ready, delaying pre-expansion of ValueSet[url={}, version={}]",
+						url,
+						version);
+				throw new RetryChunkLaterException(Msg.code(3047), getRetryLaterDelay());
```
tadgh, line 105: this is a pretty clean fix, nice!

tadgh (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/batch2/jobs/term/valueset/preexpand/Step1InitiateJob.java
codeforgreen: thank you @tadgh happy to work on it

PR #8337, opened 2026-08-31
8318 mdm matching skip placeholders

### LINKED ISSUE #8318: Add configuration to allow ignoring placeholder resources in mdm matching
By default mdm matching will try and match placeholder resources before they'e been populated.

This is sometimes not desired behaviour.

Allowing an ability to skip said resources is desireable.

TipzCM:
closes #8318


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/validation/RepositoryValidatingInterceptorR4Test.java
```diff
@@ -530,6 +533,30 @@ public void testImpliedProfileAlways_ValidResource_Success() {
...
+		List<IRepositoryValidatingRule> rules = newRuleBuilder()
+			.forResourcesOfType("Patient")
+			.requireAtLeastOneProfileOf("http://foo/Profile1", "http://foo/Profile2")
+			.build();
+		myValInterceptor.setRules(rules);
+
+		Binary binary = new Binary();
+		binary.setContentType("text/plain");
+		binary.setContent("hello".getBytes(StandardCharsets.UTF_8));
+		assertThat(myBinaryDao.create(binary).getId().getVersionIdPart()).isEqualTo("1");
```
jmarchionatto: Why version 1 is asserted? Add a comment explaining your knowledge about this test assertions

hapi-fhir-storage/src/test/java/ca/uhn/fhir/storage/PlaceholderResourceUtilTest.java
```diff
@@ -42,4 +50,89 @@ void testBuildPlaceholderResource_typeWithoutExtensionSupportSkipsMarker() {
...
+		}
+		// 3 basic resource with multiple extensions, but not a placeholder one
+		{
+			Patient patient = new Patient();
+			patient.addExtension()
+				.setUrl(EXT_RESOURCE_PLACEHOLDER + "/not")
+				.setValue(new BooleanType(true));
+			args.add(patient);
+		}
+		// 4 basic resource with wrong data type
```
jmarchionatto: ... with extension with wrong data type

jmarchionatto (changes requested):
I cant see an exception implemented for placeholders with matching EIDs. Also the placeholder metadata definition was narroed from just existence to having a true value which isn't part of the requirement, and no visible changelog or documentation update.

hapi-fhir-storage/src/test/java/ca/uhn/fhir/storage/PlaceholderResourceUtilTest.java
TipzCM: updated

hapi-fhir-jpaserver-test-r4/src/test/java/ca/uhn/fhir/jpa/interceptor/validation/RepositoryValidatingInterceptorR4Test.java
TipzCM: uodated

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmResourceFilteringSvc.java
dotasek, line 54: question: Does the JavaDoc here need to be updated to include the ignorePlaceHolderResources behaviour?

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmPlaceholderResourceIT.java
```diff
@@ -0,0 +1,352 @@
...
+
+	// placeholder as candidate
+	@ParameterizedTest
+	@ValueSource(booleans = {
+		true,
+		false
+	})
+	public void getMatchedTargets_realResource_ignoresPlaceholdersInDb(boolean theHasEidIdentifier) throws InterruptedException {
+		// setup
+		IIdType placehodlerId;
```
dotasek: nitpick: typo on placehodler

dotasek (changes requested):
Nitpicks and documentation questions.

Not a codeowner review; I defer to any codeowner comments.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmPlaceholderResourceIT.java
TipzCM: updated

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmResourceFilteringSvc.java
TipzCM: updated

jmarchionatto (approved):

jmarchionatto (changes requested):
Revising approval. IT seems to have issues...

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/api/IMdmSettings.java
```diff
@@ -33,6 +33,12 @@ public interface IMdmSettings {
 
 	boolean isEnabled();
 
+	/**
+	 * Whether or not placeholder resources will be ignored during MDM matching.
+	 * By default this is 'false'.
+	 */
+	boolean isIgnorePlaceholderResources();
```
jmarchionatto, line 40: Is a breaking change as is. Add default method returning false for backward compatibility.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmPlaceholderResourceIT.java
```diff
@@ -0,0 +1,399 @@
...
+	private MdmRulesJson myExistingRules;
+
+	@BeforeEach
+	public void before() throws Exception {
+		super.before();
+		myExistingRules = myMdmSettings.getMdmRules();
+
+		MdmRulesJson rules = JsonUtil.deserialize(MDM_RULES_JSON, MdmRulesJson.class);
+
+		myMdmSettings.setMdmRules(rules);
```
jmarchionatto, line 94: This is cached at construction, so ...matchesRealOnly (:264) and ...matchesPlaceholderToo(theIsEid=false) (:449) assert only link counts (2 and 3), which come out the same whether the resources merge into one golden resource or each get their own. Today each gets its own, and neither test notices.

jmarchionatto (approved):
Approved pending a couple of fixes

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/api/IMdmSettings.java
TipzCM: There is a version bump for current implementations.

I'm not sure it matters if someone who has their own rolled implementation doesn't have this.

In fact, i'd argue it's probably better than their own rolled version *does* break so they're forced to implement it the way our implementations do, instead of it being completely hidden with a default.

hapi-fhir-jpaserver-mdm/src/test/java/ca/uhn/fhir/jpa/mdm/svc/MdmPlaceholderResourceIT.java
TipzCM: I think claude is out of date on the code. The tests don't just assert link count for one thing.

But also, they're not testing match logic other than whether or not placeholders are included in the matches (which *is* tested).

Truthfully i don't think this is valuable in this test (it makes them dependent on the logic of matching when they're not actually testing that).

I'll add it to the test that takes the ied bool (because that one is checking that matches *exist*). But i didn't add it to the "matchesRealOnly" one because the thing under test is that *no* match for placeholder is created (not how the other resources are matched, which isn't relevant to the placeholder itself).

TipzCM: Not sure there's much value in checking the match logic (i added it; even though all i care about is *if* the placeholders are matched)

i did update the settings though to not cache

dotasek (approved):
Looks good to me.

AD1306 (approved):

PR #8346, opened 2026-09-02
Consider a CodeableConcept valid if any coding is in the bound ValueSet

### LINKED ISSUE #8345: CodeableConcept validation requires ALL codings to be in the bound ValueSet, spec requires only one
### Problem

`ValidationSupportChain` validates a CodeableConcept against a ValueSet binding successfully only if **every** coding is a member of the ValueSet.

Per FHIR spec, one matching coding is sufficient ([Terminology Bindings](https://hl7.org/fhir/R4/terminologies.html#strength)):

> one of the Coding values SHALL be from the specified value set

The HL7 reference validator accepts the concept as soon as one coding matches.

### Example

A `jurisdiction` with two codings:

- `urn:iso:std:iso:3166#DE` — member of the bound ValueSet
- a German DEÜV country code — not a member

`$validate` fails with:

```
None of the codes provided are in the value set 'Jurisdiction'
```

validator_cli accepts the same resource.

### Cause

`WorkerContextValidationSupportAdapter#validateCode` requires `validationResultsOk.size() == code.getCoding().size()` unless `ValidationSupportChain#setCodeableConceptValidationSuccessfulIfNotAllCodingsAreValid(true)` is set; default is `false`.

In addition, with the ANY behavior enabled, the adapter returns the first successful result and discards all issues collected for the codings that did not validate — including a code that does not exist in its CodeSystem, which the reference validator reports as an error even when another coding satisfies the binding (verified with validator_cli against tx.fhir.org).

### Proposal

- Flip the default to `true` so out-of-the-box behavior matches spec and reference validator
- Keep the issues of the non-matching codings on the successful result: pure ValueSet-membership findings become warnings, everything else (e.g. an unknown code in its CodeSystem) keeps its severity — an invalid sibling coding still fails validation, as it does with validator_cli
- Keep the flag for the stricter check
- Mention the change in the release notes

**Environment:** HAPI FHIR 8.10.0

patrick-werner:
Closes #8345.

Flips the default of `isCodeableConceptValidationSuccessfulIfNotAllCodingsAreValid` from `false` to `true`, so a CodeableConcept validated against a ValueSet binding is accepted when at least one coding is a member — matching the FHIR spec ("one of the Coding values SHALL be from the specified value set") and the HL7 reference validator. The stricter ALL-codings check stays available via `ValidationSupportChain#setCodeableConceptValidationSuccessfulIfNotAllCodingsAreValid(false)`.

With the ANY behavior, issues collected for the codings that did not validate were previously discarded entirely. They are now retained on the successful result: ValueSet-membership findings are downgraded to warnings, while other findings — such as a code that does not exist in its CodeSystem — keep their severity, so an invalid sibling coding still fails validation. This matches the reference validator (verified with validator_cli against tx.fhir.org).

- `IValidationSupport`: interface default flipped, javadoc updated
- `ValidationSupportChain`: field default flipped, javadoc updated
- `WorkerContextValidationSupportAdapter`: non-matching codings' issues retained (membership findings as warnings, everything else with original severity)
- Tests for the new default and both severity behaviors
- Changelog entry (8_14_0)

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
@@ -1638,24 +1638,26 @@ public interface IValidationSupport {
 	}
 
 	/**
-	 * When validating a CodeableConcept containing multiple codings, this method can be used to control whether
-	 * the validator requires all codings in the CodeableConcept to be valid in order to consider the
-	 * CodeableConcept valid.
+	 * When validating a CodeableConcept containing multiple codings against a ValueSet binding,
+	 * this setting controls whether all codings need to be members of the ValueSet in order to
+	 * consider the CodeableConcept valid, or whether a single matching coding is sufficient.
 	 * <p>
-	 * See VersionSpecificWorkerContextWrapper#validateCode in hapi-fhir-validation, and the refer to the values below
-	 * for the behaviour associated with each value.
+	 * See WorkerContextValidationSupportAdapter#validateCode in hapi-fhir-validation for the
+	 * behaviour associated with each value:
 	 * </p>
-	 * <p>
-	 *   <ul>
-	 *     <li>If <code>false</code> (default setting) the validation for codings will return a positive result only if
-	 *     ALL codings are valid.</li>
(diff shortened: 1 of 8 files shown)


tadgh:
Formatting check succeeded!

PR #8349, opened 2026-09-03
Making reproducer

jkiddo:
This pull request adds a new test to ensure that a CodeSystem stored in the database at a core R4 URL does not override the built-in definition during instance validation when database validation override is disabled. This addresses a bug where custom CodeSystems could unintentionally shadow core definitions.

Test coverage improvements:

* Added `CodeSystemDatabaseShadowingR4Test` to verify that a database CodeSystem at a core URL does not shadow the built-in CodeSystem during validation when `allowDatabaseValidationOverride` is false, reproducing and guarding against a known issue.


tadgh:
Formatting check succeeded!

PR #8351, opened 2026-09-03
fix(8350): Prevent stored CodeSystems from shadowing built-in core terminology when allowDatabaseValidationOverride=false

### LINKED ISSUE #8350: Storing a CodeSystem with a core canonical URL overrides built-in terminology in instance validation, despite allowDatabaseValidationOverride=false
Hi HAPI FHIR team.

I ran into a peculiar issue when I tried to install the SDC 4.0.0 IG and its dependencies: `hl7.fhir.uv.xver-r5.r4#0.1.0` and `hl7.fhir.uv.extensions.r4#5.3.0`.

**The Problem:**
After installing `hl7.fhir.uv.xver-r5.r4#0.1.0` I started getting validation errors when attempting to create a simple Questionnaire that conform to the R4 base profile:

```
Unable to validate code http://hl7.org/fhir/item-type#choice - Code is not found in CodeSystem: http://hl7.org/fhir/item-type
```
After a long debugging session and lots of head scratching, I finally figured out what was happening: The `hl7.fhir.uv.xver-r5.r4` contains https://hl7.org/fhir/uv/xver-r5.r4/0.1.0/CodeSystem-item-type.html which has same url `http://hl7.org/fhir/item-type` as the R4 core CodeSystem https://hl7.org/fhir/R4/codesystem-item-type.html (although different version).

**Analysis** 
There is a setting `JpaStorageSettings.setAllowDatabaseValidationOverride(boolean)` that is set to `false` by default, but it seems the setting is not respected during instance validation and ValueSet pre-expansion (although it _is_ respected when calling `CodeSystem/$validate-code` without a version).

The problem also leaks into ValueSet pre-expansion: Stored ValueSets that include codes from a core system that has been overridden produce expansions based on the database stored CodeSystem (`$expand` returns the diverging code set).

**Action**
I have produced a regression test and a fix and will open a PR shortly
Update: PR opened https://github.com/hapifhir/hapi-fhir/pull/8351

AI generated analysis that includes a (tested) reproducer:
-------
`JpaStorageSettings.setAllowDatabaseValidationOverride(boolean)` (since 8.8.0) is documented as:
> Sets whether database-stored validation resources should take precedence over built-in default profile validation resources.

With the default `false`, built-in definitions win on the resource-fetch path. However, a `CodeSystem` stored in the database at a canonical URL that the built-in definitions also define (e.g. `http://hl7.org/fhir/item-type`) still takes precedence during **instance validation**: The jpaserver's terminology service `TermReadSvcImpl.isCodeSystemSupported` finds the CodeSystem in the db and uses it during instance validation. If the content of the CodeSystem diverges from R4 core (as R5 terminology backports at R4-core URLs do), validation of perfectly valid R4 codes fails.

### To Reproduce
Note: Two things make this easy to misdiagnose:

1. `ValidationSupportChain`'s result cache (default 10 minutes) can mask the problem when attempting to reproduce: If the same validation ran before the CodeSystem was stored, the cached "valid" result keeps being served for up to 10 minutes afterwards.
2. The behaviour is inconsistent: type-level operations (`CodeSystem/$validate-code` without a version + `$lookup`) answer correctly from the built-in definition, while instance level `$validate` operations fail.

Verified using `docker run -p 8080:8080 hapiproject/hapi:latest` (HAPI FHIR 8.10.0, R4, default configuration).

**Step 1 — baseline: a valid R4 Questionnaire validates cleanly.**

```bash
curl -s -X POST 'http://localhost:8080/fhir/Questionnaire/$validate' \
  -H 'Content-Type: application/fhir+json' \
  -d '{"resourceType":"Questionnaire","status":"active","item":[{"linkId":"1","text":"Pick one","type":"choice"}]}'
```

→ only the dom-6 best-practice warning; `choice` is a valid R4 `item-type` code. (Note: this call also warms the validation cache — see step 3.)

**Step 2 — store a diverging CodeSystem at the built-in canonical URL.** The core URL is the essential ingredient: the bug is that database content *shadows a core-defined system*. This mirrors what installing `hl7.fhir.uv.xver-r5.r4#0.1.0` does — it ships R5 terminology backports at R4-core URLs, and R5 removed `choice` from `item-type`:

```bash
curl -s -X POST 'http://localhost:8080/fhir/CodeSystem' \
  -H 'Content-Type: application/fhir+json' \
  -d '{"resourceType":"CodeSystem","url":"http://hl7.org/fhir/item-type","version":"5.0.0",
       "name":"QuestionnaireItemTypeR5Backport","status":"active","content":"complete",
       "concept":[{"code":"group"},{"code":"display"},{"code":"question"}]}'
```

**Step 3 — repeat the validation from step 1.**

Actual result:

```json
{ "severity": "error",
  "diagnostics": "Unable to validate code http://hl7.org/fhir/item-type#choice - Code is not found in CodeSystem: http://hl7.org/fhir/item-type" }
```

⚠️ If step 3 runs within ~10 minutes of step 1, the **cached** step-1 result may still be returned as "valid" — wait out `ValidationSupportChain`'s default 10-minute cache, restart the server, or skip step 1 entirely (on a fresh server the failure is immediate on the first validation).

Expected result (reading the `allowDatabaseValidationOverride=false` contract literally): The built-in R4 `item-type` definition — which contains `choice` — should take precedence, and validation should keep passing.

### Analysis (based on v8.12.0 code)

- `ValidationSupportChain.validateCode(...)` (~line 811) walks the chain and returns the outcome of the first module that **claims** the code system and returns a non-null result.
- `DefaultProfileValidationSupport` implements neither `isCodeSystemSupported` nor `validateCode` — it serves resources but can never *answer* a code-validation question. Its position in the chain (first, with `allowDatabaseValidationOverride=false`) is therefore irrelevant to this path.
- `TermReadSvcImpl.isCodeSystemSupported(...)` claims **any** system with a current `TermCodeSystemVersion` in the terminology tables. Its `validateCode` then resolves the code via `findCode(...)` against that stored content and, on a miss, returns the error above (`TermReadSvcImpl#createMessageAppendForCodeNotFoundInCodeSystem`). `InMemoryTerminologyServerValidationSupport` — which would have validated correctly against the built-in CodeSystem — is never consulted.
- `allowDatabaseValidationOverride` only permutes `[DefaultProfileValidationSupport, TermReadSvc, JpaPersistedResourceValidationSupport]` in `JpaValidationSupportChain.postConstruct`; it cannot influence which module answers `validateCode`.

### Possible remedies

1. **Honor the setting on the validation path**: when `allowDatabaseValidationOverride=false`, `TermReadSvc` declines to claim code systems whose URL the built-in definitions can serve (`DefaultProfileValidationSupport.fetchCodeSystem(url) != null`). This keeps `TermReadSvc` authoritative for externally loaded terminology (LOINC/SNOMED, `content=not-present`/fragment uploads, cf. #7961 and #7798) while restoring built-in precedence exactly where built-ins have content. Deployments that deliberately override built-ins can keep doing so with `allowDatabaseValidationOverride=true`.
2. If current behavior is intended, a javadoc/documentation note on `setAllowDatabaseValidationOverride` (that it governs resource fetching only, not code validation against terminology-table content) would prevent others from relying on it for this scenario.

ttnTrifork:
Make TermReadSvcImpl respect when JpaStorageSettings#isAllowDatabaseValidationOverride() is `false` - which is the default setting.

This fixes #8350

### DIFF
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8350-core-terminology-shadowed-by-database-codesystems.yaml
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/8350-core-terminology-shadowed-by-database-codesystems.yaml
@@ -1,0 +1,13 @@
+---
+type: fix
+issue: 8350
+title: "Previously, a CodeSystem stored in the JPA server with same canonical URL as a built-in FHIR definition was used for code validation and for
+ValueSet pre-expansion even when JpaStorageSettings#isAllowDatabaseValidationOverride() was false. 
+This meant that if the stored CodeSystem diverges from the built-in (as happens when installing e.g. hl7.fhir.uv.xver-r5.r4 that contains R5 terminology 
+that exist at same core canonical URLs as in R4 (e.g. http://hl7.org/fhir/item-type),
+then instance validation and pre-expanded ValueSets were based on the stored backported R5 CodeSystem rather on the built-in R4 version. 
+This has been fixed such that when allowDatabaseValidationOverride is false (which is the default), the JPA terminology
+service no longer answers for code systems that is part of the core specification namespace (http://hl7.org/fhir/*) of the active FHIR version the server is running (e.g. R4). 
+Terminology stored at other URLs (e.g. LOINC, SNOMED CT, terminology.hl7.org content), 
+or are not defined in the active FHIR context (e.g. the R5 CodeSystem http://hl7.org/fhir/version-algorithmon an R4 server), are unaffected. 
+Fix submitted by Tue Toft Nørgård (@ttnTrifork)"
--- a/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/introduction/changelog.md
+++ b/hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/docs/introduction/changelog.md
@@ -1,4 +1,0 @@
-# Changelog: 2025
-
-<th:block th:insert="~{fragment_changelog.md :: changelog('2025', '')}"/>
-
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
@@ -22,6 +22,7 @@ package ca.uhn.fhir.jpa.term;
 import ca.uhn.fhir.context.FhirContext;
 import ca.uhn.fhir.context.FhirVersionEnum;
 import ca.uhn.fhir.context.support.ConceptValidationOptions;
+import ca.uhn.fhir.context.support.DefaultProfileValidationSupport;
 import ca.uhn.fhir.context.support.IValidationSupport;
 import ca.uhn.fhir.context.support.LookupCodeRequest;
 import ca.uhn.fhir.context.support.ValidationSupportContext;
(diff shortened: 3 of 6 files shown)


tadgh:
Formatting check succeeded!

ttnTrifork:
Hmm according to Copilot, the build error is not related to my changes, but happens because the Maven surefire plugin version is not pinned in hapi-fhir-jpaserver-searchparam/pom.xml

PR #8352, opened 2026-09-03
Fail the SNAPSHOT deploy job if the version is not a SNAPSHOT

jdar8:
SNAPSHOT pipelines never verified whether the artifact it was about to publish was a SNAPSHOT or not. This means if the version was, for example `8.12.0`, the pipeline would publish this full release.

This PR fixes this by adding a check for the `-SNAPSHOT` suffix before proceeding with the pipeline. If it is absent the pipeline now exits and logs an error message.

### DIFF
--- a/snapshot-deploy-job.yml
+++ b/snapshot-deploy-job.yml
@@ -39,6 +39,22 @@ jobs:
         displayName: 'Free up host disk space before build'
         target: host
 
+      # This job deploys whatever version the branch currently declares, and Maven decides
+      # release-vs-snapshot from the version string alone. If a release branch has already been
+      # bumped to a non-SNAPSHOT version, a scheduled run here would publish a real, immutable
+      # release to Maven Central rather than a snapshot. Fail fast instead.
+      - bash: |
+          version=$(mvn -q -N help:evaluate -Dexpression=project.version -DforceStdout)
+          echo verifying that project version $version is a SNAPSHOT before deploying...
+          if [[ "$version" == *-SNAPSHOT ]]
+          then
+            echo "version $version is a SNAPSHOT, proceeding..."
+          else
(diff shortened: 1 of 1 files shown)


tadgh:
Formatting check succeeded!

dotasek (approved):
I copied the script here onto my own machine and ran it in two states:

Once with the master branch as `8.13.2-SNAPSHOT`, which resulted in:

```
Project version: 8.13.2-SNAPSHOT
Version is a SNAPSHOT, proceeding with deploy.
```

And once after removing the snapshot with `mvn versions:set -DnewVersion=8.13.2`, which resulted in:

```
Project version: 8.13.2
##vso[task.logissue type=error]Refusing to deploy non-SNAPSHOT version '8.13.2' from the SNAPSHOT pipeline. Releases must go through the release pipeline.
```

The latter showed up in the console as a failed execution (exit code != 0) which from the PR description appears to be the desired behaviour.

snapshot-deploy-job.yml
```diff
@@ -39,6 +39,22 @@ jobs:
         displayName: 'Free up host disk space before build'
         target: host
 
+      # This job deploys whatever version the branch currently declares, and Maven decides
+      # release-vs-snapshot from the version string alone. If a release branch has already been
+      # bumped to a non-SNAPSHOT version, a scheduled run here would publish a real, immutable
+      # release to Maven Central rather than a snapshot. Fail fast instead.
+      - bash: |
```
jmarchionatto, line 46: suggestion: adding `-e` would, on error, abort the step right away with Maven's own error on the log, which is the desired diagnosis.  

```bash
  - bash: |                     
      set -e                           
      version=...
```

jmarchionatto (approved):

PR #8355, opened 2026-09-05
[Draft ]Fix ValueSet pre-expansion runs on a node that has not stored the terminology - failing tests

### LINKED ISSUE #8354: ValueSet pre-expansion runs on a node that has not stored the terminology (deferred concepts are node-local)
## Summary

On a clustered server a `ValueSet` can be pre-expanded while the `CodeSystem` it references is still being written to the database, because the check meant to prevent that only sees the node it runs on. The expansion is stored with a subset of the codes it should contain, and marked `EXPANDED`.

**Waiting does not fix it.** When the terminology finishes loading, nothing re-runs the expansion. The incomplete result is served by `$expand` and `$validate-code` indefinitely, across restarts, until someone explicitly invalidates it. Nothing in the logs or the `ValueSet`'s status says it happened.

Follow-up to #8321. Unlike #8321, this is not a regression introduced by the move of terminology operations to batch2 (#8127). It is a newly discovered bug, present before that work and unaffected by it.

## Background: how concepts are stored, and why not all at once

Storing a `CodeSystem` resource larger than `deferIndexingForCodesystemsOfSize` (default 100 concepts) does not write all of its concepts in the request's transaction. `TermCodeSystemStorageSvcImpl.persistChildren()` hands the excess to `ITermDeferredStorageSvc.addConceptToStorageQueue()`, and the request returns with the `CodeSystem` only partially stored.

The rest is written afterwards, in the background. `TermDeferredStorageSvcImpl` keeps the pending concepts in ordinary `java.util.List` fields — `myDeferredConcepts`, `myConceptLinksToSaveLater` and their siblings — and its `scheduleJobs()` registers a *local* job that calls `saveDeferred()` every 5 seconds. Each `saveDeferred()` makes up to ten passes of `processDeferredConcepts()`, and each pass writes at most 1000 concepts. So one node works through roughly 10,000 concepts every 5 seconds, and a `CodeSystem` of that size stays partially stored for several seconds after the request that created it has returned.

`isStorageQueueEmpty(..)` answers whether those lists are empty. Because the lists are fields on a Spring singleton, and there is one instance of that singleton per node, the answer describes the node the call runs on and nothing else.

## Problem

`Step1InitiateJob.run()` calls `isStorageQueueEmpty(false)` before pre-expanding, so that a `ValueSet` is not expanded against a `CodeSystem` that is still being written.

That job is a batch2 job, and batch2 hands work to whichever node picks it up. `BaseBatch2Config` publishes work notifications through `batch2ProcessingChannelProducer` and consumes them through `batch2ProcessingChannelConsumer`, both on the channel named by `BaseBatch2Config.CHANNEL_NAME` (`batch2-work-notification`). When that channel is backed by a real broker, every node of the cluster is a consumer, so a job started on node A may execute on node B.

Node B then asks its own `TermDeferredStorageSvcImpl` whether anything is pending. Node A holds the concepts, so node B's lists are empty, `isStorageQueueEmpty(false)` returns `true`, and `Step1InitiateJob` proceeds. The `ValueSet` is expanded from whatever concepts happen to be in the database at that moment, and the result is stored with status `EXPANDED`.

### Waiting for the terminology to finish does not fix it

The window in which this can be triggered is short: only while the remaining concepts are being written. But once a short expansion has been stored, no amount of waiting corrects it.

- Only two things start a `PRE_EXPAND_VALUESET` job: the `afterCommit` hook in `TermValueSetStorageSvcImpl`, which runs when a `ValueSet` resource is stored, and `$invalidate-expansion` in `ValueSetOperationProvider`. `TermDeferredStorageSvcImpl` does not reference pre-expansion at all, so finishing the deferred work starts nothing.
- `invalidatePreCalculatedExpansionOfValueSetsContainingCodeSystem(..)`, which resets dependent `ValueSet`s to `NOT_EXPANDED`, has one caller: `JpaResourceDaoCodeSystem.updateEntity(..)`. It therefore runs only when the `CodeSystem` **resource** is updated. Writing the deferred concepts goes through `TermConceptDaoSvc.saveConcept(..)`, which is not a resource update, so it invalidates nothing.

The row keeps its `EXPANDED` status and its expansion timestamp, and both `$expand` and `$validate-code` go on serving the short list. Recovering takes a deliberate act: `$invalidate-expansion` on the `ValueSet`, or re-saving the `ValueSet` resource. Re-saving the `CodeSystem` resource is a trap — it invalidates dependent expansions to `NOT_EXPANDED` but starts no job, so it removes the wrong answer without producing a right one.

## Conditions required

All four must hold:

1. A cluster whose batch2 channel is backed by a real broker. One node, or an in-memory channel, is not affected.
2. A `CodeSystem` with **nested child** concepts, larger than `deferIndexingForCodesystemsOfSize`, stored as a **FHIR resource**.
   - Only `persistChildren()` calls `addConceptToStorageQueue()`, so a `CodeSystem` with a flat list of concepts never defers anything, whatever its size.
   - `$upload-external-code-system` does not defer either: it goes through `addConceptsToCodeSystemVersion()` → `addConceptInHierarchy()` → `saveConcept()`, writing as it goes. LOINC and SNOMED CT loaded that way are not affected.
3. A `ValueSet` that references the `CodeSystem` is created or activated during the seconds it takes the background job to write the remaining concepts.
4. The pre-expansion job is picked up by a node other than the one that stored the `CodeSystem` — on an N-node cluster, roughly (N-1)/N of the time.

Conditions 1, 3 and 4 are ordinary in a cluster. Condition 2 is the one that rarely holds, and is what makes this rare rather than routine.

## Scenario

A three-node deployment behind a load balancer, with Kafka backing the batch2 channel. An implementation guide is installed whose package contains a hierarchical `CodeSystem` of about 8,000 concepts nested under a handful of roots, plus the `ValueSet`s that draw from it.

1. The install request lands on **node A**. Storing the `CodeSystem` writes the roots and the first 100 children in the request's transaction, and leaves the remaining 7,900 or so on node A's in-memory list. At 10,000 concepts per 5 seconds, node A needs a few seconds to finish.
2. Within milliseconds, the same install stores an ACTIVE `ValueSet` that includes the `CodeSystem`. Committing that resource starts a `PRE_EXPAND_VALUESET` job.
3. Kafka delivers the work notification to **node B**, whose own list is empty. The check passes and the expansion proceeds.
4. The expansion resolves the `ValueSet`'s compose rules against the concepts present in the database at that moment, stores that subset, and sets the status to `EXPANDED`.

**Impact.** The `ValueSet` is now missing codes, with nothing to indicate it, and by the section above it stays that way. `$validate-code` reports codes that do belong to the `ValueSet` as not present. On a server with validation enabled, requests carrying such a code are rejected, naming a code that is visible in the server's own `CodeSystem`. `$expand` returns fewer codes than it should, so terminology-driven pickers are missing options. Nothing fails at install time, and the problem surfaces later as rejected data, which makes it expensive to trace back.

## Failing test

PR #8355, which also carries the proposed fix. It stands alone: it does not depend on the #8321 fix in PR #8335, and fails on `master` today.

`TermValueSetPreExpansionLifecycleR4Test#isStorageQueueEmpty_conceptStorageDeferredOnAnotherNode_reportsQueueNotEmpty`

A JPA test cannot start a second node, and does not need to: node B differs from node A only in that it reads the same database while holding its own empty lists. The test stores a `CodeSystem` larger than `deferIndexingForCodesystemsOfSize`, so concepts are left pending and the database holds a partial load, then puts the same readiness question to a second `TermDeferredStorageSvcImpl`. That instance is built through Spring's `AutowireCapableBeanFactory`, so it is wired against the same DAOs, but it is not registered as a bean, so `BaseSchedulerServiceImpl` never finds it and it writes nothing of its own.

```java
ITermDeferredStorageSvc otherNode = newDeferredStorageSvcOnAnotherNode();

assertFalse(otherNode.isStorageQueueEmpty(false),
    "A node that did not store the CodeSystem still has to see that its concepts are not fully stored");
```

It reports nothing pending, because the only record that concepts are still being written is node A's memory. The test is expected to fail until that fact is recorded somewhere both nodes can read.

codeforgreen:
Fixes #8354

Carries the failing test only. The fix needs a schema change; the proposal is below.

## What is broken

Pre-expansion has to know whether the `CodeSystem` it is about to expand is fully stored. The only signal available is `ITermDeferredStorageSvc.isStorageQueueEmpty(..)`, and the service behind it keeps its pending concepts in memory, with one instance per node. The answer therefore describes the node the call runs on and nothing else.

Batch2 runs a job on whichever node picks it up, so pre-expansion can run on a node other than the one that stored the `CodeSystem`. That node has nothing pending, the check reports ready, and the `ValueSet` is expanded from a partially stored `CodeSystem` and marked `EXPANDED`. Nothing re-runs it afterwards, so the incomplete expansion is served until someone calls `$invalidate-expansion`. Full analysis in #8354.

## The test

`TermValueSetPreExpansionLifecycleR4Test#isStorageQueueEmpty_conceptStorageDeferredOnAnotherNode_reportsQueueNotEmpty`

A JPA test cannot start a second node and does not need to: the other node differs only in that it reads the same database while holding its own empty lists. The test stores a `CodeSystem` larger than `deferIndexingForCodesystemsOfSize`, so concepts are left pending and the database holds a partial load, then puts the readiness question to a second `TermDeferredStorageSvcImpl`. That instance is built through Spring's `AutowireCapableBeanFactory`, so it is wired against the same DAOs, but it is not registered as a bean, so `BaseSchedulerServiceImpl` never finds it and it writes nothing of its own.

It reports nothing pending, because the only record that concepts are still being written is the storing node's memory. Expected to fail until that fact is recorded where every node can read it.

The test stands alone: it does not depend on the #8321 fix in PR #8335, and fails on `master` today.

## Proposed fix

**Preferred — record load state in the database.** `TermCodeSystemVersion` has no field saying that concept loading is in progress. Add one, for example a `CS_LOAD_STATE` column on `TRM_CODESYSTEM_VER`, set when concepts are first deferred and cleared when the last one is written. `Step1InitiateJob` reads that instead of the in-memory lists, which makes the check correct on any node. This needs a `MigrationTask`, which is why it could not be part of #8335.

**Alternative — persist the pending concepts.** Store the queue in the database rather than in memory. A larger change, and it also fixes a separate problem: a restart currently discards whatever is pending.

**Not viable — ask every node.** Polling each node for the state of its lists adds a cluster round trip to a hot path, and still races.

The fix in #8335 rules out none of these: replacing the condition with one read from the database is a one-line change at the same call site in `Step1InitiateJob`.


tadgh:
Formatting check succeeded!

PR #8356, opened 2026-09-07
Fix mass ingestion mode with Patient ID Partitioning

nathandoef:
When a `PatientCompartmentEnforcingInterceptor` was registered with mass ingestion mode enabled, resource 
updates could fail with a NullPointerException (`HAPI-2223: theResource must not be null`). This affected registered 
subclasses of the interceptor and the history rewrite path. This has been fixed.

### DIFF
--- a/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
+++ b/hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
@@ -29,6 +29,7 @@ import ca.uhn.fhir.context.RuntimeResourceDefinition;
 import ca.uhn.fhir.i18n.Msg;
 import ca.uhn.fhir.interceptor.api.HookParams;
 import ca.uhn.fhir.interceptor.api.IInterceptorBroadcaster;
+import ca.uhn.fhir.interceptor.api.IInterceptorService;
 import ca.uhn.fhir.interceptor.api.Pointcut;
 import ca.uhn.fhir.interceptor.model.RequestPartitionId;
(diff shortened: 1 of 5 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
```diff
@@ -1337,7 +1349,7 @@ public IBasePersistedResource<?> updateHistoryEntity(
 			ResourceTable entity = (ResourceTable) theEntity;
 
 			IBaseResource oldResource;
-			if (getStorageSettings().isMassIngestionMode()) {
+			if (getStorageSettings().isMassIngestionMode() && !isPatientCompartmentEnforcingInterceptorRegistered()) {
```
codeforgreen, line 1352: the fix looks good. But this makes me wonder if there are other interceptors that need to be kept even in mass ingestion mode. For example validations such as `PatientLockingInterceptor` and the custom one ending `PatientIdentifierPreResolutionInterceptor`. Maybe we need  a mechanism to mark interceptors as by-passable or non-bypassable in mass ingestion?

codeforgreen (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/dao/BaseHapiFhirDao.java
nathandoef: That's a good point. Going to add a tech debt ticket to look into this more.

PR #8367, opened 2026-09-08
Evaluate custom concept-property filters on content=complete CodeSystems

### LINKED ISSUE #8366: In-memory ValueSet expansion cannot evaluate custom concept-property filters on content=complete CodeSystems
### Summary

Validating a code against a ValueSet whose `compose.include.filter` references a **custom**
concept property (e.g. LOINC `SCALE_TYP = Doc`, as used by the core ValueSet
`http://hl7.org/fhir/ValueSet/doc-typecodes`) fails with:

    HAPI-3004: In-memory ValueSet expansion does not support filter with property 'SCALE_TYP' and operator '='

even when the CodeSystem is supplied inline with `content: complete` — i.e. when every input
needed to evaluate the filter is already in memory.

### Steps to reproduce

1. Register a `http://loinc.org` CodeSystem with `content: complete`:

```json
{
  "resourceType": "CodeSystem",
  "url": "http://loinc.org",
  "status": "active",
  "content": "complete",
  "caseSensitive": true,
  "concept": [
    { "code": "29550-1", "display": "Glucose [Mass/volume] in Serum or Plasma" }
  ]
}
```

2. Register a ValueSet that filters that system by `SCALE_TYP = Doc` (the shape of the shipped
   `doc-typecodes` ValueSet):

```json
{
  "resourceType": "ValueSet",
  "url": "http://example.org/fhir/ValueSet/doc-typecodes",
  "status": "active",
  "compose": {
    "include": [
      {
        "system": "http://loinc.org",
        "filter": [
          { "property": "SCALE_TYP", "op": "=", "value": "Doc" }
        ]
      }
    ]
  }
}
```

3. Validate the code against the ValueSet:

```
POST [base]/ValueSet/$validate-code
```
```json
{
  "resourceType": "Parameters",
  "parameter": [
    { "name": "url",    "valueUri":  "http://example.org/fhir/ValueSet/doc-typecodes" },
    { "name": "system", "valueUri":  "http://loinc.org" },
    { "name": "code",   "valueCode": "29550-1" }
  ]
}
```

Observed: `HAPI-3004`. Expected: the filter is evaluated in memory.
(The same surfaces via `Composition.type`, whose preferred binding is `doc-typecodes`, during
`Bundle/$validate`.)

### Root cause

`ValueSetExpansionFilterContext` only indexes/evaluates `code`, `display`, and a fixed set of
standard concept-properties (`buildIndexes`). A custom property such as `SCALE_TYP` reaches the
blanket reject in `passesFilter` and throws `UnsupportedFilterException` (`Msg.code(3004)`).

This is **not** a request to revert the change that surfaced this (the previous behaviour
fabricated a "not a member" result for a filter that was never evaluated, producing an empty
expansion reported as `not-in-vs`, which `InstanceValidator.isIgnoredTxIssueType` discards
unconditionally — so a check that never ran silently "passed"). Making the failure explicit was
correct. This issue asks to **implement the missing capability**.

### Why this belongs in the in-memory support

`content: complete` asserts the resource is self-contained: the ValueSet (shipped in
`DefaultProfileValidationSupport`), the inline CodeSystem, and the concept/properties are all
present. No terminology backend is involved or required — the DB-backed expander is inapplicable
because a core-spec ValueSet is never stored (`TermReadSvcImpl.isValueSetSupported`). The reference
implementation already evaluates custom property filters in memory
(`org.hl7.fhir.r5.terminologies.validation.ValueSetValidator#isDefinedProperty` /
`#codeInPropertyFilter`); HAPI's filter context is strictly less capable here.

### Scope

Applies **iff** `CodeSystem.content == complete`. That mode makes in-memory evaluation both
*possible* (all concepts present) and *sound* (a missing property value is meaningful). Behaviour
for `fragment` / `example` / `not-present` / unset is unchanged and out of scope.
(`InMemoryTerminologyServerValidationSupport` currently only distinguishes `not-present`; this
requires distinguishing `complete` specifically.)

### Two determinations that must be kept apart

- **Undetermined** — the property is unknown to the code system; membership can be neither
  established nor refuted → issue type **`not-found`**. Not on the `isIgnoredTxIssueType` discard
  list, so it always surfaces; severity is recalculated from binding strength (`ERROR` for
  `required`, `WARNING` for `extensible`/`preferred`). Must **not** be reported as `not-in-vs`
  (discarded unconditionally) or `vs-invalid` (fatal at every strength).
- **Determined negative** — the property resolved and the concept is not a member → **`not-in-vs`**.

### Expected behaviour (validating `loinc|29550-1` against filter `SCALE_TYP = Doc`, content=complete)

| | `CodeSystem.property[]` declares SCALE_TYP | concept's SCALE_TYP value | Expected |
|---|---|---|---|
| A  | no  | absent  | issue `not-found`, ERROR, returned to caller |
| B1 | yes | `"Doc"`   | no issue — validates |
| B2 | yes | `"wrong"` | `not-in-vs` — fails |
| C  | yes | absent    | `not-in-vs` — fails |
| D1 | no  | `"Doc"`   | no issue — validates |
| D2 | no  | `"wrong"` | `not-in-vs` — fails |

**Req 1** — the filter must actually be evaluated per `ConceptSetFilterComponent.op` (today B1/B2/D1/D2 are indistinguishable, all `HAPI-3004`).
**Req 2** — a concept's property value must be used whether or not `CodeSystem.property[]` declares it (scenario D); no invariant requires the declaration. Intentionally more permissive than the R5 reference, which gates on `isDefinedProperty`.
**Req 3** — scenario A must be reported as `not-found` and must reach the client (precedent: `Msg.code(2646)` already reports `not-found` for the `content: not-present` filter case).

jorgepinilla-smile:
Fixes #8366.

## Problem

In-memory ValueSet expansion never evaluated a filter that selects on a custom
`CodeSystem.concept.property`, such as the LOINC `SCALE_TYP` filter the core `doc-typecodes`
ValueSet uses. It returned a hard failure instead, with `HAPI-3004`. But `content: complete` asserts
that the CodeSystem carries every concept it defines, so when such a CodeSystem is supplied inline
and the concept carries the property value, the expansion already holds every input the filter
needs. The evaluation is feasible without any terminology backend, and is what should happen in
place of a rejection.

This is not a regression in capability. Before 8.12 the same case returned a fabricated "not a
member" for a filter that was never evaluated; the resulting empty expansion was reported as
`not-in-vs`, which the validator discards unconditionally. As a result, a check that never ran
silently passed. #8181 replaced that with an explicit failure. This PR implements the missing
capability rather than reverting it.

## Change

All of it applies only when `CodeSystem.content == complete`.

**`ValueSetExpansionFilterContext`**

- Indexes custom `CodeSystem.concept.property` values for the property codes the ValueSet filters
  reference.
- Evaluates the filter for operation (`ConceptSetFilterComponent.op`) `=`, `in`, `not-in` and
  `regex`. The structural operators (`is-a`, `descendent-of`, `child-of`, …) resolve against concept
  codes and remain unsupported.
- Adds `UndeterminedFilterException`, raised when membership can be neither established nor ruled
  out.

**`InMemoryTerminologyServerValidationSupport`**

- Maps `UndeterminedFilterException` to a `not-found` issue (`Msg.code(3049)`), beside the existing
  `UnsupportedFilterException` → `vs-invalid` mapping.

## Behaviour

Validating `http://loinc.org|29550-1` against a filter of `SCALE_TYP = Doc`, with
`content: complete`:

| `CodeSystem.property[]` | concept's value | determination | result |
|---|---|---|---|
| declares `SCALE_TYP` | `"Doc"` | matches | validates |
| declares `SCALE_TYP` | `"wrong"` | does not match | `not-in-vs` |
| declares `SCALE_TYP` | absent | no value, so not a member | `not-in-vs` |
| undeclared | `"Doc"` | matches | validates |
| undeclared | `"wrong"` | does not match | `not-in-vs` |
| undeclared | absent | cannot be made | `not-found` |

Before this change every row produced `HAPI-3004`. Five of them now reach a real determination; the
last still cannot be determined, and what changes there is how that is reported. It previously
produced `vs-invalid`, which is fatal at every binding strength and blocks a client on a condition
it cannot remedy. It now produces `not-found`, whose severity gets recalculated from the binding
strength: `WARNING` for `extensible` and `preferred`, and `ERROR` for `required` when the issue
concerns an HL7 system or `unknownCodeSystemsCauseErrors` is set. A `preferred` binding is therefore
no longer blocked, but the client is still told the check could not be completed.

## Testing

`InMemoryTerminologyServerValidationSupportTest` covers the six rows end to end through
`validateCodeInValueSet`; `ValueSetExpansionFilterContextTest` covers the filter context directly,
including the other operators and a guard proving the gate is on the content mode rather than on the
data being present. Full `hapi-fhir-validation` module: 1190 tests green.


tadgh:
Formatting check succeeded!

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValueSetExpansionFilterContext.java
```diff
@@ -598,6 +722,82 @@ private static boolean parseRequiredBoolean(ValueSet.ConceptSetFilterComponent t
...
+
+		Set<String> values = myCodeSystemCustomPropertyIndex
+				.getOrDefault(property, Collections.emptyMap())
+				.getOrDefault(conceptCode, Set.of());
+
+		if (values.isEmpty()) {
+			if (myCodeSystemDeclaredCustomProperties.contains(property)) {
+				// 'complete' makes the absence meaningful: the concept has no value, so it is not a member.
+				// This holds for every operator, including the negative ones - a concept with nothing to
+				// compare is not admitted by a filter it was never measured against.
```
TipzCM, line 765: I'm guessing this is here because (as stated above), custom codes that are absent are different than those that are skipped; but that this is only true of *custom* codes.

If so, i think the comment should include this. Because the 'custom code' part is what's relevant.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValueSetExpansionFilterContext.java
```diff
@@ -695,4 +895,23 @@ public UnsupportedFilterException(String theMessage) {
...
+
+	/**
+	 * Thrown when a ValueSet filter references a concept property that cannot be resolved for the concept
+	 * under evaluation, so membership can be neither established nor refuted (an <em>undetermined</em>
+	 * result, as opposed to a determined negative). This is distinct from {@link UnsupportedFilterException}:
+	 * the property/operator combination is one the in-memory expansion could evaluate, but the data needed to
+	 * do so is absent. Callers should surface this as a {@code not-found} issue (recalculated by binding
+	 * strength) rather than a fatal {@code vs-invalid}, so an undetermined check is never silently dropped.
+	 */
+	public static class UndeterminedFilterException extends RuntimeException {
```
TipzCM, line 909: does this have to be a subclass here? can it not be its own class?

This file is already very long, so embedding it is not actually helping discoverability.

TipzCM (approved):

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValueSetExpansionFilterContext.java
```diff
@@ -43,37 +48,85 @@ public class ValueSetExpansionFilterContext {
...
+	// that is the ValueSet.compose.include.filter (or compose.exclude.filter) entries being applied.
+	//
+	// Indexing walks every CodeSystem.concept.property and keeps only those whose code appears in this set,
+	// storing them in myCodeSystemCustomPropertyIndex; the rest are skipped to avoid indexing properties that
+	// can never be accessed.
+	//
+	// Mirrors myValueSetStandardPropertiesUsedInFilters, but stores codes verbatim: custom property codes are
+	// case-sensitive
+	// and must not be lowercased the way the standard-property dispatch does.
+	private final Set<String> myValueSetCustomPropertiesUsedInFilters = new HashSet<>();
```
codeforgreen, line 97: Consider trimming down or removing javadoc for private fields added to thsi class. They should be self-explanatory by behaviour.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValueSetExpansionFilterContext.java
epeartree: it could but it would be just an empty marker class.  it was added here since UnsupportedFilterException was already declared above.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValueSetExpansionFilterContext.java
epeartree: usually, i would not add such lengthy comments on declarations but since the processing in this class is quite complex to grasp, i felt like it was better to have more than not enough.

PR #8371, opened 2026-09-09
Preserve extensions on a valueless repeating primitive (#8370)

### LINKED ISSUE #8370: Extensions on a repeating primitive with no value are silently dropped when parsing JSON
### Describe the bug

An extension on a **repeating** primitive that carries no value is silently dropped when parsing JSON.

FHIR represents a repeating primitive as two parallel arrays, `foo` and `_foo`, padded with `null`. When none of the repetitions has a value, some tooling emits only the `_foo` array (see FHIR/sushi#1631 for one producer of this shape). HAPI then discards the extensions entirely.

The **non-repeating** equivalent (`"_family": { ... }`, an object) has always parsed correctly. Only the array form is affected.

### To Reproduce

```java
FhirContext ctx = FhirContext.forR4();

String input = """
	{
	  "resourceType": "Patient",
	  "address": [ {
	    "_line": [ {
	      "extension": [ { "url": "http://example.org/line", "valueString": "line-value" } ]
	    } ]
	  } ]
	}""";

Patient p = ctx.newJsonParser().parseResource(Patient.class, input);
System.out.println(p.getAddressFirstRep().getLine().size());   // prints 0
```

### Expected behavior

One `Address.line` repetition with no value but with the extension preserved, re-encoded in the spec-conformant null-padded form:

```json
"address": [ { "line": [ null ], "_line": [ { "extension": [ ... ] } ] } ]
```

i.e. HAPI normalises the input to the representation the FHIR JSON rules call for, rather than dropping data.

### Root cause

`JsonParser.parseChildren(...)` has a branch for an `_`-prefixed element that has no corresponding value element. It handles `nextValue.isObject()`, and sends everything else to `getErrorHandler().incorrectJsonType(...)` — which the default `LenientErrorHandler` logs and discards. An orphan alternate that is an **array** therefore never reaches `parseAlternates(...)`.

There is a matching gap on the RDF side. In `RDFParser`, the `PRIMITIVE_DATATYPE` case guards with `if (value != null || !hasNoExtensions(pd))`, but every statement inside is nested in a further `if (value != null)`, so a valueless primitive carrying extensions produces no RDF output at all.

That second gap is not observable today, because such elements never survive JSON parsing in the first place. Fixing the JSON parser exposes it: 9 `RDFParserTest.testRDFRoundTrip` cases begin to fail, e.g. `activitydefinition-administer-zika-virus-exposure-assessment.json`, which contains

```json
"timingTiming": { "_event": [ { "extension": [ { "url": ".../cqf-expression", ... } ] } ] }
```

### Real-world impact

We hit this with SDC template-based `$extract` Questionnaires, where `sdc-questionnaire-templateExtractValue` extensions sit on `Address.line` and `HumanName.given` of a contained template Bundle. Storing and reading back `ChEkmQuestionnaireGonorrhoea` loses 4 of its 51 directives, so `$extract` silently stops populating address lines and given names. Nothing in the server logs indicates data was discarded.

### Environment

- HAPI FHIR 8.13.5-SNAPSHOT (master @ d5841d0d6ce), also reproduced on 8.8.0
- R4, `LenientErrorHandler` (the default)

### Related

- #8238 / #8260 — a different bug in the same area. #8260 is a **precondition** for reproducing the full scenario above: without it, encoding a resource with a contained Bundle that has an extension on a root-level primitive throws `HAPI-2223` before the extension loss can be observed. The two fixes are independent, but only both together make the SDC questionnaires above round-trip losslessly (verified: 51 in / 51 out, and 107 in / 107 out for `ChEkmQuestionnaireMpox`).
- FHIR/sushi#1631 — the producer side of the non-null-padded shape.

oliveregger:
Fixes #8370.

## What

An extension on a **repeating** primitive that carries no value is emitted by some tooling as an `_element` array with no corresponding value array:

```json
"address": [ { "_line": [ { "extension": [ { "url": "...", "valueString": "..." } ] } ] } ]
```

`JsonParser` handled the **non-repeating** (object) form of such an orphan alternate — `"_family": { "extension": [...] }` — but sent the array form to `ErrorHandler.incorrectJsonType(...)`, so the default `LenientErrorHandler` logged it and discarded the extensions. Each entry of the array is now treated as one repetition of the primitive. An element is entered for every index, including `null` entries, so extensions stay aligned with the repetitions they belong to.

The parsed resource re-encodes to the spec-conformant null-padded form, so HAPI normalises the input rather than dropping it:

```json
"address": [ { "line": [ null ], "_line": [ { "extension": [ ... ] } ] } ]
```

## Why RDFParser is also touched

`RDFParser`'s `PRIMITIVE_DATATYPE` case is guarded by `value != null || !hasNoExtensions(pd)`, but every statement inside was nested in a further `if (value != null)`, so a valueless primitive carrying extensions produced no RDF output at all.

That was not observable before this PR, because such elements never survived JSON parsing. With the parser fix alone, **9 `RDFParserTest.testRDFRoundTrip` cases fail** — for example `activitydefinition-administer-zika-virus-exposure-assessment.json`, which contains `"_event": [ { "extension": [ { "url": ".../cqf-expression", ... } ] } ]`. The encoder now emits a blank node without a `fhir:value` predicate to carry the extensions. The bundled ShEx schemas accept this, and all 2696 `RDFParserTest` cases pass.

I kept it in the same PR because the parser fix cannot be merged green without it. Happy to split if you would rather review them separately.

## Relationship to #8260 (precondition)

These are independent bugs in the same area, but **#8260 is a precondition** for the real-world scenario: without it, encoding a resource with a contained Bundle carrying an extension on a root-level primitive throws `HAPI-2223` before the extension loss can be reached. This PR is based on `master` and does not include #8260.

With both applied, the SDC template-extract Questionnaires that motivated this round-trip losslessly:

| resource | `templateExtractValue` in | out |
|---|---|---|
| `ChEkmQuestionnaireGonorrhoea` | 51 | 51 |
| `ChEkmQuestionnaireMpox` | 107 | 107 |

On a build with #8260 only, the first is 51 in / 47 out — the four lost are all on `Address.line` and `HumanName.given`.

## Tests

Four tests added to `JsonParserR4Test`:

- `testParseOrphanExtensionOnRepeatingPrimitive_ArrayFormIsPreserved`
- `testParseOrphanExtensionOnRepeatingPrimitive_MultipleEntriesArePreserved`
- `testParseOrphanExtensionOnRepeatingPrimitive_RoundTrips`
- `testParseOrphanExtensionOnNonRepeatingPrimitive_StillPreserved` (regression guard for the object form)

The first three fail on `master` and pass with this change; the fourth passes on both.

Verified locally: `ca.uhn.fhir.parser.*Test` in `hapi-fhir-structures-r4` — 2833 run, 0 failures (5 pre-existing skips), including the full 2696-case `RDFParserTest` round-trip suite.

## Note, not addressed here

While restructuring the RDF primitive encoder I noticed the extension loop writes `fhir:index` from a counter `int i = 0` that is never incremented, so every extension on a primitive gets index 0. That is pre-existing and orthogonal, so I have left it alone — say the word if you would like it fixed here or in a separate issue.

### DIFF
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/JsonParser.java
@@ -1405,6 +1405,23 @@ public class JsonParser extends BaseParser implements IJsonLikeParser {
 								parseAlternates(nextValue, theState, alternateName, alternateName);
 								theState.endingElement();
 							}
+						} else if (nextValue.isArray()) {
+							/*
+							 * The same thing, but for a repeating primitive: e.g. an "_line" array
+							 * with no corresponding "line" array at all. Each entry of the alternate
+							 * array is one repetition of the primitive, so an element is entered for
+							 * every index in order to keep the extensions aligned with the
+							 * repetitions they belong to.
+							 */
+							BaseJsonLikeArray array = nextValue.getAsArray();
+							for (int i = 0; i < array.size(); i++) {
+								BaseJsonLikeValue nextEntry = array.get(i);
+								theState.enteringNewElement(null, nextName);
+								if (nextEntry != null && !nextEntry.isNull()) {
+									parseAlternates(nextEntry, theState, alternateName, alternateName);
+								}
+								theState.endingElement();
+							}
 						} else {
 							getErrorHandler()
 									.incorrectJsonType(
--- a/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/RDFParser.java
+++ b/hapi-fhir-base/src/main/java/ca/uhn/fhir/parser/RDFParser.java
@@ -304,6 +304,24 @@ public class RDFParser extends BaseParser {
 	}
 
 	/**
+	 * Utility method to create a blank node for a primitive that carries no value but does carry
+	 * extensions. Such an element is legal in FHIR - the value is simply absent - so it still needs
+	 * a node for the extensions to hang off, just without a fhir:value predicate.
+	 * @param rdfModel Model to create node within
+	 * @param cardinalityIndex if a collection, this value is written as a fhir:index predicate
+	 * @return Blank node resource containing possibly fhir:index, but no fhir:value
+	 */
+	private Resource createFhirValuelessBlankNode(Model rdfModel, Integer cardinalityIndex) {
+		Resource blankNodeResource = rdfModel.createResource();
+		if (cardinalityIndex != null && cardinalityIndex > -1) {
+			blankNodeResource.addProperty(
+					rdfModel.createProperty(FHIR_NS + FHIR_INDEX),
+					rdfModel.createTypedLiteral(cardinalityIndex, XSDDatatype.XSDinteger));
+		}
+		return blankNodeResource;
+	}
+
+	/**
 	 * Builds the predicate name based on field definition
 	 * @param resource Resource being interrogated
 	 * @param definition field definition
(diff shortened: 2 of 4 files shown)


tadgh:
Formatting check succeeded!

jamesagnew (approved):
Looks good, thanks!

PR #8375, opened 2026-09-09
8342 too many candidates match

### LINKED ISSUE #8342: Update MDM to allow handling of too many matching candidates without failing
Currently if there's too many matching candidates, the MDM matching service throws a TooManyCandidatesException.

Update to allow skipping and logging and flagging for later human review.

TipzCM:
closes #8342

### DIFF
--- a/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/broker/MdmMessageListener.java
+++ b/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/broker/MdmMessageListener.java
@@ -29,7 +29,6 @@ import ca.uhn.fhir.interceptor.api.Pointcut;
 import ca.uhn.fhir.jpa.mdm.svc.IMdmModelConverterSvc;
 import ca.uhn.fhir.jpa.mdm.svc.MdmMatchLinkSvc;
 import ca.uhn.fhir.jpa.mdm.svc.MdmResourceFilteringSvc;
-import ca.uhn.fhir.jpa.mdm.svc.candidate.TooManyCandidatesException;
 import ca.uhn.fhir.jpa.subscription.model.ResourceModifiedMessage;
 import ca.uhn.fhir.jpa.topic.SubscriptionTopicUtil;
 import ca.uhn.fhir.mdm.api.IMdmSettings;
@@ -131,15 +130,6 @@ public class MdmMessageListener implements IMessageListener<ResourceModifiedMess
 				default:
 					ourLog.trace("Not processing modified message for {}", theMsg.getOperationType());
 			}
-		} catch (Exception e) {
-			if (e instanceof TooManyCandidatesException) {
-				ourLog.debug(
-						"Failed to handle MDM Matching for resource: {} since candidate matches exceeded the "
-								+ "candidate search limit",
-						theSourceResource.getIdElement());
-			}
-			log(mdmContext, "Failure during MDM processing: " + e.getMessage(), e);
-			mdmContext.addTransactionLogMessage(e.getMessage());
 		} finally {
 			// Interceptor call: MDM_AFTER_PERSISTED_RESOURCE_CHECKED
 			HookParams params = new HookParams()
--- a/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchFinderSvcImpl.java
+++ b/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchFinderSvcImpl.java
@@ -78,8 +78,10 @@ public class MdmMatchFinderSvcImpl implements IMdmMatchFinderSvc {
 	@Nonnull
 	@Transactional
 	public List<MatchedTarget> getMatchedTargets(
-		String theResourceType, IAnyResource theResource, RequestPartitionId theRequestPartitionId,
-		MdmTransactionContext theContext) {
+			String theResourceType,
+			IAnyResource theResource,
+			RequestPartitionId theRequestPartitionId,
+			MdmTransactionContext theContext) {
 
 		List<MatchedTarget> retval = matchBasedOnEid(theResourceType, theResource, theRequestPartitionId, theContext);
 		if (!retval.isEmpty()) {
@@ -99,7 +101,9 @@ public class MdmMatchFinderSvcImpl implements IMdmMatchFinderSvc {
 	}
 
 	private List<MatchedTarget> matchBasedOnEid(
-			String theResourceType, IAnyResource theResource, RequestPartitionId theRequestPartitionId,
+			String theResourceType,
+			IAnyResource theResource,
+			RequestPartitionId theRequestPartitionId,
 			MdmTransactionContext theContext) {
 
 		List<CanonicalEID> eidsFromResource = myEIDHelper.getExternalEid(theResource);
@@ -120,8 +124,7 @@ public class MdmMatchFinderSvcImpl implements IMdmMatchFinderSvc {
 			List<CanonicalEID> theEids,
 			String theResourceType,
 			RequestPartitionId theRequestPartitionId,
-			MdmTransactionContext theContext
-	) {
+			MdmTransactionContext theContext) {
 		final SearchParameterMap map = SearchParameterMap.newSynchronous();
 		final TokenOrListParam tokenOrListParam = new TokenOrListParam();
 		final String eidSystemForResourceType =
--- a/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchFinderSvcImpl.java
+++ b/hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchFinderSvcImpl.java
@@ -30,6 +30,7 @@ import ca.uhn.fhir.mdm.api.MatchedTarget;
 import ca.uhn.fhir.mdm.api.MdmMatchOutcome;
 import ca.uhn.fhir.mdm.log.Logs;
 import ca.uhn.fhir.mdm.model.CanonicalEID;
+import ca.uhn.fhir.mdm.model.MdmTransactionContext;
 import ca.uhn.fhir.mdm.rules.svc.MdmResourceMatcherSvc;
 import ca.uhn.fhir.mdm.util.EIDHelper;
 import ca.uhn.fhir.rest.api.server.IBundleProvider;
@@ -77,15 +78,16 @@ public class MdmMatchFinderSvcImpl implements IMdmMatchFinderSvc {
 	@Nonnull
 	@Transactional
 	public List<MatchedTarget> getMatchedTargets(
-			String theResourceType, IAnyResource theResource, RequestPartitionId theRequestPartitionId) {
+		String theResourceType, IAnyResource theResource, RequestPartitionId theRequestPartitionId,
+		MdmTransactionContext theContext) {
 
-		List<MatchedTarget> retval = matchBasedOnEid(theResourceType, theResource, theRequestPartitionId);
+		List<MatchedTarget> retval = matchBasedOnEid(theResourceType, theResource, theRequestPartitionId, theContext);
 		if (!retval.isEmpty()) {
 			return retval;
 		}
 
 		Collection<IAnyResource> targetCandidates =
-				myMdmCandidateSearchSvc.findCandidates(theResourceType, theResource, theRequestPartitionId);
+				myMdmCandidateSearchSvc.findCandidates(theResourceType, theResource, theRequestPartitionId, theContext);
 
 		List<MatchedTarget> matches = targetCandidates.stream()
 				.map(candidate ->
@@ -97,7 +99,8 @@ public class MdmMatchFinderSvcImpl implements IMdmMatchFinderSvc {
 	}
 
 	private List<MatchedTarget> matchBasedOnEid(
-			String theResourceType, IAnyResource theResource, RequestPartitionId theRequestPartitionId) {
+			String theResourceType, IAnyResource theResource, RequestPartitionId theRequestPartitionId,
+			MdmTransactionContext theContext) {
 
 		List<CanonicalEID> eidsFromResource = myEIDHelper.getExternalEid(theResource);
 		if (eidsFromResource.isEmpty()) {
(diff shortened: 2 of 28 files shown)


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/CandidateSearcher.java
```diff
@@ -69,9 +73,22 @@ public Optional<IBundleProvider> search(
 		systemRequestDetails.setRequestPartitionId(partitionId);
 		IBundleProvider retval = resourceDao.search(searchParameterMap, systemRequestDetails);
 
-		if (retval.size() != null && retval.size() >= myMdmSettings.getCandidateSearchLimit()) {
-			return Optional.empty();
+		if (retval.size() != null) {
+			if (retval.size() >= myMdmSettings.getCandidateSearchLimit()) {
+				ourLog.warn(
```
codeforgreen, line 78: The search criteria can have PHI. Switch over the logger to `Logs.getMdmTroubleshootingLog` which will filter those out. 
There may be a few other places where we need to do that. There are a few other places (pre-existing). It may not be as severe if only system-like users run mdm operations.

TipzCM: switched

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/broker/MdmMessageListener.java
```diff
@@ -131,15 +130,6 @@ private void matchMdmAndUpdateLinks(IBaseResource theSourceResource, ResourceMod
...
 			}
-		} catch (Exception e) {
-			if (e instanceof TooManyCandidatesException) {
-				ourLog.debug(
-						"Failed to handle MDM Matching for resource: {} since candidate matches exceeded the "
-								+ "candidate search limit",
-						theSourceResource.getIdElement());
-			}
-			log(mdmContext, "Failure during MDM processing: " + e.getMessage(), e);
-			mdmContext.addTransactionLogMessage(e.getMessage());
```
codeforgreen, line 142: Did you intend to remove this branch completely, or just the `TooManyCandidatesException` part?

The failure is logged but the exception is re-thrown in the calling method. And the transaction log entry is gone.
This sees to be a big change as it fails the message and sends it to a DLQ instead of configured. It looks like only for Kafka the DLQ and retry limit is configured so the message would resend indefinitely for other brokers unless error handling specifically configured.

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/CandidateSearcher.java
```diff
@@ -69,9 +73,22 @@ public Optional<IBundleProvider> search(
...
+						"At least {} search candidates were returned for search criteria {}; this is the configured maximum candidates to allow. Resource will be omitted from further MDM matching.",
+						retval.size(),
+						theResourceCriteria);
+				theContext.setTooManyCandidatesMatched(true);
+				return Optional.empty();
+			} else if (retval.size() >= myMdmSettings.getCandidateSearchWarnLimit()) {
```
codeforgreen, line 84: I'm going to have to challenge the existence of this configuration. What value would be a decent default and how could you determine a good default value to set? Another thing is, this pollutes the log with a bunch of repetitive logs. 

So we have a setting which determines the existence of log statements.
I would suggest to downgrade this to debug level and have only the internal default limit set at a more reasonable value since 5 seems very low versus the total candidate max threshold (10,000).
If we really want the warning, we have to decrease it's frequency to a few times until the max is rached.

With respect to the configuration existence itself, have we considered having an Open Telemetry configuration for MDM? What could happen is, a metric could show the candidate match count for the user to monitor. The action taken would not be to update a threshold to limit it but to actually tighten the rules to make the candidate match go down. If just a setting, what is the real Production use-case value? What happens to the other matches which are beyond the threshold?

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/FindCandidateByEidSvc.java
```diff
@@ -58,10 +59,13 @@ public class FindCandidateByEidSvc extends BaseCandidateFinder {
 	MdmPartitionHelper myMdmPartitionHelper;
 
 	@Override
-	protected List<MatchedGoldenResourceCandidate> findMatchGoldenResourceCandidates(IAnyResource theIncomingResource) {
+	protected List<MatchedGoldenResourceCandidate> findMatchGoldenResourceCandidates(
```
codeforgreen, line 62: What is the purpose of this change? Do we think we're going to need other information from the full set of parameters instead of just the resource? It's better to have a concrete narrowed down purpose for this method rather than having too much responsibility.

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchFinderSvcImpl.java
```diff
@@ -141,6 +150,7 @@ private List<MatchedTarget> searchForResourceByEIDs(
 		systemRequestDetails.setRequestPartitionId(theRequestPartitionId);
 		IBundleProvider search = resourceDao.search(map, systemRequestDetails);
 		List<MatchedTarget> retval = new ArrayList<>();
+		// TODO LS - too many candidates should be handled here?
```
codeforgreen, line 153: We probably should answer this question as it seems this is what this change is about.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/config/MdmSettings.java
```diff
@@ -175,11 +177,16 @@ public void setSurvivorshipRules(String theSurvivorshipRules) {
 
 	@Override
 	public int getCandidateSearchLimit() {
-		return myCandidateSearchLimit;
+		return myCandidateSearchHardLimit;
+	}
+
+	@Override
+	public int getCandidateSearchWarnLimit() {
```
codeforgreen, line 184: It looks like this is not configurable in hapi-fhir and it is set at a very low limit. I am afraid the logs would be polluted with redundant information by default.

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmResourceDaoSvcImpl.java
```diff
@@ -163,4 +170,36 @@ private boolean carriesEid(IBaseResource theGoldenResource, CanonicalEID theEid)
...
+			MdmResourceUtil.tagResourceAsTooManyMatchCandidates(theResource);
+		} else {
+			ourLog.warn("Attempt to tag resource, but no criteria for tagging provided");
+			return;
+		}
+
+		SystemRequestDetails rd = new SystemRequestDetails();
+		RequestPartitionId partitionId = (RequestPartitionId) theResource.getUserData(Constants.RESOURCE_PARTITION_ID);
+		if (partitionId == null) {
+			partitionId = RequestPartitionId.allPartitions();
```
codeforgreen, line 194: This should not happen I believe. Maybe we should log an error and maybe throw an exception if we don't believe this is a legitimate use-case.

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmResourceDaoSvcImpl.java
```diff
@@ -163,4 +170,36 @@ private boolean carriesEid(IBaseResource theGoldenResource, CanonicalEID theEid)
 				.stream()
 				.anyMatch(candidate -> Objects.equals(candidate.getValue(), theEid.getValue()));
 	}
+
+	@SuppressWarnings({"rawtypes", "unchecked"})
+	@Override
+	public void tagResourceAsUnmatched(IBaseResource theResource, MdmTransactionContext theContext) {
```
codeforgreen, line 176: missing javadoc for public method

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/api/MdmConstants.java
```diff
@@ -34,6 +34,41 @@ public class MdmConstants {
 			"http://hapifhir.io/fhir/NamingSystem/mdm-golden-resource-enterprise-id";
 	public static final String ALL_RESOURCE_SEARCH_PARAM_TYPE = "*";
 
+	/**
+	 * The namespace (system) of the tag for MDM resources that were not
+	 * matched to potential candidates.
+	 * The reason will come in the value.
+	 */
+	public static final String MDM_UNMATCHED_TAG_NAMESPACE =
+			"http://hapifhir.io/fhir/StructureDefinition/mdm-unmatched";
```
codeforgreen, line 43: Should be `NamingSystem` not `StructureDefinition`

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/CandidateSearcher.java
codeforgreen: in addition to the above - this branch is untested, if I'm not mistaken

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmResourceDaoSvcImpl.java
```diff
@@ -163,4 +170,36 @@ private boolean carriesEid(IBaseResource theGoldenResource, CanonicalEID theEid)
...
+		if (!theResource.getIdElement().hasIdPart()) {
+			ourLog.error("Cannot tag resources that have not first been persisted!");
+			return;
+		}
+
+		if (theContext.getIsBlocked()) {
+			MdmResourceUtil.tagResourceAsBlocked(theResource);
+		} else if (theContext.isTooManyCandidatesMatched()) {
+			MdmResourceUtil.tagResourceAsTooManyMatchCandidates(theResource);
```
codeforgreen, line 185: Let's say we adjust the config and re-run mdm match rules. Does the tag get cleared if it's persisted?

codeforgreen (changes requested):

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/CandidateSearcher.java
codeforgreen: sorry, I missed to mention that there is a security requirement about what to log based on the log level. If it's debug, it's ok but at warn level we can only have resource type and resource id.

`audit-and-phi.md` says `at `info` or above, log the resource type and ID instead — always`

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/broker/MdmMessageListener.java
TipzCM: Probably a mistake?

I don't remember though so i'll add back the parts other than the TooManyResources catch (because the exception will never be that now).

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/FindCandidateByEidSvc.java
TipzCM: It's already needed.

This is an override of a base class method.

One such override requires the parameters (FindCandidateByExampleSvc).

This one just happens not to require it. But it's still required in general.

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmMatchFinderSvcImpl.java
TipzCM: removed; left over from when it was on eid matching (but is not needed).

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmResourceDaoSvcImpl.java
TipzCM: It actually can be null since it's not always set.

I'll switch to using the RequestPartitionId.getPartitionFromUserDataIfPresent to make it more explicit and obvious.

But logging it as an 'error' isn't correct here (partitioning may be disabled for instance, or it could've come from somewhere that isn't the database and isn't populated - neither of which are 'error' cases and expected to have no partition id and "all partitions" is the expected result then).

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/api/MdmConstants.java
TipzCM: updated

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmResourceDaoSvcImpl.java
TipzCM: Yup.

That's why meta is used here.

When you resave, meta is cleared and recalculated.

Also, 'meta' field exists on all resource types so it's 'convenient' that way too.

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/MdmResourceDaoSvcImpl.java
TipzCM: javadoc is on the interface, not the implementation.

If you really want it here too i can put it here though. But currently no other implementations in this class redocument.

hapi-fhir-server-mdm/src/main/java/ca/uhn/fhir/mdm/rules/config/MdmSettings.java
TipzCM: added a setter

hapi-fhir-jpaserver-mdm/src/main/java/ca/uhn/fhir/jpa/mdm/svc/candidate/CandidateSearcher.java
TipzCM: added tests for the log warn limit.

TipzCM: waiting on comments from jason and anushe for updates to AC.

PR #8385, opened 2026-09-15
Skip the npm publish when the version is already published from this commit

jdar8:
When publishing a HAPI release, the npm publish would happen before the sonatype publish. 

If the sonatype publish fails, and you need to retry, the release pipeline will fail earlier since npm will try republishing the typescript artifacts, but the npm registry forbids republishing artifacts with an existing version number. 

The new script fixes this:
- If the version and commit SHA is the same as a published version, the npm publish is skipped automatically. 
- If the version is the same as a published version, but the commit SHA is different, reject with a message.


tadgh:
Formatting check succeeded!

PR #8392, opened 2026-09-17
`$validate` uses the last-written `CodeSystem` version instead of the specified one

### LINKED ISSUE #8391: Validation rejects a code that is in the `CodeSystem` version the `ValueSet` references
If the server holds two versions of the same `CodeSystem`, `$validate` rejects codes that are valid in the
version the `ValueSet` asked for. 

This scenario is common when an IG publishes another version of a particular profile `StructureDefinition` resource. The resources written against the older version need to still validate. It looks like this does not work in some cases.

See the attached pull request failing tests in `ValidateWithMultiVersionTerminologyR4Test`. There are a total of 4 use-cases failing. The most common one is described below.

Follow-ups:
- https://github.com/hapifhir/hapi-fhir/issues/8402
- https://github.com/hapifhir/hapi-fhir/issues/8403

### To reproduce

Installing packages is what triggers this in practice, but you don't need them. Creating the resources
directly is enough:

1. Create `CodeSystem` `http://example.org/fhir/CodeSystem/colour`, `version` `1.0.0`, `content` `complete`,
   with one code `vermilion`.
2. Create another `CodeSystem` at the same url, `version` `1.0.1`, `content` `complete`, with one code
   `cerulean`. Save this one second.
3. Create `ValueSet` `http://example.org/fhir/ValueSet/colour` that asks for version 1.0.0 of that
   CodeSystem:
   `compose.include = [{ "system": "http://example.org/fhir/CodeSystem/colour", "version": "1.0.0" }]`
4. Create a `StructureDefinition` on `Observation` with a required binding on `Observation.code` to
   `http://example.org/fhir/ValueSet/colour`.
5. `POST Observation/$validate` with `meta.profile` set to that profile and
   `code.coding = [{ "system": "http://example.org/fhir/CodeSystem/colour", "code": "vermilion" }]`.

### Expected

No error. `vermilion` is in CodeSystem 1.0.0, and that's the version the ValueSet asked for. 

### Actual

```
Unable to validate code http://example.org/fhir/CodeSystem/colour#vermilion
 - Code is not found in CodeSystem: http://example.org/fhir/CodeSystem/colour
```
Notice there is no version in the error message.

codeforgreen:
Fixes #8391

Root Cause:
With two versions of an IG loaded, `$validate` used whichever `CodeSystem` version was current rather than the version specified in the canonical URL that referenced it. Specifically in `WorkerContextValidationSupportAdapter#validateCodeInValueSet` in the branch that validates against the `CodeSystem`.


The reference could be done in multiple ways: a profile named in `meta.profile` as `url|version`, an element binding, or a `compose.include` entry. Codes valid in the specified version were rejected, and codes valid only in the other version were accepted. Install order decided which version won because it determined the current version which is the latest installed.

Fix:                                                                                                                                                                                                                           
- `WorkerContextValidationSupportAdapter` re-checked the code against the `CodeSystem`, after checking against the `ValueSet` and merged the issues into the result. For the `CodeSystem` validation it did not pass the intended version, so that validation used whichever version was current.                                                                                                                                                                                                                            
-  As a consequence of the above`IValidationSupport` now has a new `validateCode` overload that takes the `CodeSystem` version as its own parameter, with a default implementation delegating to the existing method so nothing implementing that method breaks. 
-`TermReadSvcImpl.validateCodeInValueSet` re-resolved the ValueSet by bare URL, landing on whichever version was saved last. 

Alternatives considered:
- Passing the canonical url for `CodeSystem` to `IValidationSupport` is not a viable option, despite many internal implementations are able to process it. The reason being it is not a documented behaviour of the interface and therefore may break validations involving custom validators.

Refactoring:                                                        
- refactored the code to use the same utilities in `UrlUtils` for parsing and building the canonical version of a url.

Documentation:
- clarified in comment why the fallback to validate against CodeSystem as part of validate against the ValueSet is needed in the place of the bug . See `WorkerContextValidationSupportAdapter#validateCodeInValueSet`.
- Clarified in `IValidationSupport` which urls are canonical and which are not. The existing validateCode and lookupCode methods that don't have a separate parameter for version have been left unchanged as canonical support is not enforced (by this interface) or guaranteed (for internal validators).

Tests: 
- 56 brand new tests. I added test coverage to ensure use-cases are well tested in different combinations.


Follow-ups:
- https://github.com/hapifhir/hapi-fhir/issues/8402
- https://github.com/hapifhir/hapi-fhir/issues/8403


tadgh:
Formatting check succeeded!

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
```diff
@@ -2687,7 +2688,11 @@ public CodeValidationResult validateCodeInValueSet(
 		}
 		String url = urlPrimitive.getValueAsString();
 		if (isNotBlank(url)) {
-			return validateCode(theValidationSupportContext, theOptions, theCodeSystem, theCode, theDisplay, url);
+			// A URL with no version resolves to whichever version was saved last, so keep the version of the ValueSet
+			String version = CommonCodeSystemsTerminologyService.getValueSetVersion(myContext, theValueSet);
+			String canonicalUrl = isNotBlank(version) ? url + "|" + version : url;
+			return validateCode(
+					theValidationSupportContext, theOptions, theCodeSystem, theCode, theDisplay, canonicalUrl);
```
tadgh, line 2694: ah nice find

tadgh (approved):

dotasek (approved):

hapi-fhir-jpaserver-base/src/main/java/ca/uhn/fhir/jpa/term/TermReadSvcImpl.java
codeforgreen: There are so many places where we construct and de-construct canonical urls in validators. I'll open an issue for implementing some refactoring for robustness.
If you look at `IWorkerContext` from hl7.fhir you will notice the APIs don't let any ambiguity in about weather or not the version is present. However `IValidationSupport` is ambiguous. Not only does it not declare version for both `CodeSystem` and `ValueSet` for all cases, but there are internal implementations that process canonical because of other callers (not `$validate`).

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -308,6 +313,39 @@ default CodeValidationResult validateCode(
...
+	 */
+	@Nullable
+	default CodeValidationResult validateCode(
+			@Nonnull ValidationSupportContext theValidationSupportContext,
+			@Nonnull ConceptValidationOptions theOptions,
+			@Nullable String theCodeSystem,
+			@Nullable String theCodeSystemVersion,
+			@Nullable String theCode,
+			@Nullable String theDisplay,
+			@Nullable String theValueSetUrl) {
```
codeforgreen: I have not added another parameter to capture the `ValueSet` version. It is true that some implementations already interpret it as canonical. This method would become obsolete once the long-term solution is implemented where everything is captured within a request object. This is a short-term solution. The long-term solution would produce quite a larger sized PR and we need this fixed asap.

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
```diff
@@ -308,6 +313,40 @@ default CodeValidationResult validateCode(
 		return null;
 	}
 
+	/**
```
jamesagnew, line 323: Do we really need a new method for this concept? Could we not just allow (and recommend) that a versioned URL be used for `theCodeSystem`? I worry about this interfacing getting more and more bloated, it's an interface that does get implemented by users too, and the more ways to do the same thing we add, the more potential for bugs it has...

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
```diff
@@ -98,11 +105,19 @@ private static String extractCodeSystemForCodeR4(ValueSet theValueSet, String th
...
+	 * <p>
+	 * The code system URL is returned unchanged if it already carries a version, or if no version is given.
+	 * </p>
+	 *
+	 * @param theCodeSystem the code system URL, which may already be of the form <code>system|version</code>
+	 * @param theVersion    the code system version, or <code>null</code> to return the code system URL unchanged
+	 * @return the code system URL with the version appended, or the code system URL unchanged
+	 */
+	@Nullable
+	public static String getVersionedCodeSystem(@Nullable String theCodeSystem, @Nullable String theVersion) {
```
jamesagnew: This method seems to be partially duplicative (although not completely overlapping) to `UrlUtil.parseCanonicalUrl`. And that method is more fully developed, in that it understands both `|` and `%7C` as separators.

Honestly I'd recommend just replacing calls to `ValidationSupportUtils.getVersionedCodeSystem(url, version)` with a call to `UrlUtil.parseCanonicalUrl(url, version).toString()`, which is functionally equivalent.

At a minimum though, this method should be renamed to not reference CodeSystem, since there is nothing codesystem specific going on here, and it should delegate the work to UrlUtil.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
```diff
@@ -141,4 +156,142 @@ private static String extractCodeSystemForCodeR5(org.hl7.fhir.r5.model.ValueSet
...
+	 *
+	 * @param theFhirContext the FHIR context the ValueSet belongs to
+	 * @param theValueSet    the ValueSet to read the URL from
+	 * @return the ValueSet's URL
+	 * @throws IllegalArgumentException if the FHIR version is not supported
+	 * @since 8.14.0
+	 */
+	// Created by Claude Opus 5
+	@Nullable
+	public static String getValueSetUrl(@Nonnull FhirContext theFhirContext, @Nonnull IBaseResource theValueSet) {
```
jamesagnew: This whole method could be replaced by the following single line, which would automatically be compatible with future versions and would work on any resource that has a canonical version: https://github.com/hapifhir/hapi-fhir/blob/3baf7f9454621aa8b4d5fd7726fcf22c45fff031/hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java#L168

Honestly, it might be nice to have a utility somewhere that returns a `UrlUtil.CanonicalUrlParts` by pulling both the `url` and `version` properties off of an object in a single call.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
```diff
@@ -141,4 +156,142 @@ private static String extractCodeSystemForCodeR5(org.hl7.fhir.r5.model.ValueSet
...
+	 *
+	 * @param theFhirContext the FHIR context the ValueSet belongs to
+	 * @param theValueSet    the ValueSet to read the version from
+	 * @return the ValueSet's version, or <code>null</code> if the FHIR version has no version element
+	 * @since 8.14.0
+	 */
+	// Created by Claude Opus 5
+	@Nullable
+	public static String getValueSetVersion(@Nonnull FhirContext theFhirContext, @Nonnull IBaseResource theValueSet) {
+		String version;
```
jamesagnew: This could theoretically be replaced by the same suggestion as above.

jamesagnew (changes requested):

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
codeforgreen: We could do that. I did that initially. But now (this fix) we call validateCode with canonical from the $validate call. If we are sure customer implementations parse it then it is ok. Or we just enforce it? Previously I believe calls with canonical came from other calls not $validate.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
codeforgreen: Sure makes sense

hapi-fhir-base/src/main/java/ca/uhn/fhir/context/support/IValidationSupport.java
codeforgreen: I believe ultimately what we want is having URL and version separate to as this is the pattern in both hl7.fhir library and the fhir standard operations definitions. For a more elegant addition, I updated the new method to use a request object. I've raised the full refactoring as https://github.com/hapifhir/hapi-fhir/issues/8402. I intend to work on it sometime after this fix is shipped.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
codeforgreen: I decided to keep this part (refactoring) out of scope as it is a nice to have. I filed it as: https://github.com/hapifhir/hapi-fhir/issues/8403

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
codeforgreen: I decided to keep this part (refactoring) out of scope as it is a nice to have. I filed it as: https://github.com/hapifhir/hapi-fhir/issues/8403

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
codeforgreen: I decided to keep this part (refactoring) out of scope as it is a nice to have. I filed it as: https://github.com/hapifhir/hapi-fhir/issues/8403

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
codeforgreen: Sorry, I actually got rid of this and used `UrlUtils`, as per your suggestion. I also created a light wrapper around `parseCanonicalUrl` and named it `toCanonicalUrl` to return `String` instead of `CanonicalUrlParts`.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
codeforgreen: Sorry, I actually got rid of this and used `UrlUtils`, as per your suggestion. I also created a light wrapper around `parseCanonicalUrl` and named it `toCanonicalUrl` to return `String` instead of `CanonicalUrlParts`.

hapi-fhir-validation/src/main/java/org/hl7/fhir/common/hapi/validation/support/ValidationSupportUtils.java
codeforgreen: Sorry, I actually got rid of this and used `UrlUtils`, as per your suggestion. I also created a light wrapper around `parseCanonicalUrl` and named it `toCanonicalUrl` to return `String` instead of `CanonicalUrlParts`.

hapi-fhir-base/src/main/java/ca/uhn/fhir/util/UrlUtil.java
```diff
@@ -639,10 +639,11 @@ public static CanonicalUrlParts parseCanonicalUrl(@Nullable String theUrl) {
 	/**
 	 * Parses a versioned or unversioned canonical URL (e.g. <code>http://foo</code> or <code>http://foo|123</code>)
 	 * into its constituent parts. An optional version ID can also be provided, for scenarios where the version can be
-	 * provided either in the URL or in a separate parameter. If both are provided, the version from the parameter
-	 * takes precedence.
+	 * provided either in the URL or in a separate parameter. If both are provided and they agree, that version is
+	 * returned; if they disagree, neither is chosen and the call fails.
```
jamesagnew, line 643: This feels like a sensible change. Nice.

jamesagnew (approved):

PR #8393, opened 2026-09-17
Fixing 8308 changelog type

jorgepinilla-smile:
- fixed value of changelog type to "perf"
- added changelog type constants
- added assertion in existing changelog test that validates the changelog type value
- updated migration unit test to use changelog type constants


tadgh:
Formatting check succeeded!

tynerguo (commented):
Suggestion: this change feels substantial enough to have its own change log. Maybe something like "added new changelog type for performance improvement changes."

Otherwise LGTM!

tadgh (approved):

PR #8401, opened 2026-09-18
8383 bulk export output formats

### LINKED ISSUE #8383: Add pointcut and conversion option for bulk export job
Currently Bulk Export only allows ndjson output.

`outputFormat` is already a parameter; but no others are supported.

We need to add a means for other formats to be supported (and provide a pointcut to allow for customers to define their rules for conversion of resources -> data.

TipzCM:
closes #8383


tadgh:
Formatting check succeeded!

PR #8404, opened 2026-09-20
[8387] STORAGE_PARTITION_SELECTED is not invoked for partition-selecting operations without a resource type

### LINKED ISSUE #8387: STORAGE_PARTITION_SELECTED is not invoked for partition-selecting operations without a resource type
**Description**

`BaseRequestPartitionHelperSvc.validateAndNormalizePartition` only fires the `STORAGE_PARTITION_SELECTED` pointcut when a resource type is present:

```java
if (StringUtils.isNotBlank(theResourceType)) {
    validateHasPartitionPermissions(theRequest, theResourceType, retVal);
}
```

Partition-selecting operations that carry no resource type therefore never reach interceptors registered on `STORAGE_PARTITION_SELECTED`, even though the caller-chosen partition has already been resolved and is used for the query. Affected paths:

- System-level history (`GET [base]/_history`) via `determineReadPartitionForRequestForHistory(req, null, null)`
- Result paging (`_getpages`) via `ReadPartitionIdRequestDetails.forSearchUuid(...)` / `forHistory(null, null)`
- `determineGenericPartitionForRequest` for server-level requests (e.g. `POST [base]` transactions)
- `ReadPartitionIdRequestDetails.forServerOperation(...)` callers (`$expunge`, `$delete-expunge`, MDM server operations, ...)

Bulk import/export providers already work around this by calling `validateHasPartitionPermissions(req, "Binary", pid)` manually, so the gap has been patched case by case rather than at the source.

**Reproduce Steps**

1. `partitioning.enabled=true`, `RequestHeaderPartitionInterceptor` registered, named partitions `1` and `2`.
2. Register a `STORAGE_PARTITION_SELECTED` interceptor that throws `ForbiddenOperationException` when partition `2` is selected.
3. `GET /Patient/_history` with `X-Request-Partition-IDs: 2` → 403 as expected.
4. `GET /_history` with `X-Request-Partition-IDs: 2` → 200 with partition 2's full resource history. The interceptor is never invoked. Following the `next` link behaves the same.

**Expected behavior**

`STORAGE_PARTITION_SELECTED` should fire for every partition-selecting resolution regardless of resource type, with a null `RuntimeResourceDefinition` for system-level operations. This is already the documented contract of the pointcut ("Invoked before any partition aware FHIR operation"; the `RuntimeResourceDefinition` parameter is "null if no specific type is associated with the request"), `validateHasPartitionPermissions` already handles a null type, and `BaseBulkModifyOrRewriteProvider` already invokes the hook with an explicit null definition.

Proposed fix: remove the `isNotBlank(theResourceType)` guard so `validateHasPartitionPermissions` is always called. Interceptors on this pointcut must then tolerate a null `RuntimeResourceDefinition`; this should be called out in the upgrade notes.

**Environment**

- HAPI FHIR Version: master (8.13.7-SNAPSHOT), present since 5.x

YalingPeiS:
Fixes #8387

## Problem

`STORAGE_PARTITION_SELECTED` was only fired when the request carried a resource type. Partition-selecting requests without one — system-level `_history`, result paging, transactions, server-level operations — skipped the hook entirely, so a client scoped to one partition could read another partition's data.

## Change

`BaseRequestPartitionHelperSvc.validateAndNormalizePartition()` now fires the pointcut for every partition-selecting resolution. When no resource type applies, the `RuntimeResourceDefinition` hook parameter is `null`, which its Javadoc already allows.

## Upgrade note

Interceptors hooking `STORAGE_PARTITION_SELECTED` must handle a `null` `RuntimeResourceDefinition`.

## Tests

- `RequestPartitionHelperSvcTest` — hook fires with a null resource definition for system history, server operations, page fetches, and generic (transaction) resolution.
- `PartitioningSqlR4Test` — system-level history fires the hook, and an interceptor that throws blocks the query before it runs.


tadgh:
Formatting check succeeded!

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/upgrade.md
```diff
@@ -5,3 +5,9 @@
 HAPI FHIR now depends on Jackson 3 instead of Jackson 2. Jackson packages move from
 `com.fasterxml.jackson` to `tools.jackson`, and several Jackson defaults changed. See the
 [Jackson 3.0 release notes](https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0).
+
+## Partition security for system-level operations
+
+The `STORAGE_PARTITION_SELECTED` pointcut is now invoked for operations with no resource type (system-level
+`_history`, paging, transactions, server-level operations). Interceptors hooking it must handle a `null`
+`RuntimeResourceDefinition`, and these operations are now subject to partition security checks.
```
tadgh: suggestion: This change is IMO not significant enough to warrant an entry on this page. Recommend removal

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/partition/BaseRequestPartitionHelperSvc.java
```diff
@@ -390,9 +389,8 @@ private RequestPartitionId validateAndNormalizePartition(
 
 		// Note: It's still possible that the partition only has a date but no name/id
 
-		if (StringUtils.isNotBlank(theResourceType)) {
```
tadgh, line 393: question (non-blocking): Did you look into the git blame of why this line existed originally? It _looks_ like it was very much on purpose.

tadgh (approved):

hapi-fhir-docs/src/main/resources/ca/uhn/hapi/fhir/changelog/8_14_0/upgrade.md
YalingPeiS: Removed.

hapi-fhir-storage/src/main/java/ca/uhn/fhir/jpa/partition/BaseRequestPartitionHelperSvc.java
YalingPeiS: Ya, it looks that the whole if block was from 2022/08/29 written by James, likely the initial implementation.
Screenshot 2026-09-22 at 11 19 01 AM

